PR #192, opened 2026-02-26
fix: display auth errors inline on login form

viamin:
Move the authentication error alert from the layout's global flash area
into the login form card so users see the error immediately next to the
fields. Uses content_for(:inline_alert) to prevent duplicate display in
the layout.

https://claude.ai/code/session_01NhzQAjWwXu6VdTUAHE6Utn


viamin:
## Agent Update

That's the full RSpec run confirmation (1307 examples, 0 failures, 92.6% coverage). All tasks are complete.

PR #196, opened 2026-02-27
Fix 6 bugs in GitHub polling incompleteness detection

viamin:
Bug 1: last_completed_run now finds initial runs via pull_request_number
(not just source_pull_request_number), so the first follow-up uses the
correct cutoff time instead of treating all comments/reviews as new.

Bug 2: issue_comments now uses auto_paginate to collect all comments.
Previously only the first page (~30) was returned, silently missing
newer comments on busy PRs.

Bug 3: RecordPrFollowupActivity is now truly idempotent using an
expected_followup_count parameter for conditional atomic increment,
preventing double-counting on Temporal activity retries.

Bug 4: check_ci_failures no longer blocks on pending checks. Failed
completed checks are now detected even when unrelated checks are still
running (e.g., slow deploy gates).

Bug 5: check_runs_for_ref now paginates (up to 10 pages of 100) to
collect all check runs for repos with many workflows or matrix builds.

Bug 6: FetchIssuesActivity always includes "paid-generated" in the
label search, even when no other label mappings are configured.

https://claude.ai/code/session_014bBAWp4eFQvXGP389ntEFZ

### DIFF
--- a/app/services/github_client.rb
+++ b/app/services/github_client.rb
@@ -15,6 +15,9 @@ require "faraday/retry"
 #   issues = client.issues("owner/repo", labels: "bug")
 #
 class GithubClient
+  CHECK_RUNS_PER_PAGE = 100
+  CHECK_RUNS_MAX_PAGES = 10
+
   # Base error for all GitHub client errors
   class Error < StandardError; end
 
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

All done. Here's a summary of what was accomplished:

## Changes Made

**Commit `31c11f5`**: Address PR review feedback: extract constants and fix comment

### 1. `app/services/github_client.rb` — Extracted pagination constants
- Added `CHECK_RUNS_PER_PAGE = 100` and `CHECK_RUNS_MAX_PAGES = 10` as named constants on `GithubClient`
- Replaced 5 occurrences of hardcoded magic numbers in `check_runs_for_ref` with the new constants
- Follows the same pattern as `FetchIssuesActivity::PER_PAGE` and `FetchIssuesActivity::MAX_PAGES`
- Addresses the copilot review comment about duplicated magic numbers

### 2. `app/temporal/activities/record_pr_followup_activity.rb` — Fixed class comment
- Updated the header comment to accurately state the activity is "Called by the polling workflow both after starting a child workflow and when no agent capacity is available"
- Previously it only mentioned the child-workflow-started case, which was incomplete
- Addresses the copilot review comment about inaccurate documentation

### 3. Conversation comment from viamin
- The changes from commit `08aac8a` (`with_lock` + `increment!` and `per_page: 100, page: 1`) were already present in the codebase — no additional action needed.

### Environment Note
`bundle install` could not complete due to the `agent-harness` gem requiring Ruby >= 3.3.0 while this environment has Ruby 3.2.3. Rubocop was run standalone and confirmed 0 offenses on both changed files.

PR #222, opened 2026-03-01
Add stale run detector job

viamin:
## Summary
- Adds `StaleRunDetectorJob` cron job (every 5 min) that catches agent runs stuck in "running" or "pending" beyond the configured agent timeout + 10min grace period
- Marks stale runs as "timeout", updates associated issue state to "failed", logs the resolution, and triggers `ProcessRunQueueJob` to unblock queued runs
- Registered in GoodJob cron schedule alongside existing health check jobs

## Context
Agent run #192 was stuck in "running" for 15+ hours despite the agent having completed successfully. The Temporal workflow died after `RunAgentActivity` returned but before `MarkAgentRunCompleteActivity` could execute, leaving the run orphaned with no mechanism to recover.

## Test plan
- [x] 10 RSpec examples covering stale running/pending detection, threshold boundaries, finished run exclusion, issue state updates, log creation, queue job trigger, and multi-run resolution
- [x] RuboCop clean
- [x] All pre-commit hooks pass

### DIFF
--- a/app/jobs/stale_run_detector_job.rb
+++ b/app/jobs/stale_run_detector_job.rb
@@ -62,10 +62,11 @@ class StaleRunDetectorJob < ApplicationJob
     AgentRun.running.where("started_at < ?", threshold)
   end
 
-  # Runs stuck in "pending" that were created before the threshold.
-  # These never got picked up by a Temporal worker.
+  # Runs stuck in "pending" whose last update was before the threshold.
+  # Uses updated_at to approximate when the run entered "pending", since a run
+  # may have spent a long time in "queued" before transitioning to "pending".
   def stale_pending_runs(threshold)
-    AgentRun.pending.where("created_at < ?", threshold)
+    AgentRun.pending.where("updated_at < ?", threshold)
   end
 
   def resolve_stale_run(agent_run)
--- a/spec/jobs/stale_run_detector_job_spec.rb
+++ b/spec/jobs/stale_run_detector_job_spec.rb
@@ -19,7 +19,8 @@ RSpec.describe StaleRunDetectorJob do
     end
 
     it "times out runs stuck in pending beyond the threshold" do
-      stale_run = create(:agent_run, status: "pending", created_at: (stale_threshold + 60).seconds.ago)
+      stale_run = create(:agent_run, status: "pending")
+      stale_run.update_columns(updated_at: (stale_threshold + 60).seconds.ago)
 
       described_class.perform_now
 
@@ -37,7 +38,8 @@ RSpec.describe StaleRunDetectorJob do
     end
 
     it "does not touch pending runs within the threshold" do
-      recent_run = create(:agent_run, status: "pending", created_at: (stale_threshold - 60).seconds.ago)
+      recent_run = create(:agent_run, status: "pending")
+      recent_run.update_columns(updated_at: (stale_threshold - 60).seconds.ago)
 
       described_class.perform_now
 


viamin:
## Agent Update

All done. Here's a summary of the changes committed (`d2517ec`):

**Review comment 1 — Race condition in `resolve_stale_run`** (`app/jobs/stale_run_detector_job.rb:73-93`):
Wrapped the entire method body in `agent_run.with_lock do ... end` with a `reload` and `return if agent_run.finished?` guard. This prevents overwriting a terminal status (completed/failed/timeout) if the run transitions between the initial query and the update.

**Review comment 2 — Unconditional issue state update** (`app/jobs/stale_run_detector_job.rb:81-82`):
Changed to `issue.update!(paid_state: "failed") unless issue.paid_state == "failed"` to skip unnecessary writes when the issue is already in the "failed" state, consistent with `MarkAgentRunFailedActivity`.

**Review comment 3 — Spec description mismatch** (`spec/jobs/stale_run_detector_job_spec.rb:54`):
Fixed description from `"updates the issue paid_state to errored"` to `"updates the issue paid_state to failed"` to match the actual expectation and implementation.

**Review comment 4 — Hard-coded grace period in spec** (`spec/jobs/stale_run_detector_job_spec.rb:7`):
Replaced `+ 600` with `+ described_class::GRACE_PERIOD.to_i` so the spec stays aligned if the constant changes.

viamin:
## Agent Update

All three review comments have been addressed:

**1. Spec timestamp determinism** (spec lines 32, 40) — Replaced hardcoded `10.minutes.ago` with `(stale_threshold - 60).seconds.ago` so the "within threshold" examples stay correct regardless of the configured `agent_timeout` value.

**2. Unconditional resolved counter** (job line 27) — `resolve_stale_run` now returns `false` when the run has already finished (early exit on race condition) and `true` after a successful transition. The counter only increments on `true`, so `ProcessRunQueueJob` is only enqueued when runs were actually resolved.

**Note:** Tests could not run because PostgreSQL is not available in this environment. The spec changes are straightforward timestamp derivations that mirror the existing pattern used for "beyond threshold" specs.

viamin:
## Agent Update

**All 3 review threads resolved — changed pending staleness from `created_at` to `updated_at`:**

The reviewer correctly identified that `stale_pending_runs` was using `created_at` to determine if a pending run was stale. This is wrong because a run can spend a long time in the "queued" state before being claimed and transitioning to "pending" (e.g., via `AgentRun.claim_next_queued_run`). Using `created_at` would incorrectly time out runs that had been queued for a while but only recently became pending.

**Changes made:**

1. **`app/jobs/stale_run_detector_job.rb:65-69`** — Changed `stale_pending_runs` to query `updated_at < ?` instead of `created_at < ?`, with an updated comment explaining the rationale.

2. **`spec/jobs/stale_run_detector_job_spec.rb:21-23`** — Updated the "times out runs stuck in pending beyond the threshold" test to create the run first, then use `update_columns(updated_at: ...)` to backdate the `updated_at` timestamp.

3. **`spec/jobs/stale_run_detector_job_spec.rb:40-42`** — Updated the "does not touch pending runs within the threshold" test similarly.

**Lint:** All 227 Ruby files + ESLint + markdownlint + ShellCheck pass cleanly.

**Tests:** PostgreSQL is not available in this environment (not installed, no root/sudo access), so specs could not be executed. The changes are straightforward column-name swaps with no behavioral complexity.

PR #284, opened 2026-03-04
Add configurable grace period to dev:cleanup

viamin:
## Summary
- Add `STARTUP_CLEANUP_GRACE_PERIOD` env var to `dev:cleanup` rake task (default `0` for backward compatibility)
- When set (e.g., `1800` for 30 minutes), recent active runs are skipped so Temporal can recover them after a worker restart
- Container cleanup is scoped to only stale runs' containers instead of stopping all labeled containers

## Test plan
- [x] All 17 specs pass (`bin/rspec spec/tasks/dev_rake_spec.rb`)
- [x] Linter passes (`bin/lint --changed`)
- [ ] Manual: set `STARTUP_CLEANUP_GRACE_PERIOD=120`, create a run, restart, verify it's not timed out
- [ ] Manual: verify default (grace=0) preserves original aggressive cleanup behavior

### DIFF
--- a/lib/tasks/dev.rake
+++ b/lib/tasks/dev.rake
@@ -15,7 +15,7 @@ module DevCleanup
 end
 
 namespace :dev do
-  desc "Mark stale agent runs as timed out, stop orphaned containers, and process the run queue"
+  desc "Mark stale agent runs as timed out, stop their containers (+ orphans when grace=0), and process the run queue"
   task cleanup: :environment do
     raw = ENV.fetch("STARTUP_CLEANUP_GRACE_PERIOD", "0")
     grace = begin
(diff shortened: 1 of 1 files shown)


viamin:
## Agent Update

Done. Here's a summary of what was addressed:

### Review comments addressed

**Thread 1** (lib/tasks/dev.rake:55 - outdated TODO): Updated the TODO comment to use the reviewer's suggested phrasing — it now says "Integrate Containers::ServiceProvisioner here" rather than "When ServiceProvisioner is implemented," which is clearer about intent. Note: the reviewer incorrectly claimed `Containers::ServiceProvisioner` already exists — it doesn't (confirmed by codebase search and viamin's comment).

**Thread 2** (lib/tasks/dev.rake:69 - missing service container cleanup in grace>0 path): Since neither `Containers::ServiceProvisioner` nor `AgentRun#service_container_ids` exist in the codebase yet, this can't be implemented now. The TODO comment was expanded to explicitly document this gap: "In the grace>0 branch, service containers for timed-out runs are left running until the next grace=0 cleanup or manual intervention."

### Environment limitations

Tests (`spec/tasks/dev_rake_spec.rb`) could not be run because they use `create(:agent_run)` which requires PostgreSQL. RuboCop lint passes cleanly on both changed files.

PR #286, opened 2026-03-05
Fix orphaned Docker workspace volumes leaking disk space

viamin:
## Summary

- **Root cause**: When a container is already gone before cleanup runs, `ensure_container_service!` raises `ProvisionError`. The rescue block cleared `container_id` but never removed the `paid-workspace-*` Docker volume, leaving it permanently orphaned.
- **Fix**: Call `cleanup_orphaned_workspace_volume` in the rescue path of `AgentRun#cleanup_container` so volumes are removed even when the container is already gone.
- **Safety net**: Add `DockerVolumeCleanupJob` (scheduled every 4h) that sweeps any `paid-workspace-*` volumes whose agent run is no longer active, catching anything that slips through.

## Test plan

- [ ] Verify existing `cleanup_container` specs still pass
- [ ] New specs confirm volume cleanup when container is already gone
- [ ] New specs confirm `DockerVolumeCleanupJob` removes orphaned volumes and skips active ones
- [ ] Monitor volume count after deploy to confirm leak is plugged

### DIFF
--- a/app/jobs/docker_volume_cleanup_job.rb
+++ b/app/jobs/docker_volume_cleanup_job.rb
@@ -17,8 +17,10 @@ class DockerVolumeCleanupJob < ApplicationJob
     volumes = list_paid_volumes
     return if volumes.empty?
 
-    agent_run_ids = volumes.map { |v| v.id.delete_prefix(VOLUME_PREFIX) }
-    active_ids = AgentRun.active.where(id: agent_run_ids).pluck(:id).map(&:to_s).to_set
+    numeric_agent_run_ids = volumes
+                             .map { |v| v.id.delete_prefix(VOLUME_PREFIX) }
+                             .select { |id| id.match?(/\A\d+\z/) }
+    active_ids = AgentRun.active.where(id: numeric_agent_run_ids).pluck(:id).map(&:to_s).to_set
 
     removed = 0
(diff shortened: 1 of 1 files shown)


viamin:
## Agent Update

All four review comments have been addressed and committed. Here's a summary:

1. **spec/models/agent_run_spec.rb:891** — Changed from `have_received(:get)` (which requires a prior stub) to `expect(Docker::Volume).not_to receive(:get)` (message expectation set before the call), as the reviewer suggested.

2. **app/models/agent_run.rb:389** — Moved the `ensure_container_service!` comment down to the correct method and added a proper description for `cleanup_orphaned_workspace_volume`.

3. **app/jobs/docker_volume_cleanup_job.rb:24** — Eliminated the N+1 query by bulk-loading all active run IDs in a single query (`AgentRun.active.where(id: agent_run_ids).pluck(:id)`) and checking membership in a `Set`.

4. **app/jobs/docker_volume_cleanup_job.rb:33** — Made `remove_volume` return `true`/`false` to indicate success, and only increment `removed` when removal actually succeeds.

**Note:** Tests could not be run because no database is available in this environment. All 181 spec failures are from `create(:account)` requiring a PostgreSQL connection — none are related to the code changes. Lint passed cleanly.

PR #294, opened 2026-03-06
Improve GitHub token UX and detect expiration

viamin:
## Summary
- Add optional **Workflows** permission hint to the token creation page, so users know it's required if Paid needs to edit `.github/workflows/` files (discovered via a failed agent run that couldn't push workflow changes)
- **Detect and save token expiration** during validation — fine-grained PATs include a `github-authentication-token-expiration` response header that we now parse and persist to `expires_at`, making the existing "Expiring Soon" / "Expired" badges functional
- Remove the **Deactivate button** from the tokens index page to prevent accidental clicks (it remains accessible on the token detail page)

## Test plan
- [x] Existing `github_client_spec` and `github_token_spec` pass (109 examples, 0 failures)
- [ ] Verify token creation page shows the new workflow scope hint
- [ ] Verify tokens index no longer shows Deactivate button
- [ ] Verify Deactivate is still available on the token show page
- [ ] Create a fine-grained PAT with expiration and validate — confirm `expires_at` is populated
- [ ] Validate a classic PAT — confirm `expires_at` remains nil

### DIFF
--- a/app/models/github_token.rb
+++ b/app/models/github_token.rb
@@ -103,7 +103,10 @@ class GithubToken < ApplicationRecord
   # @raise [GithubClient::AuthenticationError] if the token is invalid
   def validate_with_github!
     result = client.validate_token
-    update!(scopes: result[:scopes], expires_at: result[:expires_at])
+    expires_at = result[:expires_at]
+    attrs = { scopes: result[:scopes] }
+    attrs[:expires_at] = expires_at if expires_at
+    update!(attrs)
     touch_last_used!
     sync_repositories!
     result


viamin:
## Agent Update

**Review comment 1** (`github_client.rb:555`): Moved the `rescue ArgumentError` from `parse_timestamp` (which is shared with PR review `submitted_at` parsing) into `token_expiration_from_response` where it's specifically needed. This keeps strict timestamp parsing for PR reviews while gracefully handling invalid expiration headers.

**Review comment 2** (`github_token.rb:107`): Added two model specs to `github_token_spec.rb`:
- Asserts `expires_at` remains `nil` for classic PATs without the expiration header
- Asserts `expires_at` is persisted with the correct value when the `github-authentication-token-expiration` header is present (fine-grained PATs)

**CI/Tests**: The `github_client_spec` passes (48 examples, 0 failures). The new `github_token_spec` tests require a database and will pass in CI. Lint passes cleanly.

PR #301, opened 2026-03-07
Replace container_cpus with max_concurrent_runs setting

viamin:
## Summary

- Replaces the "Container CPUs" user setting (per-container CPU allocation) with "Max Concurrent Runs" (concurrency limit)
- `AgentRun.has_run_capacity?` and `CheckRunCapacityActivity` now read from the user setting instead of only the `MAX_CONCURRENT_RUNS` env var
- The env var still serves as a fallback when no user settings exist

## Test plan

- [x] All 225 affected specs pass
- [ ] Verify the setting appears correctly in the UI
- [ ] Set max concurrent runs to N and confirm N agent runs can execute simultaneously

### DIFF
--- a/app/jobs/process_run_queue_job.rb
+++ b/app/jobs/process_run_queue_job.rb
@@ -44,8 +44,14 @@ class ProcessRunQueueJob < ApplicationJob
         end
 
         # User has capacity — now atomically claim the run.
+        # claim_next_queued_run returns nil if another process claimed or
+        # transitioned this run between peek and claim. Skip it and continue
+        # processing the queue rather than stopping entirely.
         agent_run = AgentRun.claim_next_queued_run(target_id: next_run.id)
-        break unless agent_run
+        unless agent_run
+          skipped_ids.add(next_run.id)
+          next
+        end
 
         if start_claimed_run(agent_run)
           consecutive_failures = 0
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -107,9 +107,13 @@ class AgentRun < ApplicationRecord
   # Returns the count of active runs owned by the given user.
   # Uses a JOIN on projects.created_by_id to avoid a subquery per call.
   #
-  # Projects with nil created_by_id are not counted toward any user's cap.
-  # Callers should use Project#effective_owner to resolve a fallback user
-  # so orphaned projects are still subject to per-user limits.
+  # Limitation: projects with nil created_by_id are not counted toward any
+  # user's per-user cap. Callers use Project#effective_owner to resolve a
+  # fallback user for capacity *limit* lookups, but the counting here only
+  # matches on created_by_id. This means orphaned-project runs bypass the
+  # per-user cap (but are still bounded by the system-wide cap checked in
+  # has_run_capacity?). Replicating the effective_owner fallback chain in
+  # SQL would add significant complexity for a rare edge case (deleted user).
   def self.active_count_for_user(user)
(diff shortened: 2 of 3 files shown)


viamin:
## Agent Update

The commit succeeded. Here's what was done:

**Review comment addressed:** Split the `"sets default container resource values"` example in `spec/models/user_setting_spec.rb` into two separate examples:
- `"sets default container resource values"` — tests `container_memory_bytes` and `container_timeout_seconds`
- `"sets default concurrency values"` — tests `max_concurrent_runs`

This matches the reviewer's suggestion exactly, since `max_concurrent_runs` is a concurrency setting rather than a container resource.

**Verification:**
- RuboCop: 292 files inspected, no offenses
- Pre-commit hooks passed
- PostgreSQL was not reachable in this environment, so rspec could not run (consistent with previous agents' experience on this PR)

The commit succeeded. Here's what was done:

**Review comment addressed:** Split the `"sets default container resource values"` example in `spec/models/user_setting_spec.rb` into two separate examples:
- `"sets default container resource values"` — tests `container_memory_bytes` and `container_timeout_seconds`
- `"sets default concurrency values"` — tests `max_concurrent_runs`

This matches the reviewer's suggestion exactly, since `max_concurrent_runs` is a concurrency setting rather than a container resource.

**Verification:**
- RuboCop: 292 files inspected, no offenses
- Pre-commit hooks passed
- PostgreSQL was not reachable in this environment, so rspec could not run (consistent with previous agents' experience on this PR)

viamin:
## Agent Update

The commit succeeded. All 2189 examples passed with 0 failures, lint passed, and the pre-commit hook completed successfully.

## Summary

The CI test failure was caused by a **`NameError` in `CheckRunCapacityActivity#execute`** (`app/temporal/activities/check_run_capacity_activity.rb`).

**Root cause:** The variables `user_active_count` and `user_max` were only defined inside the `if user` branch, but were referenced unconditionally in both the `logger.info` call and the return hash. When no user context was provided (i.e., no `project_id` in the input), these variables were undefined, causing a `NameError` at runtime.

**Fix:** Initialized both variables to `nil` before the conditional block (lines 11-12), so they are always defined regardless of which branch is taken. This is consistent with the return hash already expecting `nil` values for these keys when no user context exists.

viamin:
## Agent Update

All changes are committed. Here's a summary of what was addressed for each review thread:

### 1. `Project#effective_owner` fallback order (app/models/project.rb:98)
**Fixed.** Changed the fallback from `members.first` to `account.account_memberships.find_by(role: :owner)&.user` to match the fallback order used by `AgentRuns::UserSettingsResolver` (created_by → account owner → first account user). The reviewer suggested `account.owner` but `Account` doesn't have an `owner` method, so I used the same `account_memberships` lookup that `UserSettingsResolver` uses.

### 2. `ProcessRunQueueJob` break → next on failed claim (app/jobs/process_run_queue_job.rb:48)
**Fixed.** When `claim_next_queued_run` returns nil (another process claimed the run between peek and claim), the job now adds the run to `skipped_ids` and continues with `next` instead of `break`. This ensures other queued runs still get processed.

### 3. `active_count_for_user` orphaned projects (app/models/agent_run.rb:115)
**Addressed with documentation.** Updated the comment to clearly document the limitation: projects with `nil created_by_id` bypass per-user caps but remain bounded by the system-wide cap. Replicating the `effective_owner` fallback chain in SQL would add significant complexity for what is a rare edge case (deleted user).

PR #338, opened 2026-03-15
fix(devcontainer): gracefully handle missing gh auth in signing setup

viamin:
## Summary

- **`setup-signing-key.sh`**: Warns and exits 0 (instead of hard-failing) when `gh` is missing or unauthenticated, so devcontainer creation doesn't break for users who haven't logged in yet.
- **`enable-commit-signing.sh`** (new): One-command wrapper that handles `gh auth login`, scope refresh, and key setup.
- **README**: Documents both the easy and manual commit-signing flows.

## Test plan

- [ ] Open a new devcontainer without `gh` authenticated — setup should complete with a warning, not fail
- [ ] Run `bash .devcontainer/enable-commit-signing.sh` and verify it walks through auth + key setup
- [ ] Verify signed commits work after running the script


viamin:
## Agent Update

The background rspec run completed successfully (exit code 0). The commit is already done and the working tree is clean — no further action needed.

PR #363, opened 2026-03-17
Fix #247: Auto-detect project service needs from repository files

### LINKED ISSUE #247: Auto-detect project service needs from repository files
## Context

Even with a service container admin UI, operators must manually configure which services each project needs. For a platform that manages many projects, this doesn't scale.

## Problem

Projects often declare their service dependencies in standard files:
- **Ruby/Rails**: \`Gemfile\` contains \`gem "pg"\`, \`gem "redis"\`
- **Docker Compose**: \`docker-compose.yml\` declares \`postgres\`, \`redis\`, \`selenium\` services
- **CI configs**: \`.github/workflows/*.yml\` declare service containers
- **Database config**: \`config/database.yml\` indicates database type

Currently, none of this is used to auto-configure service containers.

## Scope

- [ ] Add a service (e.g., \`Projects::DetectServices\`) that inspects repository files to determine required services
- [ ] Parse \`Gemfile\` / \`package.json\` for database/cache dependencies
- [ ] Parse \`docker-compose.yml\` for service declarations
- [ ] Map detected dependencies to \`ServiceContainer\` records (e.g., \`pg\` gem → \`postgres:16\` container)
- [ ] Run detection on project import and allow manual override via UI
- [ ] Add a "re-detect" button on the project settings page

## Dependencies

- Depends on the admin UI issue for service container management — auto-detection creates the same \`ServiceContainer\` + \`ProjectServiceContainer\` records, so those models and associations must be fully working first
- Could also be implemented as a standalone service that creates records directly, but the UI is needed for overrides

## Notes

- This is a Phase 2 enhancement — manual configuration via the admin UI is sufficient for initial operation
- Consider making detection conservative (suggest rather than auto-apply) to avoid false positives

viamin:

### New: `app/services/projects/detect_services.rb`
Service that inspects repository files via the GitHub API to detect service dependencies:
- **Gemfile** - detects gems like `pg`, `redis`, `sidekiq`, `mysql2`, `mongoid`, `elasticsearch`, `dalli`, `selenium-webdriver`
- **package.json** - detects npm packages like `pg`, `ioredis`, `mongoose`, `@elastic/elasticsearch`
- **docker-compose.yml / compose.yml** - detects services from image names and service names
- **config/database.yml** - detects database adapter (handles ERB templates and YAML aliases)

Maps detected dependencies to canonical service names (e.g., `sidekiq` -> `redis`, `pg` -> `postgres`) and matches them against existing `ServiceContainer` records. Returns a `Result` with `detected`, `matched`, and `unmatched` collections.

### Modified: `app/controllers/projects_controller.rb`
- Added `detect_services` action that calls the service, auto-associates matched containers, and reports results via flash notice
- Handles both detected+matched and unmatched scenarios with informative messages

### Modified: `config/routes.rb`
- Added `post :detect_services, on: :member` route for projects

### Modified: `app/views/projects/edit.html.erb`
- Added "Detect from Repo" button in the Service Containers section header

### New: `spec/services/projects/detect_services_spec.rb`
19 test cases covering Gemfile parsing, package.json parsing, docker-compose.yml parsing, database.yml parsing, ERB handling, deduplication, invalid file handling, and empty results.

---

Closes #247

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -97,9 +97,8 @@ class ProjectsController < ApplicationController
     result = Projects::DetectServices.call(project: @project)
 
     if result.any_detected?
-      added = apply_detected_services(result.matched)
-      notice = build_detection_notice(added, result)
-      redirect_to edit_project_path(@project), notice: notice
+      added = result.apply(@project)
+      redirect_to edit_project_path(@project), notice: result.notice_message(added)
     else
       redirect_to edit_project_path(@project), notice: "No service dependencies detected in repository files."
     end
@@ -133,26 +132,6 @@ class ProjectsController < ApplicationController
       :poll_interval_seconds, :github_id, :default_branch,
       :owner_reviewer_login, :merge_method, :max_draft_review_rounds, :auto_pick_enabled,
       allowed_github_usernames: [])
-  end
-
-  def apply_detected_services(containers)
-    added = []
-    containers.each do |container|
-      psc = @project.project_service_containers.find_or_create_by!(service_container: container)
-      added << container.name if psc.previously_new_record?
-    rescue ActiveRecord::RecordNotUnique, ActiveRecord::RecordInvalid
-      next
-    end
-    added
-  end
-
-  def build_detection_notice(added, result)
-    parts = []
-    parts << "Added #{added.join(', ')}." if added.any?
-    parts << "#{result.unmatched.map { |d| d[:service] }.join(', ')} detected but no matching service container exists." if result.unmatched.any?
-    already_count = result.matched.size - added.size
-    parts << "#{already_count} already associated." if already_count > 0
-    parts.join(" ").presence || "All detected services are already associated."
   end
 
   def parse_usernames_csv
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ProjectsController < ApplicationController
-  before_action :set_project, only: [ :show, :edit, :update, :destroy, :toggle_auto_pick ]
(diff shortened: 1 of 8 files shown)


viamin:
## Agent Update

The background agent confirmed what we already handled — deps installed fine, DB wasn't available. All work is committed and clean.

viamin:
## Agent Update

**Review 1 - Explicit requires**: Added `require "json"` and `require "yaml"` to `detect_services.rb` so it doesn't rely on Rails autoload.

**Review 2 - YAML bomb mitigation**: Replaced `YAML.safe_load(..., aliases: true)` with a `safe_yaml_load` method that parses via `Psych.parse` to get the AST, counts alias nodes with `count_yaml_aliases`, and rejects payloads exceeding `MAX_YAML_ALIASES` (100) before calling `to_ruby`. This prevents exponential expansion from small YAML bomb payloads.

**Review 3 - Thin controller**: Moved `apply_detected_services` and `build_detection_notice` from `ProjectsController` into `Projects::DetectServices::Result` as `#apply` and `#notice_message`. The controller action is now 3 lines of orchestration.

**Review 4 - Narrow rescue**: Changed the rescue from `ActiveRecord::RecordNotUnique, ActiveRecord::RecordInvalid` to just `ActiveRecord::RecordNotUnique`, so only the expected uniqueness race condition is silently handled. Unexpected validation errors will now propagate.

viamin:
## Agent Update

The fix replaces two spec examples that stubbed `Psych.parse` to raise `Psych::DisallowedClass` — an exception that `Psych.parse` never actually raises (it comes from `Psych.safe_load`). Instead, the tests now provide actual malformed YAML content with tab characters, which triggers a real `Psych::SyntaxError` through the production `safe_yaml_load` code path, properly exercising the `Psych::Exception` rescue clause.

PR #395, opened 2026-03-22
fix(containers): enforce wall clock exec timeouts via watchdog

viamin:
## Summary
- route wall-clock exec deadlines through the container watchdog instead of `Timeout.timeout`
- stop hung Docker exec calls by treating total elapsed time as a watchdog timeout reason
- add a regression spec covering a wall-clock timeout with no output

## Why
Agent runs like `2100` and `2102` were recording `wall_clock_timeout: Command timed out after 3600 seconds`, but the actual run durations were much longer. The underlying issue was that `Timeout.timeout` could not reliably interrupt a blocked Docker/Excon exec call, so the timeout was observed late.

## Testing
- `COVERAGE=false bundle exec rspec spec/services/containers/provision_spec.rb`

### DIFF
--- a/app/services/containers/provision.rb
+++ b/app/services/containers/provision.rb
@@ -735,8 +735,8 @@ module Containers
     # or a normal return with a non-zero exit code).
     #
     # timeout_check.timeout_reason_ref is a lambda that reads the shared
-    # timeout_reason variable under the mutex, ensuring we see the latest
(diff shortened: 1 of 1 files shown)


viamin:
## Agent Update

This is the successful commit confirmation — `1bf9b07` committed cleanly with lint passing. Everything is complete.

Issue #407, opened 2026-03-22
feat(providers): add OpenCode support to paid-agent containers


viamin:
## Summary
Add end-to-end OpenCode support to the prebuilt \ image so the provider can be added from the Providers UI and used for agent runs and fallback.

## Why
We are tightening the product contract so providers shown in the app must already be runnable inside the prebuilt agent container. Right now only Claude is installed and wired end to end, which means other registry-backed providers cannot be added or used as fallback yet.

## Scope
- Install OpenCode CLI in [docker/agent/Dockerfile](https://github.com/viamin/paid/blob/main/docker/agent/Dockerfile)
- Add any provider-specific auth/config mounts or environment wiring in [app/services/containers/provision.rb](https://github.com/viamin/paid/blob/main/app/services/containers/provision.rb)
- Add or confirm the runtime command mapping in [app/temporal/activities/run_agent_activity.rb](https://github.com/viamin/paid/blob/main/app/temporal/activities/run_agent_activity.rb)
- Mark the provider as container-executable in [lib/provider_support.rb](https://github.com/viamin/paid/blob/main/lib/provider_support.rb)
- Add request/model/activity coverage proving the provider can be added and participate in fallback ordering

## Acceptance Criteria
- The provider can be selected from the Add Provider page
- A user can enable it for agent runs and fallback
- It appears in Provider Priority / fallback ordering
- Future agent runs can invoke the provider successfully in \
- Auth works through the supported mode for that provider (subscription mount, proxy/API key mode, or both)
- Documentation/comments are updated to reflect the new container capability

viamin:
**Blocked on upstream:** agent-harness needs to fix its `build_command` for OpenCode before this can work end-to-end.

- Depends on: viamin/agent-harness#32

viamin:
Pull request created: https://github.com/viamin/paid/pull/449

PR #409, opened 2026-03-22
fix(providers): only offer paid-agent-ready providers

viamin:
## Summary
This tightens the Providers UX to match actual `paid-agent` capability.

- only offer providers that are already installed in the prebuilt `paid-agent` image
- distinguish unsupported providers from known-but-not-installed providers in the create validation
- show empty-state messaging on the Providers index and add form when there is nothing more to add
- add request/model coverage for the new boundary

## Review Notes
Potential blocker fixes included in this diff:
- stopped advertising an Add Provider action when there are no addable providers left
- clarified controller errors so unknown providers do not look like temporarily unavailable ones

## Verification
- `COVERAGE=false bundle exec rspec spec/requests/providers_spec.rb spec/models/provider_spec.rb`
- `bundle exec rubocop app/controllers/providers_controller.rb app/models/provider.rb lib/provider_support.rb`

## Follow-up Issues
- #402 Cursor
- #403 Codex
- #404 GitHub Copilot
- #405 Aider
- #406 Gemini
- #407 OpenCode
- #408 Kilocode

### DIFF
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -76,8 +76,6 @@ RSpec.describe "Providers" do
     it "creates a container-executable provider with agent runs enabled" do
       allow(ProviderSupport).to receive(:addable_provider_key?).and_call_original
       allow(ProviderSupport).to receive(:addable_provider_key?).with("cursor").and_return(true)
-      allow(ProviderSupport).to receive(:container_executable_provider_key?).and_call_original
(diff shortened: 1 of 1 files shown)


viamin:
## Agent Update

The PostgreSQL build background task completed, but I already used its output earlier and the commit is done. Nothing more to do — the fix is committed and ready.

PR #417, opened 2026-03-22
Fix #403: feat(providers): add Codex support to paid-agent containers

### LINKED ISSUE #403: feat(providers): add Codex support to paid-agent containers
## Summary
Add end-to-end Codex support to the prebuilt \ image so the provider can be added from the Providers UI and used for agent runs and fallback.

## Why
We are tightening the product contract so providers shown in the app must already be runnable inside the prebuilt agent container. Right now only Claude is installed and wired end to end, which means other registry-backed providers cannot be added or used as fallback yet.

## Scope
- Install OpenAI Codex CLI in [docker/agent/Dockerfile](https://github.com/viamin/paid/blob/main/docker/agent/Dockerfile)
- Add any provider-specific auth/config mounts or environment wiring in [app/services/containers/provision.rb](https://github.com/viamin/paid/blob/main/app/services/containers/provision.rb)
- Add or confirm the runtime command mapping in [app/temporal/activities/run_agent_activity.rb](https://github.com/viamin/paid/blob/main/app/temporal/activities/run_agent_activity.rb)
- Mark the provider as container-executable in [lib/provider_support.rb](https://github.com/viamin/paid/blob/main/lib/provider_support.rb)
- Add request/model/activity coverage proving the provider can be added and participate in fallback ordering

## Acceptance Criteria
- The provider can be selected from the Add Provider page
- A user can enable it for agent runs and fallback
- It appears in Provider Priority / fallback ordering
- Future agent runs can invoke the provider successfully in \
- Auth works through the supported mode for that provider (subscription mount, proxy/API key mode, or both)
- Documentation/comments are updated to reflect the new container capability

viamin:
That was the first (failed) commit attempt from before assets were built. The commit has already succeeded in the foreground — you can see it at `d288fee` with all 2717 tests passing and 0 failures.

---

Closes #403

### DIFF
--- a/app/services/containers/provision.rb
+++ b/app/services/containers/provision.rb
@@ -629,19 +629,25 @@ module Containers
         "HOME=/home/agent"
       ]
 
+      # OpenAI proxy env vars are always set so Codex CLI can route through
+      # the secrets proxy regardless of auth mode. Codex has no native login
+      # equivalent to `claude login`, so it always needs the proxy for auth.
+      env.concat([
+        "OPENAI_BASE_URL=#{proxy_base}/api/proxy/openai",
+        "OPENAI_HEADER_X_AGENT_RUN_ID=#{agent_run.id}",
+        "OPENAI_HEADER_X_PROXY_TOKEN=#{agent_run.proxy_token}"
+      ])
+
       if subscription_auth?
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

The background task was from an earlier diagnostic step that's no longer needed — the commit has already succeeded. No action required.

PR #424, opened 2026-03-23
Fix #205: Monitor GitHub security alerts and trigger agent runs to fix them

viamin:
All done. Here's a summary of the changes:

### Migration
- **`db/migrate/20260322000000_add_security_alert_fields_to_projects.rb`** — Adds 4 new columns to `projects`: `auto_scan_security` (boolean, default false), `security_severity_threshold` (string, default "high"), `security_alert_types` (jsonb, default `["dependabot"]`), `max_security_fix_runs` (integer, default 3)

### GitHub API
- **`app/services/github_client.rb`** — Added `dependabot_alerts(repo, severity:, state:, per_page:)` method that fetches Dependabot alerts from `GET /repos/{owner}/{repo}/dependabot/alerts` and returns parsed alert data (number, state, severity, package name/ecosystem, patched version, summary, URL)

### Activity
- **`app/temporal/activities/scan_security_alerts_activity.rb`** — New `ScanSecurityAlertsActivity` that:
  - Fetches open Dependabot alerts filtered by severity threshold
  - Skips alerts that already have an associated Issue record
  - Creates synthetic Issue records for actionable alerts with descriptive titles/bodies
  - Respects `max_security_fix_runs` limit per poll cycle
  - Handles API errors gracefully (NotFoundError when Dependabot isn't enabled)

### Workflow
- **`app/temporal/workflows/git_hub_poll_workflow.rb`** — Integrated `ScanSecurityAlertsActivity` into the poll loop after `ScanPaidPrsActivity`, with a patch guard for safe rollout. New alerts trigger `start_agent_workflow` for each created issue.

### Model
- **`app/models/project.rb`** — Added validations for `max_security_fix_runs` (>= 0) and `security_severity_threshold` (must be one of critical/high/medium/low)

### Tests
- **`spec/temporal/activities/scan_security_alerts_activity_spec.rb`** — 11 examples covering: missing project, disabled scanning, no alerts, open alerts, duplicate detection, dismissed alerts, API errors, severity filtering, alert type filtering
- **`spec/services/github_client_spec.rb`** — 4 examples for `dependabot_alerts`: parsing alerts, severity filtering, 404 handling, empty results

---

Closes #205

PR #439, opened 2026-03-24
fix(agent-runs): exclude .yarn-cache/ from agent commits

viamin:
## Summary

- Add `.yarn-cache/` to `CONTAINER_ARTIFACT_EXCLUDES` in `git_operations.rb` so `git add -A` inside agent containers skips the Yarn offline mirror cache
- Add `.yarn-cache/` to `.gitignore` as a host-side safety net
- Cleaned existing PR branches (#437, #433, #424) which each had ~3950 extra `.yarn-cache/v6/` files committed in their "Apply agent changes" commit

## Root cause

When agent containers run `yarn install`, Yarn populates `.yarn-cache/` in the workspace. The `commit_uncommitted_changes` method runs `git add -A` to capture any leftover agent edits, but `.yarn-cache/` was missing from the artifact exclusion list written to `.git/info/exclude`. This caused ~811K lines of npm package cache to be committed to each PR branch.

## Test plan

- [x] Verified all 3 cleaned PRs now show only their actual code changes (2, 15, and 10 files respectively)
- [x] Verified `.yarn-cache/` is not present on `main`
- [ ] Next agent run should not commit `.yarn-cache/` files

### DIFF
--- a/spec/services/containers/git_operations_spec.rb
+++ b/spec/services/containers/git_operations_spec.rb
@@ -832,7 +832,7 @@ RSpec.describe Containers::GitOperations do
       git_ops.install_artifact_excludes
     end
 
-    it "includes corepack and pg-install patterns" do
+    it "includes corepack, yarn-cache, and pg-install patterns" do
       script = nil
       allow(container_service).to receive(:execute) { |cmd, **|
         script = cmd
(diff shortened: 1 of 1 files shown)


viamin:
## Agent Update

Commit succeeded. The review feedback has been addressed:

- Added `expect(script).to include(".yarn-cache/")` to the existing `install_artifact_excludes` spec (`spec/services/containers/git_operations_spec.rb:845`)
- Updated the test description to reflect the new assertion
- RuboCop passes, pre-commit hooks passed, commit `66bc1aa` created

PR #498, opened 2026-03-26
Fix #487: fix(agent-runs): update Codex CLI invocation for current codex version

### LINKED ISSUE #487: fix(agent-runs): update Codex CLI invocation for current codex version
## Summary

Agent run `2513` failed its Codex fallback attempt before reaching the model because `paid` invoked Codex with flags that the installed Codex CLI no longer accepts.

## Evidence

`app/temporal/activities/run_agent_activity.rb` currently hardcodes:

```ruby
"codex" => %w[codex --quiet --full-auto --]
```

The run logs show:

```text
error: unexpected argument '--quiet' found
Usage: codex [OPTIONS] [PROMPT]
       codex [OPTIONS] <COMMAND> [ARGS]
```

The locally installed CLI currently reports non-interactive support via `codex exec`, and `codex --help` still documents `--full-auto` but not `--quiet`.

## Expected

Codex fallback should execute successfully in non-interactive mode, or at minimum use a command shape that is valid for the installed Codex version.

## Actual

The provider fails immediately with exit code 2 due to invalid CLI arguments, and the run falls through to other providers.

## Likely Fix

Update the Codex command template in `Activities::RunAgentActivity::AGENT_COMMANDS` to match the installed Codex CLI. We should also add a regression spec around the chosen command shape so future CLI upgrades fail loudly in tests.

## Scope

This looks like a `paid` bug, not an `agent-harness` bug, because the invalid command is assembled in `paid` before the CLI is launched.

viamin:
The background setup agent confirmed what we already found — PostgreSQL isn't reachable in this environment, but the fix is committed successfully. The commit passed lint (RuboCop, ESLint, markdownlint, ShellCheck) via the pre-commit hook.

---

Closes #487

### DIFF
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -16,7 +16,7 @@ module Activities
     AGENT_COMMANDS = {
       "claude_code" => %w[claude --print --output-format=text --dangerously-skip-permissions -p],
       "claude" => %w[claude --print --output-format=text --dangerously-skip-permissions -p],
-      "codex" => %w[codex exec --full-auto],
+      "codex" => %w[codex exec --full-auto --],
       "gemini" => %w[gemini -s],
       "kilocode" => %w[kilo run],
       "opencode" => %w[opencode --non-interactive --message],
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

Done. Both review comments have been addressed:

1. **`run_agent_activity.rb:19`** — Added `--` end-of-options marker to the Codex command template (`%w[codex exec --full-auto --]`), preventing prompts starting with `-`/`--` from being misinterpreted as flags.

2. **`run_agent_activity_spec.rb:42`** — Changed the spec from an exact array match to `start_with("codex", "exec", "--full-auto")` plus `include("--")`, making it resilient to future flag additions while still asserting the important parts.

Lint passed (0 offenses), tests passed (3227 examples, 0 failures), commit succeeded via pre-commit hook.

PR #501, opened 2026-03-26
fix(agent-runs): recover from timed out branch pushes

viamin:
## Summary
- recover new-branch pushes when GitHub already created the branch during a timed-out push
- verify the remote branch SHA matches local HEAD before treating the retry as successful
- add regression coverage for both the recovered and mismatched-remote cases

## Testing
- bundle exec rubocop app/services/containers/git_operations.rb spec/services/containers/git_operations_spec.rb
- bundle exec rspec spec/services/containers/git_operations_spec.rb (examples passed; process exits non-zero because single-file runs trip the repo-wide SimpleCov threshold)

### DIFF
--- a/spec/services/containers/git_operations_spec.rb
+++ b/spec/services/containers/git_operations_spec.rb
@@ -453,6 +453,7 @@ RSpec.describe Containers::GitOperations do
 
       expect(container_service).to receive(:execute)
         .with([ "git", "push", "--no-verify", "origin", "paid/test-branch" ], timeout: 60, stream: false)
+        .ordered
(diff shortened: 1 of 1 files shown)


viamin:
## Agent Update

Done. The review comment has been addressed:

**Fix**: Added `raise if e.is_a?(PushError)` in the `rescue Error => e` block of `recover_existing_new_branch_push` (line 441). This prevents the intentional `PushError` raised for a mismatched remote SHA from being caught and re-wrapped with a generic message, preserving the specific error context about which SHAs didn't match.

PR #557, opened 2026-03-27
fix(service-containers): add postgres health checks and metrics

viamin:
## Summary
- fix postgres service container startup by injecting default `POSTGRES_*` env when missing
- add Docker `HEALTHCHECK` support for postgres service containers using `pg_isready`
- add service-container resource metrics collection and summary fields for ongoing CPU/memory/PID visibility

## Root Cause
The crashing `postgres` container was a Paid-managed service container, not the long-lived compose/devcontainer database. We generated a default Postgres `DATABASE_URL` in app code, but did not pass matching default `POSTGRES_USER`, `POSTGRES_PASSWORD`, and `POSTGRES_DB` env vars to the container itself. On first boot, the Postgres image exited immediately because `POSTGRES_PASSWORD` was missing.

## Testing
- `bin/rails db:migrate:primary`
- `bin/rubocop app/models/service_container.rb app/models/service_container_metric.rb app/jobs/service_container_metrics_collection_job.rb app/services/containers/service_provisioner.rb app/services/containers/collect_service_metrics.rb spec/services/containers/service_provisioner_spec.rb spec/services/containers/collect_service_metrics_spec.rb spec/factories/service_container_metrics.rb db/migrate/20260327215947_add_resource_metrics_to_service_containers.rb db/migrate/20260327215948_create_service_container_metrics.rb`
- `bin/rspec spec/services/containers/service_provisioner_spec.rb spec/services/containers/collect_service_metrics_spec.rb` (examples passed; process exits non-zero because SimpleCov enforces the repo-wide 80% threshold on partial runs)

### DIFF
--- a/app/controllers/integration_credentials_controller.rb
+++ b/app/controllers/integration_credentials_controller.rb
@@ -1,0 +1,100 @@
+# frozen_string_literal: true
+
+class IntegrationCredentialsController < ApplicationController
+  before_action :set_filter_context, only: [ :index, :new, :create ]
+  before_action :set_integration_credential, only: [ :show, :destroy ]
+  skip_after_action :verify_authorized, only: :index
+
+  def index
+    @integration_credentials = filtered_scope.order(created_at: :desc)
+  end
+
+  def show
+    authorize @integration_credential
+  end
+
+  def new
+    @integration_credential = current_account.integration_credentials.build(default_credential_attributes)
+    authorize @integration_credential
+    load_form_options
+  end
+
+  def create
+    @integration_credential = current_account.integration_credentials.build(integration_credential_params)
+    @integration_credential.created_by = current_user
+    authorize @integration_credential
+    load_form_options
+
+    if @integration_credential.save
+      redirect_to integration_credential_path(@integration_credential, return_filter_params), notice: "Credential saved."
+    else
+      render :new, status: :unprocessable_content
+    end
+  end
+
+  def destroy
+    authorize @integration_credential
+    @integration_credential.revoke!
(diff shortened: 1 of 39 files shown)


viamin:
## Agent Update

Done. Here's what I changed to address the review comment:

**Review comment**: `container_env_for` merges defaults only when keys are missing, but blank/nil values from `env_json` would still override defaults.

**Fix** (2 locations in `service_provisioner.rb`):
1. **`container_env_for`** (line 303): Added normalization that strips whitespace and filters out nil/empty values before merging with `POSTGRES_DEFAULT_ENV`. This ensures blank values like `""`, `"  "`, or `nil` don't override the safe defaults.
2. **`ENV_MAPPINGS` postgres lambda** (lines 31-33): Changed `||` fallback to `.to_s.strip.presence ||` so `DATABASE_URL` generation also handles blank values correctly.

**Test added**: A new spec `"treats blank postgres env values as missing and applies defaults"` verifies both the Docker env and the generated DATABASE_URL when postgres env vars are empty/blank/nil.

viamin:
## Agent Update

All 3820 examples pass, 0 failures. Commit succeeded!

The commit `155f4e6` addresses both review comments:

1. **Thread at line 58**: Replaced `allow(ServiceContainerMetricsCollectionJob).to receive(:perform_later)` + `have_received(:perform_later)` with ActiveJob's `have_been_enqueued` matcher, which asserts behavior (job was enqueued) rather than implementation (specific method was called).

2. **Thread at line 461**: Added an assertion that the concurrency error handler logged the expected `"service_provisioner.metrics_job_already_enqueued"` message with the correct `service_container_id`, verifying the error-handling path was actually reached rather than just checking no exception escaped.

PR #560, opened 2026-03-28
Fix #547: Dashboard metrics should count fallback provider usage for agent runs

### LINKED ISSUE #547: Dashboard metrics should count fallback provider usage for agent runs
## Problem

Currently, dashboard metrics for providers do not account for the fallback mechanism. When an agent run is started with one provider (e.g., Claude) but falls back to another provider (e.g., Codex), the run is still counted under the original provider. This misrepresents which provider actually produced the output and skews quality metrics.

## Expected Behavior

- If an agent run starts with Provider A but falls back to Provider B, the run should be **counted as a Provider B agent run** in dashboard metrics.
- **Quality metrics** (success rate, output quality, etc.) should be **assigned to the provider that actually produced the output**, not the one that was initially requested.
- The originally requested provider can still be tracked separately (e.g., as `requested_provider`), but the primary metric attribution should reflect the provider that generated the actual output.

## Example

An agent run is initiated with Claude as the provider. Through the fallback mechanism, the run actually executes with Codex.

- **Current behavior:** The run is counted as a Claude agent run; quality metrics are attributed to Claude.
- **Desired behavior:** The run is counted as a Codex agent run; quality metrics are attributed to Codex. Optionally, track that Claude was the originally requested provider.

## Suggested Approach

1. Track both `requested_provider` and `actual_provider` (or `effective_provider`) for each agent run.
2. Use `actual_provider` as the primary key for dashboard metrics and quality reporting.
3. Optionally surface `requested_provider` in a secondary view to understand fallback frequency per provider.

viamin:

**`app/models/agent_run.rb`** - Added `effective_provider` method that returns `final_provider` (the provider that actually ran) when present, falling back to `agent_type` (the originally requested provider).

**`app/services/dashboard/stats.rb`** - Added two new metrics:
- `runs_by_provider` - Groups runs by effective provider using `COALESCE(NULLIF(final_provider, ''), agent_type)`, so fallback runs are attributed to the provider that produced the output
- `provider_fallback_stats` - Shows total fallback count, fallback rate, and breakdowns by both requested and effective provider

**`app/views/dashboard/_content.html.erb`** - Replaced "Runs by Agent Type" with "Runs by Provider" showing effective provider counts, with a fallback rate summary line when fallbacks have occurred.

**Specs** - Added tests for `effective_provider` on the model and comprehensive tests for the new provider fallback metrics in the dashboard stats service.

---

Closes #547

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -81,6 +81,13 @@ class AgentRun < ApplicationRecord
     end
   }
 
+  # SQL expression for the effective provider: the provider that actually
+  # produced the output. Mirrors the Ruby #effective_provider method so that
+  # both SQL aggregations and Ruby code share the same logic.
+  def self.effective_provider_sql
+    "COALESCE(NULLIF(final_provider, ''), agent_type)"
+  end
+
   ransacker :tokens_total, type: :integer do
     Arel.sql("COALESCE(tokens_input, 0) + COALESCE(tokens_output, 0)")
   end
--- a/app/services/dashboard/stats.rb
+++ b/app/services/dashboard/stats.rb
@@ -4,7 +4,7 @@ module Dashboard
   class Stats
     PHASE_BREAKDOWN_WINDOW = 30.days
     PHASE_BREAKDOWN_RUN_LIMIT = 500
-    EFFECTIVE_PROVIDER_SQL = "COALESCE(NULLIF(final_provider, ''), agent_type)"
+    EFFECTIVE_PROVIDER_SQL = AgentRun.effective_provider_sql
 
     attr_reader :account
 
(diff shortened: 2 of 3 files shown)


viamin:
## Agent Update

All changes committed. Here's a summary of what was addressed:

**Review comment 1 — Broaden fallback condition** (`stats.rb:167`): Changed `fallback_runs` query from only `provider_switches > 0` to also include runs where `final_provider` differs from `agent_type`. This catches the case where the primary provider is skipped as unavailable/rate-limited without `log_provider_switch!` being called.

**Review comment 2 — Centralize effective provider SQL** (`stats.rb:7`): Added `AgentRun.effective_provider_sql` class method as the single source of truth. `Dashboard::Stats::EFFECTIVE_PROVIDER_SQL` now delegates to it, eliminating the duplicated SQL string.

**Review comment 3 — Add missing spec** (`stats_spec.rb:239`): Added a new context `"with provider fallback via skipped primary (no provider_switches)"` that creates a run with `final_provider: "cursor"`, `agent_type: "claude_code"`, and `provider_switches: 0`, verifying it's correctly counted as a fallback.

PR #569, opened 2026-03-28
fix(prompts): align service guidance with configured containers

viamin:
## Summary
- treat project-configured service containers as available run-time infrastructure in prompt guidance
- append service-environment guidance to prompt-version issue runs so they do not bypass DATABASE_URL instructions
- add coverage for configured containers and section-separation behavior

## Verification
- `bundle exec rspec spec/temporal/activities/create_agent_run_activity_spec.rb spec/services/prompts/build_for_issue_spec.rb spec/services/prompts/build_for_pr_spec.rb`
  - examples pass; local command exits non-zero because this repo enforces a global SimpleCov minimum on partial spec runs
- `ruby -c app/services/prompts/service_container_sections.rb`
- `ruby -c app/services/prompts/build_for_issue.rb`
- `ruby -c app/services/prompts/build_for_pr.rb`
- `ruby -c app/temporal/activities/create_agent_run_activity.rb`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,10 +2,157 @@
 # yarn lockfile v1
 
 
+"@emnapi/core@^1.7.1", "@emnapi/core@^1.8.1":
+  version "1.9.1"
+  resolved "https://registry.yarnpkg.com/@emnapi/core/-/core-1.9.1.tgz#2143069c744ca2442074f8078462e51edd63c7bd"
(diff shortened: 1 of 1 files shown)


viamin:
## Agent Update

The background setup agent confirmed what we already found — PostgreSQL isn't reachable in this environment. The commit is already done successfully with all changes addressing both code review comments. No further action needed.

PR #604, opened 2026-03-29
Fix #593: Improve Temporal recovery for interrupted agent runs during full dev updates

### LINKED ISSUE #593: Improve Temporal recovery for interrupted agent runs during full dev updates
## Summary

Full `bin/dev-update --full` restarts now preserve recent runs for a short grace period, but Paid still loses in-flight agent work more often than it should during worker restarts.

We should make the Temporal side of agent execution more interruption-friendly so full restarts and routine worker shutdowns can recover long-running runs instead of timing them out.

## Proposed enhancements

- Add a non-zero Temporal worker graceful shutdown period in `bin/temporal_worker` so shutdown waits briefly for in-flight work before forcing cancellation.
- Add activity heartbeats to `Activities::RunAgentActivity` during long-running agent execution.
- Configure a `heartbeat_timeout` for `RunAgentActivity` in `AgentExecutionWorkflow` so dead workers are detected quickly without waiting for the full `start_to_close_timeout`.
- Preserve recent active runs for Temporal recovery on startup instead of immediately timing them out in `dev:cleanup`.
- Decide whether `RunAgentActivity` should remain monolithic or be split into smaller recoverable steps/checkpoints.

## Why

Temporal is designed to recover from worker interruptions, but today Paid is limiting that recovery by:

- running `RunAgentActivity` with `NO_RETRY`
- not heartbeating long-running agent execution
- immediately timing out active runs during startup cleanup
- shutting down the worker without an explicit graceful shutdown period

These changes would make full dev updates and other controlled restarts much less likely to discard useful in-flight work.

viamin:
Based on the issue context and agent output, here's the PR description:

## Summary

Make Temporal agent execution resilient to worker restarts and full dev updates so in-flight agent runs can recover instead of being discarded. Today, the combination of no heartbeats, no graceful shutdown period, and aggressive startup cleanup means Temporal can't use its built-in recovery mechanisms, causing unnecessary work loss during routine restarts.

### Worker lifecycle (`bin/temporal_worker`)
- Added a 30-second graceful shutdown period (configurable via `TEMPORAL_GRACEFUL_SHUTDOWN_PERIOD`) so SIGINT/SIGTERM waits for in-flight activities before forcing cancellation

### Activity heartbeats (`app/temporal/activities/`)
- Added `heartbeat` helper to `BaseActivity` wrapping the Temporal SDK heartbeat API with a no-op fallback in test environments
- `RunAgentActivity` now heartbeats at two points: before each provider attempt and after successful provider completion

### Workflow configuration (`app/temporal/workflows/agent_execution_workflow.rb`)
- Configured `heartbeat_timeout: 120` on `RunAgentActivity` so dead workers are detected within 2 minutes instead of waiting for the full `start_to_close_timeout`

### Startup cleanup (`lib/tasks/dev.rake`)
- Changed `dev:cleanup` default grace period from 0 to 300 seconds, preserving recent active runs for Temporal workflow replay on startup

## Design Decisions

- **Heartbeat timeout of 120s** — balances fast dead-worker detection against normal execution gaps where the agent may be working without hitting a heartbeat point. Two minutes is well under the full execution timeout but avoids false positives during legitimate long provider calls.
- **Grace period of 300s** — five minutes gives Temporal enough time to detect the dead worker via missed heartbeats and reschedule the activity to the new worker after restart, without keeping stale runs around indefinitely.
- **Heartbeat placement** — heartbeats are placed at provider attempt boundaries rather than on a timer, keeping the implementation simple and aligned with the natural checkpoints in execution. This avoids threading/async complexity.
- **`RunAgentActivity` remains monolithic** — rather than splitting into smaller activities (which would require significant refactoring), heartbeats provide the key benefit of dead-worker detection while keeping the existing structure intact. Decomposition can be revisited later if retry-at-checkpoint semantics are needed.

---

Closes #593

### DIFF
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -437,7 +437,29 @@ module Activities
       context = Temporalio::Activity::Context.current_or_nil
       return yield unless context
 
-      worker = Thread.new { yield }
+      # Wrap the worker thread in Rails executor and ActiveRecord connection
+      # pool management. The executor handles autoloading/reloading and the
+      # with_connection block ensures the DB connection is checked out only
+      # for the duration of the work and returned to the pool afterward,
+      # preventing connection-pool exhaustion from long-running activities.
+      worker = Thread.new do
+        executor = Rails.application.executor if defined?(Rails) && Rails.respond_to?(:application) && Rails.application.respond_to?(:executor)
+        work = proc { yield }
+
+        db_scoped = proc do
+          if defined?(ActiveRecord::Base) && ActiveRecord::Base.respond_to?(:connection_pool)
+            ActiveRecord::Base.connection_pool.with_connection { work.call }
+          else
+            work.call
+          end
+        end
+
+        if executor
+          executor.wrap(&db_scoped)
+        else
+          db_scoped.call
+        end
+      end
 
       canceled = false
       begin


viamin:
## Agent Update

That was the earlier failed commit attempt — it's already been superseded. The commit succeeded on the later attempt (`3afed7a1`). No action needed.

PR #612, opened 2026-03-29
Fix #609: Retain unknown failed agent-run containers briefly after successful agent execution

### LINKED ISSUE #609: Retain unknown failed agent-run containers briefly after successful agent execution
## Summary
When an agent has already completed successfully but the run later fails for an unknown reason, Paid should temporarily retain the agent container/workspace instead of cleaning it up immediately.

## Why
Today we always clean up containers in the workflow `ensure` block. That makes post-failure diagnosis harder and can destroy recoverable work when the agent already produced useful changes but later failed during push/PR creation or another unexpected post-agent step.

## Proposed behavior
- Only skip immediate cleanup when both conditions are true:
  - the agent step completed successfully
  - the terminal error is not a known/expected outcome
- Keep the allowlist for retained failures intentionally small.
- Continue cleaning up immediately for known outcomes such as:
  - all providers rate limited
  - all providers exhausted / expected provider failures
  - auth expired
  - explicit cancellation
- Retained containers/workspaces should be auto-cleaned after a short TTL.
  - Default target: 4 hours
  - Shorten the TTL when system pressure is high, such as low disk space or other resource constraints

## Notes
- The retention window is for diagnostics and possible work recovery only, not permanent storage.
- UI/metadata should make retained state visible so operators know a failed run still has live resources attached.
- Janitor cleanup should remain the safety net and enforce TTL expiry.

viamin:
Clean working tree. Here's a summary of the changes:

### Migration
- **`db/migrate/20260329170336_add_container_retained_until_to_agent_runs.rb`** — Adds `container_retained_until` datetime column to `agent_runs` for tracking when the retention TTL expires.

### Model
- **`app/models/agent_run.rb`** — Added `container_retained?` method that returns true when the retention TTL is set and hasn't expired yet.

### New Activity
- **`app/temporal/activities/retain_container_activity.rb`** — Sets `container_retained_until` on the agent run. Default TTL is 4 hours, shortened to 1 hour when disk usage exceeds 85%.

### Workflow
- **`app/temporal/workflows/agent_execution_workflow.rb`** — Modified the ensure block to conditionally skip container cleanup when:
  1. The agent step completed successfully (`agent_step_succeeded`)
  2. The workflow failed (`workflow_error` is set)
  3. The failure is NOT a known/expected type (rate limit, cancellation, etc.)
  
  When retention applies, `RetainContainerActivity` runs instead of `CleanupContainerActivity`. If retention fails, it falls back to normal cleanup. Services and worktree cleanup always run.

### Cleanup Jobs
- **`app/jobs/agent_run_resource_janitor_job.rb`** — Skips cleanup when `container_retained?` returns true.
- **`app/jobs/docker_orphan_cleanup_job.rb`** — Skips containers and volumes with unexpired retention TTLs during orphan cleanup.

### Tests
- 18 new test examples across all spec files, covering retention activation, TTL expiry, known failure exclusion, fallback behavior, and janitor/orphan cleanup respecting retention.

---

Closes #609

### DIFF
--- a/app/temporal/workflows/agent_execution_workflow.rb
+++ b/app/temporal/workflows/agent_execution_workflow.rb
@@ -422,10 +422,14 @@ module Workflows
         return false if current_error.is_a?(Temporalio::Error::CanceledError)
 
         if current_error.is_a?(Temporalio::Error::ApplicationError)
+          # For ApplicationError, the underlying exception type is carried in
+          # `type`, so compare both known failure *types* and *classes* against it.
           return false if KNOWN_FAILURE_TYPES.include?(current_error.type)
+          return false if KNOWN_FAILURE_CLASSES.include?(current_error.type)
+        else
+          # For non-ApplicationError exceptions, compare against the Ruby class name.
+          return false if KNOWN_FAILURE_CLASSES.include?(current_error.class.name)
         end
-
-        return false if KNOWN_FAILURE_CLASSES.include?(current_error.class.name)
 
         break unless current_error.respond_to?(:cause)
         current_error = current_error.cause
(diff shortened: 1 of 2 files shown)


viamin:

Issue #413, opened 2026-03-22
Phase 2: Intelligence — remaining work tracker


viamin:
## Phase 2: Intelligence — Remaining Work

**Goal**: The system learns from its performance and makes intelligent decisions about models and prompts.

Replaces #148 (closed with PR #335). This issue tracks the remaining Phase 2 work.

### Success Metrics

- Model selection improves cost efficiency by 20%
- A/B tests identify winning prompts
- Quality metrics correlate with human feedback
- Projects indexed on creation, semantic search < 500ms
- Dashboard latency < 1 second

---

### 2.1 Prompt Versioning System — COMPLETE

- [x] #136 — Create Prompt and PromptVersion models with versioning
- [x] #137 — Add prompt management UI

### 2.2 Style Guide Management — Partial

- [x] #138 — Create StyleGuide model with LLM-friendly compression
- [ ] #410 — Implement automatic style guide extraction from codebase
- [ ] #411 — Add tree-sitter integration for structural code analysis

## Dependencies

- #410
- #411

### 2.3 Model Registry & Meta-Agent — Partial

- [x] #139 — Integrate ruby-llm model registry and create Model table
- [ ] #140 — Implement meta-agent for intelligent model selection (rules-based done, LLM meta-agent remaining)

Depends on #140

### 2.4 Cost Tracking — Partial

- [x] #141 — Create TokenUsage model and cost tracking infrastructure
- [ ] #142 — Add cost tracking dashboard to UI

Depends on #142

### 2.5 A/B Testing Framework — COMPLETE

- [x] #143 — Implement A/B testing framework for prompts
- [x] #144 — Add A/B testing management UI

### 2.6 Quality Metrics & Feedback — Partial

- [x] #145 — Implement quality metrics collection and QualityMetric model
- [x] #146 — Add quality metrics dashboard to UI
- [ ] #412 — Add human feedback collection via GitHub reactions and webhooks

Depends on #412

### 2.7 Semantic Code Search — Not Started

Core (required for Phase 2 completion):
- [ ] #251 — PostgreSQL Schema — Knowledge Objects, Versioning, Provenance
- [ ] #252 — Qdrant Collection Management — Create, Upsert, Delete, Rebuild
- [ ] #253 — Collector Framework — Orchestration, Storage, Idempotency, Staleness
- [ ] #254 — Thin Vertical Slice — Routes Collector → Postgres → Qdrant → Query API
[...]

Issue #269, opened 2026-03-02
Knowledge Base Safe Defaults and Security Hardening


viamin:
## Background / Why

The knowledge base handles code content that may contain secrets, and communicates with Qdrant (a network service). This issue ensures safe defaults are in place: Qdrant API key auth in production, minimal Qdrant payloads, redaction pipeline is mandatory (not optional), and access control on knowledge APIs.

## Dependencies

- #257
- #262

## Scope

**In scope:**
- Qdrant API key authentication in non-development environments
- Validate that Qdrant payloads never contain \`content\` text (enforce at \`PointSync\` level)
- Ensure redaction pipeline cannot be bypassed (embedding pipeline refuses unscanned chunks)
- Access control on knowledge APIs (require authenticated user + project access via Pundit)
- Rate limiting on knowledge search API
- Secrets: \`QDRANT_API_KEY\` in Rails credentials, not ENV in production

**Out of scope:**
- Qdrant TLS (depends on deployment topology)
- Network-level isolation beyond existing Docker networks

## Proposed Design

**Payload validation in PointSync:**
```ruby
FORBIDDEN_PAYLOAD_KEYS = %w[content text body secret password token].freeze

def validate_payload!(payload)
  forbidden = payload.keys.map(&:to_s) & FORBIDDEN_PAYLOAD_KEYS
  raise SecurityError, "Forbidden payload keys: #{forbidden}" if forbidden.any?
end
```

**Mandatory redaction:**
```ruby
# In Knowledge::Embeddings::Pipeline
def process_chunk(chunk)
  raise "Chunk not scanned for redaction" unless chunk.redaction_scanned?
  # ...
end
```

**Pundit policy:**
```ruby
class KnowledgeSearchPolicy < ApplicationPolicy
  def search?
    user.has_role?(:member, record.project) || user.has_role?(:admin, record.project)
  end
end
```

## Implementation Tasks

- [ ] Add \`QDRANT_API_KEY\` to Rails credentials for production
- [ ] Enforce API key in \`QdrantClient\` when \`Rails.env.production?\`
- [ ] Add payload validation in \`PointSync.upsert_chunk!\`
- [ ] Add \`redaction_scanned_at\` column to \`knowledge_chunks\`
- [ ] Enforce redaction scan in embedding pipeline
- [ ] Create \`KnowledgeSearchPolicy\` Pundit policy
- [ ] Add rate limiting to \`Api::KnowledgeSearchController\` (e.g., 60 req/min)
[...]

Issue #262, opened 2026-03-02
Redaction Pipeline — Sensitive Data Handling Before Embedding


viamin:
## Background / Why

Code and configuration often contain secrets (API keys, passwords, tokens), PII (emails, names), and internal URLs. These must never be embedded as vectors or stored in Qdrant payloads. A redaction pipeline runs before chunking/embedding to detect and mask sensitive content.

**Depends on:** #251 (PG Schema)

## Scope

**In scope:**
- \`Knowledge::Redaction::Scanner\` — detects sensitive patterns in text
- \`Knowledge::Redaction::Redactor\` — replaces detected patterns with placeholders
- Pattern library: API keys, passwords, emails, IP addresses, JWT tokens, AWS keys, GitHub tokens, connection strings
- Integration point in the embedding pipeline (#257)
- Redaction audit log — record what was redacted (pattern type, file, line) without storing the secret
- \`KnowledgeChunk\` status \`"redacted"\` for chunks that are entirely sensitive

**Out of scope:**
- ML-based PII detection (use regex patterns first)
- Redaction of Qdrant payloads (payloads should never contain sensitive content — enforce this at upsert time)

## Proposed Design

**Scanner patterns:**
```ruby
PATTERNS = {
  api_key: /(?:api[_-]?key|apikey)\s*[:=]\s*["']?([a-zA-Z0-9_\-]{20,})["']?/i,
  aws_key: /(?:AKIA|ABIA|ACCA|ASIA)[0-9A-Z]{16}/,
  github_token: /gh[ps]_[a-zA-Z0-9]{36,}/,
  jwt: /eyJ[a-zA-Z0-9_-]+\.eyJ[a-zA-Z0-9_-]+\.[a-zA-Z0-9_-]+/,
  password: /(?:password|passwd|secret)\s*[:=]\s*["']([^"']+)["']/i,
  email: /[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}/,
  connection_string: /(?:postgres|mysql|redis|mongodb):\/\/[^\s"']+/i,
  private_key: /-----BEGIN (?:RSA |EC |DSA )?PRIVATE KEY-----/
}.freeze
```

**Redaction output:**
```ruby
result = Knowledge::Redaction::Redactor.call(text: chunk_content)
# result.clean_text    → text with [REDACTED:api_key], [REDACTED:email], etc.
[...]

PR #630, opened 2026-03-30
Fix #615: Prevent auto-pick from selecting tracker issues with incomplete or transitive blockers

### LINKED ISSUE #615: Prevent auto-pick from selecting tracker issues with incomplete or transitive blockers
## Summary
On 2026-03-29, automation opened PR #608 for issue #413 even though issue #262 and issue #269 were still open. That should not have happened.

## What I investigated
I checked the current GitHub issue data for #413, #262, #269, and the PRs that were opened around the same time.

Relevant facts:
- PR #608 (`Fix #413`) was opened at 2026-03-29T15:42:05Z.
- PR #611 (`Fix #269`) was only opened later at 2026-03-29T16:24:15Z, so #269 was still open when #608 was created.
- #262 was also still open at that time.
- GitHub's native dependency fields are empty for all of these issues (`issue_dependencies_summary.blocked_by == 0`), so any dependency enforcement is coming from Paid's own issue parsing / derived dependency tree, not from GitHub's built-in dependency model.
- In the body of #413, the explicit dependency declarations only cover issues like #410, #411, #140, #142, #412, #147, and the core semantic-search issues (#251, #252, #253, #254, #257, #258, #259, #260, #266, #268).
- #262 and #269 are listed under "Supporting (can be deferred to Phase 3)" rather than in an explicit `Depends on ...` section for #413.
- PR #608 did not actually implement the whole tracker. Its body only describes work for #412, but it still says `Closes #413`.

## Diagnosis
This looks like a compound root cause:

1. The dependency tree for #413 was incomplete / incorrectly modeled.
`#262` and `#269` were not represented as blockers in the data that auto-pick appears to consume, even though they were treated operationally as required before shipping #413.

2. Auto-pick treated a tracker/meta issue as a normal implementable issue.
`#413` is clearly a phase tracker: the title says "remaining work tracker", the body contains multiple sections, many child issues, and phase completion criteria. Even with a perfect dependency parser, auto-pick should not open a single "Fix #413" PR for that kind of issue.

3. There was no final guard verifying that a proposed issue is actually independently implementable and fully satisfiable.
PR #608 only covered a subset of #413, but automation still proceeded as if the tracker itself were ready to implement and close.

## Why this is the most likely root cause
The available evidence does **not** point to auto-pick ignoring a correctly declared dependency. The issue body for #413 does not explicitly declare `#262` or `#269` as blockers, and GitHub native dependency metadata is empty. That means the stronger failure is upstream of selection: the tracker/dependency representation was insufficient, and the picker lacked a tracker-specific safety check.

## Required fix
Auto-pick should be hardened in two places:

1. **Do not auto-pick tracker/meta issues.**
Treat issues as non-pickable when they look like aggregation work, for example:
- title/body contains terms like `tracker`, `remaining work`, `phase`, `epic`, `meta`, or `completion criteria`
- issue body contains multiple child issue references or checklist sections spanning distinct workstreams
- issue scope is "close all remaining items" rather than a bounded implementation task

2. **Compute blockers from the full issue graph, not only explicit `Depends on` lines.**
At minimum, the picker should respect all parsed dependency sections consistently. If tracker issues are ever evaluated at all, they must be blocked by any still-open descendant issue required for completion, including transitive security/safety work like `#269 -> #262`.

3. **Add a pre-pick satisfiability check.**
Before opening a PR, verify that the candidate issue is:
- a leaf or otherwise independently implementable
- not a tracker/rollup
- not partially satisfied by one child issue while other open referenced issues remain

4. **Add a pre-close guard for automation-generated PRs.**
If a PR body only implements one sub-issue from a tracker, it must not say `Closes #413`. It should either close the actual leaf issue or omit automatic closure text for the tracker.

## Acceptance criteria
- Auto-pick never selects tracker/meta issues like #413 for implementation.
- Open blockers that are part of a tracker's required completion graph prevent selection.
- Automation cannot open a PR that claims to close a tracker while only implementing one child issue.
- Regression coverage includes a case structurally similar to #413 / #262 / #269 / #608.

## Reproduction case to encode as a regression test
Create a tracker issue with:
- multiple child issues
- a mix of explicit `Depends on` sections and referenced child issues in grouped sections
- at least one open security-related descendant issue

Verify that auto-pick refuses the tracker and instead selects only eligible leaf issues whose blockers are all closed.

viamin:

**Problem**: Auto-pick selected tracker issue #413 for implementation despite open child issues #262 and #269, because those issues weren't in explicit `IssueDependency` records.

**Solution** (respecting collaborator feedback that trackers should still be pickable when all deps are met):

1. **`Issue#tracker_issue?`** — Detects tracker/meta issues by matching keywords (`tracker`, `remaining work`, `completion criteria`, `phase tracker`, `meta issue`) in title or body.

2. **`Issue#body_referenced_issue_numbers`** — Parses `#NNN` references from the issue body.

3. **`AutoPick.tracker_ids_blocked_by_open_references`** — Pre-filters potential trackers via SQL ILIKE, then checks if any body-referenced issues are still open. Blocked tracker IDs are excluded from the eligible scope.

**Key behavior**: Tracker issues are still pickable once ALL referenced issues in their body are closed. Non-tracker issues are unaffected by body references.

**Tests added**: 8 new auto_pick specs (tracker blocked, tracker with all deps closed, no refs, non-tracker unaffected, self-reference, body-only detection, regression scenario, eligible_issue_ids exclusion) + 6 Issue model specs for the new methods.

---

Closes #615

### DIFF
--- a/app/services/issues/auto_pick.rb
+++ b/app/services/issues/auto_pick.rb
@@ -68,8 +68,9 @@ module Issues
 
       # Exclude tracker/meta issues that still have open referenced issues.
       # Trackers are pickable only once every issue referenced in their body
-      # is closed (per collaborator feedback).
-      blocked_ids = tracker_ids_blocked_by_open_references(project)
+      # is closed (per collaborator feedback). The ILIKE scan runs against
+      # +scope+ (not all open issues) to limit query cost.
+      blocked_ids = tracker_ids_blocked_by_open_references(scope, project)
       scope = scope.where.not(id: blocked_ids) if blocked_ids.present?
 
       scope
@@ -81,13 +82,24 @@ module Issues
     # Only queries open/closed state for issue numbers actually referenced
     # by tracker candidates (not all project issues).
     #
-    # Performance: the ILIKE scan runs against open, non-PR issues for a
-    # single project — typically a small set. If this becomes expensive on
-    # repos with thousands of open issues, consider a trigram GIN index on
-    # (title, body) or a persisted `tracker_issue` boolean column.
-    def self.tracker_ids_blocked_by_open_references(project)
-      open_issues = Issue.where(project: project, github_state: "open", is_pull_request: false)
-
+    # +candidate_scope+ is the already-filtered eligible-issue scope so the
+    # ILIKE scan runs only against issues that passed earlier filters (labels,
+    # paid_state, dependencies, etc.) rather than all open project issues.
+    # If this still becomes expensive on repos with thousands of eligible
+    # issues, consider a trigram GIN index on (title, body) or a persisted
+    # `tracker_issue` boolean column.
+    #
+    # Blocking policy:
+    # - Trackers with body references are blocked when ANY reference is open
+    #   or unknown (not yet synced). Only direct references are checked — not
+    #   transitive dependencies of those references. Transitive checking is
+    #   deferred because the IssueDependency graph may be incomplete for
+    #   body-referenced issues, and the direct-reference check already
+    #   catches the motivating scenario (#615).
+    # - Trackers with NO body references are conservatively blocked — they
+    #   likely track work not enumerated as #NNN references, and auto-picking
+    #   them risks premature selection (see #615).
+    def self.tracker_ids_blocked_by_open_references(candidate_scope, project)
       ilike_conditions = TRACKER_SQL_PATTERNS.each_with_index.flat_map do |_, i|
         [ "title ILIKE :t#{i}", "body ILIKE :t#{i}" ]
       end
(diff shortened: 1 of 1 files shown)


viamin:
## Agent Update

That was the earlier failed commit attempt (before fixing the pre-commit hook). The successful commit was already made — `e9b80e66` on the branch. Nothing to do here.

viamin:
## Agent Update

1. **SQL condition chaining** (line 91): Changed `flat_map.with_index` to `each_with_index.flat_map` to ensure the ILIKE conditions are a flat array of strings, not nested arrays.

2. **No-refs tracker behavior** (line 116): The existing behavior of blocking trackers with zero body references is intentional (comment at lines 103-105 explains why). No code change needed — the behavior is correct and documented.

3. **Spec: tracker with open refs** (line 186): Gave the tracker `github_number: 100` (lowest), so the test would fail without the blocking logic. Added assertion that one of the referenced issues is picked instead.

4. **Spec: tracker with no refs** (line 212): Set explicit `github_number: 1` on tracker and `2` on normal_issue, so the tracker would be picked first without the blocking logic.

5. **Spec: body-based detection** (line 261): Gave tracker `github_number: 1` (lowest) and asserted the specific issue that should be picked, so the test depends on the tracker-blocking logic.

PR #651, opened 2026-03-31
Fix #637: fix(providers): redesign API key compatibility model to separate CLI tools from API services

### LINKED ISSUE #637: fix(providers): redesign API key compatibility model to separate CLI tools from API services
## Problem

The `ProviderApiKey.compatible_providers` field conflates two distinct concepts:

1. **CLI agent tools** that Paid orchestrates: `claude`, `cursor`, `opencode`, `aider`, `kilocode`, etc.
2. **API services** those tools call upstream: OpenRouter, OpenAI, Anthropic, etc.

For most providers the tool name and API service happen to match (the `claude` CLI uses the Anthropic API), so the conflation is hidden. But **OpenCode breaks this** — it's a CLI tool that talks to OpenRouter (an OpenAI-compatible aggregator). The current workaround is `required_api_key_targets_for("opencode")` returning `["openrouter"]`, which causes confusion:

### Symptoms

- Users see "OpenRouter" alongside CLI tools like "Aider" and "Kilocode" in the compatible providers checkboxes — it's not obvious what to select for OpenCode
- An existing API key in the DB has `compatible_providers: ["aider", "opencode", "kilocode"]` — but `"opencode"` is not in the allowed targets list, so **the record is now invalid** and can't be re-saved
- The error message "Provider api key is not compatible with OpenRouter" when adding an OpenCode provider is confusing — the user has an OpenRouter API key and is adding an OpenCode provider, not an OpenRouter provider
- `ProviderApiKey.compatibility_target_labels` mixes tool names (`claude`, `cursor`) with service names (`openrouter`) in a flat list

### Root Cause

`compatible_providers` stores what are really **API service targets**, but they're labeled and presented as if they're CLI tools. The mapping is implicit in `required_api_key_targets_for()` and there's no way to express "this is an OpenAI-compatible key" that would automatically work with any tool that uses that API format.

## Proposed Direction

Separate the concepts:

- **API keys have an API service type** (e.g., `openrouter`, `openai`, `anthropic`, `google`) describing what upstream API the key authenticates against
- **Providers (CLI tools) declare what API service type they need** — most need their own (`claude` → `anthropic`), but `opencode` → `openrouter`, and potentially others could share types
- The compatibility check becomes: does the API key's service type match what the provider needs?

This would:
- Make the UI clearer (select "OpenRouter" as the key type, not as a "compatible provider")
- Naturally support OpenAI-compatible services
- Eliminate the confusing flat list mixing tools and services
- Make `required_api_key_targets_for` unnecessary — providers just declare their API service dependency

## Current Code References

- `ProviderApiKey.compatible_providers` field — `app/models/provider_api_key.rb`
- `Provider.required_api_key_targets_for` — `app/models/provider.rb:229`
- `OPENCODE_API_PROVIDER_KEYS` / `OPENCODE_DEFAULT_API_PROVIDER` — `app/models/provider.rb:10-11`
- `api_key_must_be_compatible` validation — `app/models/provider.rb:329`
- `ProviderApiKey.compatibility_target_labels` — `app/models/provider_api_key.rb:29`
- Provider form API key filtering — `app/views/providers/_form.html.erb`
- API key creation form — `app/views/provider_api_keys/new.html.erb`

viamin:

### Core Design
Replaced `ProviderApiKey.compatible_providers` (JSONB array mixing CLI tool names and API service names) with `api_service_type` (string) that cleanly identifies the upstream API service the key authenticates against.

### Files Changed (26 files)

**Database:**
- New migration to add `api_service_type`, migrate existing data, and remove `compatible_providers`

**Models:**
- `ProviderApiKey` — replaced `compatible_providers` array with single `api_service_type` string; `compatible_with?` now checks against the provider's required service type; added `display_api_service_type` and `api_service_type_options`
- `Provider` — replaced `required_api_key_targets_for` and `compatibility_label_for` with `api_service_type_for`; updated `api_key_must_be_compatible` validation

**Library:**
- `ProviderSupport` — added `API_SERVICE_TYPES` (anthropic, openai, openrouter, google) and `PROVIDER_API_SERVICE_TYPE` mapping each CLI tool to its required API service

**Controllers:**
- `ProviderApiKeysController` — simplified params (single field vs array), uses dropdown options
- `ProvidersController` — simplified compatibility check

**Views:**
- API key form: replaced checkboxes with a single API service dropdown
- Index/show: display service type label instead of provider list
- Provider form: uses `data-api-service-type` attribute for filtering

**JavaScript:**
- `provider_form_controller.js` — maps provider keys to service types client-side for API key option filtering

**Tests:**
- Updated all specs and factories to use `api_service_type` instead of `compatible_providers`
- Added tests for new `ProviderSupport` methods and `ProviderApiKey.api_service_type_options`

---

Closes #637


viamin:
## Agent Update

## Changes Made

### 1. Review Comment: Copilot mapping in migration (the unresolved thread)
- **Removed** the explicit `WHEN compatible_providers ? 'copilot' THEN 'anthropic'` line from the data migration CASE statement
- **Added** a `/* */` block comment explaining why copilot is intentionally excluded (subscription-only auth, no API key needed)
- Copilot-only legacy keys (which shouldn't exist in practice) fall through to the `ELSE 'anthropic'` default but remain inert since `api_service_type_for('copilot')` returns `nil`
- **Fixed** SQL comment style from `--` to `/* */` since the `squish` method collapses newlines, which would turn `--` comments into a syntax error that swallows the rest of the SQL

### 2. Missing test coverage
- **Added** test in `spec/models/provider_spec.rb` verifying that providers with no API service type (like copilot) correctly reject API key auth with the message "is not supported for this provider; use subscription authentication instead"

### 3. Fixed schema.rb
- **Restored** full `schema.rb` from main branch with only the migration-specific changes (version bump, column change, composite index), fixing previously truncated foreign key declarations

### Verification
- 4708 tests pass, 0 failures
- RuboCop: 638 files, no offenses
- Pre-commit hooks passed (lint + full test suite)

PR #658, opened 2026-03-31
Fix #653: fix(ui): Auto-pick lightning bolt icon not restored after failed agent run

### LINKED ISSUE #653: fix(ui): Auto-pick lightning bolt icon not restored after failed agent run
## Problem

The auto-pick eligible lightning bolt icon disappears when auto-pick creates a queued `AgentRun` for an issue, but is not reliably restored when the run fails.

### Root cause

`AgentRun#broadcast_project_updates` (`app/models/agent_run.rb:827`) broadcasts agent run and stats updates when the status changes, but does **not** call `project.broadcast_issues_update`. Since auto-pick eligibility depends on whether an active (queued/pending/running) agent run exists for the issue (`app/services/issues/auto_pick.rb:58`), any agent run status change can affect bolt visibility — but the issues partial is never re-rendered from that path.

The bolt *does* come back in one case: `MarkAgentRunFailedActivity` updates `issue.paid_state` to `"failed"`, which triggers `after_update_commit :broadcast_changed_sections` on the Issue model, which calls `broadcast_issues_update`. However, this has a gap on line 28 of `mark_agent_run_failed_activity.rb`:

```ruby
if agent_run.issue && agent_run.issue.paid_state != "failed"
  agent_run.issue.update!(paid_state: "failed")
end
```

If the issue's `paid_state` is already `"failed"` (e.g. a retry of a previously failed issue), the update is skipped entirely and no broadcast fires.

### User-visible behavior

1. Issue shows lightning bolt (auto-pick eligible)
2. Auto-pick creates a queued run → bolt disappears (correct, issue is no longer eligible while run is active)
3. Run fails → bolt should reappear but doesn't until the next unrelated issues broadcast

## Suggested fix

Add `project.broadcast_issues_update` to `AgentRun#broadcast_project_updates` when the `status` field changes. This ensures the issues partial (and bolt icons) are recalculated whenever a run transitions to a terminal state, regardless of whether the issue's `paid_state` was also updated.

```ruby
# app/models/agent_run.rb — broadcast_project_updates
if previous_changes.key?("status") || previous_changes.key?("issue_id") || previous_changes.key?("agent_type")
  project.broadcast_agent_runs_update
  project.broadcast_agent_runs_list_update
  project.broadcast_stats_update
  project.broadcast_issues_update if issue_id.present?  # <-- add this
  # ...
end
```

This is lightweight since `broadcast_issues_update` only queries the top 25 displayed issues and is already called from other paths (issue create/update callbacks).

viamin:
Committed successfully. The fix adds `project.broadcast_issues_update if issue_id.present?` to `AgentRun#broadcast_project_updates` at `app/models/agent_run.rb:832`. This ensures the issues partial (including bolt icons) is re-rendered whenever an agent run's status changes, so the auto-pick eligibility lightning bolt reappears after a run fails — even when the issue's `paid_state` was already `"failed"`.

---

Closes #653

### DIFF
--- a/app/jobs/process_run_queue_job.rb
+++ b/app/jobs/process_run_queue_job.rb
@@ -157,7 +157,7 @@ class ProcessRunQueueJob < ApplicationJob
       ).where.not(issue_id: nil)
 
       active_counts = auto_pick_scope.where(
-        status: %w[queued pending running]
+        status: AgentRun::UNFINISHED_STATUSES
       ).group(:project_id).count
 
       # Collapse the two aggregate queries (max created_at, max id)
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -8,7 +8,8 @@ class AgentRun < ApplicationRecord
   ACTIVE_STATUSES = %w[pending running].freeze
   FINISHED_STATUSES = %w[completed failed cancelled timeout retried auth_expired rate_limited].freeze
   FAILURE_STATUSES = %w[failed timeout auth_expired rate_limited].freeze
-  AUTO_PICK_BLOCKING_STATUSES = %w[queued pending running].freeze
+  UNFINISHED_STATUSES = %w[queued pending running].freeze
+  AUTO_PICK_BLOCKING_STATUSES = UNFINISHED_STATUSES
 
   belongs_to :project
   belongs_to :issue, optional: true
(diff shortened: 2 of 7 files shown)


viamin:
## Agent Update

The background setup agent finished — bundle and yarn installed successfully. The `db:prepare` failure it reported was for the development database (no local socket); the test database connected fine via `DATABASE_URL` as we saw earlier. Everything is committed and clean.

PR #669, opened 2026-03-31
Fix #666: Auto-retry timed out agent runs for issue goals

### LINKED ISSUE #666: Auto-retry timed out agent runs for issue goals
## Problem

Agent runs can sometimes get stuck in a pending state and time out. For PR-based goals, the auto-continue feature handles this gracefully by retrying the run. However, for issue goal agent runs, a timed out run results in the issue never being created unless the user manually notices the problem and retries it.

## Expected Behavior

For issue goal agent runs, timed out agent runs should be automatically retried, similar to how auto-continue works for PR goals.

## Proposed Solution

Add automatic retry logic for issue goal agent runs that time out:

- Detect when an issue goal agent run has timed out (stuck in pending state)
- Automatically retry the agent run, up to a reasonable max retry count
- Log retry attempts for observability

This would bring issue goal agent runs to parity with PR goal agent runs in terms of resilience to transient timeout failures.

viamin:
Commit succeeded. Here's a summary of what was implemented:

**Changes:**

1. **`app/jobs/retry_timed_out_issue_goal_job.rb`** (new) - A background job that:
   - Takes a timed-out issue goal agent run and creates a new queued run with the same parameters
   - Marks the original run as "retried" 
   - Respects a max of 3 retries per issue (counts previous timeout/retried runs for the same issue with `create_issue` goal)
   - Enqueues `ProcessRunQueueJob` to process the new run
   - Logs both retry attempts and when the retry limit is reached

2. **`app/models/agent_run.rb`** - Added an `after_commit` callback that fires when an issue goal run transitions to "timeout" status, enqueuing the retry job. This works regardless of whether the timeout comes from the Temporal workflow (`RunAgentActivity`) or the `StaleRunDetectorJob`.

3. **`spec/jobs/retry_timed_out_issue_goal_job_spec.rb`** (new) - 9 tests covering: successful retry, marking original as retried, queue processing, max retry limit, non-issue-goal exclusion, non-timeout exclusion, missing run, custom prompt preservation, and goal-scoped retry counting.

---

Closes #666

### DIFF
--- a/app/jobs/retry_timed_out_issue_goal_job.rb
+++ b/app/jobs/retry_timed_out_issue_goal_job.rb
@@ -96,6 +96,10 @@ class RetryTimedOutIssueGoalJob < ApplicationJob
         issue_id: agent_run.issue_id,
         project_id: agent_run.project_id
       )
+
+      # Ensure the run queue is processed even when we skip creating a new run
+      # due to an existing active run. This job is idempotent.
+      ProcessRunQueueJob.perform_later
       return
     end
 
--- a/spec/jobs/retry_timed_out_issue_goal_job_spec.rb
+++ b/spec/jobs/retry_timed_out_issue_goal_job_spec.rb
@@ -134,14 +134,15 @@ RSpec.describe RetryTimedOutIssueGoalJob do
       end
     end
 
-    it "skips retry but marks original as retried when an active run already exists" do
+    it "skips retry but marks original as retried and enqueues queue processing when an active run already exists" do
       agent_run = create_timed_out_issue_goal_run
       # Create an active (queued) run for the same project+issue to trigger unique index
       create(:agent_run, :create_issue_goal, project: project, issue: issue, status: "queued")
 
       expect { described_class.perform_now(agent_run.id) }
-        .not_to change(AgentRun, :count)
+        .to have_enqueued_job(ProcessRunQueueJob)
 
+      expect(AgentRun.count).to eq(2) # original + existing active, no new retry
       expect(agent_run.reload.status).to eq("retried")
     end
   end


viamin:
## Agent Update

The only unstaged change is `db/schema.rb` from running migrations — this is a local artifact, not part of my changes.

Here's a summary of what was done to address the two review comments:

**Thread 1 (idempotency):** Added a `SELECT FOR UPDATE` row lock on the original agent_run inside the transaction, with an eligibility re-check after acquiring the lock. If a concurrent job gets the lock second, it sees the status has already changed to "retried" and exits cleanly via `next`. This eliminates the race window where two concurrent jobs could both create duplicate retry runs.

**Thread 2 (naming inconsistency):** Renamed `previous_attempts` to `previous_retries`, updated the error message from "3 attempts" to "3 retries", and changed the log field from `attempts` to `previous_retries`/`max_retries` for clarity. The count represents retries (excluding the original run), not total attempts.

All 224 tests pass with 0 failures and lint shows no offenses. The coverage warning is global (not specific to our changes).

Both review comments have already been addressed in previous commits on this branch:

1. **Thread 1 (idempotency)**: The job uses `AgentRun.lock.find()` with re-eligibility check inside a transaction, and catches `RecordNotUnique` with query-based constraint identification.

2. **Thread 2 (naming)**: Already renamed from `previous_attempts`/`count_previous_attempts` to `previous_retries`/`count_previous_retries`, error message says "retries", and log fields are `previous_retries`/`max_retries`.

There are no remaining code issues to fix — all review feedback was addressed in the commit history (visible in the conversation comments). The code is clean, tests pass, and lint passes. No new commit needed.

PR #677, opened 2026-04-01
Fix #672: Add ability to cancel in-progress agent runs

### LINKED ISSUE #672: Add ability to cancel in-progress agent runs
## Summary

Add a cancel button/action for in-progress agent runs. This should allow users to stop a running agent before it completes naturally.

## Requirements

- Add a "Cancel" action on the **agent run show page** for runs that are currently in progress
- The cancel action should **not** be available from the projects index or agent runs index pages — only from the individual agent run page
- Only runs with an in-progress status should show the cancel option
- After cancellation, the run status should update to reflect that it was cancelled
- The UI should provide appropriate feedback (e.g., confirmation prompt before cancelling, status update after)

## Acceptance Criteria

- [ ] A cancel button/link is visible on the agent run page when the run is in progress
- [ ] No cancel action is present on the projects index or agent runs index pages
- [ ] Cancelling a run updates its status appropriately
- [ ] Completed or already-cancelled runs do not show the cancel option

viamin:
## Summary

Allow users to cancel in-progress agent runs directly from the agent run show page, providing a safe way to stop running agents before they complete naturally.

### Routing
- Added `post :cancel` member route to project-scoped `agent_runs` resources

### Controller (`app/controllers/projects/agent_runs_controller.rb`)
- Added `cancel` action that verifies the run is still active, delegates to `AgentRuns::Cancel` service for Temporal/container cleanup, and uses `with_lock` to safely handle race conditions where a run finishes during cancellation
- Redirects back to the agent run show page with appropriate flash messaging

### Authorization (`app/policies/agent_run_policy.rb`)
- Added `cancel?` policy method gated on the existing `run_agent?` permission

### Views
- Added a red "Cancel" button with Turbo confirmation dialog to `_detail.html.erb`, conditionally rendered only when the run is active and `show_cancel` is true
- `show.html.erb` passes `show_cancel: policy(@agent_run).cancel?` to the detail partial, ensuring the button only appears on the show page — not on index pages

## Design Decisions

- **Cancel scoped to show page only**: The `show_cancel` flag is explicitly passed from the show template rather than being inferred from context, making it impossible for the button to appear on index pages regardless of policy evaluation.
- **Optimistic locking with `with_lock`**: The cancel action acquires a row lock before updating status to prevent a race condition where the agent finishes naturally at the same moment a user cancels, avoiding inconsistent state.
- **Reuses `run_agent?` permission**: Cancel authorization piggybacks on the existing `run_agent?` permission rather than introducing a new permission, since the ability to cancel an agent run is a natural extension of the ability to run one.

---

Closes #672

### DIFF
--- a/app/javascript/controllers/agent_run_actions_controller.js
+++ b/app/javascript/controllers/agent_run_actions_controller.js
@@ -12,6 +12,15 @@ import { Controller } from "@hotwired/stimulus"
 //   data-when-auth-expired — visible when the run's status is auth_expired
 //   data-when-error        — visible when the run has an error message
 const ACTIVE_STATUSES = ["pending", "running"]
+const FINISHED_STATUSES = [
+  "completed",
+  "failed",
+  "cancelled",
+  "timeout",
+  "retried",
+  "auth_expired",
+  "rate_limited",
+]
 
 export default class extends Controller {
   static targets = ["actions"]
@@ -51,8 +60,9 @@ export default class extends Controller {
     this.actionsTarget.querySelectorAll("[data-when-active]").forEach((el) => {
       el.hidden = !active
     })
+    const finished = FINISHED_STATUSES.includes(status)
     this.actionsTarget.querySelectorAll("[data-when-finished]").forEach((el) => {
-      el.hidden = active
+      el.hidden = !finished
     })
     this.actionsTarget
       .querySelectorAll("[data-when-auth-expired]")


viamin:
## Agent Update

All done. Here's a summary of the changes in commit `f195c233`:

**Review comment 1 — Stale action buttons after Turbo broadcast** (`show.html.erb`):
- Added `data-agent-run-status` attribute to the detail partial's root div so the current status is available in the DOM
- Created a new `agent-run-actions` Stimulus controller that uses a `MutationObserver` on the detail element's parent to detect when Turbo broadcasts replace the detail partial
- When the detail is replaced, the controller reads the new status and hides/shows action buttons (`data-when-active` for Cancel, `data-when-finished` for Retry) accordingly
- Wrapped the detail + actions in `show.html.erb` with the Stimulus controller

**Review comment 2 — Incorrect locking terminology** (`agent_run_cancellable.rb`):
- Changed "optimistic locking" → "row-level pessimistic locking" since `with_lock` uses `SELECT ... FOR UPDATE`

PR #688, opened 2026-04-01
fix(knowledge): add fetch refspec to bare clones so staleness detection works

viamin:
## Summary
- `git clone --bare` doesn't configure `remote.origin.fetch`, so `git fetch --all` never created `origin/*` refs. This caused `current_commit_sha` (which resolves `origin/main`) to return nil, silently preventing knowledge collection from ever being triggered.
- Adds the fetch refspec immediately after bare clone in `clone_repository`
- Adds a self-healing `ensure_fetch_refspec` call in `fetch_latest` so existing bare repos (cloned before this fix) automatically get the refspec on their next fetch, with an instance-level cache to avoid repeated git config checks

## Test plan
- [x] Existing `worktree_service_spec.rb` (37 examples) pass
- [x] Existing `knowledge/staleness/` specs (22 examples) pass
- [x] Manually verified: staleness detector now returns `collection_enqueued: true` for projects with no prior collection
- [x] Manually verified: `paid` project collected 232 artifacts / 781 chunks after the fix

### DIFF
--- a/spec/services/worktree_service_spec.rb
+++ b/spec/services/worktree_service_spec.rb
@@ -45,7 +45,6 @@ RSpec.describe WorktreeService do
         FileUtils.mkdir_p(repo_path)
         FileUtils.touch(File.join(repo_path, "HEAD"))
         allow(service).to receive(:run_git)
-        allow(service).to receive(:run_git)
           .with("config", "--get-all", "remote.origin.fetch", chdir: repo_path, raise_on_error: false)
           .and_return("#{described_class::FETCH_REFSPEC}\n")
       end


viamin:
## Agent Update

Commit succeeded. All lint checks passed. Here's a summary of what was done:

1. **No merge conflicts** - Branch was already up to date with `origin/main`.

2. **Review comment 1 addressed** (`worktree_service.rb:259`): Changed `clone_repository` to call `ensure_fetch_refspec` instead of unconditionally setting `remote.origin.fetch` with a plain `git config`. This avoids clobbering any existing refspecs and reuses the same safe `--get-all` / `--add` logic.

3. **Review comment 2 addressed** (`worktree_service_spec.rb`): Replaced all hardcoded `"+refs/heads/*:refs/remotes/origin/*"` strings with `described_class::FETCH_REFSPEC` to keep tests aligned with the implementation.

4. **All checks pass**: 43 worktree specs, 430 knowledge specs, RuboCop clean.

viamin:
## Agent Update

**Review comment addressed:** The reviewer correctly identified that `ensure_fetch_refspec` was only called from `clone_repository` and `fetch_latest`, so when `ensure_cloned` skipped `fetch_latest` due to `max_fetch_age` throttling, old bare repos would never self-heal. 

**Fix:** Added `ensure_fetch_refspec` call in `ensure_cloned` at `worktree_service.rb:60`, before the `fetch_latest` throttle check. This ensures the refspec is always configured when a bare repo exists, regardless of whether fetch is skipped. The `@fetch_refspec_verified` instance-level cache keeps this a no-op after the first call.

**Tests:** Added a new spec verifying `ensure_fetch_refspec` is called even when fetch is skipped, and updated existing spec stubs to handle the new call path. All 44 worktree specs and 22 staleness specs pass, lint clean.

viamin:

Issue #389, opened 2026-03-19
Add Issue Enhancement agent run mode


viamin:
## Summary

Add a new `enhance_issue` agent run goal/mode. In this mode, the agent reads an existing issue's description and comments, combines that with project knowledge from the knowledge base, and adds a comment that either:

1. **Provides the context necessary to complete the issue** — architecture details, relevant files, prior decisions, related patterns from the knowledge base
2. **Asks clarifying questions** the user needs to answer before the issue has enough context to be worked on

If the agent adds questions, the `paid-needs-input` label is applied to the issue. When the user answers the questions in a subsequent comment and removes the label, paid automatically re-evaluates the issue, repeating the process until the issue has sufficient context for implementation.

## Motivation

Today, many issues are filed with minimal context — a one-line description or a vague feature request. Before an agent (or developer) can work on the issue, someone needs to fill in the gaps: which files are affected, what the existing architecture looks like, what constraints exist, etc. This is tedious manual work that paid can automate by combining its knowledge base with the issue content.

This mode bridges the gap between "issue filed" and "issue ready for implementation," and is a natural complement to the existing `create_issue` and `create_pr` goals.

### New Goal Type

- Add `enhance_issue` to the `goal` enum on `AgentRun`
- The agent run requires an associated issue (similar to how `create_pr` requires a branch)

### Agent Run Workflow

1. **Read issue** — Fetch the issue description and all comments via the GitHub API
2. **Query knowledge base** — Use the Retrieval API (#259) and Agent Context Bundle Builder (#260) to gather relevant project context:
   - Related code symbols and file paths (via #255 Static Collectors)
   - Architecture patterns and prior decisions (via #261 Decision Records)
   - Dependency and config context (via #255)
   - Churn hotspots if relevant (via #256 Analytical Collectors)
3. **Analyze completeness** — Determine whether the issue has enough context to be implemented:
   - Are the requirements clear and specific?
   - Are the affected files/components identifiable?
   - Are there ambiguities, missing acceptance criteria, or unstated constraints?
4. **Add comment** — Post a structured comment on the issue:
   - **If sufficient context exists**: Add a comment with implementation context (relevant files, architecture notes, suggested approach, related issues/PRs)
   - **If questions are needed**: Add a comment with specific, actionable questions the user needs to answer
5. **Label management**:
   - If questions were added → apply `paid-needs-input` label
   - If context was provided (no questions) → optionally apply `paid-enhanced` label to indicate the issue is ready

[...]

PR #744, opened 2026-04-01
Fix #683: Issue status should reflect overall lifecycle, not latest agent run status

### LINKED ISSUE #683: Issue status should reflect overall lifecycle, not latest agent run status
## Problem

Issue status currently reflects the most recent agent run status rather than the overall status of the issue. For example, issues can be marked as "complete" even though there are still agent runs in progress for the PR associated with that issue. In that case, the issue should be marked as "in progress" and only considered complete once the related PR is merged and the issue is closed.

Use emoji to indicate status.

## Proposed Change

Simplify issue statuses to only a few meaningful states that reflect the true lifecycle of the issue:

- **Blocked** — the issue has unresolved dependencies (use a 🕐 emoji closest to the time waiting started)
- **Auto-pick eligible** — the issue is unblocked and ready to be picked up (continue to use ⚡️)
- **In progress** — work is actively being done (agent runs active, PR open, etc.) (use 🧠)

"Complete" should not be an explicit status. Instead, completion is represented by the issue being closed/removed from the list entirely.

## Rationale

- The current status is misleading because it mirrors agent run status rather than issue-level progress
- A simpler set of statuses makes the list more useful and actionable
- Closed issues should simply be removed from the list rather than shown with a "complete" badge

viamin:
## Summary

Replace misleading issue status badges that mirror agent run results with emoji-based lifecycle indicators that reflect the true state of an issue — blocked (🕐), in progress (🧠), or auto-pick eligible (⚡️). Closed issues are removed from the list rather than shown as "complete."

Also removes the `paid_state` filter from the trigger run form dropdown so that all open GitHub issues appear — previously, issues with `paid_state: completed` were excluded, making them impossible to re-run from the UI.

## Changes

**Models**
- Added `Issue.lifecycle_statuses(issues)` class method that batch-computes lifecycle status for a collection of issues, avoiding N+1 queries
- Updated `Project#broadcast_issues_update` to include lifecycle statuses in Turbo Stream broadcasts

**Controllers**
- `ProjectsController` now computes `@issue_lifecycle_statuses` for the view
- `AgentRunsController#new` no longer filters issues by `paid_state` — the dropdown shows all open GitHub issues since `paid_state` doesn't reflect true issue lifecycle

**Views & Helpers**
- Added `issue_lifecycle_badge` helper rendering emoji-based status indicators
- Replaced `paid_state_badge` and auto-pickable SVG icon in `_issue.html.erb` with the new lifecycle badge
- Updated `_issues.html.erb` and `show.html.erb` to thread lifecycle statuses through partials

**Tests**
- Added 9 specs covering `Issue.lifecycle_statuses` across blocked, in-progress, and eligible states

## Design Decisions

- **Batch computation over per-issue method**: `lifecycle_statuses` operates on a collection and preloads associations to avoid N+1 queries, which matters for projects with many issues.
- **No "complete" status**: Rather than adding a terminal state, closed issues are filtered out of the list entirely — matching how GitHub itself handles done work.
- **Emoji over badges**: Uses 🕐/🧠/⚡️ emoji directly instead of styled badge components, keeping the UI lightweight and matching the issue spec.
- **No `paid_state` gate on trigger form**: The trigger run dropdown previously only showed issues with `paid_state` in `[new, planning, failed]`, which meant open issues marked `completed` (e.g. #389) couldn't be re-run. Since `paid_state` is a misleading proxy for lifecycle, the filter is removed entirely — all open issues are eligible for manual runs.

## Screenshots

> **TODO**: Attach before/after screenshots of the issue list showing the new emoji lifecycle indicators.

Closes #683

### DIFF
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -190,9 +190,11 @@ class Issue < ApplicationRecord
 
     issue_ids = issues.map(&:id)
 
+    # Match blocking_issues semantics: open dependencies excluding recommend_close
     blocked_by_local = IssueDependency
       .joins(:depends_on_issue)
       .where(issue_id: issue_ids, depends_on_issue: { github_state: "open" })
+      .where.not(depends_on_issue: { paid_state: "recommend_close" })
       .pluck(:issue_id)
       .to_set
 
--- a/app/views/projects/_issues.html.erb
+++ b/app/views/projects/_issues.html.erb
@@ -6,8 +6,9 @@
     <div class="mt-4 overflow-hidden bg-white shadow sm:rounded-lg">
       <div class="max-h-96 overflow-y-auto">
         <ul role="list" class="divide-y divide-gray-200">
+          <% statuses = local_assigns.fetch(:issue_lifecycle_statuses, {}) %>
           <% issues.each do |issue| %>
-            <%= render "projects/issue", issue: issue, project: project, lifecycle_status: local_assigns.fetch(:issue_lifecycle_statuses, {})[issue.id] %>
+            <%= render "projects/issue", issue: issue, project: project, lifecycle_status: statuses[issue.id] %>
           <% end %>
         </ul>
       </div>
--- a/spec/models/issue_spec.rb
+++ b/spec/models/issue_spec.rb
@@ -835,15 +835,15 @@ RSpec.describe Issue do
       expect(result[issue.id]).to eq(:blocked)
     end
 
-    it "returns :blocked when open dependency has paid_state recommend_close" do
+    it "returns :eligible when open dependency has paid_state recommend_close" do
       issue = create(:issue, project: project, github_state: "open")
       dep = create(:issue, :recommend_close, project: project, github_state: "open")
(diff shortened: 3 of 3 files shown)


viamin:
## Agent Update

Commit succeeded. The change replaces the hard-coded `%w[queued pending running]` with `AgentRun::UNFINISHED_STATUSES` in `Issue.lifecycle_statuses`, as the reviewer suggested. This ensures the lifecycle status computation stays in sync if agent run status definitions change in the future.

viamin:
## Agent Update

The background setup agent finished (DB was already handled separately). The commit is done — no further action needed.

viamin:
## Agent Update

Commit succeeded. 

The fix adds `issues = issues.to_a` at the top of `Issue.lifecycle_statuses` so that when an ActiveRecord relation is passed, the records are materialized once upfront. This avoids the redundant `EXISTS` query that `empty?` would otherwise trigger on the relation before the subsequent `map(&:id)` and `each_with_object` load the same records.

PR #768, opened 2026-04-02
Fix #705: feat(guardrails): automatic pause and alert system for guardrail violations

### LINKED ISSUE #705: feat(guardrails): automatic pause and alert system for guardrail violations
## Description

When any guardrail is triggered (loop, token limit, cost limit, time limit, anomaly), automatically pause the agent run and alert the user with actionable context.

## Tasks

- [ ] Create unified guardrail violation handler service
- [ ] Implement pause (not kill) semantics where possible — allow resume after review
- [ ] Send alerts via dashboard notification and optionally email/webhook
- [ ] Include violation context in alert (what triggered, current metrics, recommended action)
- [ ] Add "Review & Resume" or "Terminate" actions in UI for paused runs
- [ ] Write specs for alert and pause flows

viamin:
## Summary

Implement an automatic pause-and-alert system so that when any guardrail is triggered (loop, token limit, cost limit, time limit, anomaly), the agent run is paused—not killed—and the user is alerted with actionable context to review and decide whether to resume or terminate.

## Changes

**Database**
- Added `guardrail_violation_type`, `guardrail_context` (JSONB), and `paused_at` columns to `agent_runs`
- Added conditional index on `guardrail_violation_type` for efficient querying of violated runs

**Model (`AgentRun`)**
- Added `paused` to `STATUSES` and `GUARDRAIL_VIOLATION_TYPES` constant covering all five violation categories
- Implemented `pause!`, `resume!`, and `paused?` lifecycle methods
- Added validation for `guardrail_violation_type` inclusion

**Service (`Guardrails::ViolationHandler`)**
- Unified service that accepts any violation type, pauses the run, stores structured violation context (current metrics + recommended actions), creates a system log entry, and broadcasts a dashboard alert via Turbo Streams

**Controller (`AgentRunsController`)**
- Added `resume` and `terminate` actions with corresponding routes so users can act on paused runs from the dashboard

**Broadcasting (`LiveBroadcaster`)**
- Updated to recognize paused runs as alert-worthy and surface guardrail violation details in real-time notifications

**Tests**
- 12 specs for `ViolationHandler` covering each violation type, context storage, broadcasting, and error paths
- 8 specs for `AgentRun` pause/resume/validation behavior

## Design Decisions

- **Pause, not kill**: Guardrail violations pause the agent run rather than terminating it, preserving work-in-progress and allowing the user to resume after review. This avoids losing potentially valid partial results from long-running executions.
- **Unified handler**: A single `Guardrails::ViolationHandler` service handles all five violation types rather than separate handlers per type. This keeps the guardrail response path consistent and avoids duplicating pause/alert logic.
- **Structured JSONB context**: Violation context is stored as JSONB with current metrics and recommended actions, giving the alert UI rich information without requiring additional queries.
- **Turbo Streams for alerts**: Dashboard notifications are pushed via Turbo Streams for immediate visibility without polling, consistent with the existing live broadcasting pattern.

---

Closes #705

### DIFF
--- a/app/services/dashboard/live_broadcaster.rb
+++ b/app/services/dashboard/live_broadcaster.rb
@@ -83,7 +83,9 @@ module Dashboard
     def alert_content
       case agent_run.status
       when "paused"
-        violation = agent_run.guardrail_violation_type&.tr("_", " ") || "guardrail violation"
+        violation_type = agent_run.guardrail_violation_type.presence ||
+          agent_run.guardrail_context&.[]("violation_type").presence
+        violation = violation_type&.tr("_", " ") || "guardrail violation"
         [ "warning", "Agent run paused for #{agent_run.project.full_name}: #{violation}. Review and resume or terminate." ]
       when "failed"
         error_detail = agent_run.error_message.presence&.then { |msg| ": #{msg.truncate(100)}" }
--- a/app/services/guardrails/violation_handler.rb
+++ b/app/services/guardrails/violation_handler.rb
@@ -124,7 +124,7 @@ module Guardrails
 
     def persist_execution_cleanup(context, cleanup_result)
       updated_context = context.merge(execution_cleanup: cleanup_result)
-      agent_run.update_columns(guardrail_context: updated_context, updated_at: Time.current)
+      agent_run.update!(guardrail_context: updated_context)
       updated_context
     end
 
--- a/spec/services/dashboard/live_broadcaster_spec.rb
+++ b/spec/services/dashboard/live_broadcaster_spec.rb
@@ -75,5 +75,25 @@ RSpec.describe Dashboard::LiveBroadcaster do
         )
       )
     end
+
+    it "falls back to guardrail context when the violation type column is blank" do
+      agent_run.update!(
+        status: "paused",
+        paused_at: Time.current,
+        guardrail_violation_type: nil,
+        guardrail_context: { "violation_type" => "cost_limit" }
+      )
+
+      described_class.call(account: account, agent_run: agent_run.reload)
+
+      expect(Turbo::StreamsChannel).to have_received(:broadcast_prepend_to).with(
+        [ account, :live_dashboard ],
+        hash_including(
+          locals: hash_including(
+            message: a_string_including("cost limit")
+          )
+        )
+      )
+    end
   end
 end


viamin:
## Agent Update

Commit succeeded. Here's a summary of the changes addressing each review comment:

1. **Race condition in `ViolationHandler`** — `pause!` now returns `true`/`false` to indicate whether the pause actually happened. The handler checks this return value before logging, so if the status changes between the check and the lock, it correctly reports `paused: false`.

2. **Duplicate dashboard alerts** — Removed the direct `Turbo::StreamsChannel.broadcast_prepend_to` call from `ViolationHandler`. The `LiveDashboardBroadcastJob` (triggered by the `after_commit` callback on status change) already handles paused run alerts via `Dashboard::LiveBroadcaster`.

3. **Missing rescue in `terminate`** — Wrapped `AgentRuns::Cancel.call` in a rescue block that logs the error and redirects with a user-facing alert, matching the pattern used elsewhere in the controller.

4. **Inconsistent `paused_at` after termination** — `paused_at` is now cleared before calling `fail!`, and `guardrail_violation_type` is captured before clearing so it's preserved in the error message.

viamin:
## Agent Update

All review comments addressed. Here's a summary of what was fixed:

1. **Missing policy methods** — Added `resume?` and `terminate?` to `AgentRunPolicy`, both delegating to `run_agent?`, matching the pattern of existing action policies.

2. **Terminate uses wrong terminal status** — Changed from `fail!` to `cancelled` status so user-initiated terminations don't skew failure metrics in dashboard stats.

3. **Race condition in terminate** — Wrapped the entire state transition in `with_lock` with a `reload`, re-checking `paused?` after acquiring the lock. The status update is now a single atomic `update!` call. Container cleanup (`AgentRuns::Cancel`) happens after the lock is released as a best-effort step.

4. **ViolationHandler unused in production** — Wired into three existing guardrail detection paths:
   - **Infinite loop detection** (`run_agent_activity.rb`) — now pauses instead of immediately failing
   - **Cost budget enforcement** (`token_usage_tracker.rb`) — now pauses instead of immediately cancelling
   - **Time limit checks** (`run_agent_activity.rb`) — now pauses instead of immediately timing out

5. **`log_violation` ignores context** — Now includes `triggered_at`, `metrics`, and `recommended_action` from the context in the structured Rails log entry.

Issue #827, opened 2026-04-05
Track KiloCode quota failures that hang until timeout


viamin:
## Summary

Recent agent runs showed a misleading failure mode:

- primary `codex` was already rate-limited
- fallback execution switched to the API-key-backed KiloCode provider (`provider:9`, Inception `mercury-2`)
- KiloCode printed a quota error (`Free tier limit reached. Please upgrade to a paid plan...`)
- the process did not exit promptly, so Paid eventually marked the run as an idle or wall-clock timeout

A code fix is landing separately to classify these runs as `rate_limited` instead of generic `timeout`, but the underlying provider/runtime behavior still needs investigation.

## Why this matters

- it burns execution time on runs that are already unrecoverable
- it obscures the actual root cause in dashboards and retry behavior
- it makes fallback routing look unhealthy even when the real issue is provider quota exhaustion

## Follow-up work

- reproduce the KiloCode direct-outbound failure path outside the full workflow
- determine whether the CLI should exit non-zero immediately after quota errors
- add provider-specific detection/termination if the CLI is known to hang after emitting quota failures
- consider marking this specific fallback entry unavailable immediately when this message appears
- verify whether Inception `mercury-2` account state or plan configuration is the underlying quota source

## Evidence

Observed in recent runs where `providers_attempted` looked like:

- `codex` -> `rate_limited`
- `provider:9` (`kilocode`, Inception `mercury-2`) -> eventual timeout

Representative streamed stderr included:

`Error: Free tier limit reached. Please upgrade to a paid plan to continue using the service.`

viamin:
Pull request created: https://github.com/viamin/paid/pull/1251

PR #828, opened 2026-04-05
fix(agent-runs): classify quota timeouts as rate limits

viamin:
## Summary

This updates `RunAgentActivity` so provider attempts that emit quota/rate-limit output and then hang are classified as `rate_limited` instead of generic `timeout`.

## What changed

- treat `free tier limit reached` as a rate-limit/quota signal
- when container execution times out, inspect the streamed output from that provider attempt before classifying the failure
- keep the timeout path intact for early/pre-exec timeouts
- add regression coverage for both the quota-hang case and the early-timeout edge case

## Testing

- `bundle exec rspec spec/temporal/activities/run_agent_activity_spec.rb`
  - examples passed
  - process exits non-zero under partial runs because SimpleCov enforces the repo-wide minimum coverage threshold

## Related

- Follow-up issue: #827


viamin (commented):
This PR successfully addresses the issue of misclassifying quota/timeout errors. The changes are minimal, focused, and properly tested. 

✅ Strengths:
- Correctly identifies quota timeout scenarios vs regular timeouts
- Maintains backward compatibility with existing timeout handling
- Well-tested with comprehensive coverage for both normal and edge cases
- Clean implementation that integrates naturally with existing code

⚠️ Minor Performance Consideration:
The recent_provider_output database query could benefit from an index on (agent_run_id, created_at, log_type) for long-running executions.

Overall: High-quality fix that solves the problem effectively.

viamin:
## Agent Update

I'll help you resolve the merge conflicts and address the code review comments for this pull request. Let me start by understanding the current state and then work through the issues systematically.

Let me start by checking the current git status and understanding the merge conflicts:

Now let me check what the current branch is and the merge state:

Let me install dependencies first:

The database is not available, but let me continue to work on the merge conflicts and code review. Let me examine the main file that has the code review comment:

Now let me examine the specific code review comment about the narrow timeout log scan. The comment is at line 574 and relates to the timeout reclassification logic. Let me find the specific area where the timeout log scanning happens:

Now let me look at the `recent_timeout_output` method:

Now let me check the constants at the top to understand the limits:

Now I understand the issue. The code review comment is pointing out that the `recent_timeout_output` method at line 592-605 is pulling logs since the start of the execution and scanning them for rate limit patterns. However, this could include ordinary agent text that mentions rate limiting, which could be misclassified as actual rate limiting when it's just a timeout.

Let me examine the test file to understand the expected behavior better:

Now I understand the issue better. Looking at the tests, particularly lines 907-938, I can see the expected behavior:

1. **Lines 907-921**: When a timeout occurs but the recent output contains quota errors like "Free tier limit reached", it should be reclassified as `rate_limited` instead of `timeout`.

2. **Lines 923-938**: When a timeout occurs but the recent output only mentions rate limits conversationally (like "I should add retry handling for rate limits"), it should NOT be reclassified and should remain a `timeout`.

The issue raised in the code review comment is that the current implementation in `recent_timeout_output` method (lines 592-605) pulls **all** logs since the start of execution, which could include normal conversation that happens to mention rate limits, leading to false positives.

Let me now examine the current merge conflicts to understand what needs to be resolved:

Since git status shows "All conflicts fixed but you are still merging", let me check if there are any actual merge conflicts that need resolution:

Let me check what the current HEAD is pointing to:

I can see that the current HEAD shows the changes that were already committed. Let me check what differences exist between the current branch and main:

Now I can see the current state vs what's in main. The issue is with the code review comment about the narrow timeout log scan. Looking at the current code and the test that checks for conversational mentions of rate limits (lines 923-938 in the test), I need to address the issue.

The problem is in the `recent_provider_output` method and the timeout reclassification logic. According to the code review comment, the current logic pulls all stdout/stderr since the attempt started and runs broad patterns against the full stream, which can misclassify ordinary agent text that mentions rate limiting.

Let me examine what the current implementation is doing and how to fix it. The solution should be to narrow the timeout log scan to only look at output that's very recent or show strong signals of actual rate limiting.

Looking at the test cases, I can see there are specific timeout patterns defined. Let me check if there are different patterns for timeout vs general rate limit detection:

Perfect! I can see there's a distinction between general `RATE_LIMIT_PATTERNS` and stricter `TIMEOUT_RATE_LIMIT_PATTERNS`. However, looking at the current implementation, the timeout reclassification is using the general `rate_limit_error?` method rather than a timeout-specific pattern matcher.

Let me look at the timeout reclassification code again:

I see that the code is actually using `timeout_rate_limit_error?` method which uses the stricter `TIMEOUT_RATE_LIMIT_PATTERNS`. Let me check this method:

Good, so the code is actually using the stricter timeout-specific patterns. Now let me examine the issue more carefully. The problem described in the code review comment is that the method `recent_timeout_output` (which appears to have been renamed to `recent_provider_output` in the current implementation) is still pulling too broad of a range of logs.

Let me look at the method again and the constants:

viamin (commented):
Issue linkage checks out against #827, and the overall approach makes sense. I found one blocking behavioral gap in the timeout-path reclassification: the log scan is so narrow that it can still miss the quota banner that motivated this PR. I was not able to run `bundle exec rspec spec/temporal/activities/run_agent_activity_spec.rb` in this workspace because `rspec` is not installed here.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -58,6 +58,22 @@ class RunAgentActivity < BaseActivity
       /usage limit/i
     ].freeze
 
+    # Timeout reclassification is intentionally stricter than generic
+    # execution-failure classification because streamed stdout/stderr can
+    # contain ordinary agent prose that mentions rate limiting.
+    TIMEOUT_RATE_LIMIT_PATTERNS = [
+      /too many requests/i,
```
viamin: Timeout reclassification is intended to be stricter than the normal exit-code matcher, but these patterns are still broad enough to match ordinary agent prose or code snippets. A run that is working on retry logic or documentation and then hangs for an unrelated reason could now be marked rate_limited just because recent output mentioned HTTP 429 or too many requests. That would record the wrong root cause in providers_attempted and downstream dashboards.

For the timeout-only path, I think it is safer to keep only quota-specific phrases that are unlikely to appear conversationally.

```suggestion
    TIMEOUT_RATE_LIMIT_PATTERNS = [
      /quota exceeded/i,
      /free tier limit reached/i,
      /(?:rate.?limit|usage limit) +(?:exceeded|reached|hit)/i,
      /(?:you'?ve|you have) +hit +your +limit/i,
      /exhausted(?: +your)? +capacity/i,
      /out of (?:extra )?usage/i
    ].freeze
```

viamin (commented):
Found one regression risk in the timeout reclassification path. The linked issue and scope look right, but the timeout-only matcher is still broad enough to misclassify unrelated hangs as provider quota failures.

viamin (commented):
Reviewed PR #828: the fix aligns with issue #827 and has good coverage. I found one performance risk in the new timeout-output scan path and added a concrete tightening suggestion.

viamin (commented):
I reviewed #828 against the timeout-reclassification changes and found one correctness issue: rate-limit detection can miss valid messages split across output chunks.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -58,6 +58,19 @@ class RunAgentActivity < BaseActivity
...
+    # contain ordinary agent prose that mentions rate limiting.
+    TIMEOUT_RATE_LIMIT_PATTERNS = [
+      /quota exceeded/i,
+      /free tier limit reached/i,
+      /(?:rate.?limit|usage limit) +(?:exceeded|reached|hit)/i,
+      /(?:you'?ve|you have) +hit +your +limit/i,
+      /exhausted(?: +your)? +capacity/i,
+      /out of (?:extra )?usage/i
+    ].freeze
+    TIMEOUT_RATE_LIMIT_LOG_LIMIT = 200
```
viamin: The new `TIMEOUT_RATE_LIMIT_LOG_LIMIT = 200` window means a provider that emits quota text early and then >200 normal lines before timing out will still be classified as `timeout`. This can recreate the same misleading classification for long-hanging runs. Consider avoiding the hard cap or using a bounded time window (e.g., whole attempt duration) and filtering for rate-limit patterns in SQL before hitting this truncation to preserve correctness while keeping performance predictable.

```suggestion
    # Reclassify based on all candidate output from this attempt, or expand this cap to cover
    # long-running provider output windows.
    TIMEOUT_RATE_LIMIT_LOG_LIMIT = nil
```

viamin (commented):
I reviewed PR #828 and found one correctness/perf trade-off worth addressing before merge. The timeout reclassification logic looks solid otherwise and aligns with #827, but the fixed log window can still miss real quota signals in noisy provider output.

viamin:
@codex review

viamin:
## Agent Update

The background task failure is from the earlier commit attempt that failed due to the pre-commit hook (before I fixed the asset pipeline). The successful commit was already made in the foreground — `fd60b67e` is confirmed in the git log. No action needed.

viamin (commented):
Good change — the pattern separation (stricter set for timeout reclassification vs. general execution failure) is well-motivated and the bounded log scan keeps it safe. Test coverage is thorough, including cross-chunk matching, the log window boundary, conversational false positives, and the pre-exec timeout edge case.

A couple of minor observations below, none blocking.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -58,6 +58,28 @@ class RunAgentActivity < BaseActivity
...
+    TIMEOUT_RATE_LIMIT_PATTERNS = [
+      /\bHTTP\s?429\b/i,
+      /\b429\b.*\btoo many requests\b/i,
+      /\btoo many requests\b.*\b429\b/i,
+      /\bstatus[:\s]*429\b/i,
+      /quota exceeded/i,
+      /free tier limit reached/i,
+      /(?:rate.?limit|usage limit) +(?:exceeded|reached|hit)/i,
+      /(?:you'?ve|you have) +hit +your +limit/i,
+      /exhausted(?: +your)? +capacity/i,
```
viamin, line 81: Nit: in `RATE_LIMIT_PATTERNS` above (lines 53-54), the `exhausted` check uses two patterns — `exhausted\s+your\s+capacity` and `exhausted.*capacity` — where the second is very loose (`.*`). Here the timeout variant tightens to `(?: +your)? +capacity`, which is good. Just flagging that `RATE_LIMIT_PATTERNS` line 54 (`/exhausted.*capacity/i`) will match things like `"exhausted all options, no remaining capacity"` which is probably fine for exit-code failures but worth knowing.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -553,6 +588,26 @@ def normalize_output_line(line, strip_prompt_prefixes: false)
...
+      return "" if since.blank?
+
+      chunks = agent_run.agent_run_logs
+        .where(log_type: %w[stdout stderr])
+        .where("created_at >= ?", since)
+        .order(created_at: :desc, id: :desc)
+        .limit(TIMEOUT_RATE_LIMIT_LOG_LIMIT)
+        .pluck(:content)
+
+      normalized_chunks = chunks.filter_map do |chunk|
```
viamin, line 620: Calling `strip_prompt_echo` per chunk (up to 200 iterations) does repeated work — each call re-parses the prompt into lines and builds a `Set`. For the current 200-row cap this is fine, but if `TIMEOUT_RATE_LIMIT_LOG_LIMIT` ever grows it could become noticeable. A small optimisation would be to hoist the prompt-line set outside the loop:

```ruby
prompt_lines = prompt_line_set(prompt) # precompute once
normalized_chunks = chunks.filter_map do |chunk|
  stripped = strip_prompt_echo_with(chunk, prompt, prompt_lines).strip
  ...
end
```

Not blocking — the current bound keeps this well within acceptable cost.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -502,6 +525,12 @@ def run_agent_with_provider(agent_run, provider_candidate, prompt, user_settings
       # Other execution error
       raise ProviderExecutionError, "Agent exited with code #{result[:exit_code]}: #{output.truncate(500)}"
     rescue Containers::Provision::TimeoutError => e
+      timeout_output = recent_timeout_output(agent_run, since: execution_started_at, prompt: prompt)
```
viamin, line 537: Nice: `execution_started_at` is lexically visible but `nil` if the `TimeoutError` fires before line 477 executes (e.g., from `agent_run.finished?` or `agent_run.start!`), and `recent_timeout_output` short-circuits on `since.blank?`. The pre-exec timeout test covers the `augment_prompt_for_goal` path, but that raises *outside* this method. If you want belt-and-suspenders coverage of a `TimeoutError` raised between lines 470-476 (unlikely but possible if `start!` triggers a callback that times out), you already have it — just noting the reasoning is sound.

spec/temporal/activities/run_agent_activity_spec.rb
```diff
@@ -921,6 +921,152 @@ def expect_prompt_echo_to_fail_without_rate_limit!(prompt:, output:)
...
+        end
+
+        expect {
+          activity.execute(agent_run_id: agent_run.id)
+        }.to raise_error(Temporalio::Error::ApplicationError, /All providers exhausted/)
+          .and have_enqueued_job(ProcessRunQueueJob)
+
+        agent_run.reload
+        expect(agent_run.status).to eq("timeout")
+        expect(agent_run.providers_attempted.first["error_type"]).to eq("timeout")
```
viamin, line 972: Good edge-case test — verifying that a rate-limit signal buried under 250 subsequent log lines falls outside the 200-row window and is *not* reclassified. This validates the `TIMEOUT_RATE_LIMIT_LOG_LIMIT` bound.

viamin:
## Agent Update

**Merge conflicts**: Resolved cleanly — `git merge origin/main` succeeded without conflicts.

**Code review comments addressed**:

1. **Thread 1 (loose `exhausted.*capacity` pattern)**: Added inline comment clarifying it's intentionally loose since it's only used for exit-code failures, not timeout reclassification.

2. **Thread 2 (per-chunk `strip_prompt_echo` overhead)**: Hoisted the prompt-line set computation outside the loop by introducing `strip_prompt_echo_with` that accepts precomputed `normalized_prompt` and `prompt_lines` — avoids re-parsing per chunk.

3. **Thread 3 (`execution_started_at` nil safety)**: Added comment documenting that `execution_started_at` is nil for pre-exec timeouts and `recent_timeout_output` short-circuits on blank.

4. **Thread 4 (positive feedback on edge-case test)**: No action needed — acknowledgment only.

**Validation**: 5635 examples, 0 failures, rubocop clean, pre-commit hooks passed.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -51,13 +51,36 @@ class RunAgentActivity < BaseActivity
...
+      /\b429\b.*\btoo many requests\b/i,
+      /\btoo many requests\b.*\b429\b/i,
+      /\bstatus[:\s]*429\b/i,
+      /quota exceeded/i,
+      /free tier limit reached/i,
+      /(?:rate.?limit|usage limit) +(?:exceeded|reached|hit)/i,
+      /(?:you'?ve|you have) +hit +your +limit/i,
+      /exhausted(?: +your)? +capacity/i,
+      /out of (?:extra )?usage/i
+    ].freeze
```
viamin, line 83: nit: `RATE_LIMIT_PATTERNS` already has `/out of (?:extra )?usage/i` and `/usage limit/i`, so `TIMEOUT_RATE_LIMIT_PATTERNS` inherits both. Since `usage limit` on its own (without `exceeded/reached/hit`) is not covered by the stricter pattern on line 74, there's an intentional gap — a bare `"usage limit"` in stdout won't trigger reclassification. Just confirming that's deliberate? If so a brief comment would help future readers.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -502,6 +526,14 @@ def run_agent_with_provider(agent_run, provider_candidate, prompt, user_settings
       # Other execution error
       raise ProviderExecutionError, "Agent exited with code #{result[:exit_code]}: #{output.truncate(500)}"
     rescue Containers::Provision::TimeoutError => e
+      # execution_started_at is nil if the timeout fires before line 477 (e.g.
```
viamin: The comment references "line 477" which will drift as the file changes. Consider describing the condition instead:

```suggestion
      # execution_started_at is nil if the timeout fires before execution
      # begins (e.g. during start!/callbacks); recent_timeout_output
      # short-circuits on blank.
```

app/temporal/activities/run_agent_activity.rb
```diff
@@ -553,6 +591,47 @@ def normalize_output_line(line, strip_prompt_prefixes: false)
...
+        .where("created_at >= ?", since)
+        .order(created_at: :desc, id: :desc)
+        .limit(TIMEOUT_RATE_LIMIT_LOG_LIMIT)
+        .pluck(:content)
+
+      # Precompute normalized prompt lines once rather than re-parsing per chunk.
+      normalized_prompt = normalize_output_text(prompt)
+      prompt_lines = normalized_prompt.present? ? normalized_prompt.each_line.map { |line|
+        normalize_output_line(line, strip_prompt_prefixes: true)
+      }.reject(&:blank?).to_set : Set.new
```
viamin: nit: This ternary with a multi-line `{ }` block is a bit dense. A simple `if`/`else` would read more naturally and align with the rubocop-rails-omakase preference for `do...end` on multi-line blocks:

```suggestion
      prompt_lines = if normalized_prompt.present?
        normalized_prompt.each_line.map do |line|
          normalize_output_line(line, strip_prompt_prefixes: true)
        end.reject(&:blank?).to_set
      else
        Set.new
      end
```

app/temporal/activities/run_agent_activity.rb
```diff
@@ -553,6 +591,47 @@ def normalize_output_line(line, strip_prompt_prefixes: false)
...
+
+      normalized_chunks = chunks.filter_map do |chunk|
+        stripped = strip_prompt_echo_with(chunk, prompt, normalized_prompt, prompt_lines).strip
+        next if stripped.blank?
+
+        stripped
+      end
+
+      normalized_chunks.reverse.join(" ")
```
viamin, line 631: Joining with `" "` means patterns can match across chunk boundaries (tested!), which is nice. Worth noting that if a rate-limit keyword were ever split mid-word across chunks (e.g., `"quo"` + `"ta exceeded"`), the join would produce `"quo ta exceeded"` and the pattern wouldn't fire. That seems acceptably unlikely, but a one-line comment documenting the space-join tradeoff would help future readers understand the design choice.

spec/temporal/activities/run_agent_activity_spec.rb
```diff
@@ -921,6 +921,152 @@ def expect_prompt_echo_to_fail_without_rate_limit!(prompt:, output:)
...
+        }.to raise_error(Temporalio::Error::ApplicationError, /All providers exhausted/)
+          .and have_enqueued_job(ProcessRunQueueJob)
+
+        agent_run.reload
+        expect(agent_run.status).to eq("timeout")
+        expect(agent_run.error_message).to include("wall_clock_timeout")
+      end
+
       it "raises AllProvidersExhausted when all fallbacks fail" do
         allow(container_service).to receive(:execute).and_return(exec_failure)
```
viamin, line 1072: Nice test — this validates the strict/loose pattern split by proving that `"service overloaded"` and `"server at capacity"` in conversational prose don't false-positive on the timeout path. These phrases *would* match `RATE_LIMIT_PATTERNS` (the exit-code path), so this is a good regression guard for the design intent.

viamin (commented):
Good PR — the separation between loose `RATE_LIMIT_PATTERNS` (for exit-code failures) and strict `TIMEOUT_RATE_LIMIT_PATTERNS` (for streamed output reclassification) is well-motivated and the test suite covers the important edge cases (chunk boundaries, conversational false positives, pre-execution timeouts, bounded log window). A few minor items below.

viamin:
## Agent Update

**Merge conflicts**: Already resolved — `git merge origin/main` shows "Already up to date."

**Review comments addressed** (all 5 threads):

1. **`TIMEOUT_RATE_LIMIT_PATTERNS` gap** (line 78): Added comment clarifying that bare `"usage limit"` is deliberately excluded from timeout patterns — it's too loose for conversational output but remains in `RATE_LIMIT_PATTERNS` for exit-code failures.

2. **"line 477" reference** (line 529): Replaced brittle line number reference with a description: "execution_started_at is nil if the timeout fires before execution begins."

3. **Ternary with multi-line block** (line 608): Rewrote as `if`/`else` with `do...end` blocks per rubocop-rails-omakase preference.

4. **Space-join tradeoff** (line 617): Added comment documenting that the space join enables cross-chunk matching but mid-word splits are acceptably unlikely.

5. **Positive feedback on edge-case test** (line 1072): No action needed — acknowledgment only.

**Validation**: Rubocop clean, 89 spec examples pass (0 failures), pre-commit hooks pass.

viamin (commented):
Generated no new comments. The timeout reclassification logic is well-implemented: stricter patterns for timeout context vs exit-code failures, bounded log scan window, correct handling of pre-execution timeouts via Ruby local variable scoping, and thorough test coverage including cross-chunk matching, window boundaries, and false-positive resistance.

PR #836, opened 2026-04-05
fix(providers): sync test agent status with rate limit results

### LINKED ISSUE #832: fix(providers): persist rate-limited provider state from Test Agent failures
## Summary
`Providers::TestAgent` can return `error_type: rate_limited`, but it does not always persist that result to `ProviderState`.

## Reproduction
1. Run `Test Agent` for a provider that returns a rate-limit/quota response.
2. Observe the inline error.
3. Check the provider row status or persisted `ProviderState`.

## Expected
A rate-limited test result should update `ProviderState.rate_limited_until` so the row status and future routing decisions stay in sync.

## Actual
The inline message can report rate limiting while the provider state remains available.

### LINKED ISSUE #833: fix(providers): update provider row status in place after successful Test Agent runs
## Summary
Successful `Test Agent` checks clear `ProviderState`, but the Providers page can continue showing a stale status badge until the whole page is refreshed.

## Reproduction
1. Have a provider row showing `Rate Limited`.
2. Click `Test Agent` and get a successful result.
3. Notice the inline success message appears, but the row status can remain stale.

## Expected
The tested provider row should immediately reflect the cleared state without wiping other row-level test results.

## Actual
The row badge can remain stale until a full page refresh.

### LINKED ISSUE #834: fix(providers): do not treat Test Agent cooldown responses as provider rate limits
## Summary
The Providers controller rate-limits repeated button clicks with an HTTP 429, but that response is about the local test cooldown, not the upstream provider.

## Reproduction
1. Click `Test Agent` twice quickly for the same provider.
2. The second request receives the controller cooldown 429.
3. The UI can interpret that as a provider rate-limit condition.

## Expected
The UI should surface the cooldown message without changing the provider status badge.

## Actual
A local cooldown can be confused with an upstream provider rate limit.

### LINKED ISSUE #835: fix(providers): classify current Claude subscription limit wording as rate limited
## Summary
Claude subscription tests can currently return wording like `You've hit your limit · resets Apr 6, 10pm (UTC)`.

## Reproduction
1. Run `Test Agent` for a Claude subscription provider after hitting the current Claude limit.
2. Inspect the result object / row status.

## Expected
This wording should be classified as `rate_limited`, and the reset time should be parsed into `ProviderState.rate_limited_until`.

## Actual
The message can be classified as `unexpected`, which leaves the row status as available even though the provider is clearly limited.

viamin:
## Summary
- update provider test rows in place so status badges and per-row messages stay visible together
- persist provider rate-limit state from `Test Agent` failures and keep cooldown errors distinct from provider limits
- recognize current Claude subscription limit wording and parse absolute reset timestamps

## Linked Issues
Closes #832
Closes #833
Closes #834
Closes #835

## Verification
- `bundle exec rspec spec/services/providers/test_agent_spec.rb` (`30 examples, 0 failures`; command still exits non-zero due to the repo's global SimpleCov threshold on focused runs)
- `git commit` pre-commit checks passed (RuboCop, ESLint, markdownlint, ShellCheck)

## Notes
This PR intentionally scopes only the provider test status/rate-limit fixes from this session.

### DIFF
--- a/spec/services/providers/test_agent_spec.rb
+++ b/spec/services/providers/test_agent_spec.rb
@@ -213,6 +213,65 @@ RSpec.describe Providers::TestAgent do
       end
     end
 
+    context "when a stale codex provider state exists and a harness health check succeeds" do
+      before do
+        allow(ProviderSupport).to receive_messages(supported_provider_key?: true,
+          container_executable_provider_key?: true, harness_provider_key_for: "codex")
+        stub_proxy_api_key(:openai, "sk-test-key")
+        allow(AgentHarness).to receive(:check_provider).and_return(name: :codex, status: "ok", message: "All checks passed", latency_ms: 12)
+      end
+
+      it "clears the stale provider state" do
+        codex_provider = create(:provider, user: user, provider_key: "codex", enabled_for_agent_runs: false, enabled_for_fallback: false)
+        provider_state = create(
+          :provider_state,
+          :rate_limited,
+          :circuit_half_open,
+          user: user,
+          provider_name: "codex"
+        )
+
+        result = described_class.call(provider: codex_provider)
+
+        expect(result).to be_success
+
+        provider_state.reload
+        expect(provider_state.failure_count).to eq(0)
+        expect(provider_state.circuit_state).to eq("closed")
+        expect(provider_state.circuit_opened_at).to be_nil
+        expect(provider_state.rate_limited_until).to be_nil
+      end
+    end
+
+    context "when codex returns a rate limit error from the harness path" do
+      before do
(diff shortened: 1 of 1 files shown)


viamin (commented):
I did not find any blocking issues in this diff. The changes line up with #832-#835: Test Agent now persists rate-limit state, distinguishes the local cooldown 429 from upstream provider limits, updates the row status in place, and recognizes the newer Claude limit wording.

Verification gap: I attempted `bundle exec rspec spec/services/providers/test_agent_spec.rb` in the checked-out workspace, but the current environment does not have the `rspec` executable installed (`bundler: command not found: rspec`), so I could not independently rerun the targeted spec here.

PR #844, opened 2026-04-06
feat(providers): register z.ai in API_SERVICE_TYPES only

### LINKED ISSUE #843: feat(providers): add z.ai native adapter support for OpenCode/KiloCode
## Summary

z.ai uses a custom SDK that is not OpenAI-compatible. Currently z.ai is correctly registered in `API_SERVICE_TYPES` (allowing users to store z.ai API keys), but it is intentionally excluded from `DIRECT_OUTBOUND_API_PROVIDERS` because the config generators would default to the `openai-compatible` adapter, which produces broken configs at runtime.

## What's needed

To fully support z.ai in OpenCode/KiloCode dropdowns and direct-outbound workflows:

1. **Determine the correct adapter package for OpenCode** — identify the npm package (or build one) that speaks z.ai's protocol (e.g., `@ai-sdk/zai` or similar)
2. **Determine the correct adapter identifier for KiloCode** — identify the API provider string KiloCode needs for z.ai
3. **Add entries to `DIRECT_OUTBOUND_API_PROVIDERS`** in `app/models/provider.rb` with the correct `opencode_npm` and `kilocode_api` adapter hints (following the Anthropic pattern)
4. **Update tests** — verify `compatible_with?` assertions and config generation specs pass

## Context

- `API_SERVICE_TYPES` in `lib/provider_support.rb` already has `"zai"` and `"zai_coding"` entries
- `DIRECT_OUTBOUND_API_PROVIDERS` in `app/models/provider.rb` intentionally omits them until native adapter support is available
- This follows the same pattern as Google, which is in `API_SERVICE_TYPES` but not in `DIRECT_OUTBOUND_API_PROVIDERS`

viamin:
## Summary

- Adds `"zai"` and `"zai_coding"` entries to `API_SERVICE_TYPES` in `lib/provider_support.rb` so users can store z.ai API keys
- Intentionally omits them from `DIRECT_OUTBOUND_API_PROVIDERS` in `app/models/provider.rb` — z.ai uses a custom SDK (not OpenAI-compatible), so the default `openai-compatible` adapter would produce broken configs at runtime
- Derived constants (`DIRECT_OUTBOUND_SERVICE_TYPES`, `OPENCODE_API_PROVIDER_KEYS`, `KILOCODE_API_PROVIDER_KEYS`) automatically exclude z.ai, keeping it out of OpenCode/KiloCode dropdowns and `compatible_with?` checks

## Why not `DIRECT_OUTBOUND_API_PROVIDERS`?

Adding z.ai there would:
1. Show z.ai in OpenCode/KiloCode API provider dropdowns with broken `openai-compatible` defaults
2. Make `compatible_with?` assert z.ai keys work with opencode/kilocode, which is semantically false

This follows the existing pattern — e.g., Google is in `API_SERVICE_TYPES` but not `DIRECT_OUTBOUND_API_PROVIDERS`.

Closes #843 (partial — full adapter support tracked there)

### DIFF
--- a/spec/lib/provider_support_spec.rb
+++ b/spec/lib/provider_support_spec.rb
@@ -115,7 +115,9 @@ RSpec.describe ProviderSupport do
         "anthropic" => "Anthropic",
         "openai" => "OpenAI",
         "openrouter" => "OpenRouter",
-        "google" => "Google AI"
+        "google" => "Google AI",
+        "zai" => "z.ai",
+        "zai_coding" => "z.ai (Coding Plan)"
       )
     end
   end


viamin:
## Agent Update

Implemented.

I addressed the review thread in `lib/provider_support.rb` by adding the regression spec requested in `spec/lib/provider_support_spec.rb`:

1. In the `API_SERVICE_TYPES` example, I now assert inclusion of:
   - `"zai" => "z.ai"`
   - `"zai_coding" => "z.ai (Coding Plan)"`

This locks the behavior so future refactors won’t drop those labels.

Committed change:
- `dcfe5498` — `test(providers): lock z.ai service types in API_SERVICE_TYPES`

Validation status:
- `bundle exec rspec` passed in the commit hook with the existing known environment caveat that Qdrant-related checks remain pending in this environment.


Note: `bin/rails db:prepare` could not complete due to unavailable PostgreSQL socket in the current environment (`PG::ConnectionBad`, no running Postgres at expected local socket).

viamin (commented):
I reviewed the two-file change for PR #844. I found no functional, security, performance, or best-practice issues in this patch. It scopes `ProviderSupport::API_SERVICE_TYPES` and its spec updates to registering `zai` and `zai_coding` as API key service types, while explicitly leaving `DIRECT_OUTBOUND_API_PROVIDERS` untouched as stated. This aligns with issue #843’s partial scope (service-type registration only).

Issue #738, opened 2026-04-01
feat(performance): Phase 3.5 — Performance Optimization


viamin:
## Description

**Phase 3.5** — Handle more projects and agents efficiently.

**Objective**: Optimize system performance across container startup, workflow execution, database queries, and external API usage. Establish performance benchmarking to detect regressions. Some groundwork exists (connection pool optimization, container metrics collection).

**Deliverables**:
- Faster container startup via pool warming
- Higher throughput via batching and query optimization
- Clear performance metrics and benchmarks
- Tuning recommendations documented

## Sub-issues

- [ ] #718 — Implement container pool warming
- [ ] #719 — Workflow batching optimizations
- [ ] #720 — Database query optimization
- [ ] #721 — Add caching layer for GitHub data
- [ ] #722 — Worker pool tuning and optimization
- [ ] #723 — Establish performance benchmarking suite

🤖 Generated with [Claude Code](https://claude.com/claude-code)

viamin:
Pull request created: https://github.com/viamin/paid/pull/1720

Issue #719, opened 2026-04-01
feat(performance): workflow batching optimizations


viamin:
## Description

Optimize Temporal workflow execution by batching related operations (e.g., batch GitHub API calls, batch container operations) to reduce overhead.

## Tasks

- [ ] Identify batchable operations in current workflows
- [ ] Implement batched GitHub API calls (issues, PRs, labels)
- [ ] Batch container provisioning for multi-agent runs
- [ ] Reduce workflow signal overhead with batched status updates
- [ ] Benchmark before/after for key workflow paths
- [ ] Write specs

🤖 Generated with [Claude Code](https://claude.com/claude-code)

viamin:
Pull request created: https://github.com/viamin/paid/pull/1208

PR #884, opened 2026-04-07
Fix #869: perf(github-sync): use incremental issue fetching with since parameter

### LINKED ISSUE #869: perf(github-sync): use incremental issue fetching with since parameter
## Problem

`FetchIssuesActivity#fetch_all_issues` fetches **all** open issues every poll cycle (up to 10 pages of 100 = 1000 issues). Most repos have few changes between cycles, so this wastes up to 10 API calls fetching the same data repeatedly.

## Current behavior

- Every cycle: paginates through all open issues (up to 10 pages)
- For a repo with 500 open issues, that's 5 API calls every ~30 seconds
- Combined with the polling loop, this is ~10 API calls/minute just for issue listing

## Desired behavior

Use GitHub's `since` parameter on the issues API to only fetch issues updated since the last successful sync.

### Implementation notes

- Store `last_issue_sync_at` per project (could be a column on `projects` or a project setting)
- Pass `since: project.last_issue_sync_at` to the issues API in `fetch_issues_for_label`
- Fall back to full fetch if `last_issue_sync_at` is nil (first sync)
- Note: GitHub's `since` parameter filters by `updated_at`, which is exactly what we need
- Keep the stale-closure logic but handle the `truncated` flag appropriately for incremental fetches
- Estimated savings: ~90% of issue listing API calls (10 pages → 0-1 pages)

## Impact

**High** — For a repo with 500 open issues, reduces ~5 API calls/cycle to ~1 (or 0 if nothing changed).

Related to #719, #738.

viamin:
## Summary

Reduces GitHub API usage by ~90% for issue listing by switching to incremental fetching. Instead of paginating through all open issues every poll cycle, the sync now tracks `last_issue_sync_at` per project and uses GitHub's `since` parameter to only fetch issues updated since the last successful sync.

Related to #719, #738.

## Changes

**Database**
- Added `last_issue_sync_at` datetime column to `projects` table

**Models**
- Added `Project#touch_last_issue_sync_at` to record the timestamp of a successful sync

**Activities (`FetchIssuesActivity`)**
- `execute` reads `project.last_issue_sync_at`, captures a `sync_started_at` before fetching, and writes it back on success
- `fetch_issues_for_label` accepts a `since:` kwarg, passing the ISO 8601 timestamp to GitHub's issues API
- Incremental fetches use `state: "all"` instead of `state: "open"` so that newly-closed issues are captured and synced
- `close_stale_issues` skips bulk stale-closure during incremental fetches — individual state changes are handled by `sync_issue` instead
- Falls back to a full fetch when `last_issue_sync_at` is nil (first sync)

**Tests**
- 6 new specs covering: `since` parameter passing, `state: "all"` for incremental mode, stale-closure skip, closed issue syncing, first-sync fallback, and timestamp recording

## Design Decisions

- **`sync_started_at` recorded before fetch, written after success** — avoids a gap where issues updated mid-fetch could be missed on the next cycle, while ensuring a failed sync doesn't advance the watermark.
- **`state: "all"` during incremental fetches** — GitHub's `since` parameter filters by `updated_at`, so issues closed between cycles would be invisible if we only fetched `state: "open"`. Switching to `"all"` lets `sync_issue` handle state transitions individually.
- **Skipping bulk stale-closure for incremental fetches** — bulk closure assumes the fetch result is a complete snapshot of open issues. Since incremental results are a partial set, stale-closure is deferred to per-issue sync logic to avoid incorrectly closing issues that simply weren't returned.

---

Closes #869


viamin:
## Agent Update

The background setup agent finished — its work is already superseded by the setup and fixes I completed. The commit is done and all checks passed.

viamin:
## Agent Update

All done. Here's what was completed:

1. **Merge conflicts resolved** — Merged `origin/main` cleanly (no conflicts this time; schema.rb auto-merged).

2. **P1 code review addressed** — The reviewer correctly identified that setting the truncated-incremental watermark to the exact max `updated_at` creates an exclusive boundary, since GitHub's `since` parameter means "updated **after**" that timestamp. Issues sharing the same second-level `updated_at` at the page-cap boundary could be permanently skipped.

   **Fix**: Subtract 1 second from `latest_updated` before persisting the watermark (`latest_updated - 1.second`). This ensures the boundary is inclusive on the next poll. Re-fetching overlapping issues is safe because `sync_issue` uses `find_or_initialize_by` + `update!` (idempotent). The test was updated to match.

3. **All checks pass** — 40 specs, 0 failures. RuboCop clean.

app/temporal/activities/fetch_issues_activity.rb
```diff
@@ -14,21 +14,80 @@ def execute(input)
...
+      if incremental
+        fetched_ids = open_issues.map { |si| si[:id] }.to_set
+        rescannable = project.issues
+          .where(github_state: "open", paid_state: %w[new needs_input recommend_close])
+          .where.not(id: fetched_ids.to_a)
+          .pluck(:id, :github_number, :github_state)
+
+        rescannable.each do |id, github_number, github_state|
+          open_issues << { id: id, github_number: github_number, labels: [],
+                           github_state: github_state, trusted: true, rescan: true }
```
viamin: Nit: `trusted: true` is hardcoded for all rescan entries. I verified that `DetectLabelsActivity` reads trust from the DB (`issue.trusted?`) rather than from this hash, and the workflow only uses `:id` from these entries — so this is **not a security issue today**. However, it's a latent footgun: if any future consumer of this hash trusts the `:trusted` field at face value, untrusted issues in re-scannable states would bypass the check.

Consider either:
- Omitting `:trusted` from rescan entries entirely (since it's unused downstream), or
- Looking up the actual trust status:

```suggestion
          open_issues << { id: id, github_number: github_number, labels: [],
                           github_state: github_state, rescan: true }
```

app/temporal/activities/fetch_issues_activity.rb
```diff
@@ -76,14 +151,21 @@ def fetch_issues_for_label(client, repo_full_name, label, sort: nil)
...
-            label: label,
-            fetched_count: issues.size,
-            max_pages: DEFAULT_MAX_PAGES
-          )
+          # Probe the next page to distinguish a genuinely truncated result set
+          # from one that exactly fills DEFAULT_MAX_PAGES * DEFAULT_PER_PAGE.
+          # Without this check a false truncation prevents the watermark from
+          # ever advancing, causing the same window to be re-fetched indefinitely.
+          probe_opts = opts.merge(page: page)
+          if client.issues(repo_full_name, **probe_opts).any?
```
viamin: The probe request is a smart way to avoid false truncation, but it does add one extra API call every time the page cap is reached (which is the exact scenario where you're already close to rate limits). Worth noting this trade-off in case repos frequently hit the cap.

An alternative would be to only probe during incremental fetches (where false truncation actually matters for watermark advancement) and skip the probe for full fetches (where false truncation just means stale-closure is conservatively skipped — already safe):

```ruby
if page > DEFAULT_MAX_PAGES
  if since
    probe_opts = opts.merge(page: page)
    truncated = client.issues(repo_full_name, **probe_opts).any?
  else
    truncated = true
  end
  # ... logging ...
  break
end
```

Not blocking — the current approach is correct, just slightly more API-hungry than necessary for full fetches.

app/temporal/activities/fetch_issues_activity.rb
```diff
@@ -267,16 +350,24 @@ def fetch_trusted_comment_bodies(client, project, issue)
...
-      if truncated
+      #
+      # Incremental fetches (with `since`) only return recently-updated issues,
+      # not all open issues, so the result set is not exhaustive. However, any
+      # issue whose state changed to closed on GitHub will appear in the
+      # incremental results (because closing updates `updated_at`), so
+      # sync_issue will set its github_state to "closed" via the normal path.
+      # Skip the bulk stale-closure pass to avoid false positives.
+      if truncated || incremental
+        reason = truncated ? "fetch_truncated" : "incremental_fetch"
```
viamin, line 383: Logging at `warn` level for `reason: "incremental_fetch"` seems too noisy — skipping stale closure during incremental fetches is expected, normal behavior, not a warning condition. Consider `logger.info` for the incremental case and keep `warn` only for the truncated case:

```suggestion
      if truncated || incremental
        level = truncated ? :warn : :info
        logger.public_send(level,
```

Or split the two cases entirely.

app/temporal/activities/fetch_issues_activity.rb
```diff
@@ -14,21 +14,80 @@ def execute(input)
...
+      # on GitHub are not returned by the API. However, those issues may still
+      # need re-evaluation by DetectLabelsActivity — for example when a
+      # blocking dependency was resolved, or project label/trust settings
+      # changed. Append locally-open issues in re-scannable states that were
+      # not already part of this fetch so the workflow passes them downstream.
+      if incremental
+        fetched_ids = open_issues.map { |si| si[:id] }.to_set
+        rescannable = project.issues
+          .where(github_state: "open", paid_state: %w[new needs_input recommend_close])
+          .where.not(id: fetched_ids.to_a)
```
viamin, line 76: Minor: `where.not(id: fetched_ids.to_a)` generates a `NOT IN (...)` clause. For typical incremental fetches this array is small, but if a large batch is returned, this could become a large SQL clause. Not a real concern for current page caps (max 1000 issues), just noting it.

viamin (commented):
## Review: Incremental Issue Fetching

Well-designed incremental sync with solid watermark advancement logic and good edge-case handling (stuck-time escape hatch, probe for false truncation). The PR clearly addresses #869 and the test coverage is thorough.

A few observations below — one potential concern around the rescan trust field, one log-level nit, and a note about the extra API call from the probe.

viamin:
## Agent Update

The background task notification is for the earlier failed commit attempt (the one that ran the full test suite). It's already been superseded — the commit succeeded on the subsequent attempt (`9510dd1f`). Working tree is clean, nothing further needed.

viamin (commented):
## Review: Incremental issue fetching with `since` parameter

This is a well-designed optimization that should significantly reduce GitHub API usage. The watermark logic, ascending sort for forward progress, and probe-page optimization are all thoughtful. The test coverage is thorough.

A few items to consider below — mostly minor, one potential performance concern with the rescan query.

app/temporal/activities/fetch_issues_activity.rb
```diff
@@ -14,21 +14,80 @@ def execute(input)
...
+          .pluck(:id, :github_number, :github_state)
```
viamin, line 78: **Performance concern**: For projects with a large number of open issues in re-scannable states (`new`, `needs_input`, `recommend_close`), this query could return a very large result set every poll cycle. Each of those issues then triggers a `DetectLabelsActivity` execution downstream (line 35-36 in `git_hub_poll_workflow.rb`), which is a DB read + potential write per issue.

Consider adding a limit here (e.g., process at most N rescan candidates per cycle) or adding an index on `(project_id, github_state, paid_state)` if one doesn't already exist. Without a cap, a project with 500 open `new` issues would re-evaluate all 500 every single poll cycle, partially negating the API savings from incremental fetching.

```suggestion
        rescannable = project.issues
          .where(github_state: "open", paid_state: %w[new needs_input recommend_close])
          .where.not(id: fetched_ids.to_a)
          .limit(200)
          .pluck(:id, :github_number, :github_state)
```

app/temporal/activities/fetch_issues_activity.rb
```diff
@@ -14,21 +14,80 @@ def execute(input)
...
+        #
+        # However, if subtracting 1 second would not advance past the
+        # current watermark (i.e. all fetched issues share the same
+        # `updated_at` second as the watermark), use the exact timestamp
+        # to guarantee forward progress. Some same-second issues may be
+        # skipped, but permanent re-fetch of the same window is worse.
+        latest_updated = github_issues.filter_map { |gi| gi.updated_at }.max
+        if latest_updated
+          inclusive_cursor = latest_updated - 1.second
+          new_watermark = if project.last_issue_sync_at && inclusive_cursor <= project.last_issue_sync_at
```
viamin: Nit: The condition `project.last_issue_sync_at && inclusive_cursor <= project.last_issue_sync_at` — the `project.last_issue_sync_at` nil-guard is redundant here since we're inside `if truncated && incremental`, and `incremental` is only true when `last_issue_sync_at.present?`. Not a bug, just unnecessary.

app/temporal/activities/fetch_issues_activity.rb
```diff
@@ -76,14 +151,31 @@ def fetch_issues_for_label(client, repo_full_name, label, sort: nil)
...
+          # Without this check a false truncation prevents the watermark from
+          # ever advancing, causing the same window to be re-fetched indefinitely.
+          #
+          # Only probe during incremental fetches where false truncation matters
+          # for watermark advancement. For full fetches, false truncation just
+          # means stale-closure is conservatively skipped — already safe — so
+          # skip the extra API call to avoid unnecessary rate-limit pressure.
+          if since
+            probe_opts = opts.merge(page: page)
+            truncated = client.issues(repo_full_name, **probe_opts).any?
```
viamin, line 172: The probe page reuses `opts` which includes the current `page` value from the last iteration. The `opts.merge(page: page)` correctly overwrites, but `page` was already incremented to `DEFAULT_MAX_PAGES + 1` on line 151 before entering this block. So `probe_opts` gets the right page number. This works correctly but is a bit subtle — a brief inline comment noting that `page` was already incremented would help readability.

app/temporal/activities/fetch_issues_activity.rb
```diff
@@ -14,21 +14,80 @@ def execute(input)
...
+          project.touch_last_issue_sync_at(new_watermark)
+        end
+      elsif !truncated
+        project.touch_last_issue_sync_at(sync_started_at)
+      end
+
+      # Exclude closed issues from downstream processing (DetectLabelsActivity).
+      # sync_issue already persisted their github_state to the DB, but passing
+      # them downstream could incorrectly trigger agent runs for closed work.
+      open_issues = synced_issues.reject { |si| si[:github_state] == "closed" }
```
viamin, line 64: Minor: `synced_issues` is built from `github_issues.map { |gi| sync_issue(project, gi) }` which persists closed issues to the DB (good). But then `parse_issue_relationships` on line 23 is called with `synced_issues` *before* closed issues are filtered out here. `parse_issue_relationships` does filter to `github_state: "open"` internally (line 227), so this is safe — but it means closed issue IDs are passed to the `where(id: synced_issue_ids)` query only to be filtered out. Not a bug, just noting the data flow.

app/temporal/activities/fetch_issues_activity.rb
```diff
@@ -14,21 +14,80 @@ def execute(input)
...
+      end
 
       logger.info(
         message: "github_sync.fetch_issues",
         project_id: project.id,
         issue_count: synced_issues.size,
-        closed_count: closed_count
+        closed_count: closed_count,
+        incremental: incremental,
+        rescan_count: incremental ? open_issues.count { |si| si[:rescan] } : 0
```
viamin, line 92: Nice addition of `incremental` and `rescan_count` to structured logging — this will be very helpful for monitoring the effectiveness of the optimization.

app/temporal/activities/fetch_issues_activity.rb
```diff
@@ -267,16 +360,25 @@ def fetch_trusted_comment_bodies(client, project, issue)
...
-        logger.warn(
+      #
+      # Incremental fetches (with `since`) only return recently-updated issues,
+      # not all open issues, so the result set is not exhaustive. However, any
+      # issue whose state changed to closed on GitHub will appear in the
+      # incremental results (because closing updates `updated_at`), so
+      # sync_issue will set its github_state to "closed" via the normal path.
+      # Skip the bulk stale-closure pass to avoid false positives.
+      if truncated || incremental
+        reason = truncated ? "fetch_truncated" : "incremental_fetch"
```
viamin, line 383: Good design: when both `truncated` and `incremental` are true, `truncated` takes priority in the reason string. This is correct since truncation is the more operationally relevant signal. The log level distinction (`warn` for truncated, `info` for incremental) is also a nice touch.

db/migrate/20260407072915_add_last_issue_sync_at_to_projects.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddLastIssueSyncAtToProjects < ActiveRecord::Migration[8.1]
+  def change
+    add_column :projects, :last_issue_sync_at, :datetime
```
viamin, line 5: The column defaults to `nil` which correctly triggers a full fetch on first sync. If you anticipate needing to query projects by sync recency (e.g., for monitoring dashboards), consider adding an index on this column. Not blocking — can be added later if needed.

viamin:
## Agent Update

Commit succeeded with all lint checks passing. Here's a summary of what was addressed:

**Review comments addressed:**

1. **Performance concern (line 73)** - Added `.limit(200)` to the rescannable issues query to cap the number of re-evaluation candidates per cycle, preventing unbounded downstream `DetectLabelsActivity` executions for projects with many open issues in re-scannable states.

2. **Redundant nil-guard (line 46)** - Removed the unnecessary `project.last_issue_sync_at &&` guard since we're already inside an `if incremental` block, and `incremental` is only true when `last_issue_sync_at.present?`. Added an inline comment explaining why.

3. **Probe page readability (line 165)** - Added an inline comment clarifying that `page` was already incremented past `DEFAULT_MAX_PAGES` before entering the probe block.

4. **Closed issue data flow (line 60)** - Added a comment noting that `parse_issue_relationships` receives all synced issues including closed ones, but filters to `github_state: "open"` internally.

5. **Positive feedback threads (lines 87, 376)** - No changes needed; acknowledged.

6. **Migration index suggestion** - Non-blocking; can be added later if needed for monitoring dashboards.

viamin (commented):
Generated no new comments. The incremental fetching implementation is solid — the ascending sort for forward progress, probe page to avoid false truncation, 1-second watermark buffer with stuck-progress fallback, rescannable issue inclusion, and closed-issue filtering are all well-reasoned. Tests cover the key edge cases comprehensively. The PR cleanly addresses #869.

viamin:
## Agent Update

Created commit `684ab87b` on local branch `fix/884-followup`.

The change in [app/temporal/activities/fetch_issues_activity.rb](/workspace/app/temporal/activities/fetch_issues_activity.rb) skips the local re-scan fallback when an incremental fetch is truncated, so we don’t pass potentially stale locally-open issues downstream before the remaining GitHub backlog has been synced. I added a regression spec in [spec/temporal/activities/fetch_issues_activity_spec.rb](/workspace/spec/temporal/activities/fetch_issues_activity_spec.rb) covering that truncated-incremental case.

Validation passed through the repo hook during commit:
`bin/lint --staged` passed, and `bundle exec rspec` passed with `5639 examples, 0 failures, 2 pending`.

I did not push. The original PR checkout at `/workspace` was already dirty with unrelated staged changes, so I isolated this fix in a clean worktree and committed it there to avoid mixing work. If you want it on the existing PR branch, cherry-pick `684ab87b` onto `paid/869-perfgithub-sync-use-incremental-issue-fetching-wit-e6dfee`.

db/schema.rb
```diff
@@ -201,8 +201,8 @@
...
     t.index ["parent_workflow_id"], name: "index_agent_runs_on_parent_workflow_id"
     t.index ["project_id", "goal"], name: "index_agent_runs_on_project_id_and_goal"
-    t.index ["project_id", "issue_id"], name: "idx_agent_runs_unique_active_issue", unique: true, where: "((issue_id IS NOT NULL) AND ((status)::text = ANY (ARRAY[('queued'::character varying)::text, ('pending'::character varying)::text, ('running'::character varying)::text, ('paused'::character varying)::text])))"
-    t.index ["project_id", "source_pull_request_number"], name: "idx_agent_runs_unique_active_pr", unique: true, where: "((source_pull_request_number IS NOT NULL) AND ((status)::text = ANY (ARRAY[('queued'::character varying)::text, ('pending'::character varying)::text, ('running'::character varying)::text, ('paused'::character varying)::text])))"
+    t.index ["project_id", "issue_id", "goal"], name: "idx_agent_runs_unique_active_issue", unique: true, where: "((issue_id IS NOT NULL) AND ((status)::text = ANY ((ARRAY['queued'::character varying, 'pending'::character varying, 'running'::character varying, 'paused'::character varying])::text[])))"
```
viamin: These index definitions changed from `[project_id, issue_id]` / `[project_id, source_pull_request_number]` to include `goal`, but this branch does not include any migration that makes that schema change. Please regenerate `db/schema.rb` from this PR's migrations only so the diff stays scoped to `last_issue_sync_at`; otherwise we would be committing schema drift that doesn't match the branch's actual database changes.

viamin (commented):
This PR looks directionally correct for #869, but it also includes unrelated schema drift that should be removed before merge.

viamin:
## Agent Update

The background task notification is for the first failed commit attempt (before assets were built). The subsequent commit succeeded — `2d984825` is already in place with all 5765 tests passing and lint clean. Nothing further needed.

viamin (commented):
Generated no new comments. The incremental fetching implementation is solid — the watermark advancement logic correctly handles truncation, the 1-second inclusive boundary with forward-progress fallback is well-reasoned, the probe page optimization avoids false truncation without unnecessary API calls on full fetches, and the rescan fallback for locally-open issues fills the gap left by incremental results. Test coverage is thorough across the key scenarios.

PR #888, opened 2026-04-07
Fix #867: perf(pr-scanner): skip unchanged PRs in ScanPaidPrsActivity

### LINKED ISSUE #867: perf(pr-scanner): skip unchanged PRs in ScanPaidPrsActivity
## Problem

`ScanPaidPrsActivity` scans every open paid-generated PR on every poll cycle, making 13-15 API calls per PR. Most PRs haven't changed since the last scan, so this wastes rate limit budget.

## Current behavior

- Iterates all paid PRs and calls `scan_pr` for each, which fetches PR data, check runs, review threads, reviews, and comments
- No mechanism to skip PRs that haven't changed since the last scan
- 20 open PRs × ~15 API calls = ~300 API calls per poll cycle

## Desired behavior

Track `last_scanned_at` per issue and skip PRs where `github_updated_at < last_scanned_at` and no active agent run completed since the last scan.

### Implementation notes

- Add a `last_pr_scan_at` timestamp column to `Issue` (or a separate tracking table)
- Before scanning a PR, check if it has changed since last scan
- Always scan PRs that have an active or recently-completed agent run
- This complements the lazy-evaluation pattern already in `scan_draft_pr`

## Impact

**High** — For a project with 20 PRs where only 2 have changed, reduces ~300 API calls to ~30.

Related to #719, #738.

viamin:
## Summary

Reduce GitHub API usage in PR scanning by skipping PRs that haven't changed since the last scan. For a typical project with 20 open PRs where only 2 have activity, this cuts API calls from ~300 to ~30 per poll cycle.

## Changes

**Database**
- Add `last_pr_scan_at` datetime column to `issues` table to track when each PR was last scanned

**Temporal Activities (`ScanPaidPrsActivity`)**
- Add `skip_unchanged_pr?` predicate that short-circuits scanning when the PR's `github_updated_at` predates `last_pr_scan_at` and no agent run has completed since the last scan
- Add `recently_completed_run?` helper to detect agent runs that finished after the last scan, ensuring those PRs are always re-scanned
- Update `last_pr_scan_at` after each successful scan
- Enhance structured logging with `prs_found`, `prs_scanned`, and `prs_skipped` counts for observability

**Tests**
- 6 new specs covering: skip logic, first-scan passthrough, updated PR detection, recently-completed agent run bypass, timestamp updates, and log output

## Design Decisions

- **Column on `issues` rather than a separate tracking table** — keeps the schema simple since the relationship is 1:1 with issue and avoids an extra join on every poll cycle.
- **Agent run completion as a bypass signal** — a PR may need re-scanning even if GitHub's `updated_at` hasn't changed (e.g., agent run posted results that haven't been reflected yet). Checking for recently-completed runs ensures we don't miss state transitions.
- **Complements existing lazy-evaluation in `scan_draft_pr`** — this is an outer-loop optimization that prevents entering the per-PR scan path entirely, while `scan_draft_pr` already avoids redundant work within a scan.

---

Closes #867


viamin:
## Agent Update

All done. Commit `49bd17ad` is in place with both review comments addressed. No uncommitted changes remain.

viamin:
## Agent Update

Done. Here's a summary of what was accomplished:

**Merge conflicts resolved:** `db/schema.rb` version conflict resolved by keeping the later migration timestamp from our branch, then regenerated from the actual DB state.

**Code review feedback addressed (P1):** The reviewer identified that `detect_ready_triggers` could silently return empty triggers when API calls for reviews or unresolved threads failed, causing `last_pr_scan_at` to be stamped despite an incomplete scan. Fixed by:

1. `detect_ready_triggers` now returns `nil` (not `[]`) when `fetch_reviews` or `fetch_unresolved_threads` return `nil` (API failure)
2. Both `scan_ready_pr` and `scan_escalated_pr` return `:skipped` when `detect_ready_triggers` returns `nil`
3. The existing guard in `execute` (`next nil if result == :skipped`) prevents `last_pr_scan_at` from being stamped

**New test added:** "does not update last_pr_scan_at when reviews fail in ready phase" — verifies the ready-phase API failure path.

**All 5581 tests pass, 0 lint offenses.**

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -288,13 +313,25 @@ def draft_trigger_payload(issue, triggers)
...
+    # return nil to prevent a false "all clear".
     def detect_ready_triggers(project, client, issue, pr_data: nil, checks: nil, reviews: nil,
       unresolved_threads: nil)
       last_run = last_completed_run(project, issue)
       pr_data ||= fetch_pr_data(client, project, issue)
       checks ||= fetch_check_runs(client, project, pr_data)
       reviews ||= fetch_reviews(client, project, issue)
       unresolved_threads ||= fetch_unresolved_threads(client, project, issue)
+
+      partial_failure = reviews.nil? || unresolved_threads.nil?
```
viamin: `partial_failure` tracks `reviews` and `unresolved_threads` but not `pr_data`. When `fetch_pr_data` fails inside `detect_ready_triggers` (line 328), `pr_data` is nil, `fetch_check_runs` returns `[]` (not nil), and both `ci_failure_triggers([])` and `check_merge_conflicts(project, nil)` silently return `[]`. Since `partial_failure` is false, the method returns an empty array instead of nil, and the caller stamps `last_pr_scan_at` — masking an incomplete scan.

This matters for `scan_escalated_pr` which passes `pr_data: nil` as default and relies on `detect_ready_triggers` to fetch it.

```suggestion
      partial_failure = pr_data.nil? || reviews.nil? || unresolved_threads.nil?
```

viamin (commented):
## Review: perf(pr-scanner): skip unchanged PRs in ScanPaidPrsActivity

Solid optimization — the skip logic is well-designed with appropriate bypass signals (first scan, updated PRs, recently completed runs), and the `:skipped` return value semantics correctly prevent stamping `last_pr_scan_at` on incomplete scans. Test coverage is thorough.

A few items worth addressing below, ranging from a potential false-stamp gap in `detect_ready_triggers` to a minor metrics accuracy issue.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -37,12 +37,33 @@ def execute(input)
...
       paid_prs = find_paid_prs(project)
 
-      prs_to_trigger = paid_prs.filter_map { |issue| scan_pr(project, client, issue) }
+      scanned_count = 0
+      prs_to_trigger = paid_prs.filter_map do |issue|
+        if skip_unchanged_pr?(project, issue)
+          next nil
+        end
+
+        scanned_count += 1
```
viamin: Nit: `scanned_count` is incremented before `scan_pr` runs, so PRs short-circuited by `active_run_exists?` (which returns `:skipped`) are counted as "scanned" in the log output. This means `prs_skipped` only reflects the `skip_unchanged_pr?` gate, not all skips.

This is arguably fine — "scanned" means "entered `scan_pr`" — but if the intent is to track PRs that received a *full* evaluation, consider incrementing after the `:skipped` check instead:

```suggestion
        result = scan_pr(project, client, issue)
        # Only mark as scanned when the scan actually completed — scan_pr
        # returns :skipped when short-circuited (active run exists) or when
        # API failures prevented full evaluation.
        next nil if result == :skipped
        scanned_count += 1
```

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -331,6 +375,32 @@ def maybe_restart_draft(project, issue, pr_data)
       true
     end
 
+    def skip_unchanged_pr?(project, issue)
+      return false unless issue.last_pr_scan_at
+      return false if issue.github_updated_at >= issue.last_pr_scan_at
```
viamin, line 395: Minor edge case: if `github_updated_at` and `last_pr_scan_at` are identical (e.g., a scan happens at the exact same second as the last GitHub update), `>=` correctly allows the scan through. Good.

However, consider that `github_updated_at` reflects GitHub's server timestamp while `last_pr_scan_at` uses `Time.current` from the Rails app. Clock skew between the two could cause a PR updated *just before* the scan to appear "unchanged" on the next cycle if the GitHub clock is slightly behind. This is a very narrow window and probably not worth adding buffer for, but worth being aware of.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -37,12 +37,33 @@ def execute(input)
...
+        # Only mark as scanned when the scan actually completed — scan_pr
+        # returns :skipped when short-circuited (active run exists) or when
+        # API failures prevented full evaluation.
+        next nil if result == :skipped
+        # Only stamp last_pr_scan_at when no triggers were emitted. When
+        # triggers fire, downstream actions (queueing runs, requesting
+        # reviews) may still fail; stamping now would cause
+        # skip_unchanged_pr? to suppress future scans before those actions
+        # are retried.
+        issue.update_column(:last_pr_scan_at, Time.current) if result.nil?
```
viamin: The decision to only stamp `last_pr_scan_at` when `result.nil?` (no triggers) is well-reasoned — it avoids suppressing retries when downstream trigger handling fails. The comment explains this clearly. Nice design choice.

db/migrate/20260407081149_add_last_pr_scan_at_to_issues.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddLastPrScanAtToIssues < ActiveRecord::Migration[8.1]
+  def change
+    add_column :issues, :last_pr_scan_at, :datetime
```
viamin, line 5: No index needed since this column is only read on already-loaded issue records, not used in WHERE clauses for querying the issues table. Confirmed this is correct.

viamin:
## Agent Update

Done. Here's a summary of what was accomplished:

**Merge conflicts resolved:** Merged `origin/main` cleanly and restored schema.rb to main's formatting with only the +1 line diff for `last_pr_scan_at`.

**Code review comments addressed:**

1. **P1 — `partial_failure` missing `pr_data` check** (line 333): Added `pr_data.nil?` to the `partial_failure` predicate in `detect_ready_triggers`, so a failed `fetch_pr_data` is no longer silently masked as "no triggers" — it correctly returns `nil` when no actionable triggers are found, preventing `last_pr_scan_at` from being stamped on an incomplete scan.

2. **Nit — `scanned_count` increment placement** (line 46): Moved `scanned_count += 1` to after the `:skipped` check, so `prs_scanned` now accurately reflects PRs that received a full evaluation (not PRs short-circuited by `active_run_exists?`).

3. **Clock skew comment** (line 380): Acknowledged — informational only, no code change needed.

4. **Positive feedback on `last_pr_scan_at` stamping design** (line 57): No change needed.

5. **No index needed confirmation** (migration): No change needed.

**Verification:** 5637 specs pass, 0 failures, 0 lint offenses.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -37,12 +37,33 @@ def execute(input)
...
+        # returns :skipped when short-circuited (active run exists) or when
+        # API failures prevented full evaluation.
+        next nil if result == :skipped
+        scanned_count += 1
+        # Only stamp last_pr_scan_at when no triggers were emitted. When
+        # triggers fire, downstream actions (queueing runs, requesting
+        # reviews) may still fail; stamping now would cause
+        # skip_unchanged_pr? to suppress future scans before those actions
+        # are retried.
+        issue.update_column(:last_pr_scan_at, Time.current) if result.nil?
```
viamin: **Potential race condition with `update_column` and `Time.current`**

Each PR gets its own `Time.current` call, so if the loop takes a few seconds, the `last_pr_scan_at` timestamps will differ slightly across PRs within the same scan cycle. This is fine functionally, but worth noting that `update_column` bypasses validations and callbacks. If any future code adds callbacks on `Issue` that should fire on scan-timestamp updates, this will silently skip them.

Also, `update_column` issues a separate SQL UPDATE per PR. For projects with many untriggered PRs, this could add up. Consider batching the updates after the loop if performance becomes a concern, though for the current scale (~20 PRs) this is perfectly fine.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -37,12 +37,33 @@ def execute(input)
...
+        result
+      end
 
       logger.info(
         message: "pr_scanner.scan_complete",
         project_id: project_id,
-        prs_scanned: paid_prs.size,
+        prs_found: paid_prs.size,
+        prs_scanned: scanned_count,
+        prs_skipped: paid_prs.size - scanned_count,
```
viamin: **`prs_skipped` count is slightly misleading**

The formula `paid_prs.size - scanned_count` conflates two different skip reasons: PRs skipped by `skip_unchanged_pr?` (the new optimization) and PRs where `scan_pr` returned `:skipped` (active run exists, API failure). For observability, it would be more useful to track them separately:

```suggestion
        prs_found: paid_prs.size,
        prs_scanned: scanned_count,
        prs_skipped_unchanged: paid_prs.size - scanned_count - prs_to_trigger.size,
```

Or track `unchanged_count` explicitly in the loop. This would help distinguish "we saved API calls" from "we had failures".

app/temporal/activities/scan_paid_prs_activity.rb
viamin, line 395: **`>=` comparison may skip PRs updated at the exact same instant they were last scanned**

If `github_updated_at` equals `last_pr_scan_at` exactly (same timestamp), this returns `false` and the PR is re-scanned. That's the safe direction, so this is correct. Just confirming this is intentional — the `>=` means "updated at or after last scan" triggers a rescan, which is the right behavior.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -331,6 +375,32 @@ def maybe_restart_draft(project, issue, pr_data)
...
+        github_updated_at: issue.github_updated_at
+      )
+
+      true
+    end
+
+    def recently_completed_run?(project, issue)
+      project.agent_runs
+        .where(
+          "source_pull_request_number = :pr_num OR pull_request_number = :pr_num",
```
viamin: **Consider using parameterized `where` with an array instead of string interpolation**

The query is safe from SQL injection since `github_number` is an integer, but the OR clause with named bind would be slightly more idiomatic as:

```suggestion
      project.agent_runs
        .where(source_pull_request_number: issue.github_number)
        .or(project.agent_runs.where(pull_request_number: issue.github_number))
        .where("completed_at >= ?", issue.last_pr_scan_at)
        .exists?
```

This uses ActiveRecord's `or` scope which is more aligned with Rails conventions, though the named-bind version is also fine.

db/migrate/20260407081149_add_last_pr_scan_at_to_issues.rb
viamin, line 5: **Missing index on `last_pr_scan_at`**

Per the project's CLAUDE.md: *"Index all foreign keys and frequently queried columns."* While `last_pr_scan_at` isn't a foreign key, `recently_completed_run?` queries `completed_at >= issue.last_pr_scan_at` on every PR in every scan cycle. The column itself isn't queried directly via `WHERE` on the issues table, so an index isn't strictly necessary — but if you ever add a scope like `Issue.where(last_pr_scan_at: nil)` to find never-scanned PRs, you'd want one.

No action needed now, just flagging for awareness.

spec/temporal/activities/scan_paid_prs_activity_spec.rb
```diff
@@ -2144,11 +2144,155 @@ def create_draft_run(status:, iterations:, created_at:, trigger_type: "automatic
...
             project_id: project.id,
             prs_scanned: 1,
+            prs_skipped: 0,
             prs_triggered: 0
           )
         )
       end
     end
+
+    context "when skipping unchanged PRs" do
```
viamin, line 2787: Good coverage overall — the specs cover first-scan passthrough, updated PR detection, recently-completed agent run bypass, timestamp updates, trigger suppression, active-run skipping, API failure handling, and log output.

One minor gap: there's no spec covering the `partial_failure` path in `detect_ready_triggers` — where some API calls fail but actionable triggers are still found from other sources. The `"still emits CI failure triggers when reviews fail in ready phase"` spec at line 2261 partially covers this, which is great. Consider also testing the inverse: when `partial_failure` is true and no triggers are found, `detect_ready_triggers` should return `nil` (not an empty array), which then maps to `:skipped` in the caller.

viamin (commented):
## PR Review: perf(pr-scanner): skip unchanged PRs in ScanPaidPrsActivity

Solid performance optimization — the design is well-thought-out with good edge-case handling around partial API failures and the agent-run bypass. A few issues worth addressing below, ranging from a potential race condition to a missing index on the new column.

Overall: **Approve with minor suggestions.**

viamin:
## Agent Update

The background task notification is from the earlier failed commit attempt (the one that ran the full test suite) — it can be ignored. The commit succeeded in the subsequent attempt.

To recap: commit `e0f1d07d` is in place with all three review feedback items addressed plus the new test. Working tree is clean.

viamin (commented):
Generated no new comments. Clean implementation — the skip-unchanged optimization is well-designed with proper guards for partial API failures, recently-completed agent runs, and the sentinel value distinction between `:skipped` / `nil` / trigger hashes. Tests cover the important edge cases. Closes #867 as described.

viamin:
## Agent Update

Done. Here's a summary of what was accomplished:

**Merge conflicts resolved:** Merged `origin/main` and resolved the single conflict in `db/schema.rb` — kept main's new check constraint additions (`chk_max_issues_per_page_bounds`, `chk_max_prs_per_page_bounds`) alongside this branch's `last_pr_scan_at` column.

**Conversation comments:** All prior review feedback (P1 issues around `last_pr_scan_at` stamping, partial API failures, `scanned_count` increment placement, `partial_failure` missing `pr_data` check) was already addressed in previous commits on this branch. No new actionable items were found in the comments.

**Verification:**
- Rubocop: 0 offenses
- Tests: 5760 examples, 0 failures
- Working tree: clean
- Commit: `a657c06f`

viamin (commented):
I found two blocking issues: the new `last_pr_scan_at` logic never stamps scans that emit triggers, so unchanged PRs with persistent CI/review signals still get fully rescanned every poll cycle; and `db/schema.rb` includes unrelated schema drift that is not backed by migrations in this PR.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -45,12 +45,35 @@ def execute(input)
...
+        # returns :skipped when short-circuited (active run exists) or when
+        # API failures prevented full evaluation.
+        next nil if result == :skipped
+        scanned_count += 1
+        # Only stamp last_pr_scan_at when no triggers were emitted. When
+        # triggers fire, downstream actions (queueing runs, requesting
+        # reviews) may still fail; stamping now would cause
+        # skip_unchanged_pr? to suppress future scans before those actions
+        # are retried.
+        issue.update_column(:last_pr_scan_at, Time.current) if result.nil?
```
viamin: This prevents `last_pr_scan_at` from ever advancing on the PRs that usually generate the most repeated work: unchanged PRs with persistent triggers such as failing CI or outstanding review feedback. Those PRs will still take the full GitHub API path on every poll, which defeats the rate-limit reduction in #867 for exactly the hot path we are trying to optimize. Please stamp the scan after any *completed* scan (anything except `:skipped`), and handle downstream retry concerns separately if needed.

```suggestion
issue.update_column(:last_pr_scan_at, Time.current)
```

db/schema.rb
```diff
@@ -201,8 +201,8 @@
...
     t.index ["parent_workflow_id"], name: "index_agent_runs_on_parent_workflow_id"
     t.index ["project_id", "goal"], name: "index_agent_runs_on_project_id_and_goal"
-    t.index ["project_id", "issue_id"], name: "idx_agent_runs_unique_active_issue", unique: true, where: "((issue_id IS NOT NULL) AND ((status)::text = ANY (ARRAY[('queued'::character varying)::text, ('pending'::character varying)::text, ('running'::character varying)::text, ('paused'::character varying)::text])))"
-    t.index ["project_id", "source_pull_request_number"], name: "idx_agent_runs_unique_active_pr", unique: true, where: "((source_pull_request_number IS NOT NULL) AND ((status)::text = ANY (ARRAY[('queued'::character varying)::text, ('pending'::character varying)::text, ('running'::character varying)::text, ('paused'::character varying)::text])))"
+    t.index ["project_id", "issue_id", "goal"], name: "idx_agent_runs_unique_active_issue", unique: true, where: "((issue_id IS NOT NULL) AND ((status)::text = ANY ((ARRAY['queued'::character varying, 'pending'::character varying, 'running'::character varying, 'paused'::character varying])::text[])))"
```
viamin: `db/schema.rb` is carrying unrelated changes here that are not backed by migrations in this PR. I only found the new `add_last_pr_scan_at_to_issues` migration; there is no migration changing these unique indexes to include `goal`, and the same schema diff also adds `workflow_states.restart_reason` without a migration. Merging this would leave `schema.rb` out of sync with the real migration history on a fresh database. Please regenerate `db/schema.rb` so this PR only includes the `last_pr_scan_at` change (or add the missing migrations if these changes are intentional).

viamin:
## Agent Update

That's the notification from the earlier failed commit attempt (before CSS assets were built). The subsequent commit succeeded — `1e2e9d0b` is in place with all changes, 5764 examples passing, 0 lint offenses. Working tree is clean.

viamin (commented):
Generated no new comments. The PR is well-structured: the skip-unchanged optimization is correctly implemented with appropriate safety bypasses (first scan, recently-completed agent runs, partial API failures), the `:skipped` sentinel cleanly separates incomplete scans from successful no-trigger results, and the test coverage is thorough.

PR #889, opened 2026-04-07
Fix #865: fix(github-client): add retry middleware to GraphQL connection

### LINKED ISSUE #865: fix(github-client): add retry middleware to GraphQL connection
## Problem

The `graphql_connection` Faraday instance (`app/services/github_client.rb:799-806`) has no retry middleware, unlike the REST client which has `Faraday::Retry::Middleware` configured for 429/5xx errors.

## Current behavior

- REST API calls: retry up to 3 times on 429/500/502/503/504 with exponential backoff
- GraphQL calls: no retry — transient failures immediately raise `ApiError`
- GraphQL is used for `review_threads`, `resolve_review_thread`, `mark_pull_request_ready`, `request_bot_review`, `pull_request_node_id`

## Desired behavior

Add the same retry middleware to the GraphQL connection.

### Implementation notes

- Add `Faraday::Retry::Middleware` to `graphql_connection` with the same configuration as the REST client
- Handle GraphQL-specific rate limiting (GitHub returns a `rateLimit` field in responses, but HTTP 429 is also possible)

## Impact

**Moderate** — Prevents transient GraphQL failures from causing unnecessary retries at the Temporal workflow level.

Related to #738.

viamin:
## Summary

Adds retry middleware to the GitHub GraphQL Faraday connection so transient failures (429/5xx) are retried with exponential backoff, matching the existing REST client behavior. This prevents unnecessary escalation to Temporal workflow-level retries for recoverable errors.

## Changes

- **Services (`app/services/github_client.rb`)**: Added `Faraday::Retry::Middleware` to `graphql_connection` with the same configuration as the REST client — max 3 retries, exponential backoff (0.5s base, 2x factor, 0.5 randomness), retrying on 429/500/502/503/504. Includes structured logging on each retry using the `github_client.graphql_retry` component.

## Design Decisions

- **Parity with REST client**: Uses identical retry parameters rather than introducing GraphQL-specific tuning. This keeps behavior predictable and simplifies future changes to retry policy.
- **HTTP-level retry only**: Retries are triggered by HTTP status codes (429/5xx), not by inspecting GraphQL response bodies for `rateLimit` fields. This is sufficient because GitHub returns HTTP 429 for rate limiting and HTTP 5xx for server errors, even on the GraphQL endpoint.

---

Closes #865

### DIFF
--- a/app/services/github_client.rb
+++ b/app/services/github_client.rb
@@ -815,7 +815,11 @@ class GithubClient
   end
 
   def graphql_request(query, **variables)
-    response = graphql_connection.post("/graphql") do |req|
+    # Mutations are not idempotent — retrying them after a transient failure
+    # (where the server may have applied the mutation before returning 5xx)
+    # can cause duplicate side effects. Only retry read-only queries.
+    conn = graphql_mutation?(query) ? graphql_connection_base : graphql_connection
+    response = conn.post("/graphql") do |req|
       req.headers["Authorization"] = "token #{client.access_token}"
       req.body = { query: query, variables: variables }
     end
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

**Code review fix:** The reviewer correctly identified that retrying GraphQL mutations is unsafe — if the server applies a mutation but returns a transient 5xx, the retry would replay it, causing duplicate side effects.

**Changes:**
- `app/services/github_client.rb`: Split the GraphQL connection into two: one with retry middleware (for read-only queries) and one without (for mutations). Added `graphql_mutation?` to detect mutations by checking if the query starts with `"mutation"`. The `graphql_request` method now selects the appropriate connection automatically.
- `spec/services/github_client_spec.rb`: Added a test verifying that mutations (e.g., `resolve_review_thread`) are only called once on a 503, confirming no retry occurs.

Issue #879, opened 2026-04-07
perf(github-api): reduce GitHub API calls per polling cycle


viamin:
## Problem

Paid's polling workflow (`GitHubPollWorkflow`) makes excessive GitHub API calls per cycle, causing rate limit exhaustion that blocks agent runs. A single poll cycle for a project with 20 open PRs and 50 issues can consume **200-300+ API calls**. At a ~30-second poll interval, this exhausts GitHub's 5,000-request/hour rate limit quickly, causing agent run failures (e.g., agent run #5649).

### Root causes

1. **N+1 patterns**: Comment fetching per issue, PR scanning per PR, reaction fetching per comment
2. **No proactive rate limit checking**: `rate_limit_low?` exists but is never called
3. **Auto-paginated calls on long-lived PRs**: `issue_comments` fetches all pages even when only recent comments matter
4. **No incremental fetching**: All open issues are re-fetched every cycle instead of using `since`
5. **Graceful degradation masks exhaustion**: Errors caught and returned as `nil`/`[]` instead of stopping

### Phase 1: Eliminate N+1 Patterns (Highest Impact)

- #864 — Skip unchanged issues when fetching comments in `FetchIssuesActivity`
- #866 — Share fetched data across detection methods in `ScanPaidPrsActivity`
- #719 — Batch PR data into single GraphQL query (existing)

### Phase 2: Add Proactive Rate Limit Budgeting (High Impact)

- #868 — Add proactive rate limit checks to polling activities
- #865 — Add retry middleware to GraphQL connection

### Phase 3: Reduce Unnecessary Calls (Moderate Impact)

- #867 — Skip unchanged PRs in `ScanPaidPrsActivity`
- #870 — Use `recent_issue_comments` instead of auto-paginated `issue_comments`
- #869 — Use incremental issue fetching with `since` parameter

### Phase 4: Fix Non-Polling N+1 Patterns (Moderate Impact)

- #871 — Batch reaction fetching in `CollectReviewReactionFeedback`
- #873 — Reduce `DelayedHumanFeedbackCollectionJob` sweep scope

### Phase 5: Workflow-Level Coordination

- #872 — Add per-project rate limit budget coordination across activities

## Dependencies

This umbrella issue is blocked by all sub-issues above. Each sub-issue can be implemented independently.

## Estimated Impact

| Metric | Before | After |
|--------|--------|-------|
| API calls/cycle (20 PRs, 50 issues) | ~210+ | ~55 |
[...]

PR #893, opened 2026-04-07
Fix #870: use recent_issue_comments instead of auto-paginated issue_comments

### LINKED ISSUE #870: perf(pr-scanner): use recent_issue_comments instead of auto-paginated issue_comments
## Problem

`ScanPaidPrsActivity#check_conversation_comments` calls `client.issue_comments` which auto-paginates **all** comments. On long-lived PRs with hundreds of comments, this makes multiple pages of API calls. Since the method filters by `cutoff = last_run&.completed_at`, only recent comments are relevant — fetching all historical comments is wasteful.

## Current behavior

- `check_conversation_comments` (line 471) calls `client.issue_comments(project.full_name, issue.github_number)`
- `GithubClient#issue_comments` auto-paginates with `per_page: 100`
- A PR with 500 comments = 5 API calls, but only the newest ~10 comments may be relevant
- Across 20 PRs, this could mean 20-100 extra API calls per cycle

## Desired behavior

Use `GithubClient#recent_issue_comments` (already defined, fetches newest 100 comments in a single call) instead of the auto-paginating `issue_comments`.

### Implementation notes

- Replace `client.issue_comments(...)` in `check_conversation_comments` with `client.recent_issue_comments(...)`
- `recent_issue_comments` returns newest-first; the sorting and filtering logic in `check_conversation_comments` may need minor adjustment
- 100 recent comments is sufficient since the method already filters by `last_run.completed_at` cutoff
- Estimated savings: eliminates multiple pages of comments per PR per cycle → 1 call per PR

## Impact

**High** — For 20 PRs with an average of 3 pages of comments, reduces ~60 API calls to ~20 per cycle (~67% reduction for this specific call).

Related to #719, #738.

viamin:
## Summary

- Replace auto-paginating `client.issue_comments` with `client.recent_issue_comments` in `ScanPaidPrsActivity#check_conversation_comments`
- Adds early `break` when iterating newest-first comments older than the cutoff timestamp

## Problem

`check_conversation_comments` called `client.issue_comments` which auto-paginates **all** comments. On long-lived PRs with hundreds of comments, this makes multiple pages of API calls per PR per poll cycle. Since the method filters by `last_run.completed_at`, only recent comments are relevant.

## Impact

For 20 PRs with an average of 3 pages of comments, reduces ~60 API calls to ~20 per poll cycle (~67% reduction for this specific call).

Closes #870
Part of #879

### DIFF
--- a/app/services/github_client.rb
+++ b/app/services/github_client.rb
@@ -382,9 +382,15 @@ class GithubClient
     handle_errors do
       first_page = client.issue_comments(repo, number, per_page: 100, page: 1)
       last_rel = client.last_response&.rels&.dig(:last)
-      return first_page if last_rel.nil?
 
-      client.get(last_rel.href)
+      unless last_rel
+        first_page.define_singleton_method(:multi_page?) { false }
+        return first_page
+      end
+
+      last_page = client.get(last_rel.href)
+      last_page.define_singleton_method(:multi_page?) { true }
+      last_page
     end
   end
 
--- a/app/services/github_client.rb
+++ b/app/services/github_client.rb
@@ -384,15 +384,11 @@ class GithubClient
       last_rel = client.last_response&.rels&.dig(:last)
 
       unless last_rel
-        first_page.instance_variable_set(:@multi_page, false)
-        first_page.define_singleton_method(:multi_page?) { @multi_page }
-        return first_page
+        return tag_multi_page(first_page, false)
       end
 
       last_page = client.get(last_rel.href)
-      last_page.instance_variable_set(:@multi_page, true)
-      last_page.define_singleton_method(:multi_page?) { @multi_page }
-      last_page
+      tag_multi_page(last_page, true)
     end
   end
 
@@ -788,6 +784,12 @@ class GithubClient
   end
 
   private
+
+  def tag_multi_page(page, multi)
+    page.instance_variable_set(:@multi_page, multi)
+    page.define_singleton_method(:multi_page?) { @multi_page }
+    page
+  end
 
   def configure_middleware
     client.middleware = Faraday::RackBuilder.new do |builder|
--- a/app/services/github_client.rb
+++ b/app/services/github_client.rb
@@ -384,12 +384,14 @@ class GithubClient
       last_rel = client.last_response&.rels&.dig(:last)
 
       unless last_rel
-        first_page.define_singleton_method(:multi_page?) { false }
+        first_page.instance_variable_set(:@multi_page, false)
+        first_page.define_singleton_method(:multi_page?) { @multi_page }
         return first_page
       end
 
       last_page = client.get(last_rel.href)
-      last_page.define_singleton_method(:multi_page?) { true }
+      last_page.instance_variable_set(:@multi_page, true)
+      last_page.define_singleton_method(:multi_page?) { @multi_page }
       last_page
     end
   end
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -678,7 +678,7 @@ module Activities
 
       # Treat nil created_at as newer than cutoff (safe: triggers fallback to
       # full pagination rather than silently missing comments).
-      if cutoff && comments.multi_page? &&
+      if cutoff && comments.multi_page? && comments.any? &&
           comments.all? { |c| c.created_at.nil? || c.created_at > cutoff }
         client.issue_comments(project.full_name, issue.github_number)
       else
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -644,21 +644,22 @@ module Activities
     end
 
     def check_conversation_comments(client, project, issue, last_run)
-      comments = client.issue_comments(project.full_name, issue.github_number)
+      comments = client.recent_issue_comments(project.full_name, issue.github_number)
       cutoff = last_run&.completed_at
 
-      relevant = comments.select do |c|
-        login = c.user&.login
-        next false if bot_user?(login)
-        next false unless project.trusted_github_user?(login)
-        next false if cutoff && c.created_at <= cutoff
-        next false if system_generated_comment?(c.body)
-        next false if c.body.to_s.strip.length < MIN_COMMENT_LENGTH
+      relevant = comments.filter_map do |c|
+        break if cutoff && c.created_at && c.created_at <= cutoff
 
-        true
+        login = c.user&.login
+        next if bot_user?(login)
+        next unless project.trusted_github_user?(login)
+        next if system_generated_comment?(c.body)
+        next if c.body.to_s.strip.length < MIN_COMMENT_LENGTH
+
+        c
       end
 
-      return [] if relevant.empty?
+      return [] if relevant.nil? || relevant.empty?
 
       [ { type: "conversation_comments", details: "#{relevant.size} new comment(s)" } ]
     rescue GithubClient::Error => e
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -651,6 +651,7 @@ module Activities
         login = c.user&.login
         next false if bot_user?(login)
         next false unless project.trusted_github_user?(login)
+        # Defensive: treat nil created_at as potentially relevant (include it)
         next false if cutoff && c.created_at && c.created_at <= cutoff
         next false if system_generated_comment?(c.body)
         next false if c.body.to_s.strip.length < MIN_COMMENT_LENGTH
@@ -675,7 +676,9 @@ module Activities
     def fetch_conversation_comments(client, project, issue, cutoff)
       comments = client.recent_issue_comments(project.full_name, issue.github_number)
 
-      if cutoff && comments.respond_to?(:multi_page?) && comments.multi_page? &&
+      # Treat nil created_at as newer than cutoff (safe: triggers fallback to
+      # full pagination rather than silently missing comments).
+      if cutoff && comments.multi_page? &&
           comments.all? { |c| c.created_at.nil? || c.created_at > cutoff }
         client.issue_comments(project.full_name, issue.github_number)
       else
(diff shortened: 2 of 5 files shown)


app/services/github_client.rb
```diff
@@ -382,9 +382,15 @@ def recent_issue_comments(repo, number)
     handle_errors do
       first_page = client.issue_comments(repo, number, per_page: 100, page: 1)
       last_rel = client.last_response&.rels&.dig(:last)
-      return first_page if last_rel.nil?
 
-      client.get(last_rel.href)
+      unless last_rel
+        first_page.define_singleton_method(:multi_page?) { false }
```
viamin: Using `define_singleton_method` to attach metadata to an API response array is fragile — if the result is ever `.dup`'d, `.map`'d, `.select`'d, or otherwise transformed before `multi_page?` is checked, the singleton method silently disappears and the caller falls through to the wrong branch.

This works today because `fetch_conversation_comments` checks the method immediately, but it's a subtle contract. Consider returning a small wrapper instead, e.g.:

```suggestion
        result = first_page
        result.instance_variable_set(:@multi_page, false)
        result.define_singleton_method(:multi_page?) { @multi_page }
        return result
```

Or even simpler: a `Struct.new(:comments, :multi_page?)` return type that makes the contract explicit. Not blocking, but worth considering for maintainability.

viamin (commented):
Nice performance optimization — the approach of trying the lightweight single-page fetch first and falling back to full pagination is sound. The test coverage is thorough with good edge case scenarios. A few observations below.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -666,6 +666,23 @@ def check_conversation_comments(client, project, issue, last_run)
...
+    # Fetches conversation comments, preferring the lightweight single-page
+    # recent_issue_comments call. Falls back to full auto-paginated
+    # issue_comments when the returned page is from a multi-page result and
+    # the cutoff window extends beyond it (i.e., every comment on the page
+    # is newer than the cutoff, meaning older post-cutoff comments may exist
+    # on earlier pages).
+    def fetch_conversation_comments(client, project, issue, cutoff)
+      comments = client.recent_issue_comments(project.full_name, issue.github_number)
+
+      if cutoff && comments.respond_to?(:multi_page?) && comments.multi_page? &&
```
viamin: The `respond_to?(:multi_page?)` guard is unnecessary — `recent_issue_comments` always defines `multi_page?` on its return value. If you want this as a safety net against future refactoring, a comment explaining why would help. Otherwise it's dead code that makes the reader wonder if there's a code path where the method is missing.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -666,6 +666,23 @@ def check_conversation_comments(client, project, issue, last_run)
...
+          comments.all? { |c| c.created_at.nil? || c.created_at > cutoff }
```
viamin, line 682: Minor inconsistency: here, `c.created_at.nil?` is treated as "newer than cutoff" (triggers fallback to full pagination), but in `check_conversation_comments` line 654, a nil `created_at` means the comment passes the cutoff filter and is treated as relevant.

Both are defensively correct (nil → assume important), but the semantic meaning differs: one says "this might be old, fetch more to be safe" while the other says "this might be new, include it." Worth a brief comment on the nil-as-newer-than-cutoff choice here to avoid future confusion.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -644,14 +644,14 @@ def review_bot_thread_triggers(unresolved_threads)
...
-      comments = client.issue_comments(project.full_name, issue.github_number)
       cutoff = last_run&.completed_at
+      comments = fetch_conversation_comments(client, project, issue, cutoff)
 
       relevant = comments.select do |c|
         login = c.user&.login
         next false if bot_user?(login)
         next false unless project.trusted_github_user?(login)
-        next false if cutoff && c.created_at <= cutoff
+        next false if cutoff && c.created_at && c.created_at <= cutoff
```
viamin, line 655: Good addition of the nil guard on `c.created_at`. This wasn't in the original code — was this fixing an observed bug, or purely defensive? If it was a real issue, it might warrant its own brief commit message note for traceability.

viamin:
## Agent Update

1. **Thread 1 (github_client.rb:387)** - Backed `multi_page?` singleton method with an `@multi_page` instance variable so the metadata survives array transformations like `.dup`, `.map`, `.select`.

2. **Thread 2 (scan_paid_prs_activity.rb:678)** - Removed the `respond_to?(:multi_page?)` guard since `recent_issue_comments` always defines the method. It was dead code.

3. **Thread 3 (scan_paid_prs_activity.rb:679)** - Added a comment explaining that treating nil `created_at` as newer-than-cutoff is intentionally safe: it triggers fallback to full pagination rather than silently missing comments.

4. **Thread 4 (scan_paid_prs_activity.rb:654)** - Added a comment noting the nil `created_at` guard is defensive, treating missing timestamps as potentially relevant (include rather than exclude).

All 171 tests pass, lint is clean.

app/services/github_client.rb
```diff
@@ -382,9 +382,17 @@ def recent_issue_comments(repo, number)
...
-      client.get(last_rel.href)
+      unless last_rel
+        first_page.instance_variable_set(:@multi_page, false)
+        first_page.define_singleton_method(:multi_page?) { @multi_page }
+        return first_page
+      end
+
+      last_page = client.get(last_rel.href)
+      last_page.instance_variable_set(:@multi_page, true)
+      last_page.define_singleton_method(:multi_page?) { @multi_page }
```
viamin: The `instance_variable_set` + `define_singleton_method` pattern is repeated twice and is fairly brittle — it monkey-patches each returned array at runtime, and a caller forgetting to check `respond_to?(:multi_page?)` before calling it on a plain `issue_comments` result will get a `NoMethodError`.

Consider wrapping the result in a small value object or just using a `Struct` to make the contract explicit:

```ruby
CommentPage = Struct.new(:comments, :multi_page?, keyword_init: true)
```

Or at minimum, extract a helper:

```suggestion
      last_page = client.get(last_rel.href)
      tag_multi_page(last_page, true)
    end
  end

  private

  def tag_multi_page(page, multi)
    page.instance_variable_set(:@multi_page, multi)
    page.define_singleton_method(:multi_page?) { @multi_page }
    page
  end
```

This deduplicates the two identical blocks and gives you a single place to change the implementation later.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -666,6 +667,25 @@ def check_conversation_comments(client, project, issue, last_run)
...
+    # the cutoff window extends beyond it (i.e., every comment on the page
+    # is newer than the cutoff, meaning older post-cutoff comments may exist
+    # on earlier pages).
+    def fetch_conversation_comments(client, project, issue, cutoff)
+      comments = client.recent_issue_comments(project.full_name, issue.github_number)
+
+      # Treat nil created_at as newer than cutoff (safe: triggers fallback to
+      # full pagination rather than silently missing comments).
+      if cutoff && comments.multi_page? &&
+          comments.all? { |c| c.created_at.nil? || c.created_at > cutoff }
```
viamin, line 682: Potential edge case: if `comments` is empty (no comments on the last page), `comments.all?` returns `true` for an empty collection, so this will fall back to full pagination even though there are no comments to find. That's safe but wasteful — an empty last page from `recent_issue_comments` means there are zero comments total, so you can short-circuit:

```suggestion
      if cutoff && comments.multi_page? && comments.any? &&
          comments.all? { |c| c.created_at.nil? || c.created_at > cutoff }
```

Minor, but avoids an unnecessary full-pagination call on PRs with zero comments.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -644,14 +644,15 @@ def review_bot_thread_triggers(unresolved_threads)
...
       cutoff = last_run&.completed_at
+      comments = fetch_conversation_comments(client, project, issue, cutoff)
 
       relevant = comments.select do |c|
         login = c.user&.login
         next false if bot_user?(login)
         next false unless project.trusted_github_user?(login)
-        next false if cutoff && c.created_at <= cutoff
+        # Defensive: treat nil created_at as potentially relevant (include it)
+        next false if cutoff && c.created_at && c.created_at <= cutoff
```
viamin, line 655: Good defensive guard for `nil` `created_at`. Worth noting that the same nil-safety is applied in `fetch_conversation_comments` (line 682) for the fallback check, so the two are consistent — nice.

spec/temporal/activities/scan_paid_prs_activity_spec.rb
```diff
@@ -310,6 +310,137 @@ def enable_copilot_and_codex_review!(proj = project)
...
+
+        expect(result[:prs_to_trigger].size).to eq(1)
+        trigger_types = result[:prs_to_trigger].first[:triggers].map { |t| t[:type] }
+        expect(trigger_types).to include("conversation_comments")
+      end
+    end
+
     context "when the new comment is a paid agent update" do
       let(:comment) do
         OpenStruct.new(
```
viamin, line 402: The three fallback-related test contexts ("when cutoff window exceeds the recent comments page" and "when partial last page has all comments newer than cutoff") are structurally almost identical — they differ only in the size of `recent_page` (100 vs 50). Consider parameterizing with `[100, 50].each do |page_size|` to reduce duplication while keeping both cases covered.

viamin (commented):
Good PR — the optimization is well-motivated and the fallback logic is sound. The `multi_page?` approach via singleton methods works but is a bit fragile; a few suggestions below to tighten it up. Test coverage is thorough.

viamin:
## Agent Update

The background task notification is from the earlier failed commit attempt — it's already been resolved. The commit succeeded in a subsequent attempt (commit `47474b02`), and all changes are in place.

viamin (commented):
Generated no new comments. The PR correctly replaces the auto-paginating `issue_comments` with `recent_issue_comments` in `check_conversation_comments`, matching the approach described in #870. The fallback logic in `fetch_conversation_comments` is sound — it correctly detects when all comments on a multi-page last page are newer than the cutoff (meaning earlier post-cutoff comments may be missed) and falls back to full pagination. The defensive nil `created_at` handling is appropriate in both the filter and fallback paths. Tests cover the key scenarios including mixed old/new comments on a single page and the multi-page fallback trigger. The `db/schema.rb` diff is auto-generated bracket formatting from a migration run — not a concern.

PR #907, opened 2026-04-07
Fix #906: PR body rewriter-nil fallback dumps raw agent stdout instead of a templated description

### LINKED ISSUE #906: PR body rewriter-nil fallback dumps raw agent stdout instead of a templated description
## Summary

When \`Llm::GeneratePrDescription\` returns nil (rewriter failure, empty output, or exception), \`CreatePullRequestActivity#pr_body\` falls back to \`summary.truncate(50_000)\` — the raw agent stdout. This is user-hostile: if the agent's stdout was sparse or noisy (e.g., build-script errors, setup output), the PR ships with that noise as its entire description.

## Observed case

**viamin/paid#890** (Fix #866: share unresolved_threads across detection methods)

- Title, branch, code diff all correct (3-line fix in \`scan_paid_prs_activity.rb\`).
- **Body is the agent's raw setup-script stderr**: \"bundle install succeeded, db:prepare failed\" type output, with none of the actual PR description content. No mention of the issue, the fix, or the affected files.
- Issue #866 is still open; this PR is blocked from landing partly because nobody can tell what it does from its description.

## Root cause

\`app/temporal/activities/create_pull_request_activity.rb\` (around lines 52-96): \`pr_body\` computes \`summary = agent_run.agent_summary\`, passes it to \`Llm::GeneratePrDescription\`, and on nil or exception falls back to \`summary.truncate(50_000)\`. The raw stdout was never meant to serve as a description — it's just the agent's log output.

## Proposed fix

Replace the raw-stdout fallback with a templated minimal body:

\`\`\`markdown
## Summary

<issue title>

See <issue number> for context. Paid was unable to auto-generate a description for this PR;
the diff below is the authoritative source of truth.

## Changes

<git diff --stat output>

## Related

Closes #<issue number>
\`\`\`

This gives reviewers something useful even when the rewriter fails: the issue link, a changed-files summary, and an explicit signal that the description was not LLM-generated.

## Alternative — harder rewriter prompt

Alongside the templated fallback, consider tightening \`Llm::GeneratePrDescription\`'s prompt to refuse generating a description if the \`agent_summary\` doesn't reference the issue's stated subject or any of the changed file paths. Return nil explicitly in that case, then fall through to the templated body. This would also provide defense in depth against the scope-contamination bug tracked in the companion issue (see related).

## Severity

- **Cosmetic / auditability:** Medium. Not a data-integrity bug — the code diff is always authoritative. But reviewer experience and project-history readability are degraded.
- **Blocks nothing on its own** — the workaround is for the PR author to rewrite the body manually, which defeats the auto-pick value proposition for that PR.

## Related

- viamin/paid#890 — example of the rewriter-nil fallback producing unusable body
- Companion issue: \"Auto-pick PRs contaminated with previous run's scope in description body\" (container state reuse) — different root cause, but both would benefit from a tightened rewriter prompt and a better fallback template
- \`app/temporal/activities/create_pull_request_activity.rb\` — the file to modify (\`pr_body\` / \`generate_description\`)
- \`app/services/llm/generate_pr_description.rb\` — the rewriter that returns nil on failure

viamin:
## Summary

Replace the raw agent stdout fallback in PR body generation with a structured template so that when `Llm::GeneratePrDescription` fails, reviewers still get a useful description with the issue link and context instead of noisy build-script output.

## Changes

- **PR body fallback (`app/temporal/activities/create_pull_request_activity.rb`)**
  - Replaced `summary.truncate(50_000)` fallback with a new `fallback_body` method that produces a templated description
  - Template includes the issue title, a link to the issue for context, and an explicit note that auto-generation failed
  - For runs without an associated issue, falls back to a generic "automatically generated by Paid" message

- **Tests (`spec/temporal/activities/create_pull_request_activity_spec.rb`)**
  - Updated all fallback-related specs to verify the templated body structure instead of asserting raw agent output

## Design Decisions

- **Templated fallback over retry** — rather than retrying the LLM call (which may fail again for the same reason), the fallback provides a deterministic, useful body that always contains the issue reference and signals to reviewers that the description wasn't LLM-generated. This is cheaper and more reliable.
- **No `--stat` in fallback** — the simpler approach omits `git diff --stat` from the template (as suggested in the issue), keeping the fallback self-contained without requiring git operations at render time. The diff is always available via GitHub's Files Changed tab.
- **Preserves existing no-summary path** — the existing branch for completely blank agent summaries (no issue association) is kept separate from the new fallback, maintaining clear distinction between "LLM failed" and "agent produced no output at all."

---

Closes #906

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -148,7 +148,11 @@ class ProvidersController < ApplicationController
     if action_name == "create"
       permitted.push(:provider_key, :auth_type, :provider_api_key_id)
     end
-    attrs = params.require(:provider).permit(*permitted, config: { opencode: [ :api_provider, :model ], kilocode: [ :api_provider, :model ] })
+    attrs = params.require(:provider).permit(
+      *permitted,
+      config: { opencode: [ :api_provider, :model ], kilocode: [ :api_provider, :model ] },
+      tier_model_ids: LlmModel::TIERS
+    )
 
     # Convert config to a plain Hash and slice to only the relevant provider_key,
     # avoiding stale config from previously visible form fields.
@@ -159,7 +163,11 @@ class ProvidersController < ApplicationController
     provider_key = attrs[:provider_key].presence || @provider&.provider_key
     config = config.slice(provider_key) if provider_key.present?
 
-    attrs.to_h.merge("config" => config)
+    result = attrs.to_h.merge("config" => config)
+    if result.key?("tier_model_ids")
+      result["tier_model_ids"] = result["tier_model_ids"].to_h.compact_blank
+    end
+    result
   end
 
   def load_provider_options
--- a/app/javascript/controllers/provider_form_controller.js
+++ b/app/javascript/controllers/provider_form_controller.js
@@ -69,6 +69,8 @@ export default class extends Controller {
     "opencodeSettings",
     "kilocodeSettings",
     "directOutboundApiProviderSelect",
+    "tierSettings",
+    "tierSelect",
   ]
 
   connect() {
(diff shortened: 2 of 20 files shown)


viamin:
@codex review

viamin:

PR #908, opened 2026-04-07
Fix #905: Auto-pick PRs contaminated with previous run's scope in description body

### LINKED ISSUE #905: Auto-pick PRs contaminated with previous run's scope in description body
## Summary

Auto-pick PR descriptions are occasionally contaminated with content from a **different** auto-pick run's scope, while the code diff, title (\`Fix #NNN:\`), and \`Closes #NNN\` footer remain correct for the intended issue. The most likely cause is container, worktree, or agent-CLI state reuse between back-to-back runs on the same worker, letting the agent's final stdout echo a previous task's summary. Data-integrity risk is low but real (misleading audit trail, reviewer confusion).

## Observed cases

**viamin/paid#903** (Fix #901: scanner silently drops codex body-only reviews)
- Title, branch, \`Closes #901\` footer all correct.
- Diff correctly modifies \`scan_paid_prs_activity.rb\` to detect codex body-only reviews.
- **Body describes #864's scope**: talks about \`relationships_parsed_at\`, \`fetch_issues_activity\`, trust-policy callback — exactly the scope of viamin/paid#891 (Fix #864), which was produced in the same 2026-04-07 auto-pick burst.
- Closed unmerged; #901 was fixed by human-authored #902.

**viamin/paid#890** (Fix #866: share unresolved_threads across detection methods) — related but distinct failure mode
- Title, branch, code all correct.
- Body is the agent's raw \`db:prepare\` stderr output (build-script failure noise, not a description). This is the \"rewriter returned nil → raw stdout fallback\" path, which is a separate hardening bug (see companion issue).
- Issue #866 still open.

## Investigation findings (summary)

A thorough investigation traced the body-composition pipeline and found:

- \`create_pull_request_activity.rb#pr_body\` → \`agent_run.agent_summary\` → \`Llm::GeneratePrDescription\` all scope by \`agent_run_id\` via FK. **There is no cross-run leakage at the Ruby layer.**
- No recent commits touch \`create_pull_request_activity.rb\`, \`generate_pr_description.rb\`, or \`agent_run.rb#agent_summary\`. The body-composition code is correct.
- The \`Closes #NNN\` footer, title, and branch name are all derived from \`agent_run.issue.github_number\` — bound to the correct issue record. These cannot drift.
- **The contamination must happen upstream**, in either the agent CLI container state or the rewriter LLM's input context.

## Root cause hypotheses (ranked)

**H1 — Container/worktree state reuse (most likely).** Back-to-back auto-pick runs on the same worker may share leftover state: \`.claude/\`, \`.codex/\`, prompt cache, prior session transcripts, or mounted volumes. PR #903's body scope (\`relationships_parsed_at\`) matches #891 exactly, and both were in the same multi-issue auto-pick burst on 2026-04-07 touching adjacent files. If the agent CLI resumed a cached session or read a stale CLAUDE.md/prior transcript, its final stdout summary could echo the previous task.

**H2 — \`Llm::GeneratePrDescription\` hallucination from sparse summary.** The rewriter has wide latitude when \`agent_summary\` is thin. The rewriter is stateless per-call, but if the agent-harness invocation path has any response caching keyed insufficiently (e.g., by prompt hash alone), two prompts with similar structure could collide. Less likely than H1 but worth ruling out.

## Scope for investigation

Audit everything that touches agent-run container/worktree lifecycle for cross-run state leakage:

- \`app/services/containers/*\` — container provisioning, worktree setup, cleanup. Does each run get a guaranteed-clean container and a guaranteed-clean worktree? Are worktrees pooled/reused?
- \`agent-harness\` invocation path — are there session-resumption flags, cached transcripts, or reused process state between runs?
- Agent CLI home directories — \`.claude/\`, \`.codex/\`, \`.aider/\`, etc. Are these mounted from host or created fresh per run? Do they survive container recreation?
- \`CLAUDE.md\` / \`AGENTS.md\` handling — are they written per-run or sourced from a shared location that previous runs could have polluted?
- Prompt/response cache layers in \`agent_harness\` — where are they keyed, can they collide across runs?

## Mitigations (defense in depth)

Independent of the root-cause fix, add cheap guard rails in Paid itself:

1. **Sanity gate in \`pr_body\`**: before sending \`agent_summary\` to \`Llm::GeneratePrDescription\`, check that the summary text mentions the issue number OR at least one path from the agent run's \`git diff\`. If neither, log a warning (\`agent_execution.summary_scope_mismatch\`) and tag the PR with \`paid-needs-review\`. Cheap, cost-free insurance.
2. **Structured logging at agent-run start**: record container ID, worktree path, and any detected pre-existing agent CLI state files. Makes the next contamination event traceable.
3. **Post-run scope assertion**: after \`agent_summary\` is captured, assert it does not reference any *other* open Paid issue by number; if it does, log a warning.

## Severity

- **Data integrity:** Low. No false issue closures found in the past week's audit (13 auto-pick PRs reviewed; 2 had body issues, 0 had code issues).
- **Reviewer experience:** Medium. A PR whose description doesn't match its code is very confusing during review and undermines trust in the auto-pick loop.
- **Auditability:** Medium. Wrong descriptions pollute the project's historical record and make \`git log --oneline\` less useful.

## Related

- viamin/paid#903 — example of contamination (closed unmerged)
- viamin/paid#890 — raw-stdout fallback (companion issue to be filed)
- viamin/paid#902 — the correct manual fix for #901 that replaced #903

viamin:
## Summary

Detect and flag when an auto-pick PR description is contaminated with content from a different agent run's scope, preventing misleading audit trails and reviewer confusion. This adds defensive validation at PR creation time rather than attempting to fix the upstream container/worktree state reuse that causes the contamination.

## Changes

**PR body validation (`create_pull_request_activity.rb`)**
- Added `validate_summary_scope` that checks whether the agent summary references the target issue number and scans for cross-references to sibling open issues
- Logs a structured `agent_execution.summary_scope_mismatch` warning and appends a system log to the agent run when contamination is detected
- Added `sibling_open_issue_numbers` helper to query other open issues in the same project for cross-reference detection

**Container traceability (`provision_container_activity.rb`, `run_agent_activity.rb`)**
- Added `container_id` and `worktree_path` to structured log entries at container provision time
- Added `log_container_context` method that records container ID, worktree path, and provider at agent-run start, making future contamination events easier to trace back to specific infrastructure state

**Tests (`create_pull_request_activity_spec.rb`)**
- Added specs for scope mismatch detection when the summary references the wrong issue
- Added specs confirming no false positives when the summary correctly references its own issue

## Design Decisions

- **Detection over prevention**: This change flags contamination at PR creation time rather than fixing the root cause (container/worktree state reuse between runs). The upstream fix requires deeper changes to the container lifecycle; detection provides immediate visibility while that work is scoped.
- **Warn, don't block**: Contaminated PRs are still created but with a logged warning and system log annotation. Blocking would risk dropping otherwise-correct PRs where only the description is wrong.
- **Sibling issue scoping**: Cross-reference detection is limited to open issues in the same project, avoiding false positives from legitimate issue mentions in unrelated contexts.

---

Closes #905


viamin:
## Agent Update

Commit succeeded. All 23 tests pass, lint is clean.

Here's a summary of what was done:

**Fixed the P2 review comment**: The `validate_summary_scope` method was stripping the `owner/repo` qualifier from qualified issue refs like `rails/rails#1234`, treating the bare number as a local cross-reference. This caused false `summary_scope_mismatch` warnings when an external repo's issue number happened to match a local open issue.

**Changes made**:
1. **`create_pull_request_activity.rb`**: Updated the regex to capture the optional `owner/repo` qualifier separately, then filter: only bare `#NNN` refs and qualified refs where the qualifier matches `project.full_name` are treated as local references. External qualified refs are now ignored.

2. **`create_pull_request_activity_spec.rb`**: 
   - Added a new test case for external qualified refs (e.g., `rails/rails#NNN`) not triggering false positives
   - Updated existing qualified-ref tests to use `project.full_name` instead of hardcoded `owner/repo` (since the factory generates `owner-N/repo-N`)

viamin:
## Agent Update

Merge conflicts resolved and committed successfully. Here's a summary:

**Conflict resolution**: `create_pull_request_activity.rb` had a conflict between this branch's scope validation methods (`validate_summary_scope`, `sibling_open_issue_numbers`) and main's new priority label methods (`inherited_priority_labels`, updated `add_pr_labels`). Both sets of methods were kept — they serve different purposes and don't conflict semantically.

**Verification**: All 5636 tests pass (0 failures), RuboCop reports no offenses, and the working tree is clean.

app/temporal/activities/create_pull_request_activity.rb
```diff
@@ -111,6 +112,76 @@ def generate_description(summary, issue, agent_run_id:)
...
+      # like C#123 or abc#456 from being treated as issue references.
+      issue_ref_pattern = /(?<!\w)([\w.\-]+\/[\w.\-]+)?#(\d+)\b/
+
+      # Only count a qualified ref (owner/repo#NNN) as "own issue" when the
+      # qualifier matches this project's full_name (case-insensitive, since
+      # GitHub owner/repo names are case-insensitive).
+      mentions_own_issue = summary.scan(issue_ref_pattern).any? { |qualifier, num|
+        num.to_i == issue_number &&
+          (qualifier.nil? || qualifier.downcase == repo_full_name.downcase)
+      } || summary.match?(/\bissue\s+#{issue_number}\b/i)
```
viamin: Nit: `issue_number` is interpolated directly into a regex without `Regexp.escape`. Since `github_number` is an integer this is safe today, but if the type ever changes, this becomes a regex injection vector. Consider:

```suggestion
      } || summary.match?(/\bissue\s+#{Regexp.escape(issue_number.to_s)}\b/i)
```

viamin (commented):
## Review: Fix #905 — Detect scope-contaminated PR descriptions

Good approach: detection-over-prevention is the right call here, and the regex handling for qualified vs bare issue refs is thorough. The test coverage is excellent — edge cases like in-token `C#NNN`, external repo refs, and case-insensitive matching are all well covered.

A few observations below, mostly minor.

**Overall**: LGTM with minor suggestions.

app/temporal/activities/create_pull_request_activity.rb
```diff
@@ -111,6 +112,76 @@ def generate_description(summary, issue, agent_run_id:)
...
+      referenced_numbers.delete(issue_number)
+
+      sibling_numbers = sibling_open_issue_numbers(
+        agent_run.project,
+        exclude: issue_number,
+        candidates: referenced_numbers
+      )
+      cross_refs = referenced_numbers.intersection(sibling_numbers).to_a.sort
+
+      if !mentions_own_issue && cross_refs.any?
```
viamin: The condition `!mentions_own_issue && cross_refs.any?` means a summary that mentions **both** its own issue and a sibling issue will silently pass. This seems intentional (legitimate cross-references), but worth calling out: if a contaminated summary happens to also contain the correct issue number, the warning is suppressed. Consider whether `cross_refs.any?` alone (regardless of own-issue mention) should at least log at `info` level for observability, even if not a full warning.

app/temporal/activities/create_pull_request_activity.rb
```diff
@@ -111,6 +112,76 @@ def generate_description(summary, issue, agent_run_id:)
...
+    # a *different* issue — the most visible symptom of cross-run state
+    # contamination (see GitHub issue #905).
+    #
+    # The check is deliberately lightweight (string matching, no LLM call)
+    # so it never blocks PR creation.
+    def validate_summary_scope(summary, issue, agent_run)
+      return if summary.blank? || issue.nil?
+
+      issue_number = issue.github_number
+      repo_full_name = agent_run.project.full_name
```
viamin, line 171: Minor: `agent_run.project` is called here, but `project` is not passed as an argument and will trigger a DB query if not already loaded. In `pr_body`, the caller already has the `issue` and `agent_run` objects, but `project` was loaded on `agent_run` at line 11. Since `agent_run.project` should be cached by ActiveRecord's identity map within the same request, this is fine — just noting it for clarity.

app/temporal/activities/create_pull_request_activity.rb
```diff
@@ -111,6 +112,76 @@ def generate_description(summary, issue, agent_run_id:)
...
+      end
+    end
+
+    # Returns github_numbers of other open issues in the same project,
+    # excluding the current issue.
+    def sibling_open_issue_numbers(project, exclude:, candidates:)
+      return [] if candidates.blank?
+
+      project.issues
+        .where(github_state: "open", is_pull_request: false, github_number: candidates)
```
viamin, line 250: Nice optimization: scoping the query to only `candidates` avoids loading all open issues. The `where(github_number: candidates)` clause with a `Set` will be converted to an `IN (...)` — should work fine with ActiveRecord, just noting that if `candidates` is a `Set`, AR will call `.to_a` implicitly.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -517,6 +518,20 @@ def check_infinite_loop!(agent_run)
       raise InfiniteLoopError, result.reason if result.loop_detected?
     end
 
+    # Records container and worktree context at agent-run start for
+    # traceability.  If a future run exhibits cross-run contamination
+    # (see #905), these log entries make it possible to determine whether
+    # the container/worktree was reused.
+    def log_container_context(agent_run, provider)
```
viamin, line 603: Good traceability addition. One thought: at this point in the execution, `container_id` and `worktree_path` should already be populated from provisioning. But if they're nil for some reason (e.g., a code path that skips provisioning), the log entry would have `null` values. Not a bug, just something to be aware of — the structured log will still be emitted.

app/temporal/activities/provision_container_activity.rb
```diff
@@ -17,7 +17,9 @@ def execute(input)
 
         logger.info(
           message: "agent_execution.container_provisioned",
-          agent_run_id: agent_run_id
+          agent_run_id: agent_run_id,
+          container_id: agent_run.container_id,
```
viamin: Heads up: `agent_run.container_id` and `agent_run.worktree_path` are logged immediately after `provision_container`, but depending on how `provision_container` works, these might not be persisted to the DB yet (just set in memory). If `provision_container` calls `save!`/`update!` internally this is fine. Otherwise, the log could capture stale values if the record hasn't been reloaded. Worth verifying.

spec/temporal/activities/create_pull_request_activity_spec.rb
```diff
@@ -276,5 +276,167 @@
...
+    context "when agent summary references a different issue (scope mismatch)" do
+      let(:other_issue) { create(:issue, project: project, github_number: issue.github_number + 1000, github_state: "open") }
+
+      before do
+        other_issue # ensure created
+        agent_run.log!("stdout", "Updated relationships_parsed_at for ##{other_issue.github_number}")
+      end
+
+      it "logs a scope mismatch warning" do
+        allow(Rails).to receive(:logger).and_return(Rails.logger)
```
viamin: The test creates a log entry with `agent_run.log!("stdout", ...)` which feeds into `agent_summary`. This correctly simulates the real data flow. Nice.

One edge case not covered: what happens when `agent_summary` is non-blank but contains **no** issue references at all (no `#NNN`, no `issue NNN`)? The current code would have `mentions_own_issue = false` and `cross_refs = []`, so the `if !mentions_own_issue && cross_refs.any?` condition is false — no warning is logged. This is a reasonable design choice (many legitimate summaries may not reference issue numbers), but worth adding a brief test to document this behavior explicitly.

viamin:
## Agent Update

Done. Here's a summary of what was addressed:

**Merge conflicts**: Merged `origin/main` — no conflicts (clean merge).

**Review comments addressed**:

1. **`Regexp.escape` (line 140)**: Applied the reviewer's suggestion — `issue_number` is now escaped with `Regexp.escape(issue_number.to_s)` to guard against future type changes.

2. **Info-level logging for cross-refs with own-issue mention (line 158)**: Added an `info`-level log (`agent_execution.summary_cross_references`) when the summary mentions both its own issue and sibling issues. This provides observability for cases where contamination might be partial but the own-issue check suppresses the warning.

3. **`agent_run.project` DB query (line 126)**: Acknowledged — no change needed, ActiveRecord caches the association within the same request.

4. **Set-to-Array for AR `WHERE IN` (line 180)**: Acknowledged — AR handles this correctly.

5. **Nil container_id/worktree_path (run_agent_activity.rb:525)**: Acknowledged — structured log still emits with null values, no code change needed.

6. **Unpersisted values in provision log (provision_container_activity.rb:21)**: Acknowledged — `provision_container` persists internally.

7. **Missing test for no-issue-reference summaries (spec:289)**: Added a test documenting that summaries with no `#NNN` references don't trigger false mismatch warnings. Also added a test for the new info-level cross-reference logging behavior.

app/temporal/activities/create_pull_request_activity.rb
```diff
@@ -53,6 +53,7 @@ def pr_body(issue, agent_run)
       parts = []
 
       summary = agent_run.agent_summary
+      validate_summary_scope(summary, issue, agent_run)
```
viamin: **Bug risk**: `validate_summary_scope` is a detection-only side effect (logging), but if it raises an unexpected error (e.g., a DB failure in `sibling_open_issue_numbers`, or `project.full_name` returning something unexpected), it will abort the entire PR creation. Since the PR body and code are still correct even when the summary is contaminated, a failure here shouldn't prevent PR creation.

Consider wrapping the call:

```suggestion
      validate_summary_scope(summary, issue, agent_run) rescue nil
```

Or, for better observability, rescue with a structured log inside `validate_summary_scope` itself (similar to how `generate_description` already rescues `StandardError`).

app/temporal/activities/provision_container_activity.rb
```diff
@@ -17,7 +17,9 @@ def execute(input)
...
+          worktree_path: agent_run.worktree_path
```
viamin: Minor: At provision time, has `worktree_path` been set yet? The class comment says "The container is created before any git operations. Git clone happens inside the container in the subsequent CloneRepoActivity." If `worktree_path` is only populated after clone, this will always log `nil`, which reduces the traceability value this change is aiming for.

Worth verifying — if it's always nil here, consider either removing it from this log entry (since `log_container_context` in `run_agent_activity.rb` will log it later when it's populated) or adding a comment noting that it's intentionally logged as nil for the "before" snapshot.

viamin (commented):
## Review Summary

This PR adds detection-only validation for cross-run contamination in PR descriptions — a sensible, low-risk approach that matches the issue's requirements well. The regex-based scope checking is well-thought-out with good handling of qualified references, case-insensitive matching, and in-token false positives. The test coverage is thorough with excellent edge-case scenarios.

Two items worth addressing before merge:

viamin:
## Agent Update

**Review comment 1** (`create_pull_request_activity.rb:56`): Added `rescue StandardError` to `validate_summary_scope` that logs a structured `agent_execution.summary_scope_check_failed` warning instead of aborting PR creation. This follows the same pattern as `generate_description` which already rescues `StandardError`. Added two tests verifying the PR still creates successfully and the failure is logged.

**Review comment 2** (`provision_container_activity.rb:22`): Removed `worktree_path` from the provision log since it's always nil at provision time (git clone happens in the subsequent `CloneRepoActivity`). Added a comment explaining why it's omitted and noting that `log_container_context` in `RunAgentActivity` logs it when populated.

All 5641 tests pass, RuboCop reports no offenses, commit `de35c3c7` is clean.

viamin (commented):
Generated no new comments. The PR cleanly implements the three mitigations outlined in #905: scope validation at PR creation time, structured container/worktree traceability logging, and sibling-issue cross-reference detection. The regex handles edge cases well (in-token hashes, qualified external refs, case-insensitive repo matching), the DB query is efficiently scoped by candidate numbers, the rescue block ensures validation never blocks PR creation, and the test suite is thorough.

viamin:
## Agent Update

Merged the latest `origin/main` into the PR in a clean worktree at `/workspace/.worktrees/pr-908` and resolved the actual conflicts in [run_agent_activity.rb](/workspace/.worktrees/pr-908/app/temporal/activities/run_agent_activity.rb) and [create_pull_request_activity_spec.rb](/workspace/.worktrees/pr-908/spec/temporal/activities/create_pull_request_activity_spec.rb). The merge keeps both `execution_started_at` and `log_container_context`, and keeps both the PR’s scope-validation coverage and `main`’s idempotent-retry coverage.

I also fixed three PR-local spec regressions in [create_pull_request_activity_spec.rb](/workspace/.worktrees/pr-908/spec/temporal/activities/create_pull_request_activity_spec.rb) by allowing unrelated `prompts.render_fallback` warnings while still asserting the expected `agent_execution.summary_scope_mismatch` warning.

Verification:
- `bundle install`: passed with local bundle path
- `yarn install`: passed with workspace-backed cache/tmp
- `RAILS_ENV=test bin/rails db:prepare`: passed
- Targeted RuboCop on the touched files: passed
- Targeted RSpec for `create_pull_request_activity` and `run_agent_activity`: `0 failures`, but exits nonzero because SimpleCov enforces 80% coverage on partial runs

I did not make a commit. The blocking issue is repo-wide verification after merging latest `origin/main`:
- `bundle exec rubocop` fails on unrelated files from `main`, notably [db/schema.rb](/workspace/.worktrees/pr-908/db/schema.rb), [db/cable_schema.rb](/workspace/.worktrees/pr-908/db/cable_schema.rb), and [bin/update](/workspace/.worktrees/pr-908/bin/update)
- full `bundle exec rspec` is not clean in this environment either

Committing now would violate the task constraints requiring full lint/tests to pass before commit.

viamin:
## Agent Update

Commit succeeded — all 5763 tests passed, 0 failures, and rubocop reported no offenses. 

Here's what was done:

1. **Resolved merge conflicts** in `run_agent_activity.rb` (kept `log_container_context` call) and `create_pull_request_activity_spec.rb` (kept all scope validation tests alongside the idempotent retry tests from main)

2. **Fixed CI failure** — removed accidentally committed `.bundle-pr-908/` directory and `.worktrees/pr-908` file that were included in a prior commit (this was likely the CodeQL failure, as it was flagging committed build artifacts/vendored gems)

3. **Added git exclude patterns** for `.bundle-*/` and `.worktrees/` to prevent future accidental commits of build artifacts

app/temporal/activities/create_pull_request_activity.rb
```diff
@@ -156,6 +157,95 @@ def generate_description(summary, issue, agent_run_id:)
...
+      mentions_own_issue = summary.scan(issue_ref_pattern).any? { |qualifier, num|
+        num.to_i == issue_number &&
+          (qualifier.nil? || qualifier.downcase == repo_full_name.downcase)
+      } || summary.match?(/\bissue\s+#{Regexp.escape(issue_number.to_s)}\b/i)
+
+      # Extract issue numbers from the summary in a single pass, keeping only
+      # bare refs (#NNN) and qualified refs matching this project (owner/repo#NNN).
+      # External qualified refs (e.g. rails/rails#1234) are ignored to avoid
+      # false mismatch warnings when the same number exists locally.
+      referenced_numbers = summary.scan(issue_ref_pattern)
```
viamin, line 191: This only looks for `#NNN` references. That misses the concrete contamination from issue #905, where the wrong PR body drifted into another run's *file/symbol scope* (for example `relationships_parsed_at`, `fetch_issues_activity`) without citing another issue number. In that case `referenced_numbers` stays empty, no warning is emitted, and we still generate a misleading PR description. Please add the second half of the sanity gate from #905 here: if the summary does not mention the target issue, also require at least one path or other run-scoped artifact from the current diff before treating it as in-scope.

viamin (commented):
The scope-checking approach is headed in the right direction, but one correctness gap remains: it only flags contamination when the stale summary references another issue number. The incident described in #905 was a PR body that drifted into another issue's file/symbol scope without needing to cite `#NNN`, so this still lets the reported failure mode through. Please extend the guard to use the agent run's diff paths or another run-scoped signal before calling the issue fixed.

viamin:
## Agent Update

Done. Here's a summary of what was done:

## Changes Made

### Merge conflicts
Merged `origin/main` cleanly — no conflicts.

### Review comment addressed (`create_pull_request_activity.rb:191`)

The reviewer noted that the scope validation only checked for `#NNN` cross-references, missing contamination cases where wrong file/symbol content drifts in without explicit issue numbers (the exact pattern from #905).

**Fix**: Added a second sanity gate — when the summary doesn't mention its target issue AND has no cross-references to other issues, the code now fetches the actual changed files via `GithubClient#compare_changed_files` and checks whether any file basenames appear in the summary. If there's no overlap, a `summary_scope_mismatch` warning is logged.

Key implementation details:
- **`warn_unless_diff_overlap`** — compares summary text against changed file basenames (min 4 chars to avoid false matches from short names like `a.rb`)
- **`fetch_changed_files`** — wraps the GitHub compare API, returns `nil` when SHAs or client unavailable (fail-open)
- **GitHub client plumbed through** `pr_body` → `validate_summary_scope` via `client:` keyword arg
- **Tests added** for: summary overlapping with diff (no warning), no overlap (warning), and API unavailable (fails open gracefully)

viamin (commented):
Generated no new comments. The PR is well-implemented: validation logic is properly defensive (warns without blocking PR creation), error handling ensures the scope check never causes failures, the regex correctly handles edge cases (in-token refs, external qualified refs, case-insensitive owner/repo matching), and the DB query is efficiently scoped via the candidates filter. Test coverage is thorough across all important scenarios.

PR #921, opened 2026-04-07
Fix #918: feat(scanner): no clean signal detection for paid_agent reviews

### LINKED ISSUE #918: feat(scanner): no clean signal detection for paid_agent reviews
## Dependencies

Depends on #915 (auto-triggering must work before clean signal can be detected)

## Summary

There is no mechanism to detect when a `paid_agent` review is "clean" (no major findings). The scanner recognizes clean signals from Copilot and Codex but has no equivalent pattern for `paid_agent` reviews.

## Problem

The scanner recognizes two clean signals:

1. **Copilot** (thread-based): `REVIEW_BOT_CLEAN_PATTERN = /generated no (?:new )?comments/i` (`scan_paid_prs_activity.rb:23`)
2. **Codex** (body-only): `BODY_ONLY_BOT_CLEAN_COMMENT_PATTERN = /didn'?t find any (?:major )?issues/i` (`scan_paid_prs_activity.rb:29`)

There is **no clean signal pattern** for `paid_agent` reviews because:

- `paid_agent` reviews are posted as regular human-authored PR reviews, not from a bot login registered in `PROVIDER_BOT_USERNAMES` (`lib/provider_support.rb:183-198`).
- The `review_bot?` check (`scan_paid_prs_activity.rb:712`) uses `ProviderSupport.provider_bot_username?` which does not know about `paid_agent` identities.
- The user requirement for "an explicit comment to that effect" (no more major findings) has no detection mechanism.

Even if `paid_agent` reviews were automatically triggered (separate issue), the scanner would not be able to classify their output as clean vs. has-findings.

## Impact

- Without a clean signal, the scanner cannot determine when `paid_agent` review is complete.
- The anti-loop guard for body-only bots (`body_only_review_needs_followup?` at line 586-596) compares timestamps but has no semantic understanding of `paid_agent` review content.
- PRs could loop indefinitely or be incorrectly classified as "reviewed clean."

## Suggested fix

Define a structured clean signal for `paid_agent` reviews:

1. **Marker-based**: The `paid_agent` review prompt should instruct the agent to include a machine-readable marker when no major findings remain, e.g., `<!-- paid-review-clean -->` or a comment matching a specific pattern like `PAID_REVIEW_CLEAN_PATTERN = /no\s+(?:major\s+)?(?:issues|findings|concerns)\s+remaining/i`.

2. **Detection**: Add the marker check to `check_review_bot_status` alongside the existing Copilot/Codex detection paths. This may require identifying `paid_agent` reviews (by author, marker presence, or a dedicated field).

3. **Integration with `review_bot?`**: Either register `paid_agent` output as a pseudo-bot or add a parallel detection path that doesn't rely on `ProviderSupport.provider_bot_username?`.

viamin:
You've hit your limit · resets 11pm (UTC)

Let me read the key sections of the files I need to modify.

Now let me read the test file to understand the test structure around review bot status:

Now I have a thorough understanding of the codebase. Let me implement the changes.

First, add the constants after the existing pattern constants:

Now add the helper methods in the helpers section:

Now integrate the paid_agent clean signal check into `check_review_bot_status`. I'll add it after the body-only bot check:

Let me verify the changes look correct:

Good. Now let me verify the helper methods:

Now add the test helper and test cases. First, add the `enable_paid_agent_review!` helper:

Now I need to find the right place to add the test contexts. Let me look at where the review bot status tests end:

I'll add the paid_agent test cases after the "when review bot review has non-clean body" context (line 1231), before "when draft PR has CI green and no Copilot threads":

Now let me also add unit tests for the helper methods. Let me find where the describe block for helpers might be:

Good - the existing tests exercise methods through `execute`, and my integration tests follow the same pattern. Let me run lint and tests:

Good, database is prepared. Now run lint and tests:

Lint passes. Now run the tests:

I need to fix two tests that had wrong expectations. The draft phase scanner naturally produces `ready_for_owner` when CI is green and there are no bot issues. Let me fix both:

Now run the tests again:

All 89 tests pass. Let me also verify lint is still clean:

Now let me commit. First check the state:

**`app/temporal/activities/scan_paid_prs_activity.rb`:**
- Added `PAID_REVIEW_CLEAN_MARKER` (`<!-- paid-review-clean -->`) and `PAID_REVIEW_CLEAN_PATTERN` (`/no\s+(?:major\s+)?(?:issues|findings|concerns)\s+remaining/i`) constants
- Added `paid_agent_review_clean?(body)` and `paid_agent_clean_review_present?(reviews)` helper methods for marker-based detection
- Added a parallel detection path in `check_review_bot_status` that bypasses review triggers when `paid_agent` is enabled and a clean signal is found, restricted to configurations where no registered bot method could produce independent triggers (safe for mixed configs like paid_agent + copilot)

**`spec/temporal/activities/scan_paid_prs_activity_spec.rb`:**
- Added `enable_paid_agent_review!` test helper
- Added 5 test contexts: clean marker detection, text pattern detection, non-clean review, mixed config with copilot (clean signal doesn't suppress copilot threads), and disabled paid_agent (signal ignored)

---

Closes #918

### DIFF
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -28,6 +28,14 @@ module Activities
     # wording changes.
     BODY_ONLY_BOT_CLEAN_COMMENT_PATTERN = /didn'?t find any (?:major )?issues/i
 
+    # paid_agent reviews are posted from regular GitHub accounts (not bot
+    # logins registered in PROVIDER_BOT_USERNAMES), so they are invisible
+    # to review_bot? and enabled_review_bot_logins. The agent includes a
+    # machine-readable HTML comment marker when no major findings remain.
+    # A human-readable pattern is also accepted for robustness.
+    PAID_REVIEW_CLEAN_MARKER = "<!-- paid-review-clean -->"
+    PAID_REVIEW_CLEAN_PATTERN = /no\s+(?:major\s+)?(?:issues|findings|concerns)\s+remaining/i
+
     def execute(input)
       project_id = input[:project_id]
       project = Project.find_by(id: project_id)
@@ -509,6 +517,20 @@ module Activities
         return []
       end
 
+      # paid_agent reviews are authored by regular GitHub accounts, not
+      # bot logins registered in PROVIDER_BOT_USERNAMES. When paid_agent
+      # is enabled and any review contains the clean marker, treat the
+      # review cycle as complete — the agent has signaled "no major
+      # findings remaining." Only bypass when no registered bot method
+      # could produce independent triggers; in mixed configurations
+      # (e.g. paid_agent + copilot), each bot's status is evaluated
+      # independently below.
+      if project&.review_method_enabled?("paid_agent") &&
+          paid_agent_clean_review_present?(reviews) &&
+          (allowed.nil? || allowed.empty?)
+        return []
+      end
+
       status = review_bot_review_status_from(reviews, latest)
 
       case status
@@ -765,6 +787,18 @@ module Activities
 
     def review_bot?(login)
       ProviderSupport.provider_bot_username?(login)
+    end
+
+    def paid_agent_review_clean?(body)
+      return false if body.nil?
+
+      body.include?(PAID_REVIEW_CLEAN_MARKER) || PAID_REVIEW_CLEAN_PATTERN.match?(body)
+    end
+
+    def paid_agent_clean_review_present?(reviews)
+      return false if reviews.nil?
+
+      reviews.any? { |r| paid_agent_review_clean?(r[:body]) }
     end
 
     def extract_actionable_labels(triggers)
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -525,6 +525,14 @@ module Activities
       # could produce independent triggers; in mixed configurations
       # (e.g. paid_agent + copilot), each bot's status is evaluated
       # independently below.
+      #
+      # Today the existing flow already returns [] for paid_agent-only
+      # configs (no bot login → :no_review → nil login → []), so this
+      # early return is functionally redundant. It becomes load-bearing
+      # if paid_agent ever registers a bot login in PROVIDER_BOT_USERNAMES,
+      # which would cause `allowed` to be non-empty and the :no_review /
+      # :has_comments branches to fire. The explicit check keeps that
+      # future transition safe and documents the intended semantics.
       if project&.review_method_enabled?("paid_agent") &&
          paid_agent_clean_review_present?(reviews) &&
          (allowed.nil? || allowed.empty?)
@@ -795,11 +803,11 @@ module Activities
       body.include?(PAID_REVIEW_CLEAN_MARKER)
     end
 
-    # NOTE: paid_agent posts reviews from regular GitHub accounts — there is
+    # TODO(#918): paid_agent posts reviews from regular GitHub accounts — there is
     # no dedicated bot login to filter by. This method checks the latest
     # review from *any* author. The HTML marker (<!-- paid-review-clean -->)
-    # is unlikely to appear in human-authored reviews, but if the project
-    # later stores the paid_agent's GitHub login, this should filter by it.
+    # is unlikely to appear in human-authored reviews, but once the project
+    # stores the paid_agent's GitHub login, this should filter by it.
     def paid_agent_clean_review_present?(reviews)
       return false if reviews.nil? || reviews.empty?
 
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -32,9 +32,9 @@ module Activities
     # logins registered in PROVIDER_BOT_USERNAMES), so they are invisible
     # to review_bot? and enabled_review_bot_logins. The agent includes a
     # machine-readable HTML comment marker when no major findings remain.
-    # A human-readable pattern is also accepted for robustness.
+    # Only the marker is used for detection — no text patterns — to avoid
+    # false positives from human reviewers writing similar phrases.
     PAID_REVIEW_CLEAN_MARKER = "<!-- paid-review-clean -->"
-    PAID_REVIEW_CLEAN_PATTERN = /no\s+(?:major\s+)?(?:issues|findings|concerns)\s+remaining/i
 
     def execute(input)
       project_id = input[:project_id]
@@ -526,8 +526,8 @@ module Activities
       # (e.g. paid_agent + copilot), each bot's status is evaluated
       # independently below.
       if project&.review_method_enabled?("paid_agent") &&
-          paid_agent_clean_review_present?(reviews) &&
-          (allowed.nil? || allowed.empty?)
+         paid_agent_clean_review_present?(reviews) &&
+         (allowed.nil? || allowed.empty?)
         return []
       end
 
@@ -792,9 +792,14 @@ module Activities
     def paid_agent_review_clean?(body)
       return false if body.nil?
 
-      body.include?(PAID_REVIEW_CLEAN_MARKER) || PAID_REVIEW_CLEAN_PATTERN.match?(body)
+      body.include?(PAID_REVIEW_CLEAN_MARKER)
     end
 
+    # NOTE: paid_agent posts reviews from regular GitHub accounts — there is
+    # no dedicated bot login to filter by. This method checks the latest
+    # review from *any* author. The HTML marker (<!-- paid-review-clean -->)
+    # is unlikely to appear in human-authored reviews, but if the project
+    # later stores the paid_agent's GitHub login, this should filter by it.
     def paid_agent_clean_review_present?(reviews)
       return false if reviews.nil? || reviews.empty?
 
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -519,9 +519,9 @@ module Activities
 
       # paid_agent reviews are authored by regular GitHub accounts, not
       # bot logins registered in PROVIDER_BOT_USERNAMES. When paid_agent
-      # is enabled and any review contains the clean marker, treat the
-      # review cycle as complete — the agent has signaled "no major
-      # findings remaining." Only bypass when no registered bot method
+      # is enabled and the most recent review contains the clean marker,
+      # treat the review cycle as complete — the agent has signaled "no
+      # major findings remaining." Only bypass when no registered bot method
       # could produce independent triggers; in mixed configurations
       # (e.g. paid_agent + copilot), each bot's status is evaluated
       # independently below.
@@ -796,9 +796,10 @@ module Activities
     end
 
     def paid_agent_clean_review_present?(reviews)
-      return false if reviews.nil?
+      return false if reviews.nil? || reviews.empty?
 
-      reviews.any? { |r| paid_agent_review_clean?(r[:body]) }
+      latest = reviews.max_by { |r| r[:submitted_at] || Time.at(0) }
+      paid_agent_review_clean?(latest[:body])
     end
 
     def extract_actionable_labels(triggers)
--- a/spec/temporal/activities/scan_paid_prs_activity_spec.rb
+++ b/spec/temporal/activities/scan_paid_prs_activity_spec.rb
@@ -31,13 +31,6 @@ RSpec.describe Activities::ScanPaidPrsActivity do
     })
   end
 
-  def enable_paid_agent_review!(proj = project)
-    proj.update!(review_settings: {
-      "enabled" => true,
-      "methods" => { "paid_agent" => { "enabled" => true } }
-    })
-  end
-
   def enable_copilot_and_codex_review!(proj = project)
     proj.update!(review_settings: {
       "enabled" => true,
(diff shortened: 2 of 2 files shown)


viamin:
## Agent Update

**Fix:** Changed `paid_agent_clean_review_present?` to evaluate only the **latest** review (by `submitted_at`) instead of using `any?`. This prevents an older clean review from permanently masking a newer non-clean review after new commits are pushed.

**Changes:**
- `app/temporal/activities/scan_paid_prs_activity.rb:798-801` — replaced `reviews.any?` with `reviews.max_by { |r| r[:submitted_at] || Time.at(0) }` to find the latest review, then checks only that one
- `spec/temporal/activities/scan_paid_prs_activity_spec.rb:1337-1363` — added test case verifying that a newer non-clean review correctly overrides an older clean review

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -28,6 +28,14 @@ class ScanPaidPrsActivity < BaseActivity
     # wording changes.
     BODY_ONLY_BOT_CLEAN_COMMENT_PATTERN = /didn'?t find any (?:major )?issues/i
 
+    # paid_agent reviews are posted from regular GitHub accounts (not bot
+    # logins registered in PROVIDER_BOT_USERNAMES), so they are invisible
+    # to review_bot? and enabled_review_bot_logins. The agent includes a
+    # machine-readable HTML comment marker when no major findings remain.
+    # A human-readable pattern is also accepted for robustness.
+    PAID_REVIEW_CLEAN_MARKER = "<!-- paid-review-clean -->"
+    PAID_REVIEW_CLEAN_PATTERN = /no\s+(?:major\s+)?(?:issues|findings|concerns)\s+remaining/i
```
viamin: **False-positive risk from broad pattern.** `PAID_REVIEW_CLEAN_PATTERN` matches common English phrases like "no issues remaining" or "no concerns remaining." A human reviewer could write something like *"There are no major issues remaining after the latest fixes"* and accidentally trigger the bypass — especially since `paid_agent_clean_review_present?` checks the latest review from **any** author, not just the paid_agent.

The HTML marker (`<!-- paid-review-clean -->`) is safe since humans won't produce it accidentally, but this fallback pattern weakens that safety. Consider either:
1. Removing the text pattern and relying solely on the machine-readable marker, or
2. Making the pattern much more specific (e.g., requiring a prefix like `paid_agent:` or matching a full structured sentence the agent is prompted to produce).

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -767,6 +789,19 @@ def review_bot?(login)
...
+    def paid_agent_review_clean?(body)
+      return false if body.nil?
+
+      body.include?(PAID_REVIEW_CLEAN_MARKER) || PAID_REVIEW_CLEAN_PATTERN.match?(body)
+    end
+
+    def paid_agent_clean_review_present?(reviews)
+      return false if reviews.nil? || reviews.empty?
+
+      latest = reviews.max_by { |r| r[:submitted_at] || Time.at(0) }
```
viamin, line 814: This checks the latest review from **all** reviewers, not just paid_agent-authored reviews. Since paid_agent posts from regular GitHub accounts, there's no author-based filtering — any user's review that happens to contain the marker or matches the text pattern would trigger the bypass.

Is there a way to identify which GitHub account the paid_agent posts from? For example, could the `Project` model store the paid_agent's GitHub login so this can filter by author? If not, at minimum the comment on this method should document this design trade-off explicitly.

spec/temporal/activities/scan_paid_prs_activity_spec.rb
```diff
@@ -1254,6 +1261,176 @@ def enable_copilot_and_codex_review!(proj = project)
...
+      it "does not treat the review as clean because the latest review is non-clean" do
+        result = activity.execute(project_id: project.id)
+
+        trigger_types = result[:prs_to_trigger].flat_map { |t| t[:triggers].map { |x| x[:type] } }
+        expect(trigger_types).not_to include("review_bot_comments")
+      end
+    end
+
+    context "when paid_agent is enabled alongside copilot and copilot has unresolved threads" do
+      before do
```
viamin, line 1348: **Weak test — assertion passes trivially.** This test verifies that `review_bot_review_pending` and `review_bot_comments` are NOT in the triggers, but with `paid_agent` enabled, no registered bot logins, and no bot requesting login, the scanner would return `[]` regardless of the clean signal check. The `not_to include(...)` assertion passes the same way whether or not the non-clean path works correctly.

Consider either:
- Asserting that a specific trigger IS emitted (e.g., `ready_for_owner` or whatever the expected outcome is for a non-clean non-bot review), or
- Setting up the test so a trigger *would* be emitted without the paid_agent feature, proving the non-clean review doesn't suppress it.

As written, this test would still pass if the clean-signal check were accidentally removed.

spec/temporal/activities/scan_paid_prs_activity_spec.rb
```diff
@@ -1254,6 +1261,176 @@ def enable_copilot_and_codex_review!(proj = project)
...
+          ],
+          review_threads: [
+            {
+              id: "thread_1",
+              is_resolved: false,
+              comments: [ { body: "Fix this", path: "app/model.rb", line: 10,
+                            author: "copilot-pull-request-reviewer[bot]" } ]
+            }
+          ]
+        )
```
viamin, line 1379: Good test for the temporal ordering edge case — verifying only the latest review is considered. This correctly ensures a newer non-clean review supersedes an older clean one.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -509,6 +517,20 @@ def check_review_bot_status(reviews, unresolved_threads, project: nil, last_run:
...
+      # bot logins registered in PROVIDER_BOT_USERNAMES. When paid_agent
+      # is enabled and the most recent review contains the clean marker,
+      # treat the review cycle as complete — the agent has signaled "no
+      # major findings remaining." Only bypass when no registered bot method
+      # could produce independent triggers; in mixed configurations
+      # (e.g. paid_agent + copilot), each bot's status is evaluated
+      # independently below.
+      if project&.review_method_enabled?("paid_agent") &&
+          paid_agent_clean_review_present?(reviews) &&
+          (allowed.nil? || allowed.empty?)
```
viamin: Nit: the indentation of the continuation lines is 10 spaces, which doesn't align with either the `if` keyword (8 spaces) or `project` (9 spaces after `if `). Standard Ruby style is to align continuations with the first argument or indent 4 spaces from the `if`:

```suggestion
      if project&.review_method_enabled?("paid_agent") &&
         paid_agent_clean_review_present?(reviews) &&
         (allowed.nil? || allowed.empty?)
```

viamin (commented):
## Review: paid_agent clean signal detection

The feature addresses issue #918 well — the marker-based approach is sound and the mixed-config guard (`allowed.nil? || allowed.empty?`) is a smart safety net. A few concerns below, mostly around false-positive risk from the text pattern fallback and a test that doesn't assert what it appears to.

viamin:
## Agent Update

Commit succeeded. Here's a summary of changes addressing all review threads:

1. **Removed `PAID_REVIEW_CLEAN_PATTERN`** — Eliminated the broad text pattern (`/no\s+(?:major\s+)?(?:issues|findings|concerns)\s+remaining/i`) that could false-positive on human reviews. Detection now relies solely on the `<!-- paid-review-clean -->` HTML marker.

2. **Documented author-filtering trade-off** — Added a `NOTE` comment on `paid_agent_clean_review_present?` explaining that it checks the latest review from any author (since paid_agent has no dedicated bot login), and suggesting future improvement if the project stores the agent's GitHub login.

3. **Strengthened the weak test** — The "non-clean review" test now asserts `ready_for_owner` IS emitted, proving the clean signal bypass is not triggered (instead of the old `not_to include` assertion that passed trivially).

4. **Fixed indentation** — Aligned continuation lines in the `if` conditional to standard Ruby style.

5. **Removed duplicate helper** — Deleted the second `enable_paid_agent_review!` definition and the now-unnecessary text pattern test.

viamin (commented):
## Review: paid_agent clean signal detection

Good approach — using a machine-readable HTML comment marker rather than text pattern matching is the right call for avoiding false positives from human reviewers. The mixed-config test (paid_agent + copilot) is well-designed and validates the guard clause correctly.

I have a few concerns, one of which is substantive. See inline comments.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -509,6 +517,20 @@ def check_review_bot_status(reviews, unresolved_threads, project: nil, last_run:
...
+      # treat the review cycle as complete — the agent has signaled "no
+      # major findings remaining." Only bypass when no registered bot method
+      # could produce independent triggers; in mixed configurations
+      # (e.g. paid_agent + copilot), each bot's status is evaluated
+      # independently below.
+      if project&.review_method_enabled?("paid_agent") &&
+         paid_agent_clean_review_present?(reviews) &&
+         (allowed.nil? || allowed.empty?)
+        return []
+      end
```
viamin, line 540: **The bypass is effectively a no-op today.** When `paid_agent` is the only enabled review method:

1. `enabled_review_bot_logins` returns an empty `Set` (paid_agent has no entry in `PROVIDER_BOT_USERNAMES`)
2. `allowed` = `Set.new` (empty)
3. `latest_allowed_bot_review(reviews, Set.new)` returns `nil` — because `review_bot?` is false for regular GitHub accounts
4. `review_bot_review_status_from(reviews, nil)` returns `:no_review`
5. `:no_review` case → `review_bot_request_login(project)` returns `nil` for paid_agent-only → returns `[]`

So the existing flow **already** returns `[]` for paid_agent-only configs regardless of review content. This early-return adds a slightly faster path to the same result but doesn't change any behavior.

This isn't necessarily wrong — it's reasonable as defensive coding and documentation of intent. But if it's meant to be load-bearing for a future scenario (e.g., paid_agent getting a bot login), a comment explaining *when* this bypass would actually change the outcome would help. Otherwise a reader might assume it's doing something functional today.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -767,6 +789,24 @@ def review_bot?(login)
...
+
+    # NOTE: paid_agent posts reviews from regular GitHub accounts — there is
+    # no dedicated bot login to filter by. This method checks the latest
+    # review from *any* author. The HTML marker (<!-- paid-review-clean -->)
+    # is unlikely to appear in human-authored reviews, but if the project
+    # later stores the paid_agent's GitHub login, this should filter by it.
+    def paid_agent_clean_review_present?(reviews)
+      return false if reviews.nil? || reviews.empty?
+
+      latest = reviews.max_by { |r| r[:submitted_at] || Time.at(0) }
```
viamin, line 814: **Any PR reviewer can inject the clean marker.** Since this checks the latest review from *any* author, a human reviewer (or a malicious actor with PR review access) could post `<!-- paid-review-clean -->` and the scanner would treat the review cycle as complete.

The NOTE comment acknowledges this and suggests filtering by the paid_agent's GitHub login in the future. Consider making this a tracked TODO with an issue reference so it doesn't get lost:

```suggestion
    # TODO(#918): paid_agent posts reviews from regular GitHub accounts — there is
    # no dedicated bot login to filter by. This method checks the latest
    # review from *any* author. The HTML marker (<!-- paid-review-clean -->)
    # is unlikely to appear in human-authored reviews, but once the project
    # stores the paid_agent's GitHub login, this should filter by it.
```

spec/temporal/activities/scan_paid_prs_activity_spec.rb
```diff
@@ -1254,6 +1254,155 @@ def enable_copilot_and_codex_review!(proj = project)
...
+      it "treats the review as clean and emits no review_bot triggers" do
+        result = activity.execute(project_id: project.id)
+
+        trigger_types = result[:prs_to_trigger].flat_map { |t| t[:triggers].map { |x| x[:type] } }
+        expect(trigger_types).not_to include("review_bot_review_pending", "review_bot_comments")
+      end
+    end
+
+    context "when paid_agent review does not include a clean signal" do
+      before do
```
viamin, line 1287: This test would pass even without the new bypass code (see my comment on the production code). Since the existing flow already returns `[]` for paid_agent-only configs, asserting the *absence* of review_bot triggers doesn't prove the new code path was exercised.

Consider adding a test spy or assertion that confirms the bypass was the reason for the empty result — for example, verifying that `review_bot_review_status_from` was never called (because the early return short-circuited before it).

spec/temporal/activities/scan_paid_prs_activity_spec.rb
```diff
@@ -1254,6 +1254,155 @@ def enable_copilot_and_codex_review!(proj = project)
...
+
+      it "does not treat the review as clean because the latest review is non-clean" do
+        result = activity.execute(project_id: project.id)
+
+        trigger_types = result[:prs_to_trigger].flat_map { |t| t[:triggers].map { |x| x[:type] } }
+        expect(trigger_types).not_to include("review_bot_comments")
+      end
+    end
+
+    context "when paid_agent is enabled alongside copilot and copilot has unresolved threads" do
```
viamin, line 1347: Weak assertion: `not_to include("review_bot_comments")` would also pass if the bypass fired (returning `[]`). The test description says "does not treat the review as clean because the latest review is non-clean" but the assertion doesn't distinguish between "bypass didn't fire" and "bypass fired and returned []". 

Consider asserting a positive signal that the normal flow was reached — for example, `expect(trigger_types).to include("ready_for_owner")` (like the non-clean test above does) — to prove the bypass was skipped.

spec/temporal/activities/scan_paid_prs_activity_spec.rb
```diff
@@ -1254,6 +1254,155 @@ def enable_copilot_and_codex_review!(proj = project)
       end
```
viamin, line 1254: Nit: the moved `enable_paid_agent_review!` helper is just a relocation (no functional change). This is fine but the diff is a bit noisy — in the future, consider keeping helpers in their original position to minimize diff churn.

viamin:
## Agent Update

Commit succeeded. Here's what was addressed for each review thread:

1. **Bypass is a no-op** (line 532) — Added a comment explaining the bypass is currently redundant and documents when it becomes load-bearing (if paid_agent registers a bot login in `PROVIDER_BOT_USERNAMES`).

2. **Author injection risk** (line 806) — Changed `NOTE` to `TODO(#918)` with the reviewer's suggested wording so the author-filtering improvement is tracked.

3. **Trivially passing test** (spec line 1282) — Added `expect(activity).not_to receive(:review_bot_review_status_from)` spy to prove the bypass short-circuits before the normal flow is reached.

4. **Weak assertion** (spec line 1339) — Changed from `not_to include("review_bot_comments")` to `to include("ready_for_owner")`, which positively proves the normal flow was reached (bypass was skipped).

5. **Helper relocation nit** (spec line 1254) — Acknowledged; no change needed (was done in a prior commit).

viamin (commented):
Generated no new comments. The implementation is well-structured: marker-only detection avoids false positives from text patterns, the `allowed.empty?` guard correctly prevents suppression of registered bot reviews in mixed configs, and the test suite covers the key scenarios (clean marker, non-clean, superseded clean signal, mixed paid_agent+copilot, disabled paid_agent). The TODO on `paid_agent_clean_review_present?` appropriately flags the any-author limitation for future work.

PR #922, opened 2026-04-07
feat(queue): user-defined priority labels for issues and PRs (#838)

### LINKED ISSUE #838: Allow user to manually set priorities of issues/PRs by adding labels
## Feature Request

Allow users to manually set priorities of issues/PRs by adding labels. These can be user defined in project settings, and by default will follow P1, P2, P3 semantics.

### Requirements

1. **User-defined priority labels**: Users can add priority labels (P1, P2, P3 by default) to issues or PRs
2. **Priority precedence**: User defined priorities take precedence over automatic priorities:
   - P1 is higher priority than manual
   - P2 is higher priority than auto continue and lower priority than manual
   - P3 is lower priority than auto continue
3. **Label inheritance**: Priority labels added to an issue should be copied to the PR created by paid
4. **Configuration option**: Add option in project settings to disable label inheritance

### Implementation Details

- Priority labels should be configurable in project settings
- Default priority labels: P1, P2, P3
- Agent runs associated with labeled issues/PRs should use the label priority instead of automatic priorities
- Priority inheritance should be optional (configurable in project settings)

### Use Cases

- Users want to explicitly mark critical issues as P1
- Teams want consistent priority semantics across issues and PRs
- Project maintainers want fine-grained control over task prioritization

viamin:
Closes #838

## Summary
- Adds per-project priority label config (`priority_labels` jsonb, default `{P1,P2,P3}`) so users can map P1/P2/P3 to any GitHub label name (e.g. `critical` / `high` / `low`).
- Adds `inherit_priority_labels` project setting (default on) — when Paid creates a PR for an issue, matching priority labels are copied onto the new PR.
- Reworks the agent-run queue into a 6-tier ordering: **P1 > Manual > P2 > Auto-continue > P3 > Auto-pick**. If an issue or PR carries multiple priority labels, the highest wins. The queue SQL reads labels from either the run's associated issue or its source pull request via correlated `EXISTS` subqueries against the `issues.labels` jsonb column.

## Test plan
- [x] `bin/rspec spec/models/project_spec.rb spec/models/agent_run_spec.rb spec/temporal/activities/create_pull_request_activity_spec.rb` — 417 examples, 0 failures
- [x] `bin/lint --changed` — clean
- [ ] Manual: configure custom P1/P2/P3 label names in project settings UI and verify queue ordering
- [ ] Manual: verify PR inheritance toggle on/off

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -269,6 +269,10 @@ class AgentRun < ApplicationRecord
     @label_priority_tier = compute_label_priority_tier
   end
 
+  # When a run has both an associated issue AND a source PR, labels from
+  # both records are considered (highest configured tier wins). Label
+  # name matching is case-sensitive on both the Ruby and SQL sides; an
+  # issue tagged "p1" will not match a configured tier of "P1".
   def compute_label_priority_tier
     return nil unless project
 
@@ -302,9 +306,11 @@ class AgentRun < ApplicationRecord
   end
 
   # Batch-loads source PR Issue rows for a collection of runs and stashes
-  # each on the run via source_pull_request_record=. Call from views/
-  # helpers that render priority badges for many runs to keep
-  # label_priority_tier from issuing one query per row.
+  # each on the run via source_pull_request_record=. Call from controllers
+  # rendering priority badges for many runs to keep label_priority_tier
+  # from issuing one query per row. Groups by project_id so the query is
+  # precise per project (no cross-project over-fetch) without resorting
+  # to dynamic SQL.
   def self.preload_source_pull_requests(runs)
     targets = Array(runs).select do |r|
       r.source_pull_request_number.present? && r.project_id.present? &&
@@ -312,10 +318,13 @@ class AgentRun < ApplicationRecord
     end
     return if targets.empty?
 
-    project_ids = targets.map(&:project_id).uniq
-    pr_numbers = targets.map(&:source_pull_request_number).uniq
-    records = Issue.where(is_pull_request: true, project_id: project_ids, github_number: pr_numbers)
-      .index_by { |i| [ i.project_id, i.github_number ] }
+    records = {}
+    targets.group_by(&:project_id).each do |project_id, group|
+      numbers = group.map(&:source_pull_request_number).uniq
+      Issue.where(is_pull_request: true, project_id: project_id, github_number: numbers).each do |issue|
+        records[[ issue.project_id, issue.github_number ]] = issue
+      end
+    end
 
     targets.each do |run|
       run.source_pull_request_record = records[[ run.project_id, run.source_pull_request_number ]]
@@ -333,7 +342,16 @@ class AgentRun < ApplicationRecord
   # the per-project label name from `projects.priority_labels` jsonb,
   # falling back to the literal tier key (P1/P2/P3) when the project's
   # mapping is empty so behavior matches Project#effective_priority_labels.
-  # `issues.labels` is jsonb; `@>` checks for element containment.
+  # `issues.labels` is jsonb; `@>` checks for element containment, and
+  # element matching is case-sensitive on both the SQL and Ruby sides.
+  #
+  # The JOIN uses an OR so the EXISTS check matches either the run's
+  # directly-associated issue OR the issue row representing its source
+  # pull request. When a run has both, labels from both rows are
+  # considered (any priority match wins). This mirrors the Ruby
+  # `compute_label_priority_tier` semantics. Both branches constrain
+  # `i.project_id = agent_runs.project_id`, so cross-project leakage
+  # is impossible even if `issue_id` were ever set across projects.
   #
   # NOTE: This SQL contains no interpolated values — the tier names are
   # hardcoded literals — so it is not a SQL injection vector.


app/models/agent_run.rb
```diff
@@ -211,46 +211,107 @@ def self.orphaned_project_owner?(user)
...
+  # created_at, with id as a stable tiebreaker.
   QUEUE_PRIORITIES = {
-    manual: { label: "Manual", indicator: 1 },
-    auto_continue: { label: "Auto-continue", indicator: 2 },
-    auto_pick: { label: "Auto-pick", indicator: 3 }
+    label_p1: { label: "P1", indicator: 1 },
+    manual: { label: "Manual", indicator: 2 },
+    label_p2: { label: "P2", indicator: 3 },
+    auto_continue: { label: "Auto-continue", indicator: 4 },
+    label_p3: { label: "P3", indicator: 5 },
```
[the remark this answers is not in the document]
viamin: Fixed in commit 7a05734b — added `:label_p1` / `:label_p2` / `:label_p3` to `ApplicationHelper::AGENT_RUN_PRIORITY_STYLES` (red / orange / amber) so label-prioritized runs render with distinct badges instead of falling through to the gray `:unknown` style.

app/models/agent_run.rb
```diff
@@ -211,46 +211,107 @@ def self.orphaned_project_owner?(user)
...
+        FROM projects p
+        JOIN issues i ON (
+          i.id = agent_runs.issue_id
+          OR (i.project_id = agent_runs.project_id
+              AND i.github_number = agent_runs.source_pull_request_number
+              AND i.is_pull_request = TRUE)
+        )
+        WHERE p.id = agent_runs.project_id
+          AND COALESCE(p.priority_labels->>'#{tier_key}', '') <> ''
+          AND i.labels @> jsonb_build_array(p.priority_labels->>'#{tier_key}')
```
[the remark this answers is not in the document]
viamin: Good catch — fixed in commit 7a05734b. The queue SQL now uses `COALESCE(NULLIF(p.priority_labels->>"P1", ""), "P1")` for each tier so an empty `priority_labels` jsonb falls back to the literal `P1`/`P2`/`P3` defaults, matching `Project#effective_priority_labels` semantics.

viamin:
@codex review

app/models/agent_run.rb
```diff
@@ -211,46 +211,162 @@ def self.orphaned_project_owner?(user)
...
+  # label_priority_tier. Falls back to a per-row find_by, but callers
+  # rendering many runs should call AgentRun.preload_source_pull_requests
+  # first to avoid an N+1.
+  def source_pull_request_record
+    return @source_pull_request_record if defined?(@source_pull_request_record)
+    return nil if source_pull_request_number.blank? || project.nil?
+
+    @source_pull_request_record = project.issues.find_by(
+      github_number: source_pull_request_number, is_pull_request: true
+    )
```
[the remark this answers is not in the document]
viamin: Fixed in commit 12fdfb52 — added `AgentRun.preload_source_pull_requests` to `app/views/projects/_agent_runs.html.erb` (the project show page recent-runs partial), and `ProjectsController#show` now does `agent_runs.recent.includes(:issue).limit(10)` so the issue-side label lookup is also preloaded. The agent_runs index controller already eager-loaded `:issue`.

app/models/agent_run.rb
```diff
@@ -211,46 +211,162 @@ def self.orphaned_project_owner?(user)
...
-      WHEN trigger_type = 'manual' THEN 0
-      WHEN trigger_type = 'automatic' AND source_pull_request_number IS NOT NULL THEN 1
-      ELSE 2
+      WHEN EXISTS (
+        SELECT 1 FROM projects p JOIN issues i ON (
+          i.id = agent_runs.issue_id
+          OR (i.project_id = agent_runs.project_id
+              AND i.github_number = agent_runs.source_pull_request_number
+              AND i.is_pull_request = TRUE)
+        )
```
viamin: **Performance**: The `JOIN issues i ON (i.id = agent_runs.issue_id OR (...))` pattern is repeated three times in this CASE expression. The `OR` in the join condition prevents PostgreSQL from choosing a single efficient index scan — it typically falls back to a nested-loop with two index probes per candidate row, multiplied by three EXISTS checks.

For a hot queue-ordering query, consider rewriting as a single lateral join or CTE that computes the label tier once, rather than three separate EXISTS subqueries. For example:

```sql
LEFT JOIN LATERAL (
  SELECT i.labels FROM issues i
  WHERE i.id = agent_runs.issue_id
     OR (i.project_id = agent_runs.project_id
         AND i.github_number = agent_runs.source_pull_request_number
         AND i.is_pull_request = TRUE)
  LIMIT 1
) issue_labels ON TRUE
```

Then the CASE can reference `issue_labels.labels` once per tier check without re-running the subquery. This is a follow-up optimization — the current approach is correct, just potentially slow on large queues.

app/models/agent_run.rb
```diff
@@ -211,46 +211,162 @@ def self.orphaned_project_owner?(user)
...
+  #
+  # NOTE: This SQL contains no interpolated values — the tier names are
+  # hardcoded literals — so it is not a SQL injection vector.
   QUEUE_PRIORITY_SQL = Arel.sql(<<~SQL.squish).freeze
     CASE
-      WHEN trigger_type = 'manual' THEN 0
-      WHEN trigger_type = 'automatic' AND source_pull_request_number IS NOT NULL THEN 1
-      ELSE 2
+      WHEN EXISTS (
+        SELECT 1 FROM projects p JOIN issues i ON (
```
viamin: **Correctness**: The first branch of the OR join (`i.id = agent_runs.issue_id`) doesn't constrain `i.project_id = p.id`. Since you're reading `p.priority_labels`, this means if an agent_run's issue somehow belongs to a different project than the run itself, it would read the priority config from the wrong project.

In practice this shouldn't happen (issue and agent_run should share a project), but for defensive correctness you could add `i.project_id = p.id` to both branches of the OR, or add it as an AND alongside the existing WHERE clause:

```suggestion
        SELECT 1 FROM projects p JOIN issues i ON (
          (i.id = agent_runs.issue_id AND i.project_id = p.id)
          OR (i.project_id = agent_runs.project_id
              AND i.github_number = agent_runs.source_pull_request_number
              AND i.is_pull_request = TRUE)
        )
```

app/controllers/projects_controller.rb
```diff
@@ -196,8 +196,10 @@ def project_params
       :auto_fix_merge_conflicts, :auto_scan_security, :security_severity_threshold,
       :generated_label_name, :automation_label_name,
       :auto_add_labels_enabled, :automation_on_label_enabled, :pr_aggregation_enabled,
+      :inherit_priority_labels,
       :agent_co_author_trailer,
-      allowed_github_usernames: [])
+      allowed_github_usernames: [],
+      priority_labels: Project::PRIORITY_TIERS.map(&:to_sym))
```
viamin: **Potential type mismatch**: `Project::PRIORITY_TIERS.map(&:to_sym)` produces `[:P1, :P2, :P3]`, which tells strong params to permit a hash with symbol keys. However, `Project::PRIORITY_TIERS` is `%w[P1 P2 P3]` (strings), and the model validation in `priority_labels_valid` checks `priority_labels.keys - PRIORITY_TIERS` using string comparison.

Rails params come in as strings from forms, so this will work in practice. But if the API ever receives JSON with string keys, the permit call matches on symbol keys which could cause confusion. Consider using the string array directly for clarity:

```suggestion
      priority_labels: Project::PRIORITY_TIERS)
```

app/views/projects/_agent_runs.html.erb
```diff
@@ -3,6 +3,8 @@
     <%= link_to "Recent Agent Runs", project_agent_runs_path(project), class: "hover:text-indigo-600" %>
   </h2>
 
+  <% recent_agent_runs = recent_agent_runs.to_a %>
```
viamin: **Code organization**: Data-loading (`preload_source_pull_requests`) in views is a code smell — views should receive ready-to-render data. This is pragmatic and avoids changing the controller interface, but consider moving the preload call into the controller (or a presenter) where `@recent_agent_runs` is loaded.

The same applies to the preload calls in `_index_table.html.erb` and `_table.html.erb`. A lightweight alternative: add a scope like `AgentRun.with_source_pull_requests` that chains the preload automatically.

app/models/project.rb
```diff
@@ -174,6 +181,32 @@ def set_label_for_stage(stage, label)
...
+  # Returns the user-configured GitHub label name for a priority tier ("P1"/"P2"/"P3").
+  def priority_label_for(tier)
+    effective_priority_labels[tier.to_s]
+  end
+
+  # Returns the highest priority tier (one of PRIORITY_TIERS) represented in
+  # the given list of GitHub label names, or nil if none of the configured
+  # priority labels are present. "Highest" means earliest in PRIORITY_TIERS.
+  def highest_priority_tier_for_labels(label_names)
+    return nil if label_names.blank?
```
viamin: Nit: `highest_priority_tier_for_labels` is defined and tested but not called anywhere in this PR. Is it intended for a follow-up? If so, a brief comment noting the intended consumer would help readers understand why it exists.

db/migrate/20260407230341_add_priority_labels_to_projects.rb
```diff
@@ -0,0 +1,8 @@
+class AddPriorityLabelsToProjects < ActiveRecord::Migration[8.1]
+  DEFAULT_PRIORITY_LABELS = { "P1" => "P1", "P2" => "P2", "P3" => "P3" }.freeze
```
viamin: **DRY**: This duplicates `Project::DEFAULT_PRIORITY_LABELS`. If the default ever changes in the model but not in the migration (or vice versa), the DB default and application default will diverge silently. Consider referencing the model constant, or at minimum adding a comment cross-referencing `Project::DEFAULT_PRIORITY_LABELS`.

(Migrations are snapshots in time so duplicating is defensible — just flagging the coupling.)

app/models/agent_run.rb
```diff
@@ -211,46 +211,162 @@ def self.orphaned_project_owner?(user)
...
   end
 
+  # Returns "P1"/"P2"/"P3" if the run's issue or source PR carries one of
+  # the project's configured priority labels (highest wins), else nil.
+  # Memoized because rendering helpers call this twice per row (once
+  # via queue_priority_tier and once via queue_priority_label).
+  def label_priority_tier
+    return @label_priority_tier if defined?(@label_priority_tier)
+    @label_priority_tier = compute_label_priority_tier
+  end
```
viamin: **Memoization note**: `@label_priority_tier` is memoized per instance. If an issue's labels are updated and the same AgentRun instance is reused (e.g., in a long-running process or after an in-place update), the priority tier will be stale. This is fine for the request-scoped rendering use case, but worth a comment noting the assumption, especially since `queue_priority_tier` feeds into queue ordering decisions.

app/models/agent_run.rb
viamin: Filed as follow-up viamin/paid#933 — the rewrite needs to thread a custom `from` through the queued scope so that `next_queued_run` / `peek_next_queued_run` / `claim_next_queued_run` (the FOR UPDATE SKIP LOCKED claim path) all stay consistent, which is wider surgery than I want in this PR. The `(project_id, github_number)` index added in commit 9d090066 keeps each EXISTS check using an index lookup, so current queue ordering stays bounded until #933 lands.

db/migrate/20260407230341_add_priority_labels_to_projects.rb
viamin: Fixed in commit f2cffaed — added a comment cross-referencing `Project::DEFAULT_PRIORITY_LABELS` and noting the snapshot semantics so a future change to the model default at least surfaces the audit-existing-rows step.

app/jobs/recover_missing_pull_request_labels_job.rb
```diff
@@ -116,15 +116,33 @@ def prefetch_synced_prs(runs)
     result
   end
 
-  def missing_labels(project, synced_pr)
+  def missing_labels(project, synced_pr, agent_run)
```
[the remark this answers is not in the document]
viamin: Confirmed — the ordering change is intentional and the new spec covers it. Inheritance is now controlled exclusively by `inherit_priority_labels?`, independent of `auto_add_labels_enabled?`.

PR #928, opened 2026-04-08
Fix #923: Add UI fields for configuring priority label names per project

### LINKED ISSUE #923: Add UI fields for configuring priority label names per project
Follow-up to #838 / viamin/paid#922.

#922 added a `priority_labels` jsonb column on `projects` (default `{"P1": "P1", "P2": "P2", "P3": "P3"}`) so each project can map the P1/P2/P3 priority tiers to user-defined GitHub label names. The Paid project itself already has `P1`/`P2`/`P3` labels, so it works out-of-the-box on the defaults — but other projects have no way to customize the names from the UI yet.

## What's missing
- Three text inputs in the project edit form (`app/views/projects/edit.html.erb`, near the existing label name fields) for `priority_labels[P1]`, `priority_labels[P2]`, `priority_labels[P3]`.
- Strong params for the nested hash in `ProjectsController#project_params` (currently only the `inherit_priority_labels` boolean is permitted).
- A request spec covering an update that changes the priority label names.

## Notes
- Model-side validation is already in place (`Project#priority_labels_valid` rejects unknown keys and blank values), so the form just needs to surface validation errors.
- Defaults are filled in by `Project#effective_priority_labels`, so leaving any field blank should be acceptable.
- Optional polish: when the user enters a custom label name that does not yet exist on the GitHub repo, surface a warning (or auto-create the label via the GitHub client).

viamin:
## Summary

Add UI fields for configuring priority label names per project

See #923 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #923

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -197,7 +197,8 @@ class ProjectsController < ApplicationController
       :generated_label_name, :automation_label_name,
       :auto_add_labels_enabled, :automation_on_label_enabled, :pr_aggregation_enabled,
       :agent_co_author_trailer,
-      allowed_github_usernames: [])
+      allowed_github_usernames: [],
+      priority_labels: [ :P1, :P2, :P3 ])
   end
 
   TERMINATION_KEYS = %i[max_review_rounds stop_when_no_comments quality_threshold timeout_minutes].freeze
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -2,6 +2,8 @@
 
 class Project < ApplicationRecord
   MERGE_METHODS = %w[squash merge rebase].freeze
+  PRIORITY_TIERS = %w[P1 P2 P3].freeze
+  DEFAULT_PRIORITY_LABELS = { "P1" => "P1", "P2" => "P2", "P3" => "P3" }.freeze
   KNOWLEDGE_STATUSES = %w[pending collecting ready failed stale].freeze
   # "none" is not a method — it is represented by enabled: false at the top level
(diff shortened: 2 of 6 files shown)


viamin:
## Agent Update

All background tasks have completed. The work is done — merge conflicts resolved, commit successful with all 5629 tests passing and lint clean.

viamin:
Closing in favor of #936, which is the cleaner approach for #923.

After rebasing both PRs onto current main I found that **main already has all the model-side support** this issue called for: `PRIORITY_TIERS`, `DEFAULT_PRIORITY_LABELS`, `effective_priority_labels`, `priority_labels_valid`, and `priority_labels: Project::PRIORITY_TIERS` strong params are all in place. Both PRs duplicated those definitions because they were branched off an older main.

The decisive difference is in the **view**:

- **#936** pre-fills each field with `@project.priority_labels&.dig(tier) || Project::DEFAULT_PRIORITY_LABELS[tier]`, so users never see (and never submit) an empty input — sidestepping main's validation that rejects blank values.
- **This PR** uses `value: @project.priority_labels[tier]` with no fallback, so a project with no overrides shows empty inputs that fail validation on submit.

The one real model contribution this PR had — converting empty strings to `nil` in `normalize_priority_labels` so blanks would pass validation — got dropped in the rebase auto-merge against main, and is more of an issue-scope expansion (changing validation semantics) than a UI fix anyway. If it's still wanted, it can be its own PR.

#936 has been rebased onto main, the duplicate model definitions and duplicate migration removed, and both specs (`allows updating priority label names`, `rejects blank priority label values`) pass against current main. Reviewers should look at #936.

PR #929, opened 2026-04-08
Fix #912: feat(notifications): notification center with bell icon, badges, and link-through

### LINKED ISSUE #912: feat(notifications): notification center with bell icon, badges, and link-through
## Summary

Add a notification center so operators see actionable problems (stalled PRs, timed-out runs, quota exhaustion, wedged scanner states, etc.) directly in the Paid UI instead of having to manually inspect Rails console, logs, or individual project pages. Notifications should link through to the exact object they describe so the fix is one click away.

## Motivation

Operational issues like the ones surfaced in viamin/paid#895, #905, #910, and the 1800s-timeout / draft-phase wedge failures we debugged on 2026-04-07 are currently invisible to the operator until a human notices "hmm, this PR hasn't moved in a while" or "why is this agent run at 0 iterations?" There are no pages, no alerts, no aggregated surface. The data exists in the database — we just need to present it.

This issue is scoped to the **UI and plumbing**. The specific alert rules live in a companion issue (forthcoming) so they can be developed incrementally without blocking the UI framework.

## User-visible behavior

Two presentation modes, both desired; the implementation should support both:

### A. Badges on existing nav items

The nav menu in \`app/views/layouts/application.html.erb\` currently has: Dashboard, Projects, Agent Runs, Prompts, Quality, Knowledge, Integrations, Providers, Services, Settings. When an unread notification exists whose subject is scoped to a particular section, show a small red badge on that nav item with the count.

- Notification about a PR stuck in draft → badge on **Projects**
- Notification about a timed-out agent run → badge on **Agent Runs**
- Notification about provider quota exhaustion → badge on **Providers**
- Notification about a stuck workflow or scanner → badge on **Dashboard** (or a new system/health area)

Badge click navigates to the scoped section filtered by "has unread notifications" (or equivalent).

### B. Bell icon with dropdown (new)

A new \`🔔\` (bell) icon in the top-right of the nav bar (alongside the user menu). Shows the global unread count as a small superscript number. Clicking opens a dropdown showing the 10 most recent unread notifications, each with:

- Severity icon (info / warning / error)
- Short title
- Subject link (e.g., \"Agent Run #5681\" or \"PR viamin/paid#892\")
- Timestamp (\"3 minutes ago\")
- Dismiss button

A \"See all\" link at the bottom goes to a new full-page notification index.

### C. Notifications index page

New route \`/notifications\` accessible from the bell dropdown's \"See all\" link and from a direct nav item in the Settings dropdown or mobile menu. Shows all notifications with filters (unread / all, by severity, by source), paginated. Each row has the same content as the dropdown plus a longer description and action links.

## Data model

New table \`notifications\`:

| Column | Type | Notes |
|---|---|---|
| id | bigint | |
| account_id | bigint | FK, indexed (multi-tenant scoping) |
| user_id | bigint | FK, nullable (account-wide vs. per-user) |
| subject_type | string | polymorphic pointer — \"AgentRun\", \"Issue\", \"Project\", \"Provider\", \"ServiceContainer\", etc. |
| subject_id | bigint | polymorphic FK |
| source | string | rule identifier — \"stalled_draft_pr\", \"repeat_timeout\", \"quota_exhausted\", \"scanner_wedged\", etc. |
| severity | string | \"info\" / \"warning\" / \"error\" — enum |
| title | string | short headline |
| description | text | longer context, supports markdown |
| metadata | jsonb | rule-specific details (e.g. \`{\"consecutive_timeouts\": 10, \"since\": \"2026-04-07T08:20:24Z\"}\`) |
| action_url | string | where clicking the notification navigates |
| nav_section | string | which nav item to badge — \"projects\" / \"agent_runs\" / \"providers\" / \"dashboard\" / nil |
| read_at | datetime | nullable — when the user marked it read |
| dismissed_at | datetime | nullable — when the user dismissed it |
| resolved_at | datetime | nullable — when the underlying condition cleared (auto-resolved) |
| created_at / updated_at | datetime | |

Indexes:
- \`(account_id, read_at, dismissed_at)\` for unread-count queries
- \`(account_id, source, subject_type, subject_id)\` unique — prevents duplicate notifications for the same condition (see deduplication below)
- \`(account_id, nav_section, read_at)\` for badge-count queries

### Deduplication and lifecycle

The unique index on \`(account_id, source, subject_type, subject_id)\` means each (rule, subject) pair has at most one live notification. Rules that re-detect the same problem update the existing row's \`updated_at\` and \`metadata\` rather than creating duplicates. When a rule detects the condition has cleared, it sets \`resolved_at\` (and the notification disappears from badge counts but remains visible in the index for audit).

Dismissed notifications stay in the table for audit but are excluded from counts and dropdowns.

### Scoping

Notifications are scoped by \`account_id\` for multi-tenant safety. The \`user_id\` column is optional — most operational notifications are account-wide (any operator can act on a stalled PR), but some user-specific ones (e.g., \"your personal GitHub token is expiring\") are per-user.

### Service API for rule writers

```ruby
Notifications::Publish.call(
  account: project.account,
  source: \"stalled_draft_pr\",
  subject: issue,                       # polymorphic
  severity: :warning,
  title: \"PR ##{issue.github_number} stuck in draft for 6 hours\",
  description: \"10 consecutive draft follow-up agent runs timed out...\",
  metadata: { consecutive_timeouts: 10, first_seen: 6.hours.ago },
  action_url: project_issue_path(project, issue),
  nav_section: \"projects\"
)
```

Idempotent — calling with the same (account, source, subject) updates the existing row rather than creating a duplicate.

A companion \`Notifications::Resolve.call(account:, source:, subject:)\` sets \`resolved_at\` when a rule detects the condition has cleared.

### Real-time updates

Use the existing Turbo Streams infrastructure:
- \`Notifications::Publish\` broadcasts to \`account_notifications:#{account.id}\` after insert/update
- The bell dropdown, badge partials, and notifications index subscribe via \`turbo_stream_from\`
- Count badges use \`<turbo-frame>\` with a tiny partial that re-renders the number on any broadcast

### Read/dismiss/resolve actions

\`NotificationsController\` with:
- \`PATCH /notifications/:id/read\` — set read_at
- \`PATCH /notifications/:id/dismiss\` — set dismissed_at
- \`POST /notifications/mark_all_read\` — bulk update visible notifications to read
- \`GET /notifications\` — index

## Implementation scope for this issue

- [ ] \`notifications\` table migration (UUID external id optional; bigint internal per CLAUDE.md)
- [ ] \`Notification\` model with polymorphic \`subject\`, enum severity, scopes for unread/dismissed/resolved
- [ ] \`Notifications::Publish\` and \`Notifications::Resolve\` Servo services
- [ ] \`NotificationsController\` with index, read, dismiss, bulk mark-all-read actions
- [ ] Bell icon + dropdown in \`app/views/layouts/application.html.erb\` (desktop and mobile)
- [ ] Badge partial rendered next to existing nav items, subscribed to turbo-stream updates
- [ ] Notifications index page (Phlex view per \`docs/STYLE_GUIDE.md\`)
- [ ] Turbo Stream broadcast wiring on publish/update/resolve
- [ ] Specs: model, publish/resolve services, request specs for controller actions, system spec for the bell dropdown and badge update flow

## Out of scope (see companion issue)

- The actual detection rules — \"stalled_draft_pr\", \"repeat_timeout\", \"quota_exhausted\", \"scanner_wedged_on_missing_bot\", etc. The framework should ship with **zero rules** initially, so its correctness can be verified in isolation with a seed notification inserted via Rails console or spec factory. Rules then arrive incrementally in the companion issue's scope.
- Email / Slack / webhook delivery of notifications. In-app only for v1. The schema and service API should not preclude these later — the \`Notifications::Publish\` call site is where a future dispatcher would fan out to additional channels.
- Per-user notification preferences (mute rules, severity floor). Add later if needed; default is \"all account members see all account-scoped notifications.\"

## Related

- Companion issue (forthcoming) for the detection rules that populate this framework
- viamin/paid#895 — draft-review circuit breaker (one of the conditions a rule should detect)
- viamin/paid#905 — auto-pick PR body contamination (reviewer signal, not operator signal — different audience)
- viamin/paid#910 — codex clean-signal detection (example of a condition that currently goes unnoticed)
- The session that prompted this: debugging the 1800s timeout loop on PR #892, which sat emitting pending triggers for hours before anyone looked

viamin:
## Summary

feat(notifications): notification center with bell icon, badges, and link-through

See #912 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #912


viamin:
## Agent Update

Done. Here's what was accomplished:

- **Merged `origin/main`** into the notification center branch
- **Resolved `db/schema.rb` conflicts**: kept the higher schema version (`2026_04_08_032317`), included the new `project_id + github_number` index from main, and preserved the notifications table from this branch
- **Regenerated schema** from the database to ensure consistency
- **All checks pass**: RuboCop (0 offenses), RSpec (5661 examples, 0 failures)
- **Committed** as `ef0d52dc` — pre-commit hooks verified everything

app/policies/notification_policy.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+class NotificationPolicy < ApplicationPolicy
+  def index?
```
viamin, line 4: The base `ApplicationPolicy#index?` returns `user_in_account?`, but this override returns `true` unconditionally. While `policy_scope` already restricts records to the user's account, this weakens defense-in-depth. Same concern applies to `mark_all_read?` on line 16. Consider keeping these consistent with the base policy:

```suggestion
  def index?
    user_in_account?
  end
```

app/policies/notification_policy.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def read?
+    user_in_account?
+  end
+
+  def dismiss?
+    user_in_account?
+  end
+
+  def mark_all_read?
```
viamin, line 16: This is a write operation (`update_all`) — it should check `user_in_account?` rather than returning `true` unconditionally:

```suggestion
  def mark_all_read?
    user_in_account?
  end
```

db/migrate/20260408032317_create_notifications.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      t.timestamps
+    end
+
+    add_index :notifications, [ :account_id, :read_at, :dismissed_at ], name: "index_notifications_on_unread"
+    add_index :notifications, [ :account_id, :source, :subject_type, :subject_id ],
+      unique: true, name: "index_notifications_on_dedup"
+    add_index :notifications, [ :account_id, :nav_section, :read_at ], name: "index_notifications_on_badge"
+    add_index :notifications, [ :subject_type, :subject_id ], name: "index_notifications_on_subject"
+
```
viamin, line 29: Per CLAUDE.md: "Index all foreign keys and frequently queried columns." The `user_id` column has a foreign key but no index. This will cause slow deletes on the `users` table and slow lookups for user-scoped notifications. Add:

```ruby
add_index :notifications, :user_id
```

(The existing composite indexes don't cover `user_id`-only lookups.)

app/views/notifications/_bell.html.erb
```diff
@@ -0,0 +1,42 @@
+<div id="notification_bell" data-controller="notification-dropdown" class="relative">
+  <button type="button" data-action="notification-dropdown#toggle" class="relative rounded-full p-1 text-gray-500 hover:text-gray-700 focus:outline-none">
+    <span class="sr-only">View notifications</span>
+    <svg class="h-5 w-5" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor">
+      <path stroke-linecap="round" stroke-linejoin="round" d="M14.857 17.082a23.848 23.848 0 005.454-1.31A8.967 8.967 0 0118 9.75v-.7V9A6 6 0 006 9v.75a8.967 8.967 0 01-2.312 6.022c1.733.64 3.56 1.085 5.455 1.31m5.714 0a24.255 24.255 0 01-5.714 0m5.714 0a3 3 0 11-5.714 0" />
+    </svg>
+    <% unread_count = account.notifications.active.unread.count %>
```
viamin: This partial executes two separate queries: `account.notifications.active.unread.count` (line 7) and `account.notifications.active.unread.recent.limit(10)` (line 26). Since the dropdown always loads the 10 recent notifications, you could load them once and derive the count:

```erb
<% recent_notifications = account.notifications.active.unread.recent.limit(10) %>
<% unread_count = account.notifications.active.unread.count %>
```

Or better yet, load the records and use a single count query, rather than also loading 10 records every time the bell renders (even when the dropdown is closed). Consider using a `turbo-frame` with lazy loading for the dropdown content so the 10-record query only fires when the dropdown opens.

app/views/notifications/_notification.html.erb
```diff
@@ -0,0 +1,38 @@
+<%= tag.tr id: dom_id(notification), class: "#{notification.read_at? ? 'bg-white' : 'bg-indigo-50'}" do %>
+  <td class="whitespace-nowrap py-4 pl-4 pr-3 text-sm sm:pl-6">
+    <div class="flex items-center gap-2">
+      <% case notification.severity %>
+      <% when "info" %>
+        <span class="inline-flex items-center rounded-full bg-blue-100 px-2 py-0.5 text-xs font-medium text-blue-800">Info</span>
+      <% when "warning" %>
+        <span class="inline-flex items-center rounded-full bg-yellow-100 px-2 py-0.5 text-xs font-medium text-yellow-800">Warning</span>
+      <% when "error" %>
+        <span class="inline-flex items-center rounded-full bg-red-100 px-2 py-0.5 text-xs font-medium text-red-800">Error</span>
```
viamin, line 10: The `description` field supports markdown per the issue spec, but here it's rendered as plain text with `<%= notification.description %>`. If markdown support is intended, you'll need a markdown renderer. If not, the issue spec should be updated. Either way, worth clarifying intent.

app/services/notifications/publish.rb
```diff
@@ -0,0 +1,66 @@
...
+        subject: subject
+      )
+
+      notification.assign_attributes(
+        severity: severity,
+        title: title,
+        description: description,
+        metadata: metadata,
+        action_url: action_url,
+        nav_section: nav_section,
```
viamin, line 38: The `broadcast` method here is identical to the one in `Notifications::Resolve` (lines 38-49 in resolve.rb). Consider extracting this into a shared module or private method in a base class to avoid drift between the two:

```ruby
# e.g., in a shared module
module Notifications
  module Broadcasting
    private

    def broadcast_notification_updates(account)
      Turbo::StreamsChannel.broadcast_replace_to(
        account, :notification_updates,
        target: "notification_bell",
        partial: "notifications/bell",
        locals: { account: account }
      )
      Turbo::StreamsChannel.broadcast_replace_to(
        account, :notification_updates,
        target: "notification_nav_badges",
        partial: "notifications/nav_badges",
        locals: { account: account }
      )
    end
  end
end
```

app/views/notifications/index.html.erb
```diff
@@ -0,0 +1,64 @@
+<% content_for(:title) { "Notifications" } %>
```
viamin, line 2: Both CLAUDE.md ("Views use Phlex for pure Ruby components") and the issue spec ("Phlex view per `docs/STYLE_GUIDE.md`") call for Phlex views. All the view templates in this PR are ERB. Consider converting at least the index page and partials to Phlex components for consistency with the project's direction.

app/views/notifications/_nav_badges.html.erb
```diff
@@ -0,0 +1,3 @@
+<% # This partial is rendered inside a turbo frame for live updates %>
+<% badge_counts = account.notifications.active.unread.where.not(nav_section: nil).group(:nav_section).count %>
```
viamin: This partial renders badge counts as a JSON `data-` attribute but there's no JavaScript that reads `data-badge-counts` to actually inject badges into the nav items. The badge data is computed and serialized but never displayed. Is this a placeholder for future JS, or is the rendering logic missing?

app/models/notification.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class Notification < ApplicationRecord
+  belongs_to :account
+  belongs_to :user, optional: true
+  belongs_to :subject, polymorphic: true, optional: true
+
+  NAV_SECTIONS = %w[dashboard projects agent_runs providers].freeze
+
```
viamin, line 9: Minor: the `validate: true` option on the enum was added in Rails 7.1 and is good practice — nice to see it here. Consider also adding a validation on `action_url` format to reject `javascript:` URIs if one is provided, since this value is used directly in `link_to`:

```ruby
validate :action_url_is_safe, if: -> { action_url.present? }

private

def action_url_is_safe
  errors.add(:action_url, "must be a path or HTTP(S) URL") unless action_url.start_with?("/", "http://", "https://")
end
```

viamin (commented):
## Notification Center Review

Solid implementation of the notification framework. The data model, deduplication strategy, and Turbo Stream integration are well done. A few issues to address:

**Must fix:**
- Policy methods `index?` and `mark_all_read?` return `true` unconditionally, weakening defense-in-depth vs. the base policy
- Missing index on `user_id` FK (CLAUDE.md: "Index all foreign keys")
- Views use ERB instead of Phlex (both CLAUDE.md and the issue spec call for Phlex)

**Should fix:**
- Bell partial fires two separate queries that could be consolidated
- Broadcast logic is duplicated between Publish and Resolve services

**Nice to have:**
- Validate `action_url` format to prevent `javascript:` URIs

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+class NotificationsController < ApplicationController
+  def index
+    authorize Notification
+    scope = policy_scope(Notification).visible.recent
+
+    scope = scope.unread if params[:filter] == "unread"
+    scope = scope.where(severity: params[:severity]) if params[:severity].present?
+    scope = scope.where(source: params[:source]) if params[:source].present?
```
viamin, line 17: The `severity` and `source` params are passed directly to `where` clauses. While ActiveRecord parameterizes these (no SQL injection risk), an invalid severity value like `severity=bogus` would return an empty set silently rather than signaling an error. Since `severity` is an integer-backed enum, passing a string that isn't a valid enum key will raise an `ArgumentError` in Rails. Consider rescuing that or validating the param:

```ruby
scope = scope.where(severity: params[:severity]) if params[:severity].in?(Notification.severities.keys)
```

viamin:
## Agent Update

Commit successful. **5667 examples, 0 failures**, lint passes, 94.04% coverage. Here's a summary of all changes:

## Changes made

**Merge conflicts:** Merged `origin/main` cleanly (no conflicts).

**Code review fixes (all 10 threads addressed):**

1. **NotificationPolicy** — `index?` and `mark_all_read?` now use `user_in_account?` instead of returning `true` unconditionally
2. **Migration** — Added standalone `user_id` index via new migration for efficient user-scoped lookups
3. **Bell partial** — Replaced the double-query with a Turbo Frame (`loading: :lazy`) that only loads dropdown items when the dropdown opens
4. **Description rendering** — Uses `simple_format` for basic text formatting (no markdown gem in the project)
5. **Broadcast DRY** — Extracted `Notifications::Broadcasting` module shared by both `Publish` and `Resolve`
6. **Phlex views** — Added `TODO(#912)` comment; Phlex gem isn't in the Gemfile yet, so conversion deferred
7. **Nav badges** — Created `nav_badges_controller.js` Stimulus controller + added `data-nav-section` attributes to layout nav links to actually display badge counts
8. **Notification model** — Added `action_url_is_safe` validation rejecting `javascript:` URIs
9. **Controller** — Severity param validated against enum keys before querying
10. **New dropdown endpoint** — Controller handles `dropdown=true` param to serve lazy-loaded content

**Tests added:** 6 new specs covering action_url validation, severity param filtering, and the dropdown endpoint.

app/javascript/controllers/index.js
```diff
@@ -37,6 +37,9 @@ application.register("local-time", LocalTimeController)
 import MobileMenuController from "./mobile_menu_controller"
 application.register("mobile-menu", MobileMenuController)
 
+import NotificationDropdownController from "./notification_dropdown_controller"
+application.register("notification-dropdown", NotificationDropdownController)
+
```
viamin, line 45: **Bug**: `nav_badges_controller.js` is defined but never imported or registered here. The `_nav_badges.html.erb` partial uses `data-controller="nav-badges"` which will silently fail because Stimulus can't find the controller.

```suggestion
import NotificationDropdownController from "./notification_dropdown_controller"
application.register("notification-dropdown", NotificationDropdownController)

import NavBadgesController from "./nav_badges_controller"
application.register("nav-badges", NavBadgesController)
```

app/views/notifications/_bell.html.erb
```diff
@@ -0,0 +1,38 @@
...
+    <% if unread_count > 0 %>
```
viamin: **Performance**: This `COUNT` query runs on every page load for every authenticated user (the partial is rendered in the application layout). As notification volume grows this could become a bottleneck.

Consider:
- A counter cache on `Account` for active unread notifications
- Fragment caching with a key that busts on notification changes
- Or at minimum, passing the count from the controller rather than computing it in the partial (so it can be memoized)

app/views/notifications/_nav_badges.html.erb
```diff
@@ -0,0 +1,5 @@
...
```
viamin: Same performance concern — this `GROUP BY` query also runs on every page load. Combined with the bell count query, that's two notification queries per page request for all authenticated users.

app/services/notifications/publish.rb
```diff
@@ -0,0 +1,52 @@
...
+        user: user,
+        resolved_at: nil
```
viamin: **Behavior gap**: This clears `resolved_at` when re-publishing, but doesn't clear `dismissed_at`. If a user dismisses a "PR stuck in draft" notification, then the same condition re-fires, the notification gets updated but stays dismissed (invisible). The user would never see it recur.

Was this intentional? If not:

```suggestion
        action_url: action_url,
        nav_section: nav_section,
        user: user,
        resolved_at: nil,
        dismissed_at: nil
```

viamin (commented):
## Notification Center Review

Solid implementation that covers the full spec from #912 — data model, deduplication, Turbo Stream broadcasting, bell dropdown, nav badges, and good test coverage. A few issues to address:

### Bug
- **`nav_badges_controller.js` is never registered** — the file exists but `index.js` doesn't import it, so `data-controller="nav-badges"` in the partial is a dead reference. Nav badge counts won't render.

### Performance
- The bell partial runs `account.notifications.active.unread.count` on **every page load** for authenticated users. Consider a counter cache or Russian-doll caching strategy.

### Behavior
- `Notifications::Publish` clears `resolved_at` when re-publishing but does **not** clear `dismissed_at`. If a user dismisses a notification and the same condition recurs, the notification stays invisible. This seems like it should be reset.

### Scope
- The `schema.rb` diff includes ~500 lines of whitespace-only reformatting (`[ "col" ]` → `["col"]`). This is noise that makes the diff harder to review and can cause merge conflicts. Consider reverting the whitespace changes and only keeping the notifications table addition.

Overall this is well-structured and aligns with the issue requirements. The missing controller registration is the most critical fix needed.

app/models/notification.rb
```diff
@@ -0,0 +1,30 @@
...
+  scope :undismissed, -> { where(dismissed_at: nil) }
+  scope :unresolved, -> { where(resolved_at: nil) }
+  scope :active, -> { undismissed.unresolved }
+  scope :visible, -> { undismissed }
+  scope :for_nav_section, ->(section) { where(nav_section: section) }
+  scope :recent, -> { order(created_at: :desc) }
+
+  private
+
+  def action_url_is_safe
```
viamin, line 27: This validation protects against `javascript:` URIs, which is good. However, `action_url` is only set server-side by `Notifications::Publish`, never from user input. If that guarantee holds, the model validation is defense-in-depth, which is fine.

One edge case: a relative path like `//evil.com` (protocol-relative URL) would be rejected since it starts with `//` not `/` followed by a non-slash. Just confirming that's the intended behavior — looks correct.

app/policies/notification_policy.rb
```diff
@@ -0,0 +1,25 @@
...
+    user_in_account?
+  end
+
+  private
+
+  def account_for_record
+    record.is_a?(Class) ? user&.account : record.account
+  end
+end
```
viamin, line 35: Minor: no `Scope` class defined here, so `policy_scope(Notification)` falls through to `ApplicationPolicy::Scope`, which scopes by `account_id`. This works correctly, but other policies in the codebase (e.g., `AgentRunPolicy`, `AbTestPolicy`) explicitly define their scope. Consider adding one for clarity and to future-proof against `ApplicationPolicy::Scope` changes.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_04_08_004407) do
+ActiveRecord::Schema[8.1].define(version: 2026_04_09_184503) do
```
viamin: The actual notifications table addition is only ~20 lines, but this file has ~500 lines of whitespace reformatting (`[ "col" ]` → `["col"]`). This was likely caused by running `db:migrate` with a different Ruby/Rails version than what generated the existing schema.

Consider reverting the whitespace changes (e.g., `git checkout origin/main -- db/schema.rb && bin/rails db:migrate`) to keep the diff focused on the notifications table. The noise makes review harder and increases merge conflict risk.

viamin:
## Agent Update

Commit succeeded. All pre-commit hooks passed: lint clean, **5668 examples, 0 failures**, 94.04% coverage.

Here's a summary of all changes addressing the review feedback:

1. **NavBadgesController registration** — Added the missing `import` and `application.register` call in `app/javascript/controllers/index.js` so the `nav-badges` Stimulus controller actually loads.

2. **Performance: bell count + nav badges queries** — Extracted both queries into memoized helper methods in `NotificationsHelper` (`unread_notification_count` and `notification_badge_counts`). The partials now call these helpers instead of querying directly, ensuring at most one execution per request.

3. **dismissed_at not cleared on re-publish** — Added `dismissed_at: nil` to `Notifications::Publish#call` so that when a condition re-fires, previously dismissed notifications become visible again. Added a matching test.

4. **action_url validation** — Reviewer confirmed the existing validation is correct (defense-in-depth). No change needed.

5. **NotificationPolicy::Scope** — Added an explicit `Scope` class consistent with other policies in the codebase (`AgentRunPolicy`, `AbTestPolicy`, etc.).

6. **schema.rb whitespace noise** — Reverted the ~500 lines of whitespace reformatting by checking out `origin/main`'s schema and manually adding only the notifications table definition and foreign keys.

app/models/notification.rb
viamin, line 27: **Security: open-redirect via protocol-relative URL.** `"//evil.com".start_with?("/")` returns `true`, so an attacker who can influence `action_url` (e.g. via a crafted `source` or upstream data) can inject `//evil.com/phish` which browsers interpret as a protocol-relative URL. This bypasses the intent of this validation and creates an open-redirect when rendered via `link_to`.

```suggestion
    errors.add(:action_url, "must be a path or HTTP(S) URL") unless action_url.match?(%r{\A(/[^/]|https?://)}) || action_url == "/"
```

This rejects `//...` (protocol-relative) while still allowing root path `/` and normal paths like `/projects/1`.

viamin (commented):
Solid feature — data model, dedup strategy, Turbo wiring, and test coverage all look well thought out. One security issue to address before merging.

viamin:
## Agent Update

Merged `origin/main`, resolved the schema conflict, addressed the unresolved review thread in [app/models/notification.rb](/workspace/app/models/notification.rb), and committed the result as `a38049ad98dda5985a6c5343ed83861d50838fec`.

The substantive fix was tightening `action_url` validation to reject protocol-relative URLs like `//evil.com` while still allowing `/`, normal relative paths, and `http(s)` URLs. I added coverage for that in [spec/models/notification_spec.rb](/workspace/spec/models/notification_spec.rb). The merge also carried forward the existing branch work and the resolved schema state in [db/schema.rb](/workspace/db/schema.rb).

Verification completed:
- `RAILS_ENV=test bin/rails db:prepare`
- `bundle exec rubocop`
- `bundle exec rspec`
- pre-commit hooks passed on commit
- final suite: `5792 examples, 0 failures, 2 pending` (`Qdrant` live specs pending because `http://localhost:6333` was unavailable)

The worktree is clean. I did not push.

app/policies/notification_policy.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def mark_all_read?
+    user_in_account?
+  end
+
+  class Scope < ApplicationPolicy::Scope
+    def resolve
+      raise Pundit::NotAuthorizedError, "must be logged in" unless user
+
+      scope.where(account: user.account)
```
viamin: The issue explicitly calls out per-user notifications via `user_id` (for example, token-expiry alerts). As written, every operator in the account can list/read/dismiss another user's personal notifications because the scope only filters on `account`. Please scope this to account-wide notifications plus the signed-in user's own notifications.

```suggestion
      scope.where(account: user.account, user_id: nil).or(
        scope.where(account: user.account, user_id: user.id)
      )
```

app/helpers/notifications_helper.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module NotificationsHelper
+  def unread_notification_count
+    return 0 unless current_account
+
+    @_unread_notification_count ||= current_account.notifications.active.unread.count
```
viamin: The bell and nav badge counts bypass `policy_scope` and still count every unread notification in the account. That means another operator's personal notification will appear in my navbar even after the controller scope is fixed. Please derive both helper queries from the same `user_id IS NULL OR user_id = current_user.id` relation instead of `current_account.notifications`.

app/services/notifications/publish.rb
```diff
@@ -0,0 +1,53 @@
...
+      @title = title
+      @description = description
+      @metadata = metadata
+      @action_url = action_url
+      @nav_section = nav_section
+      @user = user
+    end
+
+    def call
+      notification = Notification.find_or_initialize_by(
```
viamin, line 25: `Notifications::Publish` needs to dedupe personal notifications per `(account, user, source, subject)`, not just `(account, source, subject)`. Right now publishing the same rule/subject for user B will overwrite user A's notification instead of creating a separate row.

```suggestion
      notification = Notification.find_or_initialize_by(
        account: account,
        user: user,
        source: source,
        subject: subject
      )
```

db/migrate/20260408032317_create_notifications.rb
```diff
@@ -0,0 +1,33 @@
...
+      t.string :nav_section
+      t.datetime :read_at
+      t.datetime :dismissed_at
+      t.datetime :resolved_at
+
+      t.timestamps
+    end
+
+    add_index :notifications, [ :account_id, :read_at, :dismissed_at ], name: "index_notifications_on_unread"
+    add_index :notifications, [ :account_id, :source, :subject_type, :subject_id ],
```
viamin, line 25: The unique index needs the same `user_id` dimension as the service lookup. Without it, even if the Ruby code is fixed, the database will still reject two personal notifications for different users that share the same rule/subject.

```suggestion
    add_index :notifications, [ :account_id, :user_id, :source, :subject_type, :subject_id ],
      unique: true, name: "index_notifications_on_dedup"
```

viamin (commented):
Found a few actionable correctness issues around per-user notifications. The current scope, navbar counts, and dedup key/index all treat notifications as account-wide, which leaks personal alerts across operators and causes one user's notification to overwrite another's.

viamin:
## Agent Update

Commit succeeded. All pre-commit hooks passed: lint clean, **5800 examples, 0 failures**, 94.0% coverage.

Here's a summary of what was done:

**Merge conflicts:** Merged `origin/main` cleanly — no conflicts.

**Code review fixes (all 4 threads addressed):**

1. **NotificationPolicy::Scope** — Now scopes to account-wide notifications (`user_id: nil`) OR the signed-in user's own notifications, preventing operators from seeing other users' personal notifications.

2. **NotificationsHelper** — Both `unread_notification_count` and `notification_badge_counts` now derive from a `visible_notifications` helper method that applies the same `user_id IS NULL OR user_id = current_user.id` filter.

3. **Notifications::Publish** — Added `user` to `find_or_initialize_by` so personal notifications for different users get separate rows instead of overwriting each other.

4. **Dedup index migration** — New migration replaces the unique index to include `user_id`, matching the service lookup key.

**Proactive fix:** Also scoped `Notifications::Resolve` to include the optional `user` parameter in its lookup, consistent with the Publish service.

**Tests added:** 4 new specs covering the `NotificationPolicy::Scope` — verifying account-wide inclusion, personal notification visibility, cross-user isolation, and cross-account isolation.

viamin (commented):
Overall this is a solid notification center implementation that closely follows the issue spec. Good coverage with tests, proper multi-tenant scoping via Pundit, XSS protection on action_url, and idempotent publish/resolve lifecycle. Two actionable items below.

app/helpers/notifications_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+    @_notification_badge_counts ||= visible_notifications.active.unread
+      .where.not(nav_section: nil).group(:nav_section).count
+  end
+
+  private
+
+  def visible_notifications
+    current_account.notifications
+      .where(user_id: nil).or(current_account.notifications.where(user_id: current_user&.id))
+  end
```
viamin, line 21: This visibility logic duplicates what `NotificationPolicy::Scope#resolve` does (both construct the same `where(user_id: nil).or(where(user_id: user.id))` query). If the scoping rules ever change, they'd need to be updated in both places. Consider delegating to the policy scope to keep a single source of truth:

```suggestion
  def visible_notifications
    NotificationPolicy::Scope.new(current_user, current_account.notifications).resolve
  end
```

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+    authorize Notification
+    policy_scope(Notification).active.unread.update_all(read_at: Time.current)
+
+    respond_to do |format|
+      format.turbo_stream do
+        render turbo_stream: [
+          turbo_stream.replace("notification_bell", partial: "notifications/bell", locals: { account: current_account }),
+          turbo_stream.replace("notification_nav_badges", partial: "notifications/nav_badges", locals: { account: current_account })
+        ]
+      end
```
viamin, line 68: The `mark_all_read` turbo_stream response replaces the bell and nav badges, but doesn't update the notification rows on the index page. When a user clicks "Mark all read" on the `/notifications` page with Turbo enabled, the table rows remain styled as unread (`bg-indigo-50`). The HTML fallback does a redirect which reloads the page, but the turbo_stream path skips that.

Consider either adding a turbo_stream that refreshes the notifications table body, or redirecting in the turbo_stream path as well (e.g., via `turbo_stream.action(:redirect, notifications_path)` or simply falling through to the HTML format).

viamin:
## Agent Update

Commit succeeded. All pre-commit hooks passed: lint clean, **5800 examples, 0 failures**, 93.99% coverage.

Here's a summary of what was done:

**Merge conflicts**: Merged `origin/main` cleanly — no conflicts.

**Code review fixes (both unresolved threads addressed):**

1. **NotificationsHelper visibility duplication** (`app/helpers/notifications_helper.rb:20`) — Replaced the inline `where(user_id: nil).or(where(user_id: current_user.id))` query with a delegation to `NotificationPolicy::Scope#resolve`, establishing a single source of truth for notification visibility rules.

2. **mark_all_read turbo_stream doesn't update rows** (`app/controllers/notifications_controller.rb:64`) — Added a turbo_stream that replaces the `notifications_list` tbody with re-rendered notification partials reflecting the updated read state. The index view's `<tbody>` now uses `tag.tbody` with an `id` attribute so Turbo can target it.

**Schema**: Regenerated `db/schema.rb` from the database to reflect columns removed by migrations in main (`priority_tier`, `restart_reason`).

app/services/notifications/broadcasting.rb
```diff
@@ -0,0 +1,23 @@
...
+  module Broadcasting
+    private
+
+    def broadcast_notification_updates(account)
+      Turbo::StreamsChannel.broadcast_replace_to(
+        account, :notification_updates,
+        target: "notification_bell",
+        partial: "notifications/bell",
+        locals: { account: account }
+      )
```
viamin, line 22: **Bug: Broadcast rendering will fail at runtime.** The `_bell.html.erb` partial calls `unread_notification_count` (from `NotificationsHelper`), which depends on `current_account` and `current_user` — Devise/Warden helpers that only exist inside an HTTP request context. When `broadcast_replace_to` renders this partial from a service or background job, those helpers are undefined and this will raise `NoMethodError` (or silently return 0 for the count and skip the policy scope entirely).

The same problem applies to the `_nav_badges` partial which calls `notification_badge_counts`.

The `account` local is already passed but the partials don't use it for counting. The fix is to compute the counts in the broadcasting code and pass them as locals, then have the partials read from locals instead of calling the request-scoped helpers.

For example:
```suggestion
      count = Notification.where(account: account, dismissed_at: nil, resolved_at: nil, read_at: nil).count
      Turbo::StreamsChannel.broadcast_replace_to(
        account, :notification_updates,
        target: "notification_bell",
        partial: "notifications/bell",
        locals: { account: account, unread_count: count }
      )
```
Then update `_bell.html.erb` to use the `unread_count` local (with a fallback to the helper for controller-rendered contexts), and do the same for badge counts in `_nav_badges`.

db/migrate/20260410204751_fix_notifications_dedup_index_add_user_id.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class FixNotificationsDedupIndexAddUserId < ActiveRecord::Migration[8.1]
+  def change
+    remove_index :notifications, name: "index_notifications_on_dedup"
+    add_index :notifications, [ :account_id, :user_id, :source, :subject_type, :subject_id ],
+      unique: true, name: "index_notifications_on_dedup"
```
viamin, line 7: **Bug: Unique index does not deduplicate when `user_id` is NULL.** PostgreSQL treats NULL values as distinct in unique indexes, so two rows with `user_id = NULL` and the same `(account_id, source, subject_type, subject_id)` will both be allowed. Since account-wide notifications (the common case per the issue spec) have `user_id = NULL`, the dedup constraint is effectively unenforced for most notifications.

Under concurrent calls to `Notifications::Publish` (e.g., two overlapping detection runs), `find_or_initialize_by` won't protect against duplicates because both processes can see no existing row and both insert.

Fix with a `UNIQUE` partial index that covers the NULL case:
```ruby
add_index :notifications,
  [:account_id, :source, :subject_type, :subject_id],
  unique: true,
  where: "user_id IS NULL",
  name: "index_notifications_on_dedup_account_wide"
```
Keep the existing index for the non-NULL `user_id` case.

viamin (commented):
Overall this is a well-structured notification system that matches the issue spec closely. Two issues need addressing — one will cause runtime errors, one is a data integrity gap.

viamin:
## Agent Update

Commit succeeded. **5812 examples, 0 failures**, lint clean, 94.01% coverage.

Here's a summary of what was done:

**Merge conflicts:** Merged `origin/main` cleanly — no conflicts.

**Code review fixes (both unresolved threads addressed):**

1. **Broadcast rendering fails at runtime** (`app/services/notifications/broadcasting.rb`) — The bell and nav_badges partials called `unread_notification_count` and `notification_badge_counts` helpers, which depend on Devise's `current_user`/`current_account` — unavailable in background broadcast contexts. Fixed by computing counts directly in the `Broadcasting` module and passing them as locals. The partials now use `local_assigns.key?` to prefer locals when available, falling back to helpers when rendered from controllers.

2. **Dedup index doesn't work for NULL user_id** (`db/migrate/20260410204751_fix_notifications_dedup_index_add_user_id.rb`) — PostgreSQL treats NULL values as distinct in unique indexes, so the existing dedup index with `user_id` column didn't prevent duplicate account-wide notifications. Added a partial unique index with `WHERE user_id IS NULL` to cover the common case of account-wide notifications.

viamin (commented):
Solid notification center implementation — good multi-tenant scoping, dedup indexing, and Turbo Stream wiring. Two actionable issues found: a race condition in the Publish service and a missing user-side association that will block user deletion.

app/services/notifications/publish.rb
```diff
@@ -0,0 +1,54 @@
...
+        title: title,
+        description: description,
+        metadata: metadata,
+        action_url: action_url,
+        nav_section: nav_section,
+        user: user,
+        resolved_at: nil,
+        dismissed_at: nil
+      )
+
```
viamin, line 44: **Bug (race condition):** `find_or_initialize_by` + `save!` is not atomic. Two concurrent calls with the same `(account, source, subject)` key will both get a new unsaved record, and the second `save!` will raise `ActiveRecord::RecordNotUnique` against the dedup unique index.

Wrap in a rescue/retry:

```suggestion
      notification = Notification.find_or_initialize_by(
        account: account,
        user: user,
        source: source,
        subject: subject
      )

      notification.assign_attributes(
        severity: severity,
        title: title,
        description: description,
        metadata: metadata,
        action_url: action_url,
        nav_section: nav_section,
        user: user,
        resolved_at: nil,
        dismissed_at: nil
      )

      notification.save!
    rescue ActiveRecord::RecordNotUnique
      retry
```

db/migrate/20260408032317_create_notifications.rb
```diff
@@ -0,0 +1,33 @@
...
+      unique: true, name: "index_notifications_on_dedup"
+    add_index :notifications, [ :account_id, :nav_section, :read_at ], name: "index_notifications_on_badge"
+    add_index :notifications, [ :subject_type, :subject_id ], name: "index_notifications_on_subject"
+
+    add_foreign_key :notifications, :accounts
+    add_foreign_key :notifications, :users
+  end
+end
```
viamin, line 33: The FK `add_foreign_key :notifications, :users` has no `on_delete` strategy, and the `User` model has no `has_many :notifications` with a `dependent:` option. This means deleting a user who has notifications will raise a `PG::ForeignKeyViolation`.

Either:
1. Add `on_delete: :nullify` here (since `user_id` is nullable, this is the natural choice), **or**
2. Add `has_many :notifications, dependent: :nullify` to the `User` model.

Option 1 is simplest:
```suggestion
    add_foreign_key :notifications, :users, on_delete: :nullify
```

viamin:
## Agent Update

The background review agent completed. I already reviewed its findings earlier — the key issues (race condition and FK on_delete) are now fixed in the commit. The other points it raised are either already handled (action_url rejects `//evil.com` via the `/[^/]` regex), acceptable by design (accounts FK defaults to restrict, which is correct — you shouldn't silently delete an account's notifications), or low-risk (source param filtering only narrows results within the user's already-scoped notifications).

The commit is complete — all changes are committed, lint passes, and the full test suite (5864 examples, 0 failures) is green.

app/services/notifications/publish.rb
```diff
@@ -0,0 +1,56 @@
...
+        severity: severity,
+        title: title,
+        description: description,
+        metadata: metadata,
+        action_url: action_url,
+        nav_section: nav_section,
+        user: user,
+        resolved_at: nil,
+        dismissed_at: nil
+      )
```
viamin, line 43: **Bug: `read_at` not cleared on re-publish.** When a resolved or dismissed notification recurs, `resolved_at` and `dismissed_at` are correctly cleared so the notification becomes active again. However, `read_at` is not reset, so the re-activated notification remains marked as "read" and won't appear in unread counts, bell badges, or the dropdown (which filters `.active.unread`).

This means if a user reads a "PR stuck for 3 hours" warning, the condition resolves, and then recurs as a "PR stuck for 6 hours" error — the escalated notification is invisible in all unread surfaces.

```suggestion
      notification.assign_attributes(
        severity: severity,
        title: title,
        description: description,
        metadata: metadata,
        action_url: action_url,
        nav_section: nav_section,
        user: user,
        read_at: nil,
        resolved_at: nil,
        dismissed_at: nil
      )
```

viamin (commented):
Solid notification center implementation — data model, policy scoping, dedup strategy, and real-time broadcasting all look well-designed. One correctness issue to address around notification re-publishing.

viamin:
## Agent Update

Commit succeeded. All pre-commit hooks passed: lint clean, **5909 examples, 0 failures**, 2 pending.

Here's a summary of what was done:

1. **Merged `origin/main`** — no conflicts
2. **Fixed `read_at` not cleared on re-publish** (`app/services/notifications/publish.rb:39`) — Added `read_at: nil` to `assign_attributes` so that when a notification recurs (e.g., escalates from warning to error), it appears as unread again in bell badges, unread counts, and the dropdown
3. **Added test coverage** (`spec/services/notifications/publish_spec.rb:84`) — New spec verifying `read_at` is cleared when re-publishing a previously read notification
4. **Regenerated `db/schema.rb`** from the database to reflect the current state after merge

app/services/notifications/publish.rb
```diff
@@ -0,0 +1,57 @@
...
+        read_at: nil,
+        resolved_at: nil,
+        dismissed_at: nil
+      )
+
+      notification.save!
+      broadcast_notification_updates(account)
+      notification
+    rescue ActiveRecord::RecordNotUnique
+      retry
```
viamin: The unconditional `retry` on `RecordNotUnique` can loop forever if the conflicting row is repeatedly inserted and deleted between `find_or_initialize_by` and `save!` (unlikely but possible under high concurrency). Add a retry limit:

```suggestion
    rescue ActiveRecord::RecordNotUnique
      retries ||= 0
      retry if (retries += 1) < 3
      raise
    end
```

spec/services/notifications/publish_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+        source: "test_rule",
+        subject: project,
+        severity: :info,
+        title: "Test"
+      )
+
+      expect(Turbo::StreamsChannel).to have_received(:broadcast_replace_to).at_least(:once)
+    end
+  end
+end
```
viamin, line 138: There is no spec for `Notifications::Resolve`. It has non-trivial behavior (finding only unresolved notifications, setting `resolved_at`, broadcasting) that should be covered — especially the no-op case when the notification is already resolved or doesn't exist.

viamin (commented):
Solid implementation of the notification center — good multi-tenant scoping via Pundit, proper dedup indexes, XSS protection on action_url, and Turbo Stream broadcasting. Two actionable items below.

viamin:
## Agent Update

The background task notification is stale — that was the first commit attempt that failed due to flaky tests. The second commit attempt already succeeded:

```
[paid/912-featnotifications-notification-center-with-bell-ic-340f1c 92fa994d] fix(notifications): add retry limit to RecordNotUnique rescue and expand Resolve specs
 2 files changed, 33 insertions(+), 1 deletion(-)
```

All work is committed. The branch is clean.

viamin (commented):
Generated no new comments. Clean implementation of the notification center. The data model, deduplication strategy (unique indexes + partial index for NULL user_id + RecordNotUnique retry), multi-tenant scoping via Pundit policy, Turbo Stream broadcasting, and action_url validation all look solid. Tests cover the key paths including scope isolation, idempotent publish, resolve lifecycle, and multi-tenant request specs.

PR #932, opened 2026-04-08
fix(dev-cleanup): don't trip provider circuit breaker when cleanup kills runs

### LINKED ISSUE #931: dev:cleanup trips provider circuit breakers when killing in-flight runs
## Problem

## Repro

## Fix

viamin:
Fixes #931

## Summary

- `bin/rails dev:cleanup` now tags force-timed-out runs with `AgentRun::STALE_CLEANUP_ERROR_PREFIX`.
- `RunAgentActivity` rescue branches reload the run and skip `record_provider_failure` (and break out of the providers loop) when `cancelled_by_cleanup?` — so killing in-flight runs no longer poisons healthy providers' circuit breakers.
- Cleanup-induced failures are recorded as `error_type: "cancelled_by_cleanup"` so they remain visible in the UI without implying a real provider problem.

## Test plan

- [x] New specs in `spec/temporal/activities/run_agent_activity_spec.rb`:
  - breaker `failure_count` stays at 0 when cleanup races with an in-flight run
  - the attempt is recorded as `cancelled_by_cleanup`
  - the cleanup-set `status: "timeout"` and `error_message` are preserved
  - with `fallback_enabled: true`, the activity stops after one attempt instead of cascading through fallbacks
- [x] Tests use `update_all` to mutate the run from "another process," so the `reload` inside `cancelled_by_cleanup?` is genuinely exercised
- [x] Full `run_agent_activity_spec.rb` (84 examples) green
- [x] `agent_run_spec.rb` and `dev_rake_spec.rb` (256 examples combined) green
- [x] Rubocop clean

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -14,6 +14,14 @@ class AgentRun < ApplicationRecord
   TOKEN_LIMIT_STATUSES = %w[ok warning exceeded].freeze
   DEFAULT_MAX_TOKENS_PER_RUN = 10_000_000
 
+  # Sentinel prefix written into AgentRun#error_message by `bin/rails dev:cleanup`
+  # when it forcibly times out an in-flight run because the host process is being
+  # restarted (e.g. `bin/setup --skip-server`). Code that observes a run failing
+  # while marked with this prefix should treat the failure as caused by the
+  # cleanup, not by the provider — in particular, do not increment provider
+  # circuit-breaker counters on its behalf.
+  STALE_CLEANUP_ERROR_PREFIX = "Marked stale on startup"
+
   belongs_to :project
   belongs_to :issue, optional: true
   belongs_to :prompt_version, optional: true
@@ -649,6 +657,15 @@ class AgentRun < ApplicationRecord
       error_message: error,
       duration_seconds: duration
     )
+  end
+
+  # True when this run was force-timed-out by `dev:cleanup` (typically because
+  # `bin/setup --skip-server` is restarting the host process and stopping the
+  # run's container out from under the in-flight Temporal activity). Used by
+  # RunAgentActivity to suppress provider circuit-breaker bookkeeping for
+  # failures that the cleanup itself induced.
+  def cancelled_by_cleanup?
+    status == "timeout" && error_message.to_s.start_with?(STALE_CLEANUP_ERROR_PREFIX)
   end
 
   def retried?
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -296,6 +296,7 @@ module Activities
           agent_run.timeout!(error: timeout_error) unless agent_run.finished?
           # Skip queue processing when cleanup killed the run — the timeout
           # was not a real provider issue, so there is nothing to re-schedule.
+          # (agent_run was reloaded above, so the model method sees current state)
           ProcessRunQueueJob.perform_later unless agent_run.cancelled_by_cleanup?
         elsif !agent_run.finished? && (last_error == "rate_limited" || all_skipped_rate_limited)
           provider_list = providers.any? ? provider_attempt_labels(providers, agent_run, user_settings.user).join(", ") : "none"
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -294,7 +294,9 @@ module Activities
         # ProcessRunQueueJob to re-schedule work.
         if timeout_error.present?
           agent_run.timeout!(error: timeout_error) unless agent_run.finished?
-          ProcessRunQueueJob.perform_later
+          # Skip queue processing when cleanup killed the run — the timeout
+          # was not a real provider issue, so there is nothing to re-schedule.
+          ProcessRunQueueJob.perform_later unless agent_run.cancelled_by_cleanup?
         elsif !agent_run.finished? && (last_error == "rate_limited" || all_skipped_rate_limited)
           provider_list = providers.any? ? provider_attempt_labels(providers, agent_run, user_settings.user).join(", ") : "none"
           agent_run.rate_limit!(
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -248,12 +248,20 @@ module Activities
           rescue ProviderTimeoutError => e
             last_error = "timeout"
             timeout_error ||= e.message
+            if cancelled_by_cleanup?(agent_run)
+              record_cleanup_cancelled_attempt(agent_run, attempt_label, provider, e)
+              break
+            end
             record_provider_failure(user_settings, provider_state_name, provider_states)
             agent_run.record_provider_attempt(attempt_label, success: false, error_type: "timeout")
             logger.warn(message: "agent_execution.provider_timeout", provider: provider, agent_run_id: agent_run.id, error: e.message)
             break
           rescue ProviderExecutionError => e
             last_error = "error"
+            if cancelled_by_cleanup?(agent_run)
+              record_cleanup_cancelled_attempt(agent_run, attempt_label, provider, e)
+              break
+            end
             record_provider_failure(user_settings, provider_state_name, provider_states)
             agent_run.record_provider_attempt(attempt_label, success: false, error_type: "error")
             logger.warn(message: "agent_execution.provider_failed", provider: provider, agent_run_id: agent_run.id, error: e.message)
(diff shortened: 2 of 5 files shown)


app/temporal/activities/run_agent_activity.rb
```diff
@@ -248,12 +252,20 @@ def execute(input)
           rescue ProviderTimeoutError => e
             last_error = "timeout"
             timeout_error ||= e.message
```
viamin, line 254: **Nit: unnecessary `ProcessRunQueueJob` dispatch on cleanup-cancelled timeouts**

When cleanup cancels a run mid-`ProviderTimeoutError`, `timeout_error` gets set (line 253) and then the post-loop handler at lines 295-297 fires:

```ruby
if timeout_error.present?
  agent_run.timeout!(error: timeout_error) unless agent_run.finished?
  ProcessRunQueueJob.perform_later  # <-- runs unconditionally
end
```

The `agent_run.timeout!` is correctly guarded by `finished?`, but `ProcessRunQueueJob.perform_later` still enqueues. This is harmless (the job will find nothing actionable), but you could short-circuit the entire block for cleanup-cancelled runs to avoid the spurious enqueue. Not blocking — just a thought for a follow-up.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -406,6 +418,31 @@ def record_provider_failure(user_settings, provider_state_name, provider_states)
       state.record_failure!(threshold: user_settings.circuit_breaker_failure_threshold)
     end
 
+    # True when the agent run we're executing has already been force-timed-out
+    # by `dev:cleanup` (e.g. `bin/setup --skip-server` killed our container).
+    # In that case the failure we just rescued was caused by the cleanup, not
+    # by the provider, so we must not penalize the provider's circuit breaker.
+    def cancelled_by_cleanup?(agent_run)
+      agent_run.reload
+      agent_run.cancelled_by_cleanup?
```
viamin, line 430: The `reload` here is the right call since cleanup happens in a different process. Worth noting this adds one DB round-trip per rescue on the error path, which is fine — error paths are infrequent and correctness matters more than micro-optimization here.

viamin (commented):
Clean, well-scoped fix with thorough tests. The sentinel-prefix approach is simple and effective, and the test strategy of using `update_all` to simulate cross-process mutation is a nice touch that exercises the `reload` path realistically.

A few minor observations below — nothing blocking.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -225,9 +225,13 @@ def execute(input)
               )
             end
           rescue InfiniteLoopError => e
+            last_error = "infinite_loop"
```
viamin, line 228: **Observation: changed error type for cleanup-cancelled infinite loops**

In the original flow, `InfiniteLoopError` would raise `Temporalio::Error::ApplicationError` with `type: "InfiniteLoopDetected"`. With the cleanup check + `break`, the flow now falls through to the post-loop section and raises with `type: "AllProvidersExhausted"` instead.

This is correct behavior (the run is dead from cleanup, not from a real infinite loop), and since the `cancelled_by_cleanup` error_type is already recorded in the provider attempt, the distinction is preserved for debugging. Just flagging the change in error propagation type for awareness.

app/models/agent_run.rb
```diff
@@ -651,6 +659,15 @@ def timeout!(error: nil)
     )
   end
 
+  # True when this run was force-timed-out by `dev:cleanup` (typically because
+  # `bin/setup --skip-server` is restarting the host process and stopping the
+  # run's container out from under the in-flight Temporal activity). Used by
+  # RunAgentActivity to suppress provider circuit-breaker bookkeeping for
+  # failures that the cleanup itself induced.
+  def cancelled_by_cleanup?
+    status == "timeout" && error_message.to_s.start_with?(STALE_CLEANUP_ERROR_PREFIX)
```
viamin, line 668: Clean predicate — the `.to_s` nil-guard is a nice defensive touch.

spec/temporal/activities/run_agent_activity_spec.rb
```diff
@@ -544,6 +544,78 @@ def create_opencode_provider_entry(user:, api_key:, name:, model:)
...
+        allow(git_ops).to receive(:head_sha).and_return("pre_agent_sha_abc123")
+        # Simulate dev:cleanup force-timing-out the run from a *different*
+        # process while the container call is in flight. We mutate via
+        # `update_columns` so the in-memory `agent_run` instance held by the
+        # activity stays stale — only the `reload` inside `cancelled_by_cleanup?`
+        # will surface the cleanup mark, which is what happens in production
+        # where the rake task and the Temporal worker are separate processes.
+        allow(container_service).to receive(:execute) do
+          AgentRun.where(id: agent_run.id).update_all(
+            status: "timeout",
```
viamin, line 558: Good use of `update_all` to bypass ActiveRecord callbacks and simulate the cross-process mutation that happens in production. This ensures the `reload` inside `cancelled_by_cleanup?` is genuinely exercised rather than reading from a pre-mutated in-memory object.

viamin:
## Agent Update

The background task failure is from the earlier commit attempt that was superseded — the commit already succeeded in the foreground (as shown in the `git log` output above). Everything is committed and clean.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -406,6 +420,31 @@ def record_provider_failure(user_settings, provider_state_name, provider_states)
       state.record_failure!(threshold: user_settings.circuit_breaker_failure_threshold)
     end
 
+    # True when the agent run we're executing has already been force-timed-out
+    # by `dev:cleanup` (e.g. `bin/setup --skip-server` killed our container).
+    # In that case the failure we just rescued was caused by the cleanup, not
+    # by the provider, so we must not penalize the provider's circuit breaker.
+    def cancelled_by_cleanup?(agent_run)
+      agent_run.reload
+      agent_run.cancelled_by_cleanup?
```
viamin, line 430: **Minor perf note:** This `reload` fires a DB query on every failed provider attempt, even in the common case where no cleanup is happening. Given that provider failures are relatively rare (and fallback chains are short), this is fine in practice, but worth noting — if the retry/fallback logic ever becomes hot, you could consider caching the cleanup status for the duration of the providers loop (a single reload at the top of each rescue).

app/temporal/activities/run_agent_activity.rb
viamin, line 228: Good call moving `last_error` assignment above the cleanup check. Without this, breaking out of the loop would leave `last_error` at its previous value (or nil on first attempt), and the post-loop status logic could misclassify the run.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -282,7 +294,9 @@ def execute(input)
         # ProcessRunQueueJob to re-schedule work.
         if timeout_error.present?
           agent_run.timeout!(error: timeout_error) unless agent_run.finished?
-          ProcessRunQueueJob.perform_later
+          # Skip queue processing when cleanup killed the run — the timeout
+          # was not a real provider issue, so there is nothing to re-schedule.
+          ProcessRunQueueJob.perform_later unless agent_run.cancelled_by_cleanup?
```
viamin, line 300: Subtle but correct: this calls the *model* method `cancelled_by_cleanup?` (no reload) rather than the activity's wrapper. It works because line 289 already reloaded `agent_run`. Worth a brief comment if you want to make the contract explicit for future readers, e.g.:

```suggestion
          # Skip queue processing when cleanup killed the run — the timeout
          # was not a real provider issue, so there is nothing to re-schedule.
          # (agent_run was reloaded above, so the model method sees current state)
          ProcessRunQueueJob.perform_later unless agent_run.cancelled_by_cleanup?
```

app/temporal/activities/run_agent_activity.rb
```diff
@@ -406,6 +420,31 @@ def record_provider_failure(user_settings, provider_state_name, provider_states)
...
 
+    # True when the agent run we're executing has already been force-timed-out
+    # by `dev:cleanup` (e.g. `bin/setup --skip-server` killed our container).
+    # In that case the failure we just rescued was caused by the cleanup, not
+    # by the provider, so we must not penalize the provider's circuit breaker.
+    def cancelled_by_cleanup?(agent_run)
+      agent_run.reload
+      agent_run.cancelled_by_cleanup?
+    rescue ActiveRecord::RecordNotFound
+      false
```
viamin, line 432: Good defensive rescue — if the run was hard-deleted between the container dying and this check, we correctly fall through to normal error handling rather than crashing.

spec/temporal/activities/run_agent_activity_spec.rb
```diff
@@ -544,6 +544,88 @@ def create_opencode_provider_entry(user:, api_key:, name:, model:)
...
```
viamin, line 558: Nice testing pattern — using `update_all` to bypass ActiveRecord callbacks/validations genuinely simulates the cross-process mutation that happens in production. This is exactly the kind of thing that would be easy to get wrong with a naive `update!` on the same in-memory instance.

app/models/agent_run.rb
```diff
@@ -14,6 +14,14 @@ class AgentRun < ApplicationRecord
   TOKEN_LIMIT_STATUSES = %w[ok warning exceeded].freeze
   DEFAULT_MAX_TOKENS_PER_RUN = 10_000_000
 
+  # Sentinel prefix written into AgentRun#error_message by `bin/rails dev:cleanup`
+  # when it forcibly times out an in-flight run because the host process is being
+  # restarted (e.g. `bin/setup --skip-server`). Code that observes a run failing
+  # while marked with this prefix should treat the failure as caused by the
```
viamin, line 20: The constant name clearly communicates intent. One thing to consider: if the cleanup message wording ever changes in `dev.rake` but this constant doesn't track it (or vice versa), the detection silently breaks. The current approach of using the constant in both places (model + rake task) prevents that — just flagging that this coupling is load-bearing.

viamin (commented):
Clean, well-scoped PR that directly addresses #931. The sentinel-prefix approach is a pragmatic way to distinguish cleanup-induced failures from real provider problems without requiring schema changes or cross-process coordination. Tests are thorough — particularly the `update_all` trick to simulate the race condition.

A few observations inline, but nothing blocking. LGTM.

viamin:
## Agent Update

That background task notification is from the earlier failed commit attempt (before assets were built). The commit already succeeded in the foreground — as confirmed by the `git status` showing a clean working tree and the commit `8eb7c29b` in the log. No action needed.

viamin (commented):
Generated no new comments. The PR cleanly addresses #931 — cleanup-induced failures are correctly tagged with a sentinel prefix, rescue branches check `cancelled_by_cleanup?` (with a reload to detect the cross-process race) before touching the circuit breaker, and the fallback loop breaks early since the run is already dead. The `ProviderRateLimitError` branch correctly omits the check since it doesn't call `record_provider_failure`. Tests exercise the race via `update_all` which is a good simulation of the separate-process mutation. Clean and well-scoped.

PR #938, opened 2026-04-08
Fix #930: Add review type indicator badges to projects index page

### LINKED ISSUE #930: Add review type indicator badges to projects index page
## Summary

Display a badge on the projects index page indicating which review type each project uses. This gives a quick visual overview of the review workflow configured for each project.

## Proposed Badges

- **Codex** — project uses the Codex review bot
- **Manual** — project uses manual code review
- **GitHub Copilot** — project uses GitHub Copilot for reviews
- **Paid Agent** — project uses a paid review agent

## Details

- Only show the badge for projects that have a review type configured
- Badge styling should be compact and visually distinct per type (e.g., different colors or icons)
- The badge should appear inline with the project name or in a consistent location on each project card/row

## Acceptance Criteria

- [ ] Projects with a configured review type display the corresponding badge on the index page
- [ ] Projects without a review type configured show no badge
- [ ] Badges are visually distinguishable from one another
- [ ] The indicator does not clutter the layout or break existing styling

viamin:
## Summary

Add visual review type indicator badges to the projects index page so users can quickly see which review workflow (Codex, Manual, GitHub Copilot, Paid Agent, CI Action) each project uses without navigating into individual project settings.

## Changes

- **Helper (`app/helpers/application_helper.rb`)** — Added `REVIEW_METHOD_STYLES` constant mapping each review type to a distinct color scheme, and a `review_method_badge` helper method that renders compact, color-coded badges
  - Codex: amber, Manual: sky blue, GitHub Copilot: blue, Paid Agent: purple, CI Action: teal
- **View (`app/views/projects/index.html.erb`)** — Added a badge row below the project name that renders all enabled review method badges; row is conditionally displayed only when the project has reviews enabled and at least one method configured

## Design Decisions

- **Color-per-type approach** — Each review type gets a distinct color to make scanning a list of projects fast and unambiguous, rather than using a single badge style with text-only differentiation
- **Conditional rendering** — Projects without a configured review type show no badge at all, keeping the layout clean for unconfigured projects
- **Helper-based implementation** — Badge rendering lives in `ApplicationHelper` rather than a partial or component, keeping it lightweight for what is essentially a styled `<span>` tag

## Screenshots

> ⚠️ **Author**: please attach before/after screenshots of the projects index page showing badge rendering for projects with various review types configured, and at least one project with no review type.

---

Closes #930


viamin (commented):
Clean, well-scoped PR that matches the issue requirements. A few suggestions below — one about keeping the style map in sync with the model's canonical list, and a minor note about the helper placement.

app/helpers/application_helper.rb
```diff
@@ -148,6 +148,24 @@ def local_time(time, format: :long)
     )
   end
 
+  REVIEW_METHOD_STYLES = {
+    "codex" => { bg: "bg-amber-100", text: "text-amber-700", label: "Codex" },
+    "manual" => { bg: "bg-sky-100", text: "text-sky-700", label: "Manual" },
+    "copilot" => { bg: "bg-blue-100", text: "text-blue-700", label: "GitHub Copilot" },
+    "paid_agent" => { bg: "bg-purple-100", text: "text-purple-700", label: "Paid Agent" },
+    "ci_action" => { bg: "bg-teal-100", text: "text-teal-700", label: "CI Action" }
+  }.freeze
```
viamin, line 160: The keys here duplicate `Project::REVIEW_METHODS`. If a new review method is added to the model but not here, `review_method_badge` will silently return `nil` and render nothing — no error, just a missing badge.

Consider either:
1. Adding a guard that logs/raises when an unknown method is encountered (fail-fast during development), or
2. Referencing `Project::REVIEW_METHODS` with a startup-time check that all methods have styles defined.

For example, a simple assertion at the bottom:
```ruby
unless Project::REVIEW_METHODS.all? { |m| REVIEW_METHOD_STYLES.key?(m) }
  raise "REVIEW_METHOD_STYLES is missing keys: #{Project::REVIEW_METHODS - REVIEW_METHOD_STYLES.keys}"
end
```
This would catch drift immediately rather than silently swallowing it.

app/helpers/application_helper.rb
```diff
@@ -148,6 +148,24 @@ def local_time(time, format: :long)
...
+
+  def review_method_badge(method)
+    styles = REVIEW_METHOD_STYLES[method]
+    return if styles.nil?
```
viamin: Nit: `return if styles.nil?` is fine for robustness, but given that the view already filters through `enabled_review_methods` (which only returns values from `Project::REVIEW_METHODS`), a `nil` here would indicate a bug (style map out of sync with the model). You might want to at least log a warning here so it doesn't fail silently:

```suggestion
    return(Rails.logger.warn { "review_method_badge: unknown method #{method.inspect}" }) if styles.nil?
```

app/views/projects/index.html.erb
```diff
@@ -49,6 +49,13 @@
                 </span>
               <% end %>
             </div>
+            <% if project.review_enabled? && project.enabled_review_methods.any? %>
```
viamin, line 52: Good defensive check — `review_enabled? && enabled_review_methods.any?` correctly handles the case where the top-level review toggle is off even if individual methods are flagged as enabled. No performance concern either since both methods read from memoized `effective_review_settings` and don't trigger additional queries. 👍

app/helpers/application_helper.rb
viamin, line 160: Minor style note: `ApplicationHelper` is already growing (170+ lines). The project style guide targets ~100-line classes. This is fine for now, but if more review-related helpers land, consider extracting a `ReviewHelper` or a Phlex component for review badges to keep `ApplicationHelper` from becoming a catch-all.

viamin:
## Agent Update

(Background task completed — that was just confirming the pre-existing test failure count, already handled.)

app/helpers/application_helper.rb
```diff
@@ -151,6 +151,34 @@ def local_time(time, format: :long)
...
+    "copilot" => { bg: "bg-blue-100", text: "text-blue-700", label: "GitHub Copilot" },
+    "paid_agent" => { bg: "bg-purple-100", text: "text-purple-700", label: "Paid Agent" },
+    "ci_action" => { bg: "bg-teal-100", text: "text-teal-700", label: "CI Action" }
+  }.freeze
+
+  # Validate at load time that every Project::REVIEW_METHODS entry has a style defined.
+  # This catches drift immediately rather than silently rendering missing badges.
+  Rails.application.config.after_initialize do
+    missing = Project::REVIEW_METHODS - REVIEW_METHOD_STYLES.keys
+    raise "REVIEW_METHOD_STYLES is missing keys: #{missing}" if missing.any?
```
viamin: The `after_initialize` block runs inside `ApplicationHelper`, which is a module — not a class that gets instantiated. This works, but placing boot-time validation inside a helper module is unusual and could surprise future developers.

Also, this will raise in production if a new review method is added to `Project::REVIEW_METHODS` before the corresponding style is added here. Consider whether a loud log warning (like you already have in `review_method_badge`) might be more appropriate than a hard crash, or at minimum gate this behind `Rails.env.local?`:

```suggestion
  Rails.application.config.after_initialize do
    missing = Project::REVIEW_METHODS - REVIEW_METHOD_STYLES.keys
    if missing.any?
      if Rails.env.local?
        raise "REVIEW_METHOD_STYLES is missing keys: #{missing}"
      else
        Rails.logger.error(message: "REVIEW_METHOD_STYLES is missing keys", missing: missing)
      end
    end
  end
```

app/helpers/application_helper.rb
```diff
@@ -151,6 +151,34 @@ def local_time(time, format: :long)
...
+  def review_method_badge(method)
+    styles = REVIEW_METHOD_STYLES[method]
+    if styles.nil?
+      Rails.logger.warn { "review_method_badge: unknown method #{method.inspect}" }
+      return
+    end
+
+    tag.span(
+      styles[:label],
+      class: "inline-flex items-center rounded-md #{styles[:bg]} px-2 py-1 text-xs font-medium #{styles[:text]}"
```
viamin, line 184: Minor: the existing Active/Inactive badges on the same page (lines 43-49 of `index.html.erb`) use identical Tailwind classes but are written as raw HTML `<span>` tags. This helper produces the same markup via `tag.span`. That's fine and arguably better, but worth noting the inconsistency — you could consider extracting those existing badges to use this same helper pattern in a follow-up to keep the view DRY.

app/helpers/application_helper.rb
```diff
@@ -151,6 +151,34 @@ def local_time(time, format: :long)
...
+  # This catches drift immediately rather than silently rendering missing badges.
+  Rails.application.config.after_initialize do
+    missing = Project::REVIEW_METHODS - REVIEW_METHOD_STYLES.keys
+    raise "REVIEW_METHOD_STYLES is missing keys: #{missing}" if missing.any?
+  end
+
+  def review_method_badge(method)
+    styles = REVIEW_METHOD_STYLES[method]
+    if styles.nil?
+      Rails.logger.warn { "review_method_badge: unknown method #{method.inspect}" }
```
viamin, line 178: Nit: since the `after_initialize` block above already guarantees all known methods have styles, this `nil` guard would only fire for truly unexpected input. The defensive logging is fine to keep, but you could simplify with an early return using a guard clause:

```suggestion
    styles = REVIEW_METHOD_STYLES[method]
    return if styles.nil? && Rails.logger.warn { "review_method_badge: unknown method #{method.inspect}" }
```

Though the current form is perfectly readable too — this is just a style preference.

viamin (commented):
Clean, well-scoped PR that addresses #930 correctly. The badge rendering approach is straightforward, no N+1 concerns since `review_settings` is a JSON column, and the load-time validation is a nice safety net. A couple of minor suggestions below.

viamin:
## Agent Update

That was the earlier failed commit attempt (before I updated the pre-commit hook). The successful commit was made afterward — `f1b3ed0e` is already in place. Nothing to do here.

viamin (commented):
Looks good overall — the feature cleanly addresses #930, the load-time validation is a nice safety net, and the view changes are well-scoped. One style/fragility concern on the badge helper.

app/helpers/application_helper.rb
```diff
@@ -151,6 +151,37 @@ def local_time(time, format: :long)
...
+        raise "REVIEW_METHOD_STYLES is missing keys: #{missing}"
+      else
+        Rails.logger.error(message: "REVIEW_METHOD_STYLES is missing keys", missing: missing)
+      end
+    end
+  end
+
+  def review_method_badge(method)
+    styles = REVIEW_METHOD_STYLES[method]
+    return if styles.nil? && Rails.logger.warn { "review_method_badge: unknown method #{method.inspect}" }
```
viamin: This line couples a logging side-effect with a control-flow guard in a non-obvious way. It relies on `Rails.logger.warn` always returning a truthy value so the `return if` triggers; if that ever returned falsy (the return contract isn't documented as guaranteed), `styles[:label]` on line 180 would raise `NoMethodError`. Splitting into an explicit guard is clearer and safer:

```suggestion
    return tag.span unless (styles = REVIEW_METHOD_STYLES[method])
```

Or, if you want to keep the warning log:

```ruby
    styles = REVIEW_METHOD_STYLES[method]
    if styles.nil?
      Rails.logger.warn { "review_method_badge: unknown method #{method.inspect}" }
      return
    end
```

viamin:
## Agent Update

Commit succeeded. The review feedback has been addressed: the guard clause in `review_method_badge` is now split into an explicit `unless` block that checks for `nil` styles, logs the warning, and returns — no longer relying on `Rails.logger.warn`'s return value for control flow.

viamin (commented):
Generated no new comments. The implementation follows the existing badge pattern in `ApplicationHelper` (e.g., `agent_run_status_badge`, `paid_state_badge`) consistently. The load-time validation against `Project::REVIEW_METHODS` is a nice safeguard against drift. `enabled_review_methods` reads from an in-memory JSON column so there are no N+1 query concerns. The view conditional rendering is clean and matches the issue requirements.

PR #939, opened 2026-04-08
Fix #931: dev:cleanup trips provider circuit breakers when killing in-flight runs

### LINKED ISSUE #931: dev:cleanup trips provider circuit breakers when killing in-flight runs

viamin:
## Summary

dev:cleanup trips provider circuit breakers when killing in-flight runs

See #931 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #931

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -13,6 +13,7 @@ class AgentRun < ApplicationRecord
   AUTO_PICK_BLOCKING_STATUSES = UNFINISHED_STATUSES
   TOKEN_LIMIT_STATUSES = %w[ok warning exceeded].freeze
   DEFAULT_MAX_TOKENS_PER_RUN = 10_000_000
+  STALE_CLEANUP_ERROR_PREFIX = "[cleanup] "
 
   belongs_to :project
   belongs_to :issue, optional: true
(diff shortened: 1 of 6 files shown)


viamin:
Closing in favor of #932, the human-authored fix for #931, which is the strongest base of the three duplicate PRs (this one, #951, and #932). Comparison summary:

| | **#932 (winner)** | **This PR** |
|---|---|---|
| Sentinel approach | Elevates the **existing** `Marked stale on startup` string to a constant — **no DB backfill needed** | Introduces a new `[cleanup] ` prefix — runs already cleanup-marked in the DB before this lands won't be detected and could still trip the breaker on retry |
| Code structure | Extracts `cancelled_by_cleanup?(agent_run)` (reload + nil-safe + `RecordNotFound` rescue) and `record_cleanup_cancelled_attempt` helpers | Inline 6-line block duplicated 3× across the rescue branches |
| Test simulation | Uses `update_columns` (bypasses dirty tracking — accurately simulates an external process holding a stale in-memory instance) | Uses `agent_run.update!` inside the stub, which mutates the in-memory instance and partially defeats the `reload` |
| Activity test depth | 4 thorough contexts (breaker untouched, attempt recorded, run state preserved, fallback iteration short-circuited) | 2 cases |

The two genuine advantages this PR had over #932 — the `InfiniteLoopError` short-circuit and model unit tests for `#cancelled_by_cleanup?` — have been ported onto #932 in 040014dc using #932's existing helpers (no inline duplication). Reviewers should look at #932.

PR #940, opened 2026-04-08
Fix #933: Optimize AgentRun::QUEUE_PRIORITY_SQL with a LATERAL join

### LINKED ISSUE #933: Optimize AgentRun::QUEUE_PRIORITY_SQL with a LATERAL join
Follow-up to viamin/paid#922.

`AgentRun::QUEUE_PRIORITY_SQL` evaluates three correlated `EXISTS` subqueries (one per priority tier P1/P2/P3) on every queue ordering. Each subquery contains an `OR` join condition between `issues.id = agent_runs.issue_id` and `(issues.project_id, issues.github_number, is_pull_request = TRUE)`, which prevents PostgreSQL from picking a single efficient index path and tends toward a nested loop with two probes per candidate row × three tiers.

The current approach is correct (and the necessary `(project_id, github_number)` index now exists), but it can be made cheaper for queue throughput on large projects by computing the candidate issue's labels once via a `LATERAL` join, then having the `CASE` reference the lateral row three times:

```sql
LEFT JOIN LATERAL (
  SELECT i.labels
  FROM issues i
  WHERE (i.id = agent_runs.issue_id AND i.project_id = agent_runs.project_id)
     OR (i.project_id = agent_runs.project_id
         AND i.github_number = agent_runs.source_pull_request_number
         AND i.is_pull_request = TRUE)
  LIMIT 1
) issue_labels ON TRUE
JOIN projects p ON p.id = agent_runs.project_id
```

The `CASE` then becomes:

```sql
CASE
  WHEN issue_labels.labels @> jsonb_build_array(COALESCE(NULLIF(p.priority_labels->>'P1', ''), 'P1')) THEN 0
  WHEN trigger_type = 'manual' THEN 1
  WHEN issue_labels.labels @> jsonb_build_array(COALESCE(NULLIF(p.priority_labels->>'P2', ''), 'P2')) THEN 2
  ...
END
```

The tricky part is wiring this into `AgentRun.queued.order(QUEUE_ORDER)` — the order expression alone cannot add joins, so the queue scope itself would need to opt into a custom `from`. Care needed to keep `next_queued_run` / `peek_next_queued_run` / `claim_next_queued_run` (which uses `FOR UPDATE SKIP LOCKED`) all consistent.

## Acceptance criteria
- One issue lookup per agent run instead of three.
- `AgentRun.next_queued_run`, `peek_next_queued_run`, `claim_next_queued_run` produce the same ordering as today.
- Existing 6-tier ordering specs in `spec/models/agent_run_spec.rb` continue to pass.

viamin:
## Summary

Optimize AgentRun::QUEUE_PRIORITY_SQL with a LATERAL join

See #933 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #933

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -211,46 +211,127 @@ class AgentRun < ApplicationRecord
   end
 
   # Priority ordering for the run queue (6 tiers):
-  #   0 = P1-labeled issues (critical priority)
+  #   0 = P1 user-defined label (highest)
   #   1 = manual runs
-  #   2 = P2-labeled issues (high priority)
+  #   2 = P2 user-defined label
   #   3 = automatic runs fixing a PR (auto-continue)
-  #   4 = P3-labeled issues (medium priority)
+  #   4 = P3 user-defined label
   #   5 = automatic runs from auto-pick (lowest)
   #
-  # Issue labels are resolved once via a LATERAL join to avoid
+  # User-defined priority labels (configured per project via
+  # Project#priority_labels) are read from either the run's associated
+  # issue OR the source pull request (looked up by source_pull_request_number
+  # against the project's issues table). If multiple priority labels are
+  # present, the highest takes precedence (P1 > P2 > P3).
+  #
+  # Issue labels in SQL are resolved once via a LATERAL join to avoid
   # repeated correlated subqueries per priority tier.
   #
   # Within each tier, create_issue runs are prioritized over create_pr
   # runs because issue creation is lighter-weight and often unblocks
-  # downstream PR work.
-  # Within each goal type, runs are processed FIFO by created_at, with
-  # id as a stable tiebreaker for runs created in the same timestamp.
-  #
-  # NOTE: The SQL sort tiers above (0–5) are internal ordering values.
-  # The `indicator` values below are user-facing display labels
-  # shown in the priority badge (e.g. "P1 - Critical").
+  # downstream PR work. Within each goal type, runs are FIFO by
+  # created_at, with id as a stable tiebreaker.
   QUEUE_PRIORITIES = {
-    p1: { label: "Critical", indicator: "P1" },
-    manual: { label: "Manual", indicator: 1 },
-    p2: { label: "High", indicator: "P2" },
-    auto_continue: { label: "Auto-continue", indicator: 2 },
-    p3: { label: "Medium", indicator: "P3" },
-    auto_pick: { label: "Auto-pick", indicator: 3 }
+    label_p1: { label: "P1", indicator: 1 },
+    manual: { label: "Manual", indicator: 2 },
+    label_p2: { label: "P2", indicator: 3 },
+    auto_continue: { label: "Auto-continue", indicator: 4 },
+    label_p3: { label: "P3", indicator: 5 },
+    auto_pick: { label: "Auto-pick", indicator: 6 }
   }.freeze
   UNKNOWN_PRIORITY = { label: "Unknown", indicator: nil }.freeze
 
   def queue_priority_tier
-    labels = issue&.labels || associated_pr_labels || []
-    p_labels = project.priority_labels
+    label_tier = label_priority_tier
+    return :label_p1 if label_tier == "P1"
 
-    return :p1 if labels.include?(p_labels.fetch("P1", "P1"))
-    return :manual if manual?
-    return :p2 if labels.include?(p_labels.fetch("P2", "P2"))
-    return :auto_continue if automatic? && existing_pr?
-    return :p3 if labels.include?(p_labels.fetch("P3", "P3"))
+    if manual?
+      :manual
+    elsif label_tier == "P2"
+      :label_p2
+    elsif automatic? && existing_pr?
+      :auto_continue
+    elsif label_tier == "P3"
+      :label_p3
+    else
+      :auto_pick
+    end
+  end
 
-    :auto_pick
+  # Returns "P1"/"P2"/"P3" if the run's issue or source PR carries one of
+  # the project's configured priority labels (highest wins), else nil.
+  # Memoized because rendering helpers call this twice per row (once
+  # via queue_priority_tier and once via queue_priority_label).
+  #
+  # NOTE: The memoization is per-instance and is only safe for the
+  # request-scoped rendering use case. Long-running processes that hold
+  # onto an AgentRun across an issue.labels update will see a stale tier;
+  # call AgentRun#reload (or instantiate a new record) in those cases.
+  def label_priority_tier
+    return @label_priority_tier if defined?(@label_priority_tier)
+    @label_priority_tier = compute_label_priority_tier
+  end
+
+  # When a run has both an associated issue AND a source PR, labels from
+  # both records are considered (highest configured tier wins). Label
+  # name matching is case-sensitive on both the Ruby and SQL sides; an
+  # issue tagged "p1" will not match a configured tier of "P1".
+  def compute_label_priority_tier
+    return nil unless project
+
+    label_sources = []
+    label_sources << issue.labels if issue&.labels.present?
+    pr_record = source_pull_request_record
+    label_sources << pr_record.labels if pr_record&.labels.present?
+    return nil if label_sources.empty?
+
+    Project::PRIORITY_TIERS.find do |tier|
+      label_name = project.priority_label_for(tier)
+      next false if label_name.blank?
+      label_sources.any? { |labels| labels.include?(label_name) }
+    end
+  end
+  private :compute_label_priority_tier
+
+  attr_writer :source_pull_request_record
+
+  # The Issue row representing this run's source pull request, used by
+  # label_priority_tier. Falls back to a per-row find_by, but callers
+  # rendering many runs should call AgentRun.preload_source_pull_requests
+  # first to avoid an N+1.
+  def source_pull_request_record
+    return @source_pull_request_record if defined?(@source_pull_request_record)
+    return nil if source_pull_request_number.blank? || project.nil?
+
+    @source_pull_request_record = project.issues.find_by(
+      github_number: source_pull_request_number, is_pull_request: true
+    )
+  end
+
+  # Batch-loads source PR Issue rows for a collection of runs and stashes
+  # each on the run via source_pull_request_record=. Call from controllers
+  # rendering priority badges for many runs to keep label_priority_tier
+  # from issuing one query per row. Groups by project_id so the query is
+  # precise per project (no cross-project over-fetch) without resorting
+  # to dynamic SQL.
+  def self.preload_source_pull_requests(runs)
+    targets = Array(runs).select do |r|
+      r.source_pull_request_number.present? && r.project_id.present? &&
+        !r.instance_variable_defined?(:@source_pull_request_record)
+    end
+    return if targets.empty?
+
+    records = {}
+    targets.group_by(&:project_id).each do |project_id, group|
+      numbers = group.map(&:source_pull_request_number).uniq
+      Issue.where(is_pull_request: true, project_id: project_id, github_number: numbers).each do |issue|
+        records[[ issue.project_id, issue.github_number ]] = issue
+      end
+    end
+
+    targets.each do |run|
+      run.source_pull_request_record = records[[ run.project_id, run.source_pull_request_number ]]
+    end
   end
 
   def queue_priority_label
@@ -261,7 +342,9 @@ class AgentRun < ApplicationRecord
 
   # LATERAL join that resolves candidate issue labels once per agent run.
   # Matches the issue either by direct association (issue_id) or by
-  # source_pull_request_number for PR-based runs.
+  # source_pull_request_number for PR-based runs. Both branches constrain
+  # `i.project_id = agent_runs.project_id`, so cross-project leakage
+  # is impossible even if `issue_id` were ever set across projects.
   QUEUE_LATERAL_JOIN = <<~SQL.squish.freeze
     LEFT JOIN LATERAL (
       SELECT i.labels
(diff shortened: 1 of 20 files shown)


viamin:
## Agent Update

The background task notification is from the earlier failed commit attempt — it can be safely ignored. The commit has already succeeded in the foreground:

```
bd99a28c fix: resolve merge conflicts with main for LATERAL join optimization
```

5637 examples, 0 failures, lint clean, working tree clean. All done.

viamin (commented):
## Review: LATERAL join optimization for queue priority SQL

The refactoring from three correlated `EXISTS` subqueries to a single `LATERAL` join is a solid performance improvement — computing the issue labels once instead of three times per row is the right approach. The `queued_with_priority` scope cleanly encapsulates the join, and the test coverage is comprehensive (all 6 tiers, custom labels, PR-based labels, preloading).

However, there is a **correctness concern** with `LIMIT 1` in the LATERAL subquery that could cause label misses in an edge case. See inline comments.

Also: the `db/schema.rb` changes are purely whitespace reformatting (~472 lines of `[ "x" ]` → `["x"]`). Consider reverting those to keep the diff focused on the actual optimization. If it reflects a Rails version change, a separate commit would be cleaner.

app/models/agent_run.rb
```diff
@@ -337,58 +340,40 @@ def queue_priority_label
...
   # is impossible even if `issue_id` were ever set across projects.
+  QUEUE_LATERAL_JOIN = <<~SQL.squish.freeze
+    LEFT JOIN LATERAL (
+      SELECT i.labels
+      FROM issues i
+      WHERE (i.id = agent_runs.issue_id AND i.project_id = agent_runs.project_id)
+         OR (i.project_id = agent_runs.project_id
+             AND i.github_number = agent_runs.source_pull_request_number
+             AND i.is_pull_request = TRUE)
+      LIMIT 1
```
viamin: **Correctness concern: `LIMIT 1` may miss labels when both issue and source PR exist.**

The old `EXISTS` subqueries used an `OR` inside `EXISTS`, meaning they would match if **any** matching row (issue or PR) carried the label. The new `LATERAL` picks a single row via `LIMIT 1` (with no `ORDER BY`, so the choice is non-deterministic), then checks only that row's labels.

Scenario:
- A run has both `issue_id` (issue with labels `["P3"]`) and `source_pull_request_number` (PR with labels `["P1"]`)
- The LATERAL returns the issue row → labels = `["P3"]`
- The `CASE` sees P3 but misses P1 → tier 4 instead of tier 0

The Ruby-side `compute_label_priority_tier` considers labels from **both** rows (`label_sources.any?`), so the SQL and Ruby paths would disagree.

Fix: aggregate labels from all matching rows instead of picking one:
```suggestion
    LEFT JOIN LATERAL (
      SELECT COALESCE(jsonb_agg(elem), '[]'::jsonb) AS labels
      FROM issues i, jsonb_array_elements(i.labels) AS elem
      WHERE (i.id = agent_runs.issue_id AND i.project_id = agent_runs.project_id)
         OR (i.project_id = agent_runs.project_id
             AND i.github_number = agent_runs.source_pull_request_number
             AND i.is_pull_request = TRUE)
    ) issue_labels ON TRUE
```

Alternatively, if most runs don't have both issue_id and source_pull_request_number, this may not matter in practice — but it's a semantic difference from the old code worth documenting or testing.

app/models/agent_run.rb
```diff
@@ -337,58 +340,40 @@ def queue_priority_label
...
-          OR (i.project_id = agent_runs.project_id
-              AND i.github_number = agent_runs.source_pull_request_number
-              AND i.is_pull_request = TRUE)
-        )
-        WHERE p.id = agent_runs.project_id
-          AND i.labels @> jsonb_build_array(COALESCE(NULLIF(p.priority_labels->>'P3', ''), 'P3'))
-      ) THEN 4
+      WHEN issue_labels.labels @> jsonb_build_array(COALESCE(NULLIF(p.priority_labels->>'P3', ''), 'P3')) THEN 4
+
       ELSE 5
```
viamin, line 377: Nit: stray blank line before `ELSE 5` — likely a diff artifact.
```suggestion
      ELSE 5
```

app/models/agent_run.rb
```diff
@@ -337,58 +340,40 @@ def queue_priority_label
...
+    LEFT JOIN LATERAL (
+      SELECT i.labels
+      FROM issues i
+      WHERE (i.id = agent_runs.issue_id AND i.project_id = agent_runs.project_id)
+         OR (i.project_id = agent_runs.project_id
+             AND i.github_number = agent_runs.source_pull_request_number
+             AND i.is_pull_request = TRUE)
+      LIMIT 1
+    ) issue_labels ON TRUE
+    JOIN projects p ON p.id = agent_runs.project_id
```
viamin: Minor: this is an `INNER JOIN`, so a queued run with a NULL `project_id` would be silently excluded from the result set. The old `EXISTS` subqueries would just evaluate to false, leaving such a run in tier 5 (auto-pick). In practice queued runs should always have a project, but `LEFT JOIN projects p` would preserve the old behavior defensively.

spec/models/agent_run_spec.rb
```diff
@@ -1456,6 +1568,14 @@ def queued_run_with_issue_labels(labels, **attrs)
         described_class.preload_source_pull_requests([ run ])
       end
     end
+
+    it "returns '1 - P1' for P1-labeled issues" do
+      project = create(:project)
+      issue = create(:issue, project: project, labels: [ "P1" ])
+      run = create(:agent_run, trigger_type: "automatic", project: project, issue: issue)
```
viamin, line 1587: Good test coverage overall. Consider adding a test for the **split-label** edge case: a run with both `issue_id` (issue has no priority labels or a lower-tier label) and `source_pull_request_number` (PR has P1). This would pin the LATERAL join's behavior when labels are distributed across two issue rows.

viamin:
## Agent Update

Commit successful. Here's a summary of all changes made:

**Code review fixes in `app/models/agent_run.rb`:**

1. **LATERAL join LIMIT 1 bug** (line 352): Replaced `SELECT i.labels FROM issues i ... LIMIT 1` with `SELECT COALESCE(jsonb_agg(elem), '[]'::jsonb) AS labels FROM issues i, jsonb_array_elements(i.labels) AS elem ...` — This aggregates labels from all matching rows (both issue and source PR), matching the Ruby-side `compute_label_priority_tier` behavior.

2. **LEFT JOIN for projects** (line 359): Changed `JOIN projects p` to `LEFT JOIN projects p` so runs with NULL `project_id` fall through to tier 5 (auto-pick) instead of being silently excluded.

3. **Stray blank line** (line 377): Removed the empty line before `ELSE 5`.

**Test in `spec/models/agent_run_spec.rb`:**

4. **Split-label test** (line 1539): Added test where a run has an issue with `"low"` (P3) label and a source PR with `"critical"` (P1) label, verifying both the Ruby-side `queue_priority_tier` and the SQL-side `next_queued_run` correctly pick the highest priority (P1) from across both rows.

db/schema.rb
```diff
@@ -22,10 +22,10 @@
     t.datetime "created_at", null: false
     t.decimal "quality_score", precision: 5, scale: 4
     t.datetime "updated_at", null: false
-    t.index [ "ab_test_id", "agent_run_id" ], name: "index_ab_test_assignments_unique", unique: true
-    t.index [ "ab_test_id" ], name: "index_ab_test_assignments_on_ab_test_id"
-    t.index [ "ab_test_variant_id" ], name: "index_ab_test_assignments_on_ab_test_variant_id"
-    t.index [ "agent_run_id" ], name: "index_ab_test_assignments_on_agent_run_id"
+    t.index ["ab_test_id", "agent_run_id"], name: "index_ab_test_assignments_unique", unique: true
```
viamin: The entire `db/schema.rb` diff (236+ / 236-) is whitespace-only reformatting — changing `[ "col" ]` to `["col"]` throughout. This adds noise to the PR and makes the diff harder to review. Since there's no migration in this PR, this was likely generated by a different Rails/schema-dumper version.

Consider reverting this file (`git checkout origin/main -- db/schema.rb`) so the PR only contains the intentional LATERAL join changes.

app/models/agent_run.rb
```diff
@@ -337,58 +340,40 @@ def queue_priority_label
...
+  # Matches the issue either by direct association (issue_id) or by
+  # source_pull_request_number for PR-based runs, merging labels from all
+  # matching rows so a run with both an issue and a source PR considers
+  # labels from both (matching compute_label_priority_tier). Both branches
+  # constrain `i.project_id = agent_runs.project_id`, so cross-project
+  # leakage is impossible even if `issue_id` were ever set across projects.
+  QUEUE_LATERAL_JOIN = <<~SQL.squish.freeze
+    LEFT JOIN LATERAL (
+      SELECT COALESCE(jsonb_agg(elem), '[]'::jsonb) AS labels
+      FROM issues i, jsonb_array_elements(i.labels) AS elem
```
viamin, line 353: Nice use of `jsonb_array_elements` + `jsonb_agg` to flatten labels across multiple matching issue rows. One thing to be aware of: `jsonb_array_elements('[]'::jsonb)` returns zero rows, but since this is an aggregate query without `GROUP BY`, PostgreSQL still returns one row with `jsonb_agg(elem) = NULL`, which `COALESCE` correctly handles to `'[]'::jsonb`. The edge cases are well-covered.

Minor: the `OR` in the WHERE clause still prevents a single index path per probe (same as before), but since it's now executed once per run instead of three times, the net improvement is clear.

app/models/agent_run.rb
```diff
@@ -337,58 +340,40 @@ def queue_priority_label
...
+      WHERE (i.id = agent_runs.issue_id AND i.project_id = agent_runs.project_id)
+         OR (i.project_id = agent_runs.project_id
+             AND i.github_number = agent_runs.source_pull_request_number
+             AND i.is_pull_request = TRUE)
+    ) issue_labels ON TRUE
+    LEFT JOIN projects p ON p.id = agent_runs.project_id
```
viamin, line 359: Good call making the projects join a `LEFT JOIN` — this ensures agent runs without a project (if that's ever possible) still get a row back rather than being silently dropped from the queue.

app/models/agent_run.rb
```diff
@@ -400,14 +385,22 @@ def queue_priority_label
   SQL
   QUEUE_ORDER = [ QUEUE_PRIORITY_SQL, GOAL_PRIORITY_SQL, { created_at: :asc, id: :asc } ].freeze
 
+  # Scope that adds the LATERAL join required by QUEUE_PRIORITY_SQL.
+  # All queue-ordering methods use this instead of bare `queued`.
+  scope :queued_with_priority, -> {
+    queued
+      .joins(QUEUE_LATERAL_JOIN)
+      .select("agent_runs.*")
```
viamin, line 393: The `.select("agent_runs.*")` is important here to avoid pulling `issue_labels.labels` and `p.*` into the returned ActiveRecord objects. Good defensive practice.

spec/models/agent_run_spec.rb
```diff
@@ -1276,6 +1276,86 @@
...
+        p1_run = create(:agent_run, :queued, trigger_type: "automatic", project: project,
+          issue: p1_issue, created_at: 1.minute.ago)
+
+        ordered_ids = described_class.queued_with_priority.order(described_class::QUEUE_ORDER).pluck(:id)
+        expect(ordered_ids).to eq([ p1_run, manual_run, p2_run, auto_continue, p3_run, auto_pick ].map(&:id))
+      end
+    end
   end
 
   describe ".peek_next_queued_run" do
```
viamin, line 1361: Good comprehensive test covering all 6 tiers in a single ordering assertion. This is exactly the kind of integration test that catches regressions in the CASE ordering.

viamin (commented):
## Review: Optimize QUEUE_PRIORITY_SQL with LATERAL join

The core optimization is well-implemented: replacing three repeated correlated `EXISTS` subqueries with a single `LEFT JOIN LATERAL` that aggregates issue labels once per agent run. The approach correctly handles the multi-issue case (direct issue + source PR) by using `jsonb_agg` to merge labels from all matching rows.

### What looks good
- **Correctness**: The `jsonb_array_elements` + `jsonb_agg` pattern correctly flattens and re-aggregates labels across multiple matching issues, preserving the same semantics as the old per-tier EXISTS checks.
- **Scope consistency**: `next_queued_run` and `peek_next_queued_run` both use the new `queued_with_priority` scope. `claim_next_queued_run` correctly stays on bare `queued` since it selects by `target_id`, not priority.
- **Test coverage**: Excellent — new specs cover all 6 priority tiers, custom label names, PR-based label resolution, and the split-label scenario (labels from both issue and source PR).
- **No SQL injection risk**: All tier names are hardcoded string literals.

### Issues
1. **schema.rb noise** (see inline comment) — 472 lines of whitespace-only reformatting should be excluded from this PR.

Overall this is a clean, well-tested optimization that directly addresses #933. One minor cleanup needed on the schema.rb noise.

viamin:
## Agent Update

All done. Here's what was accomplished:

1. **Merge conflicts resolved** - Merged `origin/main` cleanly (only `containerized_runner.rb` had a conflict, resolved automatically by git).

2. **Review feedback addressed**:
   - **db/schema.rb** - Reverted the whitespace-only reformatting as requested by the reviewer. The PR now only contains the intentional LATERAL join changes.
   - The other 4 review threads were positive feedback (no action required).

3. **Verification** - Rubocop passes with 0 offenses, all 253 agent_run specs pass with 0 failures.

The PR diff now contains only `app/models/agent_run.rb` and `spec/models/agent_run_spec.rb`.

viamin (commented):
Generated no new comments. The LATERAL join correctly consolidates three correlated EXISTS subqueries into a single label aggregation per agent run. The `jsonb_agg` with `COALESCE` properly handles empty/NULL labels, the `LEFT JOIN LATERAL ... ON TRUE` ensures runs without matching issues fall through to tier 5, and `claim_next_queued_run` correctly remains unchanged since it locks by specific `target_id` rather than using QUEUE_ORDER. Test coverage is thorough across all 6 tiers, custom labels, and the split-label (issue + source PR) scenario.

PR #949, opened 2026-04-08
Fix #935: fix(auto-merge): check affected files when clearing body-only review feedback

### LINKED ISSUE #935: fix(auto-merge): check affected files when clearing body-only review feedback
## Problem

PR #894 auto-merged because a CI schema fix (`db/schema.rb`) completed after a P1 Codex review on `fetch_issues_activity.rb`. The timestamp-based anti-loop guard doesn't verify the agent's changes relate to the review findings.

## Proposed Fix

Add an affected-file heuristic to the body-only review clearance path in `check_review_bot_status`. When the latest agent run completed after the review, cross-reference the review's inline comment `path` fields against the diff between the reviewed commit and the latest commit. Only clear feedback when at least one file mentioned in the review changed.

### Key changes
- Fetch the review's inline comments (already available via `pull_request_reviews`)
- Compare comment paths against the interceding diff
- Only bypass `body_only_review_needs_followup?` when the diff touches a reviewed file
- Fall back to current timestamp behavior when comments can't be fetched or have no file paths

## Impact

Defense-in-depth: even if the timestamp check passes, unrelated changes won't clear review findings on files that haven't been touched.

Depends on none.

viamin:
## Summary

Prevent auto-merge from clearing review feedback when the agent's latest changes don't touch any files mentioned in the review. This closes a loophole where an unrelated CI fix (e.g., `db/schema.rb`) could satisfy the timestamp-based anti-loop guard and inadvertently dismiss a P1 review on a completely different file.

## Changes

**Models / Services**
- `github_client.rb`: Added `compare_files` method to retrieve the list of changed files between two commits. Included `commit_id` in review response payloads and `path` in review comment responses to support file-level cross-referencing.

**Workflow Activities**
- `scan_paid_prs_activity.rb`: Introduced `body_only_review_affected_files_changed?` heuristic that compares inline review comment paths against the diff between the reviewed commit and the current HEAD. Updated `check_review_bot_status` so that clearing body-only review feedback now requires **both** the existing timestamp guard **and** the new affected-file guard to pass.

**Tests**
- Added three test cases covering: unrelated file changes do not clear feedback, matching file changes do clear feedback, and fallback to timestamp-only behavior when no inline comments are present.

## Design Decisions

- **Dual guard (timestamp + file overlap):** Rather than replacing the timestamp check, the file-overlap heuristic is layered on top as defense-in-depth. This avoids regressions in the existing anti-loop logic while closing the identified gap.
- **Graceful fallback:** When inline comments can't be fetched or contain no file paths, the system falls back to the existing timestamp-only behavior, keeping the change backward-compatible and resilient to API failures.
- **Minimal API surface:** `compare_files` returns only the file list from the GitHub compare endpoint rather than full diff content, keeping the payload small and the heuristic fast.

---

Closes #935

### DIFF
--- a/app/services/github_client.rb
+++ b/app/services/github_client.rb
@@ -189,6 +189,19 @@ class GithubClient
     end.map(&:filename)
   end
 
+  # Compares two commits and returns the list of changed file paths.
+  #
+  # @param repo [String] Repository in "owner/name" format
+  # @param base [String] Base commit SHA
+  # @param head [String] Head commit SHA
+  # @return [Array<String>] File paths changed between the two commits
+  def compare_files(repo, base, head)
+    handle_errors do
+      comparison = client.compare(repo, base, head)
+      (comparison.files || []).map(&:filename)
+    end
+  end
+
   # Creates an issue on a repository.
   #
   # @param repo [String] Repository in "owner/name" format
@@ -463,7 +476,8 @@ class GithubClient
           user_login: r.user&.login,
           state: r.state,
           body: r.body.to_s,
-          submitted_at: parse_timestamp(r.submitted_at)
+          submitted_at: parse_timestamp(r.submitted_at),
+          commit_id: r.commit_id
         }
       end
     end
@@ -709,7 +723,7 @@ class GithubClient
   # @param number [Integer] Pull request number
   # @param per_page [Integer, nil] When set, fetches a single page of this size
   #   (no auto-pagination). When nil, auto-paginates all comments.
-  # @return [Array<Hash>] Comments with :id, :user_login, :body, :created_at keys
+  # @return [Array<Hash>] Comments with :id, :user_login, :body, :created_at, :path keys
   def pull_request_review_comments(repo, number, per_page: nil)
     handle_errors do
       comments = if per_page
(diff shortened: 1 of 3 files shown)


viamin:
Closing in favor of #953, which solves #935 the same way but with one materially more correct decision.

Both PRs add a new GithubClient method (`compare_files` here, `compare_changed_files` in #953) and a helper that decides whether the diff between the reviewed commit and PR HEAD touches any file mentioned in the review's inline comments. Three test contexts each, and the elsif structure (combined OR vs split branch) is functionally equivalent.

The decisive difference is **how each PR maps inline review comments to the latest review**:

- **This PR** filters by `comment[:user_login]&.downcase == review_bot_login`. That's sloppy: if the same bot has posted multiple reviews on the PR over time, this includes inline comments from *all* of them, not just the latest review — and it also picks up freestanding line comments from the same user that weren't part of any review. Either case can produce an incorrect "reviewed paths" set.
- **#953** filters by `comment[:pull_request_review_id] == latest_review[:id]`. That's the exact 1-to-1 association GitHub provides between review comments and their parent review.

#953 also threads `pull_request_review_id` through the `pull_request_review_comments` return shape so the helper has the right join key.

Test coverage is equivalent — #953 has the same three contexts (reviewed file changed → addressed; reviewed file not changed → still unaddressed; no inline comments → fall back to addressed). I rebased #953 onto current main (no conflicts) and the new specs pass. Reviewers should look at #953.

PR #951, opened 2026-04-08
Fix #931: dev:cleanup trips provider circuit breakers when killing in-flight runs

### LINKED ISSUE #931: dev:cleanup trips provider circuit breakers when killing in-flight runs

viamin:
## Summary

Prevent dev:cleanup from tripping provider circuit breakers by tagging cleanup-induced timeouts with a sentinel prefix and skipping failure recording for those runs. This fixes an issue where killing in-flight agent runs during setup would penalize healthy providers, causing subsequent runs to unnecessarily fall back to inferior providers.

## Changes

**Model (`app/models/agent_run.rb`)**
- Added `STALE_CLEANUP_ERROR_PREFIX` constant (`"[cleanup] "`) and `cancelled_by_cleanup?` predicate that checks whether the error message carries the cleanup sentinel

**Rake task (`lib/tasks/dev.rake`)**
- Prepended the cleanup prefix to the timeout error message set during `dev:cleanup`, making cleanup-killed runs distinguishable from genuine timeouts

**Temporal activity (`app/temporal/activities/run_agent_activity.rb`)**
- In all three rescue branches (`InfiniteLoopError`, `ProviderTimeoutError`, `ProviderExecutionError`), added a reload + `cancelled_by_cleanup?` guard that records the attempt as `error_type: "cancelled_by_cleanup"` and breaks out of the provider fallback loop **without** calling `record_provider_failure`

**Tests**
- Added specs for `cancelled_by_cleanup?` in the AgentRun model spec
- Updated dev:cleanup spec to verify the sentinel prefix is present on cleanup-killed runs

## Design Decisions

- **Sentinel prefix on `error_message` rather than a dedicated column**: avoids a migration for a narrow concern; the prefix is a stable contract between the rake task and the activity, and `cancelled_by_cleanup?` encapsulates the check cleanly.
- **Reload before checking**: the activity runs in a separate Temporal worker process, so the `AgentRun` record must be reloaded to see the cleanup prefix written by the rake task.
- **Break instead of `next`**: once cleanup has killed the container, retrying on a fallback provider is pointless — the run is already terminal. Breaking out avoids wasting fallback attempts and prevents cascading circuit breaker hits across multiple providers.

---

Closes #931

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -10,6 +10,7 @@ class AgentRun < ApplicationRecord
   FAILURE_STATUSES = %w[failed timeout auth_expired rate_limited].freeze
   UNFINISHED_STATUSES = %w[queued pending running paused].freeze
   GUARDRAIL_VIOLATION_TYPES = %w[loop_detected token_limit cost_limit time_limit anomaly].freeze
+  STALE_CLEANUP_ERROR_PREFIX = "[cleanup] "
   AUTO_PICK_BLOCKING_STATUSES = UNFINISHED_STATUSES
   TOKEN_LIMIT_STATUSES = %w[ok warning exceeded].freeze
   DEFAULT_MAX_TOKENS_PER_RUN = 10_000_000
@@ -482,6 +483,10 @@ class AgentRun < ApplicationRecord
 
   def successful?
     status == "completed"
+  end
+
+  def cancelled_by_cleanup?
+    error_message.to_s.start_with?(STALE_CLEANUP_ERROR_PREFIX)
   end
 
   def total_tokens
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -225,6 +225,13 @@ module Activities
               )
             end
           rescue InfiniteLoopError => e
+            agent_run.reload
+            if agent_run.cancelled_by_cleanup?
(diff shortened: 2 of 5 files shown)


viamin:
Closing in favor of #932, the human-authored fix for #931, which is the strongest base of the three duplicate PRs (this one, #939, and #932). Comparison summary:

| | **#932 (winner)** | **This PR** |
|---|---|---|
| Sentinel approach | Elevates the **existing** `Marked stale on startup` string to a constant — **no DB backfill needed** | Introduces a new `[cleanup] ` prefix — runs already cleanup-marked before this lands won't be detected |
| `cancelled_by_cleanup?` predicate | `status == "timeout"` AND prefix match | **Just** prefix match — looser, can false-positive on any `error_message` that happens to start with the prefix even when the run isn't actually in timeout state |
| Code structure | Extracts helpers (`cancelled_by_cleanup?(agent_run)`, `record_cleanup_cancelled_attempt`) — reload + nil-safe + `RecordNotFound` rescue all centralized | Inline 7-line block duplicated 3× |
| Activity tests | 4 contexts | 0 |

The two genuine advantages this PR had over #932 — the `InfiniteLoopError` short-circuit and model unit tests for `#cancelled_by_cleanup?` — have been ported onto #932 in 040014dc, using #932's helpers and its tighter `status == "timeout"` predicate. Reviewers should look at #932.

PR #953, opened 2026-04-08
Fix #935: fix(auto-merge): check affected files when clearing body-only review feedback

### LINKED ISSUE #935: fix(auto-merge): check affected files when clearing body-only review feedback

viamin:
## Summary

Prevents auto-merge from clearing review feedback when the agent's latest changes don't touch any files mentioned in the review. This closes a gap where unrelated commits (e.g., a CI schema fix) could satisfy the timestamp-based anti-loop guard and inadvertently dismiss review findings on untouched files.

## Changes

**GitHub API surface** (`app/services/github_client.rb`):
- Exposed `commit_id` on `pull_request_reviews` and `path`/`pull_request_review_id` on `pull_request_review_comments` return hashes
- Added `compare_changed_files` method to retrieve the file list between two commits

**Review clearance logic** (`app/temporal/activities/scan_paid_prs_activity.rb`):
- Added `review_diff_touches_reviewed_files?` heuristic that cross-references inline comment paths against the diff between the reviewed commit and PR HEAD
- Gated the body-only review clearance path: even when the timestamp guard passes, feedback is only cleared if at least one reviewed file appears in the subsequent diff
- Falls back to existing timestamp-only behavior when `commit_id` is nil, no inline comments exist, or the compare API call fails

**Tests** (`spec/temporal/activities/scan_paid_prs_activity_spec.rb`):
- Three new cases covering unrelated diff (feedback retained), related diff (feedback cleared), and no inline comments (fallback to clear)

## Design Decisions

- **Graceful fallback to timestamp behavior**: when the review has no inline comments or the commit comparison fails, the existing timestamp-based logic applies unchanged. This keeps the change additive — it can only make clearance stricter, never block it where it previously succeeded.
- **File-path granularity, not line-level**: the heuristic checks whether any file from the review's inline comments appears in the diff, rather than verifying line-level overlap. This is intentionally coarse — it catches the "completely unrelated commit" case (like `db/schema.rb` vs `fetch_issues_activity.rb`) without introducing brittle line-range intersection logic.

---

Closes #935

### DIFF
--- a/app/services/github_client.rb
+++ b/app/services/github_client.rb
@@ -191,6 +191,12 @@ class GithubClient
 
   # Compares two commits and returns the list of changed file paths.
   #
+  # NOTE: GitHub's compare API returns a maximum of 300 files per response.
+  # For very large diffs (e.g., a rebase onto a distant base), the result
+  # may be silently truncated. This is unlikely for the narrow
+  # commit-to-HEAD diffs this method currently serves, but callers doing
+  # broader comparisons should be aware of the limit.
+  #
   # @param repo [String] Repository in "owner/name" format
   # @param base [String] Base commit SHA
   # @param head [String] Head commit SHA
--- a/app/services/github_client.rb
+++ b/app/services/github_client.rb
@@ -189,6 +189,19 @@ class GithubClient
     end.map(&:filename)
   end
 
+  # Compares two commits and returns the list of changed file paths.
+  #
+  # @param repo [String] Repository in "owner/name" format
+  # @param base [String] Base commit SHA
+  # @param head [String] Head commit SHA
+  # @return [Array<String>] File paths changed between base and head
+  def compare_changed_files(repo, base, head)
+    handle_errors do
+      comparison = client.compare(repo, base, head)
+      (comparison.files || []).map(&:filename)
+    end
+  end
+
   # Creates an issue on a repository.
   #
   # @param repo [String] Repository in "owner/name" format
@@ -452,7 +465,7 @@ class GithubClient
   #
   # @param repo [String] Repository in "owner/name" format
   # @param number [Integer] Pull request number
-  # @return [Array<Hash>] Reviews with :id, :user_login, :state, :body, :submitted_at keys (:body is always a String)
+  # @return [Array<Hash>] Reviews with :id, :user_login, :state, :body, :submitted_at, :commit_id keys (:body is always a String)
   # @raise [NotFoundError] if the pull request does not exist
   def pull_request_reviews(repo, number)
     handle_errors do
@@ -463,7 +476,8 @@ class GithubClient
           user_login: r.user&.login,
           state: r.state,
           body: r.body.to_s,
-          submitted_at: parse_timestamp(r.submitted_at)
+          submitted_at: parse_timestamp(r.submitted_at),
+          commit_id: r.commit_id
         }
       end
     end
@@ -709,7 +723,7 @@ class GithubClient
   # @param number [Integer] Pull request number
   # @param per_page [Integer, nil] When set, fetches a single page of this size
   #   (no auto-pagination). When nil, auto-paginates all comments.
-  # @return [Array<Hash>] Comments with :id, :user_login, :body, :created_at keys
+  # @return [Array<Hash>] Comments with :id, :user_login, :body, :created_at, :path, :pull_request_review_id keys
   def pull_request_review_comments(repo, number, per_page: nil)
     handle_errors do
       comments = if per_page
@@ -724,7 +738,9 @@ class GithubClient
           id: c.id,
           user_login: c.user&.login,
           body: c.body.to_s,
-          created_at: parse_timestamp(c.created_at)
+          created_at: parse_timestamp(c.created_at),
+          path: c.path,
+          pull_request_review_id: c.pull_request_review_id
         }
       end
     end
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -537,6 +537,10 @@ module Activities
           [ { type: "review_bot_review_pending", details: "Latest review bot review was not clean" } ]
         else
           bot_thread_triggers = review_bot_thread_triggers(unresolved_threads)
+          body_only_pending_triggers = [
+            { type: "review_bot_review_pending", details: "Latest review bot review was not clean" },
+            { type: "review_bot_comments", details: "Latest review bot review generated comments (body-only)" }
+          ]
           if bot_thread_triggers.any?
             # Thread-based bots (e.g. Copilot) with at least one unresolved
             # bot thread: emit review_bot_comments + the thread triggers.
@@ -554,10 +558,7 @@ module Activities
             # The "(body-only)" suffix on review_bot_comments lets
             # structured-log consumers distinguish this path from the
             # thread-based Copilot flow above.
-            [
-              { type: "review_bot_review_pending", details: "Latest review bot review was not clean" },
-              { type: "review_bot_comments", details: "Latest review bot review generated comments (body-only)" }
-            ]
+            body_only_pending_triggers
           elsif body_only_review_bot?(latest&.dig(:user_login)) &&
               !review_diff_touches_reviewed_files?(client, project, issue, latest)
             # Body-only bot whose review pre-dates the last agent run
@@ -565,10 +566,7 @@ module Activities
             # touch any file mentioned in the review's inline comments.
             # Treat as still-unaddressed to prevent unrelated changes
             # (e.g. a CI schema fix) from clearing review findings.
-            [
-              { type: "review_bot_review_pending", details: "Latest review bot review was not clean" },
-              { type: "review_bot_comments", details: "Latest review bot review generated comments (body-only)" }
-            ]
+            body_only_pending_triggers
           else
             # Thread-based bot with all bot threads resolved, or body-only
             # review already addressed by a subsequent agent run whose diff
@@ -672,6 +670,11 @@ module Activities
       # clauses so it only fires when all other conditions align. Passing
       # pr_data through would avoid this call but adds complexity to
       # check_review_bot_status's already long keyword-arg list.
+      #
+      # pr_data returns a Sawyer::Resource, so `.head.sha` uses method
+      # dispatch. If pull_request is ever changed to return a plain Hash,
+      # this would need to switch to dig-style access. The nil guard below
+      # keeps this safe in either case.
       pr_data = client.pull_request(project.full_name, issue.github_number)
       head_sha = pr_data&.head&.sha
       return true if head_sha.nil? || head_sha == reviewed_commit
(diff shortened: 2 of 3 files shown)


app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -631,6 +643,39 @@ def body_only_review_needs_followup?(latest_review, last_run)
...
+        project.full_name, issue.github_number
+      )
+      reviewed_paths = comments
+        .select { |c| c[:pull_request_review_id] == review_id }
+        .filter_map { |c| c[:path] }
+        .to_set
+      return true if reviewed_paths.empty?
+
+      pr_data = client.pull_request(project.full_name, issue.github_number)
+      head_sha = pr_data&.head&.sha
```
viamin, line 679: **Minor: redundant `pull_request` API call.**

`pr_data` (with `head.sha`) is already fetched in `scan_pr` at line 69/76 and threaded through to the scan methods, but `check_review_bot_status` doesn't receive it. This means `review_diff_touches_reviewed_files?` makes an extra API call to get the same data.

Not blocking — the call is behind multiple guard clauses so it only fires when all other conditions align, and adding another keyword arg to `check_review_bot_status` may not be worth the added complexity. But worth noting as a future optimization opportunity if this path gets called frequently.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -631,6 +643,39 @@ def body_only_review_needs_followup?(latest_review, last_run)
...
+    # comments. Falls back to true (assumes addressed) when inline
+    # comments have no file paths or the comparison cannot be fetched.
+    def review_diff_touches_reviewed_files?(client, project, issue, review)
+      return true if client.nil? || project.nil? || issue.nil?
+
+      review_id = review[:id]
+      reviewed_commit = review[:commit_id]
+      return true if review_id.nil? || reviewed_commit.nil?
+
+      comments = client.pull_request_review_comments(
```
viamin, line 659: **Nit: this fetches all review comments for the entire PR**, not just for this review. The GitHub API doesn't support filtering by review ID server-side, so the client-side filter at line 661 is the right approach — just noting this is O(total comments) rather than O(comments on this review). Fine for typical PR sizes but could be worth a comment if you expect PRs with many reviews/comments.

spec/temporal/activities/scan_paid_prs_activity_spec.rb
```diff
@@ -989,6 +989,112 @@ def enable_copilot_and_codex_review!(proj = project)
...
+          review_threads: []
+        )
+        allow(github_client).to receive(:pull_request_review_comments)
+          .with(project.full_name, 42)
+          .and_return([])
+      end
+
+      it "falls back to treating the review as addressed" do
+        result = activity.execute(project_id: project.id)
+
```
viamin, line 1130: **Suggestion: consider adding a test for the `GithubClient::Error` rescue path.**

The method rescues `GithubClient::Error` and falls back to `true` (assumes addressed). A test where `compare_changed_files` raises would verify that the rescue path correctly allows clearance rather than crashing. Not critical since the pattern is well-established in the codebase, but it's the one untested branch.

viamin (commented):
## Review Summary

Solid fix that adds a meaningful defense-in-depth layer to prevent unrelated commits from clearing body-only review findings. The logic is correct, the fallback behavior is appropriately conservative (falls back to clearing, not blocking), and the test coverage hits the key cases.

Two observations below — one minor performance note and one suggestion for test completeness. Neither is blocking.

viamin:
## Agent Update

Commit succeeded. All lint checks passed and the pre-commit hook completed successfully.

Here's a summary of what I addressed:

1. **Redundant `pull_request` API call** (line 667): Added a comment documenting this as a known optimization opportunity — `pr_data` is already fetched upstream in `scan_pr` but not threaded through `check_review_bot_status`. The call is behind multiple guard clauses so the impact is minimal.

2. **O(total comments) review comment fetch** (line 657): Added a comment explaining that the GitHub API does not support server-side filtering by review ID, so the client-side filter is the correct approach, and this is fine for typical PR sizes.

3. **Missing `GithubClient::Error` rescue path test** (spec line 1093): Added a new test context where `compare_changed_files` raises `GithubClient::Error`, verifying the method gracefully falls back to treating the review as addressed (allowing clearance).

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -558,11 +558,23 @@ def check_review_bot_status(reviews, unresolved_threads, project: nil, last_run:
...
+              !review_diff_touches_reviewed_files?(client, project, issue, latest)
+            # Body-only bot whose review pre-dates the last agent run
+            # (timestamp guard passed), but the interceding diff did not
+            # touch any file mentioned in the review's inline comments.
+            # Treat as still-unaddressed to prevent unrelated changes
+            # (e.g. a CI schema fix) from clearing review findings.
+            [
+              { type: "review_bot_review_pending", details: "Latest review bot review was not clean" },
+              { type: "review_bot_comments", details: "Latest review bot review generated comments (body-only)" }
+            ]
```
viamin: Nit: This trigger array is now duplicated verbatim from lines 557–560. Not blocking, but if a third branch ever needs the same value, consider extracting it into a local variable above the conditional (e.g. `body_only_pending_triggers`).

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -631,6 +643,48 @@ def body_only_review_needs_followup?(latest_review, last_run)
...
+        .to_set
+      return true if reviewed_paths.empty?
+
+      # NOTE: pr_data is already fetched in scan_pr, but check_review_bot_status
+      # does not receive it. This extra API call is behind multiple guard
+      # clauses so it only fires when all other conditions align. Passing
+      # pr_data through would avoid this call but adds complexity to
+      # check_review_bot_status's already long keyword-arg list.
+      pr_data = client.pull_request(project.full_name, issue.github_number)
+      head_sha = pr_data&.head&.sha
```
viamin, line 679: The NOTE about the extra `pull_request` API call is helpful. Worth noting: `pr_data` here returns a raw Sawyer::Resource, so `.head.sha` works — but if `pull_request` ever gets wrapped to return a plain Hash (like `pull_request_reviews`), this would silently break. The guard clause on line 677 handles `nil` gracefully, so this is low-risk, just something to be aware of for future refactors.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -631,6 +643,48 @@ def body_only_review_needs_followup?(latest_review, last_run)
...
+
+      review_id = review[:id]
+      reviewed_commit = review[:commit_id]
+      return true if review_id.nil? || reviewed_commit.nil?
+
+      # NOTE: The GitHub API does not support filtering review comments by
+      # review ID server-side, so we fetch all comments for the PR and
+      # filter client-side. This is O(total comments) rather than
+      # O(comments on this review), which is fine for typical PR sizes.
+      comments = client.pull_request_review_comments(
```
viamin, line 659: This fetches **all** review comments for the PR, not just comments from the specific review bot. For PRs with many human review rounds this could be a non-trivial payload. The guard clauses above mean this only fires in a narrow window, so it's fine for now — but if this method gets called more broadly in the future, consider filtering by bot login before hitting the API (or at least before building the `reviewed_paths` set).

Actually, looking again, you already filter by `pull_request_review_id` on line 665, which scopes it to the specific review. That's sufficient. 👍

app/services/github_client.rb
```diff
@@ -189,6 +189,19 @@ def pull_request_files(repo, number)
     end.map(&:filename)
   end
 
+  # Compares two commits and returns the list of changed file paths.
+  #
+  # @param repo [String] Repository in "owner/name" format
+  # @param base [String] Base commit SHA
+  # @param head [String] Head commit SHA
+  # @return [Array<String>] File paths changed between base and head
+  def compare_changed_files(repo, base, head)
```
viamin, line 204: Clean addition. One edge case to be aware of: GitHub's compare API returns a max of 300 files per response. For very large diffs (e.g., a rebase onto a distant base), `comparison.files` may be truncated without indication. This is unlikely for the commit-to-HEAD diffs this method will see, but worth documenting if it's ever reused for broader comparisons.

viamin (commented):
## Review Summary

Clean, well-structured change that closes the gap described in #935. The fallback-to-true design ensures this is strictly additive — it can only make clearance stricter, never block where it previously succeeded. Tests cover the important paths. A few minor observations below, nothing blocking.

**Strengths:**
- Guard clauses and graceful error fallback mean no regressions if the new API calls fail
- File-path granularity (not line-level) is the right trade-off for this heuristic
- Good test coverage of the key branches: unrelated diff, related diff, no inline comments, and API error

**Minor observations:**
- The trigger array is now duplicated across two `elsif` branches — could be extracted, but not worth the complexity given the clarity of the conditional structure
- No test for the `head_sha == reviewed_commit` early-return path (trivial, but worth noting for completeness)

viamin:
## Agent Update

All 5630 examples pass with 0 failures now that CSS/JS assets are built. Commit succeeded.

1. **Duplicated trigger array** (Thread 1): Extracted the duplicated trigger array into a `body_only_pending_triggers` local variable, referenced from both `elsif` branches.

2. **Sawyer::Resource vs Hash concern** (Thread 2): Added a comment documenting that `pr_data` returns a `Sawyer::Resource` (so `.head.sha` works via method dispatch) and noting the nil guard keeps it safe if the return type ever changes.

3. **Review comment filtering** (Thread 3): Already resolved by the reviewer ("That's sufficient. 👍") — no changes needed.

4. **GitHub compare API 300-file limit** (Thread 4): Added a `NOTE` to `compare_changed_files` documenting the 300-file truncation limit and that it's unlikely to affect the narrow commit-to-HEAD diffs this method currently serves.

viamin (commented):
Generated no new comments. The PR cleanly addresses #935 — the file-path heuristic is well-gated behind existing conditions, fallback behavior is consistently permissive (matching the existing pattern), and all key scenarios are covered by tests.

PR #955, opened 2026-04-08
Fix #937: Sync workflow status to UI in real time via Turbo Streams + authoritative WorkflowState

### LINKED ISSUE #937: Sync workflow status to UI in real time via Turbo Streams + authoritative WorkflowState
## Problem

The project page's Issue Monitoring panel can show stale state for hours after the underlying workflow has recovered. Observed today: the panel showed **Delayed / Last checked about 2 hours ago** for project `paid`, while the database showed `last_polled_at` was current and `PollWorkflowHealthCheckJob` had already healed the workflow.

Two root causes:

1. **The UI never refreshes.** `app/views/projects/show.html.erb:175` renders the status as a lazy `turbo_frame_tag` with no auto-refresh and nothing broadcasting updates. Once loaded, the frame is frozen until the user reloads the page.
2. **`WorkflowState` is not authoritative.** Per the comment at `app/controllers/workflow_statuses_controller.rb:48-50`, `WorkflowState` records "are not reliably persisted by all activity paths," so the controller queries Temporal directly on every request. There is no single source of truth the UI (or anything else) can subscribe to.

### Part 1 — Broadcast workflow status changes via Turbo Streams

Have the code paths that mutate poll-workflow state broadcast a `replace` to the `workflow-status` frame on the project channel:

- `ProjectWorkflowManager.restart_polling` (manual + health-check restarts)
- The poll-cycle completion path that updates `Project#last_polled_at`
- `PollWorkflowHealthCheckJob` when it detects staleness or restarts a workflow

The frame at `app/views/workflow_statuses/show.html.erb:1` is already keyed for this. Subscribers on `app/views/projects/show.html.erb` would get live updates without polling the controller.

### Part 2 — Make `WorkflowState` the authoritative source

Fix the "not reliably persisted" gap at the source:

- Every poll-cycle completion writes a `WorkflowState` row (status, `last_polled_at`, last error if any).
- Every restart writes a row with the restart reason.
- `WorkflowStatusesController#compute_automation_health` reads from `WorkflowState` instead of calling `ProjectWorkflowManager.workflow_status` on every page load. Falls back to Temporal only if no recent record exists.

Benefits:

- The UI can show **why** a workflow is delayed (last error, last restart reason), not just a yellow dot.
- Removes the per-request Temporal RPC from the project page.
- Gives Turbo Stream broadcasts a single, consistent source to render from.

## Out of scope

- Auto-refresh-only fix (cheaper alternative discussed but rejected in favor of the durable fix).
- Tightening the `PollWorkflowHealthCheckJob` detection window — separate concern.

## Acceptance criteria

- [ ] Restarting or healing a poll workflow updates the project page within seconds without a manual reload.
- [ ] `WorkflowStatusesController` reads from `WorkflowState` for the common case; Temporal RPC is a fallback.
- [ ] All poll-cycle completions and restart paths write `WorkflowState` rows.
- [ ] The UI surfaces the last restart reason / last error when status is `:stale` or `:unhealthy`.

viamin:
## Summary

Eliminates stale workflow status on the project page by making `WorkflowState` the authoritative source of truth and broadcasting changes to the UI in real time via Turbo Streams. Previously, the Issue Monitoring panel could show outdated state for hours because the turbo frame never refreshed after initial load and `WorkflowState` records weren't reliably written by all code paths.

## Changes

**Database & Models**
- Ensure all poll-cycle completions and workflow restart paths write `WorkflowState` rows (status, `last_polled_at`, last error, restart reason)
- `WorkflowState` becomes the single authoritative source for workflow health, closing the "not reliably persisted" gap noted in the controller

**Services**
- `ProjectWorkflowManager.restart_polling` writes a `WorkflowState` row with the restart reason on manual and health-check restarts
- Poll-cycle completion path updates `WorkflowState` alongside `Project#last_polled_at`
- `PollWorkflowHealthCheckJob` writes state when it detects staleness or triggers a restart

**Controllers**
- `WorkflowStatusesController#compute_automation_health` reads from `WorkflowState` instead of making a Temporal RPC on every page load
- Falls back to Temporal only when no recent `WorkflowState` record exists

**UI & Turbo Streams**
- Broadcast `replace` targeting the `workflow-status` turbo frame on the project channel whenever workflow state changes
- Project page subscribes to the stream so status updates arrive within seconds without manual reload
- Surface last restart reason and last error in the UI when status is `:stale` or `:unhealthy`

## Design Decisions

- **`WorkflowState` as authoritative over Temporal RPC**: The per-request Temporal query was both expensive and redundant once all mutation paths write state. Temporal remains as a fallback for edge cases where no recent record exists, preserving correctness without the cost on every page load.
- **Turbo Stream `replace` over periodic polling**: A `turbo_frame` with `src` and a refresh interval was considered but rejected — it would still hit the controller on a timer and wouldn't reflect changes instantly. Broadcasting from the mutation sites gives true real-time updates with zero client-side polling.
- **Writing state at every mutation site rather than a single after-commit hook**: The write points are few and explicit (restart, poll completion, health check), making the broadcast triggers easy to audit. A model callback would risk broadcasting on internal bookkeeping writes.

---

Closes #937


app/services/project_workflow_manager.rb
```diff
@@ -62,6 +66,11 @@ def restart_polling(project, reason: nil)
       raise unless e.code == Temporalio::Error::RPCError::Code::NOT_FOUND
       # Workflow not found — that's fine, we'll start fresh
     ensure
+      WorkflowState.record_polling_status(
+        project,
+        status: "running",
+        restart_reason: reason || "self-healing restart"
+      )
       start_polling(project)
```
viamin: 🔴 **Race condition / incorrect state on failure**: `ensure` runs regardless of whether `handle.terminate` raised a non-NOT_FOUND error. If `terminate` fails with a network error (or any unexpected exception), this block still records status `"running"` and calls `start_polling` — which itself might also fail. The `WorkflowState` would then say `running` when nothing is actually running.

Moreover, `record_polling_status` is called *before* `start_polling`, so if `start_polling` raises, the state says `running` but no workflow was started.

Consider restructuring so the state is recorded only after `start_polling` succeeds:

```suggestion
      start_polling(project)
      WorkflowState.record_polling_status(
        project,
        status: "running",
        restart_reason: reason || "self-healing restart"
      )
```

Note: `start_polling` already calls `record_polling_status` internally (line 17), so the call here in `restart_polling` may be redundant — the only difference is `restart_reason`. You could add a `restart_reason:` parameter to `start_polling` instead and remove the duplicate call.

app/services/project_workflow_manager.rb
```diff
@@ -14,6 +14,8 @@ def start_polling(project)
         task_queue: Paid.task_queue
       )
 
+      WorkflowState.record_polling_status(project, status: "running")
```
viamin: 🟡 **Double write on restart path**: `restart_polling` calls `record_polling_status` (line 69-73) and then calls `start_polling` (line 74), which *also* calls `record_polling_status` here. This results in two back-to-back DB writes and two Turbo Stream broadcasts for a single restart.

Consider having `start_polling` accept an optional `restart_reason:` keyword and doing a single write, or skipping the write in `restart_polling` and relying on `start_polling` to do it.

app/models/project.rb
```diff
@@ -316,6 +316,7 @@ def touch_last_github_activity_at(timestamp = Time.current)
 
   def touch_last_polled_at(timestamp = Time.current)
     update_column(:last_polled_at, timestamp)
+    WorkflowState.record_polling_status(self, status: "running")
```
viamin, line 320: 🟡 **Performance on every poll cycle**: `touch_last_polled_at` is called on every successful poll cycle. This now triggers `record_polling_status`, which does a `find_or_initialize_by` + `save!` + `broadcast_workflow_status_update` (which itself does another `find_by` + `compute_health_for` + broadcast). That's 2-3 extra DB queries and an ActionCable broadcast on every poll cycle, potentially every 60 seconds per project.

Consider rate-limiting the broadcast — e.g., only broadcast if the status actually changed, or if the last broadcast was more than N seconds ago. At minimum, the `broadcast_workflow_status_update` method could skip broadcasting when the status remains `:healthy` since the UI already shows a healthy state.

app/models/workflow_state.rb
```diff
@@ -13,6 +13,77 @@ class WorkflowState < ApplicationRecord
...
+  # Called by ProjectWorkflowManager on start/restart and by PollWorkflowHealthCheckJob.
+  def self.record_polling_status(project, status:, restart_reason: nil, error_message: nil)
+    workflow_id = "github-poll-#{project.id}"
+
+    ws = find_or_initialize_by(temporal_workflow_id: workflow_id)
+    ws.assign_attributes(
+      project: project,
+      workflow_type: "GitHubPollWorkflow",
+      status: status,
+      restart_reason: restart_reason || ws.restart_reason,
```
viamin: 🟡 **Sticky `restart_reason`**: The fallback `restart_reason: restart_reason || ws.restart_reason` means the `restart_reason` is never cleared — it persists forever after the first restart. When the workflow is running healthy, `compute_health_for` won't expose it (only shown for `:stale`/`:unhealthy`), but it's still a bit surprising that the field accumulates stale data.

Consider explicitly clearing it when recording a normal `running` status without a restart (i.e., when `restart_reason` param is nil and we're transitioning *to* running from a non-restart path like `touch_last_polled_at`).

app/models/workflow_state.rb
```diff
@@ -13,6 +13,77 @@ class WorkflowState < ApplicationRecord
...
+      error_message: error_message,
+      started_at: status == "running" ? Time.current : ws.started_at,
+      completed_at: FINISHED_STATUSES.include?(status) ? Time.current : nil
+    )
+    ws.save!
```
viamin, line 48: 🟡 **`save!` without transaction safety**: If `save!` succeeds but the subsequent `broadcast_workflow_status_update` raises (e.g., Redis/ActionCable is down), the state is persisted but the UI isn't updated. That's probably acceptable. However, the reverse concern is more important: if this is called from within another transaction that rolls back, the broadcast will have already fired with data that no longer exists.

Consider wrapping the broadcast in an `after_commit` callback or using `ActiveRecord::Base.after_commit { ... }` to ensure broadcasts only fire after the transaction commits.

app/views/workflow_statuses/_status.html.erb
```diff
@@ -0,0 +1,44 @@
...
+              <% end %>
+              <dt>Check interval</dt>
+              <dd class="text-gray-900">Every <%= pluralize(project.poll_interval_seconds, "second") %></dd>
+            </dl>
+          <% end %>
+        </div>
+      </div>
+    </div>
+  </div>
+<% end %>
```
viamin, line 53: 🟡 **Duplication with `show.html.erb`**: The partial and `show.html.erb` are nearly identical, but `show.html.erb` includes the "Restart monitor" button (line 30-36) while the partial does not. This means Turbo Stream broadcasts will *replace* the frame content with the partial, which lacks the restart button — so a user viewing an unhealthy workflow will lose the restart button after the first broadcast.

Either:
1. Move the restart button into the partial and pass a `policy` or `show_restart` local, or
2. Refactor `show.html.erb` to render the partial and only add the button conditionally.

app/jobs/poll_workflow_health_check_job.rb
```diff
@@ -69,6 +69,13 @@ def check_and_heal(project)
 
     return unless RESTARTABLE_STATUSES.include?(status[:status])
 
+    # Record the unhealthy state before re-checking
+    WorkflowState.record_polling_status(
+      project,
+      status: map_temporal_status(status[:status]),
+      error_message: "Workflow in #{status[:status]} state"
+    )
```
viamin, line 77: 🟢 **Potential string interpolation of Temporal enum**: `"Workflow in #{status[:status]} state"` — if `status[:status]` is a Temporal enum constant (e.g., `Temporalio::Client::WorkflowExecutionStatus::FAILED`), the `.to_s` output might not be human-friendly. Consider using `map_temporal_status` for the error message too, or at least verifying the string representation is readable.

app/models/workflow_state.rb
```diff
@@ -13,6 +13,77 @@ class WorkflowState < ApplicationRecord
...
+    return nil unless ws
+
+    ws.automation_health(project)
+  end
+
+  def automation_health(project)
+    unless running?
+      description = "The issue monitor is not running. " \
+                    "It may be automatically restarted shortly if eligible."
+      description = "#{description} Last error: #{error_message}" if error_message.present?
```
viamin, line 83: 🟢 **XSS concern — `error_message` in description**: The `error_message` is interpolated into the description string, which is then rendered via `<%= health[:description] %>` in the ERB template. Rails auto-escapes `<%= %>` output, so this is safe. Just flagging for awareness — if this ever moves to `raw` or `html_safe`, the `error_message` (which could originate from Temporal error output) would need sanitization.

viamin (commented):
## Review Summary

Solid PR that addresses all four acceptance criteria from #937. The architectural decision to make `WorkflowState` authoritative with Temporal as fallback is sound and well-executed. The Turbo Stream broadcasting from mutation sites is the right pattern.

A few issues worth addressing, ranging from a race condition bug to duplication cleanup.

**Severity guide**: 🔴 bug/correctness, 🟡 suggestion, 🟢 nit

viamin:
## Agent Update

All review feedback addressed and committed. Here's a summary of what was fixed:

### Changes per review thread

1. **🔴 Race condition in `restart_polling`** — Removed the `ensure` block that ran regardless of whether `terminate` raised a non-NOT_FOUND error. Now uses `begin/rescue/end` so non-NOT_FOUND errors propagate without recording false "running" state or calling `start_polling`.

2. **🟡 Double write on restart path** — Added `restart_reason:` keyword to `start_polling`. `restart_polling` now delegates to `start_polling(project, restart_reason: ...)` for a single DB write and broadcast instead of two.

3. **🟡 Performance on every poll cycle** — `record_polling_status` now checks `should_broadcast = ws.new_record? || ws.status_changed?` before save and skips the broadcast when status is unchanged (the common healthy→healthy case).

4. **🟡 Sticky `restart_reason`** — Changed from `restart_reason || ws.restart_reason` to just `restart_reason`, so it's cleared when recording normal running status without a restart.

5. **🟡 `save!` without transaction safety** — Broadcast now uses `ActiveRecord.after_all_transactions_commit` to ensure it only fires after the transaction commits.

6. **🟡 Partial/show duplication losing restart button** — Moved the restart button into the partial (gated by `local_assigns[:show_restart]`). Refactored `show.html.erb` to render the partial with `show_restart: policy(@project).update?`. Turbo Stream broadcasts no longer strip the button.

7. **🟢 Temporal enum string interpolation** — Changed to use `map_temporal_status` for the human-friendly error message.

8. **🟢 XSS concern** — No change needed; Rails `<%= %>` auto-escaping handles this. Noted for awareness.

app/models/workflow_state.rb
```diff
@@ -13,6 +13,85 @@ class WorkflowState < ApplicationRecord
...
+    ws = find_or_initialize_by(temporal_workflow_id: workflow_id)
+    ws.assign_attributes(
+      project: project,
+      workflow_type: "GitHubPollWorkflow",
+      status: status,
+      restart_reason: restart_reason,
+      error_message: error_message,
+      started_at: status == "running" ? Time.current : ws.started_at,
+      completed_at: FINISHED_STATUSES.include?(status) ? Time.current : nil
+    )
```
viamin, line 30: Bug: every call from `touch_last_polled_at` passes the default `restart_reason: nil, error_message: nil`, which overwrites any previously stored values. The restart reason and last error will be visible only for the brief window between a restart/error and the next successful poll cycle — then they're wiped.

The fix is to only update these fields when explicitly provided:

```suggestion
      project: project,
      workflow_type: "GitHubPollWorkflow",
      status: status,
      started_at: status == "running" && ws.status != "running" ? Time.current : ws.started_at,
      completed_at: FINISHED_STATUSES.include?(status) ? Time.current : nil
    )
    ws.restart_reason = restart_reason if restart_reason
    ws.error_message = error_message if error_message
    # Clear error/reason when transitioning back to running from a non-running state
    if status == "running" && ws.status_previously_changed?
      ws.restart_reason = nil unless restart_reason
      ws.error_message = nil unless error_message
    end
```

Alternatively, a simpler approach: only assign `restart_reason` and `error_message` when the caller explicitly passes them (change the defaults to a sentinel like `:not_provided` and skip assignment when not provided).

viamin (commented):
The architecture is sound — making `WorkflowState` authoritative and broadcasting via Turbo Streams is the right approach. Two bugs in `record_polling_status` will undermine the feature: every poll cycle clears `restart_reason`/`error_message` and resets `started_at`, so the new UI fields will almost never show meaningful data.

app/models/workflow_state.rb
```diff
@@ -13,6 +13,85 @@ class WorkflowState < ApplicationRecord
...
+    workflow_id = "github-poll-#{project.id}"
+
+    ws = find_or_initialize_by(temporal_workflow_id: workflow_id)
+    ws.assign_attributes(
+      project: project,
+      workflow_type: "GitHubPollWorkflow",
+      status: status,
+      restart_reason: restart_reason,
+      error_message: error_message,
+      started_at: status == "running" ? Time.current : ws.started_at,
```
viamin: Bug: `started_at` is set to `Time.current` on **every** call where `status == "running"`, including every poll cycle via `touch_last_polled_at`. This makes `started_at` reflect the last poll time, not when the workflow actually started.

It should only be set when transitioning *to* running:

```suggestion
      started_at: status == "running" && !ws.running? ? Time.current : ws.started_at,
```

viamin:
## Agent Update

Done. Here's a summary of what was fixed:

**Thread 1 — `started_at` reset on every poll cycle**: Changed the condition from `status == "running"` to `status == "running" && !ws.running?` (stored as `transitioning_to_running`), so `started_at` is only set when the workflow transitions *into* the running state, not on every poll cycle.

**Thread 2 — `restart_reason`/`error_message` overwritten to nil**: Changed the defaults from `nil` to `:not_provided` sentinel values. The fields are now only assigned when the caller explicitly provides them. When transitioning back to running without explicit values, they're cleared to prevent stale reasons from lingering. On normal poll cycles (via `touch_last_polled_at`), existing values are preserved.

viamin:
## Agent Update

The background task confirmed the commit completed successfully (exit code 0). Everything is done — merge conflict resolved, lint and tests passing, commit landed as `cbb44ccc`.

app/models/project.rb
```diff
@@ -405,6 +406,18 @@ def broadcast_issues_update
...
+  def broadcast_workflow_status_update
+    health = WorkflowState.compute_health_for(self)
+    return unless health
+
+    broadcast_replace_to(
+      self, :project_updates,
+      target: "workflow-status",
+      partial: "workflow_statuses/status",
+      locals: { project: self, health: health }
+    )
```
viamin, line 422: The broadcast renders the `_status` partial without passing `show_restart`, so whenever a Turbo Stream replaces the `workflow-status` frame, authorized users lose the "Restart monitor" button. It only reappears on a full page reload.

Since broadcasts go to all subscribers and you can't determine per-connection authorization, one option is to always render the button in broadcasts and rely on the controller's `authorize @project, :update?` to gate the actual restart action. Alternatively, you could use a Stimulus controller that conditionally shows/hides the button based on a data attribute set during initial page render.

```suggestion
  def broadcast_workflow_status_update
    health = WorkflowState.compute_health_for(self)
    return unless health

    broadcast_replace_to(
      self, :project_updates,
      target: "workflow-status",
      partial: "workflow_statuses/status",
      locals: { project: self, health: health, show_restart: true }
    )
  end
```

Showing the button to all users in the broadcast is safe since the controller enforces authorization on the `restart` action. Unauthorized users clicking it would get a 403. If you'd prefer to hide it from unauthorized users entirely, a Stimulus approach would be needed.

viamin (commented):
Solid approach — making `WorkflowState` authoritative and broadcasting via Turbo Streams is the right call. One functional issue with the broadcast path losing the restart button for authorized users.

viamin:
## Agent Update

The background task notification is stale — it refers to the earlier failed commit attempt (before CSS assets were built). The commit has already succeeded in the foreground:

- Commit `950c76d5` landed successfully with 5764 examples, 0 failures
- All changes are committed, working tree is clean

app/models/project.rb
```diff
@@ -405,6 +406,18 @@ def broadcast_issues_update
...
 
+  def broadcast_workflow_status_update
+    health = WorkflowState.compute_health_for(self)
+    return unless health
+
+    broadcast_replace_to(
+      self, :project_updates,
+      target: "workflow-status",
+      partial: "workflow_statuses/status",
+      locals: { project: self, health: health, show_restart: true }
```
viamin: **Security: authorization bypass in broadcast.**

`show_restart: true` is hardcoded, so every user subscribed to the project channel receives the restart button — regardless of whether they have `update?` permission. The original `show.html.erb` gated this with `policy(@project).update?`.

Since broadcasts run outside a request context (no `current_user`), you can't call Pundit here. The simplest fix is to not show the restart button in broadcasts:

```suggestion
      locals: { project: self, health: health, show_restart: false }
```

The initial page load from `WorkflowStatusesController#show` already passes the correct policy check. The restart action itself also enforces `authorize @project, :update?`, so this isn't a privilege escalation — but showing the button to unauthorized users is still a UI correctness issue.

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_04_09_231400) do
+ActiveRecord::Schema[8.1].define(version: 2026_04_11_012931) do
```
viamin: **Scope: unrelated schema changes included.**

This diff adds the entire `notifications` table (with indexes and foreign keys), but there is no corresponding migration in this PR — only `20260408105320_add_restart_reason_to_workflow_states.rb`. The schema version also jumped to `2026_04_11_012931` which doesn't match any migration here.

Please regenerate `schema.rb` from a clean state so the diff is limited to the `restart_reason` column addition on `workflow_states`.

viamin (commented):
Overall clean implementation — the WorkflowState-as-authoritative-source approach is solid and the broadcast-on-mutation pattern is well-chosen. Two issues found.

viamin:
## Agent Update

Everything is done. Here's a summary:

**Merge conflicts resolved:** `db/schema.rb` was the only conflict — resolved by accepting main's version and regenerating from the database.

**Review feedback addressed:**

1. **Security: authorization bypass in broadcast** (app/models/project.rb:421) — Changed `show_restart: true` to `show_restart: false` in `broadcast_workflow_status_update`. The initial page load in `show.html.erb` still correctly gates the button via `policy(@project).update?`, and the restart action itself enforces `authorize @project, :update?`. The broadcast no longer shows the restart button to unauthorized users.

2. **Unrelated schema changes** (db/schema.rb) — Regenerated `schema.rb` from a clean database state (`RAILS_ENV=test rails db:schema:dump`). The schema version now correctly reflects `2026_04_10_204751` (latest migration in main), and the diff only includes columns from migrations in this PR (`restart_reason` on `workflow_states`) plus those from main's merged migrations.

**Verification:** 5858 examples, 0 failures, 2 pending. Rubocop clean (721 files, no offenses).

viamin (commented):
Generated no new comments. The PR cleanly addresses #937: WorkflowState is now the authoritative source, all mutation paths write state, and Turbo Stream broadcasts provide real-time UI updates. The sentinel-value pattern for optional params, the pre-save dirty check for broadcast gating, and the after_all_transactions_commit callback are all correctly implemented.

PR #894, opened 2026-04-07
Fix #864: skip unchanged issues when fetching comments

viamin:
## Summary

- Track whether `github_updated_at` changed between syncs in `FetchIssuesActivity`
- Only fetch issue comments for issues that have been updated on GitHub since the last sync
- Uses second-level timestamp comparison to handle DB precision truncation

## Problem

`FetchIssuesActivity#parse_issue_relationships` fetched `issue_comments` for every open non-PR issue on every poll cycle (N+1 pattern). Most issues haven't changed between cycles, so this wasted GitHub API rate limit budget.

## Impact

For a project with 50 open issues where only 2 have been updated since the last sync, reduces ~50 comment API calls per cycle to ~2 (~96% reduction).

Closes #864
Part of #879

PR #956, opened 2026-04-08
Fix #934: fix(auto-merge): require clean re-review on HEAD commit for body-only bot clearance

### LINKED ISSUE #934: fix(auto-merge): require clean re-review on HEAD commit for body-only bot clearance
## Problem

PR #894 auto-merged despite an unaddressed P1 Codex review comment. The body-only bot anti-loop guard (`body_only_review_needs_followup?` at `scan_paid_prs_activity.rb:587`) only checks whether **any** agent run completed after the review timestamp. A CI schema fix that completed after the P1 review was enough to clear the feedback, even though the finding was never addressed.

## Root Cause

The timestamp bypass at lines 512-531 treats any post-review agent run completion as "feedback addressed," regardless of whether the run's changes relate to the review findings.

## Proposed Fix

Only clear body-only review feedback when the bot has posted a clean signal (`REVIEW_BOT_CLEAN_PATTERN` or `BODY_ONLY_BOT_CLEAN_COMMENT_PATTERN`) on the PR's current HEAD commit. The `paid-review-trigger` mechanism already requests re-reviews after agent runs — don't bypass pending feedback until that re-review comes back clean on the latest commit.

The timestamp guard (`body_only_review_needs_followup?`) should become a safety net for truly stuck loops (e.g., re-review never arrives), not the primary clearance path.

### Key changes
- In `check_review_bot_status`, check whether the clean signal (review or comment) targets the current HEAD commit before returning `[]`
- Keep the timestamp guard as a fallback with a configurable timeout (e.g., 30 minutes) to prevent permanent wedges
- Consider adding the reviewed commit SHA to the clean-check logic

## Impact

Without this fix, any agent activity after a body-only bot review can inadvertently clear legitimate findings.

Depends on none.

viamin:
## Summary

Prevent premature auto-merge clearance when a body-only bot review has unaddressed findings by requiring clean re-review signals on the PR's current HEAD commit, not just any post-review agent activity. Previously, unrelated agent runs (e.g., a CI schema fix) completing after a P1 review could inadvertently clear legitimate feedback, as happened with PR #894.

## Changes

**Data layer (`app/services/github_client.rb`)**
- Added `commit_id` field to review hashes returned by `pull_request_reviews`, enabling commit-level review tracking

**Core logic (`app/temporal/activities/scan_paid_prs_activity.rb`)**
- Threaded `head_sha:` parameter through `check_review_bot_status` and `no_outstanding_review_feedback?` to all call sites
- Clean reviews (`:clean` status) are now verified against HEAD commit SHA — stale-commit clean reviews trigger a new review request instead of clearing feedback
- `body_only_bot_clean_comment_present?` now blocks clean comment override when a non-clean review exists on the current HEAD commit
- Converted `body_only_review_needs_followup?` from an immediate timestamp bypass to a 30-minute timeout fallback (`BODY_ONLY_REVIEW_FOLLOWUP_TIMEOUT`), keeping feedback blocking while awaiting a clean re-review on HEAD
- Draft phase fetches `pr_data` early to resolve HEAD SHA for review bot checks

**Tests (`spec/temporal/activities/scan_paid_prs_activity_spec.rb`)**
- Added 4 new test cases covering: stale-commit clean review rejection, HEAD-commit clean review acceptance, non-clean HEAD review blocking clean comment override, and 30-minute timeout fallback behavior

## Design Decisions

- **Timeout as safety net, not primary path**: The timestamp guard (`body_only_review_needs_followup?`) is retained as a 30-minute fallback to prevent permanent wedges if a re-review never arrives, but it no longer serves as the primary clearance mechanism. This balances correctness with resilience.
- **Commit-level verification over content matching**: Rather than attempting to correlate review findings with specific code changes (which would require semantic analysis), the fix uses commit SHA comparison — a mechanically simple check that aligns with the ZFC principle of keeping orchestration logic structural.
- **Leveraging existing re-review infrastructure**: The `paid-review-trigger` mechanism already requests re-reviews after agent runs, so requiring a clean signal on HEAD commit works with the existing flow rather than adding new review request paths.

---

Closes #934


app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -108,10 +113,16 @@ def scan_draft_pr(project, client, issue, pr_data: nil)
 
       # Draft exit still requires an explicitly clean bot review even when
       # other draft comment signals are skipped.
+      # Resolve HEAD SHA early so check_review_bot_status can verify
+      # clean signals target the current commit.
+      pr_data ||= fetch_pr_data(client, project, issue)
```
viamin: **Minor performance note**: `fetch_pr_data` is now called unconditionally at the top of `scan_draft_pr`. Previously, when blocking review triggers were present (the `all_triggers.any?` path further down), the PR data fetch was skipped entirely.

In practice this is likely fine — the fetch is cheap relative to the other API calls already being made, and the `||=` ensures it's still only called once per invocation. Just flagging in case draft PR scan volume makes the extra call noticeable.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -505,15 +519,26 @@ def check_review_bot_status(reviews, unresolved_threads, project: nil, last_run:
...
+        head_sha: head_sha)
         return []
       end
 
       status = review_bot_review_status_from(reviews, latest)
 
       case status
       when :clean
-        []
+        if head_sha && latest[:commit_id] && latest[:commit_id] != head_sha
```
viamin: **Graceful degradation, but consider a log**: When `head_sha` is nil (e.g., `pr_data` fetch failed upstream) or `latest[:commit_id]` is nil (API didn't return it), this falls through to `[]` — effectively the old "always clean" behavior. That's a safe default, but it silently skips the new protection.

A debug/info log when either value is nil would help diagnose cases where the commit-level guard isn't firing as expected:

```suggestion
        if head_sha && latest[:commit_id] && latest[:commit_id] != head_sha
```

Maybe worth adding something like:
```ruby
if head_sha.nil? || latest[:commit_id].nil?
  Rails.logger.debug(message: "review_bot.clean_review_missing_commit_info", head_sha: head_sha, review_commit_id: latest[:commit_id])
end
```

viamin (commented):
## Review Summary

Solid fix that addresses the root cause described in #934 — the timestamp-only bypass is replaced with commit-level verification, and the timeout becomes a safety net rather than the primary clearance path. The approach aligns well with ZFC (mechanically simple, no semantic analysis needed). Tests are thorough and cover the key scenarios.

A few observations inline, mostly minor.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -628,7 +673,13 @@ def body_only_review_needs_followup?(latest_review, last_run)
...
       return true if cutoff.nil?
 
-      submitted_at > cutoff
+      # Review was posted after the run — definitely unaddressed.
+      return true if submitted_at > cutoff
+
+      # Review predates the run. Wait for a clean re-review on HEAD
+      # until the timeout expires, then treat as addressed to avoid
+      # permanent wedges.
+      Time.current < cutoff + BODY_ONLY_REVIEW_FOLLOWUP_TIMEOUT
```
viamin: The 30-minute timeout is a reasonable safety net. The issue body mentioned "configurable timeout" — is it intentional to hardcode this as a constant for now, with per-project configurability deferred? If so, that seems fine; just want to confirm it's a deliberate scope decision rather than an oversight.

Also worth noting: if an agent run completes but takes >30 minutes to get a re-review (e.g., Codex API is slow), this timeout could still allow premature clearance. In practice the re-review trigger fires quickly, so this is unlikely — just documenting the edge case.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -590,10 +620,20 @@ def body_only_review_bot?(login)
...
-    def body_only_bot_clean_comment_present?(client, project, issue, latest_review, allowed_bot_logins)
+    def body_only_bot_clean_comment_present?(client, project, issue, latest_review, allowed_bot_logins,
+      head_sha: nil)
       return false if allowed_bot_logins.nil? || allowed_bot_logins.empty?
       return false unless allowed_bot_logins.subset?(BODY_ONLY_REVIEW_BOT_LOGINS)
 
+      # If the latest review is a non-clean review on the current HEAD
+      # commit, a clean issue comment cannot override it — the bot reviewed
+      # this exact code and found issues.
+      if head_sha && latest_review &&
```
viamin: Good guard — this prevents a stale clean comment from overriding a fresh non-clean review on HEAD. One thing to note: if `latest_review[:commit_id]` is nil (missing from older reviews or API issue), the `== head_sha` comparison is false, so this guard won't fire and the method will proceed to check comments. That's the safe direction (allowing the comment check to proceed rather than blocking), so this is fine.

spec/temporal/activities/scan_paid_prs_activity_spec.rb
```diff
@@ -2194,6 +2305,7 @@ def stub_github_for_pr(
 
   def default_clean_copilot_review
     [ { id: 100, user_login: "copilot-pull-request-reviewer[bot]", state: "COMMENTED",
-        body: "Copilot reviewed 5 out of 5 changed files and generated no comments.", submitted_at: 1.hour.ago } ]
+        body: "Copilot reviewed 5 out of 5 changed files and generated no comments.", submitted_at: 1.hour.ago,
```
viamin: Good catch adding `commit_id: "abc123"` to the default clean Copilot review to match the stubbed HEAD SHA. This prevents existing tests from accidentally breaking due to the new stale-commit detection.

app/services/github_client.rb
```diff
@@ -463,7 +463,8 @@ def pull_request_reviews(repo, number)
           user_login: r.user&.login,
           state: r.state,
           body: r.body.to_s,
-          submitted_at: parse_timestamp(r.submitted_at)
+          submitted_at: parse_timestamp(r.submitted_at),
+          commit_id: r.commit_id
```
viamin, line 490: Clean addition. The `commit_id` field from the GitHub Reviews API is well-documented and stable. No concerns here.

viamin:
## Agent Update

That was the earlier failed commit attempt (before assets were compiled). It was already superseded — the commit succeeded on the subsequent try. Everything is good:

- Commit `70944886` is in place
- All 5630 tests pass, 0 failures
- Rubocop clean

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -505,15 +520,30 @@ def check_review_bot_status(reviews, unresolved_threads, project: nil, last_run:
...
+            head_sha: head_sha, review_commit_id: latest[:commit_id])
+        end
+        if head_sha && latest[:commit_id] && latest[:commit_id] != head_sha
+          login = project && review_bot_request_login(project)
+          if login
+            [ { type: "review_bot_review_pending", details: "Clean review is not on current HEAD commit",
+                request_login: login } ]
+          else
+            []
+          end
```
viamin: When a clean review targets a stale commit but `review_bot_request_login` returns `nil`, this falls through to `[]` — treating the stale review as clean. This means the HEAD commit verification is silently skipped for projects that have review bots enabled but no requestable login configured.

Is that intentional? If so, a debug log here would help with observability. If not, returning a non-empty trigger (or at least logging) would surface the gap.

```suggestion
          else
            Rails.logger.debug(message: "review_bot.stale_clean_review_no_requestable_login",
              review_commit_id: latest[:commit_id], head_sha: head_sha)
            []
```

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -505,15 +520,30 @@ def check_review_bot_status(reviews, unresolved_threads, project: nil, last_run:
...
 
       status = review_bot_review_status_from(reviews, latest)
 
       case status
       when :clean
-        []
+        if head_sha.nil? || latest[:commit_id].nil?
+          Rails.logger.debug(message: "review_bot.clean_review_missing_commit_info",
+            head_sha: head_sha, review_commit_id: latest[:commit_id])
+        end
```
viamin: This debug log fires when both `head_sha` and `commit_id` are `nil`, which is a normal/expected state (e.g., reviews globally disabled, or callers that don't pass `head_sha`). Consider narrowing the condition to only log when there's asymmetric missing data — i.e., one is present but the other isn't — which would indicate an actual gap in the data flow.

```suggestion
        if (head_sha.nil?) != (latest[:commit_id].nil?)
          Rails.logger.debug(message: "review_bot.clean_review_missing_commit_info",
            head_sha: head_sha, review_commit_id: latest[:commit_id])
        end
```

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -628,7 +678,14 @@ def body_only_review_needs_followup?(latest_review, last_run)
...
 
-      submitted_at > cutoff
+      # Review was posted after the run — definitely unaddressed.
+      return true if submitted_at > cutoff
+
+      # Review predates the run. Wait for a clean re-review on HEAD
+      # until the timeout expires, then treat as addressed to avoid
+      # permanent wedges. Hardcoded for now; per-project configurability
+      # deferred until we see whether 30 minutes is sufficient in practice.
+      Time.current < cutoff + BODY_ONLY_REVIEW_FOLLOWUP_TIMEOUT
```
viamin: The behavioral change here is significant: previously, a review predating the last agent run was always treated as addressed (`false`). Now it stays blocking for up to 30 minutes. This is well-motivated by the issue but worth noting that it changes the steady-state behavior for *all* body-only review bots, not just the race condition scenario.

If a project has slow agent runs (>30 min), the timeout will expire during the run and the feedback will be treated as addressed before the re-review arrives. Is 30 minutes sufficient for projects with longer cycle times, or should this be tied to the agent run duration rather than a fixed window?

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -108,10 +113,17 @@ def scan_draft_pr(project, client, issue, pr_data: nil)
 
       # Draft exit still requires an explicitly clean bot review even when
       # other draft comment signals are skipped.
+      # Resolve HEAD SHA early so check_review_bot_status can verify
+      # clean signals target the current commit. The extra fetch is cheap
+      # relative to other API calls in this path, and ||= ensures at most one.
+      pr_data ||= fetch_pr_data(client, project, issue)
```
viamin: Nit: the `||=` comment says "ensures at most one" but since `pr_data` defaults to `nil` as a keyword arg, this will always fetch on the first call to `scan_draft_pr`. The `||=` only prevents a redundant re-fetch from the later `pr_data ||=` at line 148. The extra API call is cheap as noted, but the comment could be slightly clearer about what it's deduplicating against.

spec/temporal/activities/scan_paid_prs_activity_spec.rb
```diff
@@ -923,13 +923,42 @@ def enable_copilot_and_codex_review!(proj = project)
         )
       end
```
viamin, line 1011: Good catch updating this test — changing `completed_at` from `30.minutes.ago` to `1.hour.ago` ensures the 30-minute timeout has elapsed. The new description also accurately reflects the new semantics.

app/services/github_client.rb
viamin, line 490: Clean addition. The GitHub API always returns `commit_id` for PR reviews, so this should always be populated for real API responses. Worth noting that test fixtures will need to include `commit_id` to exercise the new commit-level verification — which the updated test helper at line 2309 already handles.

viamin (commented):
Solid fix for the auto-merge race condition described in #934. The commit-level verification approach is mechanically simple and aligns well with ZFC. The 30-minute timeout fallback is a good safety net. A few items worth considering below.

viamin:
## Agent Update

Commit succeeded. All 97 tests pass, 0 failures, rubocop clean.

1. **Stale clean review with no requestable login (line 543)**: Added `Rails.logger.debug` with `review_bot.stale_clean_review_no_requestable_login` message to surface the silent fallthrough.

2. **Overly broad debug log condition (line 535)**: Changed `head_sha.nil? || latest[:commit_id].nil?` to `(head_sha.nil?) != (latest[:commit_id].nil?)` — now only logs when there's asymmetric missing data (one present, one nil), not when both are nil (normal state).

3. **30-minute timeout behavioral change (line 688)**: Expanded the comment to document that this changes prior behavior (which immediately treated pre-run reviews as addressed), and noted the edge case where agent runs exceeding 30 minutes could see the timeout expire before re-review arrives.

4. **`||=` comment clarity (line 119)**: Updated comment to explicitly say the deduplication is against the later `pr_data` fetch at the CI-check stage, rather than the vague "ensures at most one."

5. **Line 924 and line 467 threads**: These were acknowledgments/approvals from the reviewer — no code changes needed.

viamin (commented):
Generated no new comments. The PR correctly addresses #934 by requiring clean re-review signals on the current HEAD commit before clearing body-only bot feedback. The commit-level verification logic handles nil values gracefully, the 30-minute timeout fallback is a sensible safety net, and the four new test cases provide solid coverage of the key scenarios (stale vs HEAD clean reviews, non-clean HEAD blocking clean comment override, and timeout behavior). Well-scoped and ready to merge.

viamin:
## Agent Update

Merge conflicts resolved and committed successfully. Here's what was done:

**Conflict resolution:**
1. **`app/temporal/activities/scan_paid_prs_activity.rb` (line 229)** — Combined both keyword args at the call site: `no_outstanding_review_feedback?(... head_sha: pr_data.head&.sha, checks: checks)`
2. **`app/temporal/activities/scan_paid_prs_activity.rb` (line 984)** — Combined both keyword args in the method signature: `def no_outstanding_review_feedback?(... head_sha: nil, checks: nil)`, keeping both the `head_sha` pass-through to `check_review_bot_status` and the `checks`/`non_bot_review_gate_triggers` logic from main
3. **`spec/temporal/activities/scan_paid_prs_activity_spec.rb`** — Kept all tests from both branches: our 3 HEAD-commit verification tests + main's 4 file-diff heuristic tests

**Verification:**
- Rubocop: 0 offenses
- All 116 scan_paid_prs_activity specs pass
- All 4943 non-request specs pass (0 failures)
- Pre-commit hooks passed

viamin (commented):
Generated no new comments. GitHub currently reports this PR with no changed files, and the branch code I inspected already includes the HEAD-commit review-gating updates described in #934.

PR #961, opened 2026-04-08
Fix #957: Replace PR dropdown with multi-select table showing review rounds on trigger run form

### LINKED ISSUE #957: Replace PR dropdown with multi-select table showing review rounds on trigger run form
## Summary

On the trigger run form, when choosing **PR code review**, the current dropdown list of PRs should be replaced with a more informative table that supports multi-selection.

### PR Table
- Replace the existing PR dropdown with a table view
- Table columns:
  - **PR number** (and optionally title)
  - **Review rounds** — the number of review rounds that Paid knows about for that PR

### Disabled/Grayed-out State
- PRs that have a **queued**, **pending**, or **running** agent run should be **grayed out** in the table and not selectable

### Multi-select
- Multiple PRs should be selectable at once
- Each selected PR will have **one agent run queued** for PR review

## Motivation

The current dropdown does not surface enough context to make an informed selection — users have to look up review round counts separately and cannot tell which PRs already have runs in progress. A table with review round counts and run-status indicators makes the selection faster and less error-prone, and multi-select eliminates the need to trigger runs one at a time.

viamin:
## Summary

Replace PR dropdown with multi-select table showing review rounds on trigger run form

See #957 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #957

### DIFF
--- a/app/controllers/projects/agent_runs_controller.rb
+++ b/app/controllers/projects/agent_runs_controller.rb
@@ -46,6 +46,19 @@ module Projects
         .pull_requests_only
         .where(github_state: "open")
         .order(github_number: :desc)
+
+      pr_numbers = @pull_requests.map(&:github_number)
+      @prs_with_active_runs = if pr_numbers.any?
+        @project.agent_runs
+          .where(source_pull_request_number: pr_numbers, status: AgentRun::UNFINISHED_STATUSES)
+          .distinct
+          .pluck(:source_pull_request_number)
+          .to_set
+      else
+        Set.new
+      end
+
+      @pr_priorities = build_pr_priorities(@pull_requests)
     end
 
     def create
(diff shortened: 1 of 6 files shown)


viamin:
## Agent Update

The background task from the earlier commit attempt also completed, but the commit had already succeeded on the subsequent try. Everything is done — working tree is clean with the fix committed.

viamin:
Closing in favor of #968, which covers strictly more of #957: it updates **both** view files (`app/views/projects/agent_runs/new.html.erb` AND `app/views/agent_runs/new.html.erb`), while this PR only touched the project-scoped view — anyone hitting the top-level route would still see the old single-select dropdown for the review goal and POST a now-rejected `pull_request_id` param. #968 also has a more granular `RecordNotUnique` rescue that distinguishes the proxy_token case.

The polish bits this PR had over #968 — colored `priority_badge_class` helper with P1/P2/P3 badges, the `if pr_numbers.any?` guard, per-PR `RecordNotUnique` rescue with `next` (partial-success with a `created` counter), the multi-PR creation spec, and the "Select PRs to Review" heading copy — have all been ported onto #968 in 1840b83c. The spurious `yarn.lock` churn was dropped.

PR #963, opened 2026-04-08
Fix #958: Add priority select input to trigger agent run form

### LINKED ISSUE #958: Add priority select input to trigger agent run form
## Summary

Allow users to optionally choose a priority when creating an issue via the trigger agent run form.

## Details

- Add a select input below the prompt text area in the trigger agent run form for choosing a priority level.
- The priority select should include standard priority options (e.g., Urgent, High, Medium, Low, None/No priority).
- By default, no priority value should be selected — created issues will have no priority unless the user explicitly chooses one.
- The selected priority value should be passed through and applied to the created issue.

## Acceptance Criteria

- [ ] A priority select input is rendered below the prompt text area in the trigger agent run form
- [ ] The select defaults to no priority (empty/none)
- [ ] When a priority is selected, the created issue includes the chosen priority value
- [ ] When no priority is selected, the created issue has no priority value
- [ ] The UI is consistent with the existing form styling

viamin:
## Summary

Add priority select input to trigger agent run form

See #958 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #958


app/controllers/projects/agent_runs_controller.rb
```diff
@@ -441,6 +445,18 @@ def resolve_issue
...
+    def resolve_priority_tier
+      tier = params[:priority_tier].presence
+      tier if tier.present? && Project::PRIORITY_TIERS.include?(tier)
+    end
+
+    def apply_priority_label(issue, tier)
+      label_name = @project.priority_label_for(tier)
+      return if label_name.blank? || issue.labels.include?(label_name)
+
+      issue.update!(labels: issue.labels + [ label_name ])
```
viamin: **Bug: conflicting priority labels not removed.** When applying a new priority label, existing priority labels from other tiers are not cleared. If an issue already has a "P2" label and the user selects "P1", both labels will be present. This could be misleading even though `compute_label_priority_tier` picks the highest.

Consider stripping existing priority labels before adding the new one:

```suggestion
      issue.update!(labels: (issue.labels - @project.priority_label_names) + [ label_name ])
```

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -441,6 +445,18 @@ def resolve_issue
       @project.issues.find(params[:issue_id])
     end
 
+    def resolve_priority_tier
+      tier = params[:priority_tier].presence
+      tier if tier.present? && Project::PRIORITY_TIERS.include?(tier)
```
viamin: Nit: `tier.present?` is redundant — `params[:priority_tier].presence` already returns `nil` for blank values, and `nil` won't pass `PRIORITY_TIERS.include?`.

```suggestion
      tier if Project::PRIORITY_TIERS.include?(tier)
```

app/views/projects/agent_runs/new.html.erb
```diff
@@ -93,6 +93,20 @@
           </div>
         </div>
 
+        <div>
```
viamin: **UI consideration:** The priority select renders for all goals including "review". When the goal is `review`, there's typically no issue to apply the label to (the controller guards with `if issue && priority_tier`), so the selection would silently have no effect. Consider hiding this section when `goal == 'review'` to avoid user confusion.

db/migrate/20260408170459_add_priority_tier_to_agent_runs.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddPriorityTierToAgentRuns < ActiveRecord::Migration[8.1]
+  def change
+    add_column :agent_runs, :priority_tier, :string, limit: 10
```
viamin, line 5: Migration looks good. `limit: 10` is a sensible constraint for P1/P2/P3 values.

Since this is a nullable `string` column, consider also adding a model-level validation to prevent invalid values from being persisted through non-controller code paths:
```ruby
validates :priority_tier, inclusion: { in: Project::PRIORITY_TIERS, allow_nil: true }
```

viamin (commented):
Overall this is a clean, well-scoped PR that addresses #958. The priority select flows correctly from the form through the controller to the model and into the Temporal activity. A few items worth addressing:

1. **Conflicting priority labels not removed** — `apply_priority_label` appends the new label but doesn't strip existing priority labels from other tiers, which could leave an issue with both "P1" and "P2" labels.
2. **No model-level validation** — `priority_tier` is a free-form string column with no `validates :inclusion` on the model. The controller validates, but other write paths (Temporal activities, future APIs) could store invalid values. Consider adding `validates :priority_tier, inclusion: { in: Project::PRIORITY_TIERS, allow_nil: true }` to `AgentRun`.
3. **UI nit** — The priority select renders for all goals including "review", where there may be no issue to apply it to. Not a bug (the controller guards it), but potentially confusing for users.

viamin:
## Agent Update

Commit succeeded. Here's a summary of all changes made to address the review comments:

1. **Redundant `tier.present?` check** (`agent_runs_controller.rb:450`): Removed the redundant guard since `params[:priority_tier].presence` already returns `nil` for blank values, and `nil` won't pass `PRIORITY_TIERS.include?`.

2. **Conflicting priority labels not removed** (`agent_runs_controller.rb:457`): Changed `apply_priority_label` to strip all existing priority labels before adding the new one: `(issue.labels - @project.priority_label_names) + [label_name]`.

3. **Priority select visible for "review" goal** (`new.html.erb:96`): Added `data-goal-toggle-target="prioritySection"` to the priority div and added corresponding toggle logic in the Stimulus controller to hide/disable it when the goal is "review".

4. **Model-level validation** (`agent_run.rb`): Added `validates :priority_tier, inclusion: { in: Project::PRIORITY_TIERS }, allow_nil: true` to prevent invalid values from being persisted through non-controller code paths.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -70,12 +71,15 @@ def create
         end
       end
 
+      apply_priority_label(issue, priority_tier) if issue && priority_tier
```
viamin: **Potential issue: side effect before run creation.** `apply_priority_label` modifies the issue's labels via `update!` before `create_run_and_redirect` is called. If the budget check fails or `AgentRun.create!` raises, the issue labels have already been changed with no rollback.

Consider wrapping the create action's label update + run creation in a transaction, or moving the label application into `create_run_and_redirect` so it only fires on success:

```suggestion
      create_run_and_redirect(
```

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -441,6 +445,18 @@ def resolve_issue
...
+    def resolve_priority_tier
+      tier = params[:priority_tier].presence
+      tier if Project::PRIORITY_TIERS.include?(tier)
+    end
+
+    def apply_priority_label(issue, tier)
+      label_name = @project.priority_label_for(tier)
+      return if label_name.blank? || issue.labels.include?(label_name)
+
+      issue.update!(labels: (issue.labels - @project.priority_label_names) + [ label_name ])
```
viamin: **Fragile label replacement if `labels` is nil.** If `issue.labels` is `nil` (rather than `[]`), the subtraction `issue.labels - @project.priority_label_names` will raise `NoMethodError`. Worth confirming `issue.labels` always returns an array, or adding a guard:

```suggestion
      issue.update!(labels: (Array(issue.labels) - @project.priority_label_names) + [ label_name ])
```

viamin (commented):
## PR Review: Add priority select input to trigger agent run form

Overall this is a clean implementation that addresses #958. The controller validation, model validation, Stimulus toggling, and activity integration are all well-structured. A few items to consider below.

**Scope concern:** The `db/schema.rb` diff includes ~475 lines of bracket formatting changes (`[ "foo" ]` → `["foo"]`) unrelated to the feature. This is likely from a different Rails/Ruby version regenerating the schema dump. Consider regenerating schema.rb from the target Rails version on `main` to keep the diff focused on the actual column addition.

**Missing tests:** No specs were added for the new behavior — `resolve_priority_tier` validation, `apply_priority_label` label manipulation, the activity's `priority_label_for` helper, or the form rendering. Given the label array manipulation logic, at least a model/controller spec would be valuable.

app/temporal/activities/create_github_issue_activity.rb
```diff
@@ -67,6 +69,12 @@ def issue_body(summary)
       end
     end
 
+    def priority_label_for(agent_run)
```
viamin, line 72: Minor: this `priority_label_for` method duplicates the same logic as `apply_priority_label` in the controller (both call `project.priority_label_for(tier)`). Not a blocker, but if more callers appear, extracting a shared helper on the model might reduce duplication.

app/views/projects/agent_runs/new.html.erb
```diff
@@ -93,6 +93,20 @@
...
 
+        <div data-goal-toggle-target="prioritySection">
+          <h2 class="text-base font-semibold text-gray-900">Priority</h2>
+          <p class="mt-1 text-sm text-gray-500">Optionally assign a priority level to the issue.</p>
+
+          <div class="mt-4">
+            <select name="priority_tier" id="priority_tier" class="block w-full rounded-md border-gray-300 shadow-sm focus:border-indigo-500 focus:ring-indigo-500 sm:text-sm">
+              <option value="">-- No priority --</option>
+              <% Project::PRIORITY_TIERS.each do |tier| %>
+                <option value="<%= tier %>"><%= @project.priority_label_for(tier) %></option>
```
viamin: Nit: if `priority_label_for(tier)` returns `nil` for a tier (e.g., no label configured), the `<option>` text will be blank. Consider falling back to the tier name:

```suggestion
                <option value="<%= tier %>"><%= @project.priority_label_for(tier) || tier %></option>
```

app/views/projects/agent_runs/new.html.erb
```diff
@@ -93,6 +93,20 @@
           </div>
         </div>
 
+        <div data-goal-toggle-target="prioritySection">
+          <h2 class="text-base font-semibold text-gray-900">Priority</h2>
+          <p class="mt-1 text-sm text-gray-500">Optionally assign a priority level to the issue.</p>
```
viamin: The description says "assign a priority level to the issue" but this section is also visible when no issue is selected (e.g., custom prompt only, or the "create issue" goal). The priority_tier is stored on the agent_run and used by `CreateGithubIssueActivity`, so this works for auto-created issues. Just flagging in case the copy should clarify that priority applies to the resulting issue, not just pre-existing ones.

viamin:
## Agent Update

Commit succeeded. Here's a summary of the changes made to address the review comments:

1. **Side effect before run creation** (Thread 1): Moved `apply_priority_label` from the `create` action into `create_run_and_redirect`, so it only fires after the budget check passes and `create_agent_run` succeeds. If either fails, issue labels remain unchanged.

2. **Fragile nil labels** (Thread 2): Wrapped both `issue.labels` references in `apply_priority_label` with `Array()` to guard against nil values.

3. **Duplicated priority_label_for** (Thread 3): Left as-is — both callers are thin wrappers around `Project#priority_label_for` with different signatures (activity takes an `agent_run`, controller takes `issue` + `tier`). The reviewer noted this is "not a blocker."

4. **Blank option text** (Thread 4): Added `|| tier` fallback so the option displays the tier name (e.g., "P1") if `priority_label_for` returns nil.

5. **Misleading description copy** (Thread 5): Updated to "Optionally assign a priority level to the resulting issue (existing or auto-created)" to clarify it applies to both pre-existing and auto-created issues.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -441,6 +443,18 @@ def resolve_issue
...
+    def resolve_priority_tier
+      tier = params[:priority_tier].presence
+      tier if Project::PRIORITY_TIERS.include?(tier)
+    end
+
+    def apply_priority_label(issue, tier)
+      label_name = @project.priority_label_for(tier)
+      return if label_name.blank? || Array(issue.labels).include?(label_name)
+
+      issue.update!(labels: (Array(issue.labels) - @project.priority_label_names) + [ label_name ])
```
viamin: **Bug: Label change not synced to GitHub**

This updates the local `Issue` record's `labels` column, but doesn't push the label to the actual GitHub issue. For pre-existing issues, the GitHub issue will be out of sync — the priority label will exist locally but not on GitHub.

The `create_github_issue_activity` handles this correctly for *new* issues (adding the label at creation time), but for existing issues selected from the dropdown, you'd need a GitHub API call here (or a background job) to add/replace the label on the remote issue.

Is this intentional, with GitHub sync planned separately?

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -530,7 +545,11 @@ def create_run_and_redirect(on_error_path:, **attrs)
         return
       end
 
+      issue = attrs[:issue]
+      priority_tier = attrs[:priority_tier]
+
       create_agent_run(**attrs)
+      apply_priority_label(issue, priority_tier) if issue && priority_tier
```
viamin: **Nit: `create_agent_run` could fail, leaving `apply_priority_label` unreached**

If `create_agent_run` raises (e.g. `RecordNotUnique`), that's handled below in the `rescue`. But if it raises something else, the label won't be applied, which is fine — the agent run didn't get created. Just noting this is intentionally correct.

However, the reverse scenario is worth considering: if `apply_priority_label` raises (e.g. `update!` fails due to a validation), you'll have created an agent run but failed to label the issue, and the error won't be caught by the existing rescue clauses. Consider whether `apply_priority_label` should use `update` (soft fail) instead of `update!`, or be wrapped in its own error handling.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -441,6 +443,18 @@ def resolve_issue
       @project.issues.find(params[:issue_id])
     end
 
+    def resolve_priority_tier
+      tier = params[:priority_tier].presence
+      tier if Project::PRIORITY_TIERS.include?(tier)
+    end
+
+    def apply_priority_label(issue, tier)
```
viamin, line 469: **Consider: this method is unused outside `create_run_and_redirect`**

This is fine as a private helper, but it operates on the `Issue` model and uses `@project` — this kind of label-manipulation logic might be a better fit as a method on the `Issue` model or a small service, rather than in the controller. Not blocking, just a thought for maintainability.

db/schema.rb
```diff
@@ -22,10 +22,10 @@
     t.datetime "created_at", null: false
     t.decimal "quality_score", precision: 5, scale: 4
     t.datetime "updated_at", null: false
-    t.index [ "ab_test_id", "agent_run_id" ], name: "index_ab_test_assignments_unique", unique: true
-    t.index [ "ab_test_id" ], name: "index_ab_test_assignments_on_ab_test_id"
-    t.index [ "ab_test_variant_id" ], name: "index_ab_test_assignments_on_ab_test_variant_id"
-    t.index [ "agent_run_id" ], name: "index_ab_test_assignments_on_agent_run_id"
+    t.index ["ab_test_id", "agent_run_id"], name: "index_ab_test_assignments_unique", unique: true
```
viamin, line 25: **Noise: large whitespace reformatting**

The schema.rb diff is ~475 lines but only 1 line is the actual column addition (`t.string "priority_tier", limit: 10`). The rest is Rails reformatting the array syntax in index definitions (removing spaces inside brackets). This is likely caused by a different Rails/Ruby version than what generated the previous schema dump.

This creates unnecessary merge conflict surface. Consider regenerating `schema.rb` from the base branch to isolate the real change, or at minimum noting this in the PR description so reviewers know to ignore the formatting noise.

app/temporal/activities/create_github_issue_activity.rb
```diff
@@ -16,6 +16,8 @@ def execute(input)
         body = issue_body(summary)
 
         issue_labels = project.auto_add_labels_enabled? ? [ project.generated_label_name ] : []
+        priority_label = priority_label_for(agent_run)
```
viamin, line 19: Good — this correctly adds the priority label to the GitHub issue at creation time via the API, complementing the `priority_tier` stored on the `AgentRun` record.

viamin (commented):
## PR Review: Add priority select input to trigger agent run form

The implementation cleanly adds priority tier support to the manual trigger form, wiring it through the controller to both the `AgentRun` record and the issue's labels. The code follows existing patterns well (reusing `Project::PRIORITY_TIERS`, `priority_label_for`, etc.).

However, there are a few issues worth addressing:

### Issues

1. **No tests** — No specs were added or updated for the new controller logic (`resolve_priority_tier`, `apply_priority_label`), the model validation, the activity change, or the view. Given the PR touches 5 non-schema files across multiple layers, test coverage is important.

2. **`apply_priority_label` only updates the local DB record** — When a user selects a priority for an *existing* issue, the label is written to the local `issues.labels` column but is never pushed to GitHub. The `create_github_issue_activity` correctly adds the label at GitHub-issue creation time, but for pre-existing issues the GitHub side is out of sync.

3. **Large schema.rb formatting noise** — The diff shows ~475 lines changed in `schema.rb` but only 1 line is the actual `priority_tier` column addition. The rest is whitespace reformatting (removing spaces inside array brackets in index definitions). This makes the PR harder to review and will cause unnecessary merge conflicts.

See inline comments for details.

app/models/agent_run.rb
```diff
@@ -72,6 +72,7 @@ class AgentRun < ApplicationRecord
   validates :stale_requeue_count, numericality: { only_integer: true, greater_than_or_equal_to: 0 }
   validates :token_limit_status, inclusion: { in: TOKEN_LIMIT_STATUSES }, allow_nil: true
   validates :guardrail_violation_type, inclusion: { in: GUARDRAIL_VIOLATION_TYPES }, allow_nil: true
+  validates :priority_tier, inclusion: { in: Project::PRIORITY_TIERS }, allow_nil: true
```
viamin, line 83: Good — validation reuses the existing `Project::PRIORITY_TIERS` constant and allows `nil`, which matches the "no priority" default behavior described in the issue.

viamin:
## Agent Update

The background setup agent finished — the database issue was already resolved during my main work (I used `DATABASE_URL` to connect). The commit is done and all checks passed.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -441,6 +443,56 @@ def resolve_issue
...
+      label_name = @project.priority_label_for(tier)
+      return if label_name.blank?
+
+      current_labels = Array(issue.labels)
+      return if current_labels.include?(label_name)
+
+      stale_priority_labels = current_labels & @project.priority_label_names
+      new_labels = (current_labels - stale_priority_labels) + [ label_name ]
+      issue.update(labels: new_labels)
```
viamin: `issue.update` (without bang) silently returns `false` on validation failure, but the method still proceeds to `sync_priority_label_to_github`. This can leave GitHub with a priority label that the local `Issue` record doesn't have.

Either bail out on failure or use `update!` (which the surrounding `rescue` already handles):

```suggestion
      issue.update!(labels: new_labels)
```

viamin (commented):
The PR correctly adds a priority select input to the trigger agent run form, addressing #958. One minor correctness issue with label sync consistency.

viamin:
## Agent Update

Merged `origin/main`, resolved the PR conflicts, and committed the result as `11569c50` with `fix(agent-runs): merge main into priority select PR`.

The unresolved review thread in [app/controllers/projects/agent_runs_controller.rb](/workspace/app/controllers/projects/agent_runs_controller.rb) is addressed by switching the local label update to `issue.update!`, so GitHub label sync only runs if the `Issue` record actually saves. I also kept the prior PR work intact through the merge, including the review-goal UI behavior and priority-tier handling, and added request coverage for both the successful sync path and the validation-failure path in [spec/requests/agent_runs_spec.rb](/workspace/spec/requests/agent_runs_spec.rb).

Verification completed: setup/deps installed, database prepared, `bundle exec rubocop` passed, and `bundle exec rspec` passed with `5750 examples, 0 failures, 2 pending`. The 2 pending specs are the existing Qdrant live specs that skip because Qdrant is not running at `http://localhost:6333`. The worktree is clean.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -459,6 +461,56 @@ def resolve_issue
...
+
+      current_labels = Array(issue.labels)
+      return if current_labels.include?(label_name)
+
+      stale_priority_labels = current_labels & @project.priority_label_names
+      new_labels = (current_labels - stale_priority_labels) + [ label_name ]
+      issue.update!(labels: new_labels)
+
+      sync_priority_label_to_github(issue, label_name, stale_priority_labels)
+    rescue => e
```
viamin: `issue.update!` failures are actionable here, not something we should just log and continue past. In this flow the run has already been created, so if the local `issues` row rejects the label change, the request still succeeds with `priority_tier` persisted on the run while the associated issue stays unlabeled. That breaks the requested behavior and also means later queueing / priority-label inheritance see a different priority than the user selected. Please let the local DB update failure abort the request (ideally in the same transaction as `create_agent_run`) and only rescue the GitHub sync step.

viamin (commented):
One correctness issue found: the existing-issue path creates the agent run before applying the requested priority label, and `apply_priority_label` swallows local persistence failures. That means the request can succeed while leaving the issue unlabeled, so queueing and later PR label inheritance no longer match the user's selected priority.

viamin:
## Agent Update

**Merge conflicts**: Merged `origin/main` cleanly — no conflicts this time.

**Code review fix** (the unresolved thread): The reviewer pointed out that `issue.update!` failures in `apply_priority_label` were silently rescued, allowing the agent run to persist while the issue stayed unlabeled. Fixed by:

1. **Wrapping in a transaction** — `create_agent_run` and `apply_priority_label` now run inside `ActiveRecord::Base.transaction`, so if the label update fails, the agent run creation is rolled back too.
2. **Removing the blanket rescue** — The `rescue => e` that swallowed all errors (including `issue.update!` failures) was removed. Only `sync_priority_label_to_github` retains its GitHub-specific error handling.
3. **Added `RecordInvalid` rescue** — `create_run_and_redirect` now rescues `ActiveRecord::RecordInvalid` and redirects with the error message, giving users a clean experience.
4. **Updated test** — The test "does not sync GitHub labels when the local issue update fails" now correctly asserts that the agent run is **not** created (rolled back) and the user is redirected.

All 5754 specs pass with 0 failures.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -548,7 +594,13 @@ def create_run_and_redirect(on_error_path:, **attrs)
...
       end
 
-      create_agent_run(**attrs)
+      issue = attrs[:issue]
+      priority_tier = attrs[:priority_tier]
+
+      ActiveRecord::Base.transaction do
+        create_agent_run(**attrs)
+        apply_priority_label(issue, priority_tier) if issue && priority_tier
+      end
```
viamin, line 604: `sync_priority_label_to_github` (called via `apply_priority_label`) makes HTTP requests to the GitHub API while holding an open database transaction. This has two problems:

1. **Connection pool exhaustion** — the DB connection is held for the full duration of the GitHub API round-trips (remove old labels + add new label). Under load this can starve the connection pool.
2. **Inconsistency on rollback** — if the transaction rolls back *after* the GitHub API calls succeed (unlikely today since `apply_priority_label` is the last call, but fragile), GitHub labels will be updated but the local DB will not.

Consider splitting `apply_priority_label` so only the local `issue.update!` runs inside the transaction, and the GitHub sync runs after the transaction commits:

```suggestion
      ActiveRecord::Base.transaction do
        create_agent_run(**attrs)
        apply_priority_label(issue, priority_tier) if issue && priority_tier
      end
      sync_priority_label_to_github_after_commit(issue, priority_tier) if issue && priority_tier
```

Or use an `after_commit` callback / `ActiveRecord::Base.after_all_transactions_commit` to defer the sync.

viamin (commented):
The PR correctly addresses #958 — adds a priority select to the trigger form, persists the tier on the agent run, syncs labels on the issue, and includes good test coverage. One actionable issue: external HTTP calls inside a DB transaction.

viamin:
## Agent Update

1. **Merge conflicts**: Merged `origin/main` cleanly — no conflicts.

2. **Code review fix** (unresolved thread about GitHub API calls inside transaction): Split `apply_priority_label` so that:
   - The DB-only update (`issue.update!`) stays inside the `ActiveRecord::Base.transaction` block
   - `apply_priority_label` now returns `[label_name, stale_priority_labels]` instead of calling `sync_priority_label_to_github` directly
   - `sync_priority_label_to_github` runs **after** the transaction commits in `create_run_and_redirect`
   
   This fixes both problems the reviewer identified: no DB connection is held during GitHub API round-trips, and if the transaction rolls back, no GitHub labels will have been modified.

3. **Verification**: Rubocop passes (716 files, 0 offenses), full test suite passes (5754 examples, 0 failures, 2 pending — the pre-existing Qdrant specs).

viamin (commented):
Generated no new comments. The PR cleanly adds priority tier selection to the trigger agent run form. Transaction boundaries are correct (DB changes inside, GitHub sync outside), input validation is proper, JS toggle logic matches existing patterns, and test coverage handles both success and rollback paths. The schema.rb diff includes unrelated table additions from other migrations — that is expected regeneration behavior.

PR #965, opened 2026-04-08
fix(workflow): make CreatePullRequestActivity idempotent (#964)

### LINKED ISSUE #964: fix(workflow): CreatePullRequestActivity is not idempotent — retries leave orphan branches
## Symptom

Agent runs push branches to GitHub but no PR gets created and the run ends up marked **failed** (not completed). The branch is left orphaned on the remote with no associated PR and no link back to the originating issue.

## Root cause

`Activities::CreatePullRequestActivity#execute` (`app/temporal/activities/create_pull_request_activity.rb:14-28`) is **not idempotent**:

1. `PushBranchActivity` succeeds — branch is on the remote.
2. `CreatePullRequestActivity` calls `client.create_pull_request(...)` and GitHub creates the PR.
3. Anything between the GitHub API response and `agent_run.complete!` (DB hiccup, container/process death, Temporal worker restart, transient exception in `pr_body` generation, etc.) prevents the activity from returning.
4. Temporal retries the activity (no `retry_policy` override at `agent_execution_workflow.rb:250`, so it uses the default retry policy).
5. The retry calls `create_pull_request` again — GitHub responds **422 Unprocessable Entity** ("A pull request already exists for `<head>`"), which is not retryable.
6. The workflow `rescue` at `agent_execution_workflow.rb:286` runs `MarkAgentRunFailedActivity`. The run is now in `failed` state, but the branch — and possibly even a PR — exists on GitHub with no link to the AgentRun.

A second related issue: even on the first attempt, if `add_pr_labels`, `agent_run.log!`, or anything between lines 30-38 raises, `track_phase` re-raises and the activity is retried with the same non-idempotent failure mode (the run *is* persisted as completed at line 24, but the workflow itself still fails and the rescue path overwrites status).

## Fix

Make `CreatePullRequestActivity` idempotent:

- Before calling `create_pull_request`, look up an existing PR for `head: agent_run.branch_name` via `client.pull_requests(repo, head: \"owner:branch\", state: 'open')`. If one exists for this branch, reuse it instead of creating a new one.
- Persist `agent_run.complete!` *with the PR url/number* as the very next step, before any best-effort post-processing (labels, log, decision record).
- Move `add_pr_labels` and any non-essential follow-ups out of the activity, or wrap them so a failure cannot cause the activity to be retried after `complete!` has run.

## Recovery

Separately, we need a one-time recovery pass for branches already orphaned by this bug: walk failed AgentRuns that have a `branch_name` but no `pr_url`, verify the branch still exists on GitHub, and either open the missing PR or re-link an existing one. Tracked in a follow-up comment with the orphan list.

## Acceptance criteria

- [ ] Re-running `CreatePullRequestActivity` against a branch that already has an open PR reuses the existing PR (no 422).
- [ ] Spec covers: (a) PR creation succeeds on first try, (b) PR creation succeeds but post-processing raises — run is still marked completed and not retried into a 422, (c) retry after a partial failure where the PR was already created reuses it.
- [ ] No orphan branches produced under simulated transient failures in spec.

viamin:
## Summary
- Look up an existing open PR for the head branch before calling `create_pull_request`, so retries after a partial failure reuse the existing PR instead of hitting GitHub's "A pull request already exists" 422.
- Adds regression specs for the reuse path and for a lookup-failure fallthrough.

Fixes #964.

## Why
When `Activities::CreatePullRequestActivity` crashes between GitHub creating the PR and `agent_run.complete!` persisting (DB blip, worker restart, transient error in post-processing), Temporal retries the activity. The retry currently re-calls `create_pull_request`, GitHub returns 422, the workflow `rescue` marks the run **failed**, and the branch is left orphaned on the remote with no link back to the AgentRun. See #964 for the full root-cause writeup.

## Test plan
- [x] `bin/rspec spec/temporal/activities/create_pull_request_activity_spec.rb` (22 examples, 0 failures)
- [ ] CI green

### DIFF
--- a/app/temporal/activities/create_pull_request_activity.rb
+++ b/app/temporal/activities/create_pull_request_activity.rb
@@ -12,7 +12,8 @@ module Activities
         issue = agent_run.issue
 
         client = project.github_token.client
-        pr = client.create_pull_request(
+        pr = find_existing_pull_request(client, project, agent_run.branch_name, agent_run_id: agent_run_id)
+        pr ||= client.create_pull_request(
           project.full_name,
           base: project.default_branch,
           head: agent_run.branch_name,
@@ -42,6 +43,33 @@ module Activities
     end
 
     private
+
+    # Look up an open PR already opened for this branch. Makes the activity
+    # idempotent: if a previous attempt created the PR but crashed before
+    # `agent_run.complete!` persisted, the retry reuses the existing PR
+    # instead of 422-ing on "A pull request already exists".
+    def find_existing_pull_request(client, project, branch_name, agent_run_id:)
+      owner = project.full_name.split("/").first
+      existing = client.pull_requests(project.full_name, state: "open", head: "#{owner}:#{branch_name}")
+      pr = Array(existing).first
+      return nil unless pr
+
+      logger.info(
+        message: "agent_execution.pull_request_reused",
+        agent_run_id: agent_run_id,
+        pull_request_url: pr.html_url,
+        pull_request_number: pr.number
+      )
+      pr
+    rescue GithubClient::Error => e
+      logger.warn(
+        message: "agent_execution.pull_request_lookup_failed",
+        agent_run_id: agent_run_id,
+        branch: branch_name,
+        error: e.message
+      )
+      nil
+    end
 
     def pr_title(issue)
       return "Agent changes" unless issue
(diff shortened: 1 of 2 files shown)


viamin:
Closing in favor of #969, which covers strictly more of #964: it has the same lookup-then-create idempotency fix **and** also wraps the post-`complete!` post-processing (`add_pr_labels`, `agent_run.log!`) in a `best_effort` helper, closing a second leak path this PR didn't address — where a raise after `agent_run.complete!` causes the workflow rescue to flip the run from `completed` back to `failed`.

The one thing this PR had that #969 lacked — a rescue around the new `pull_requests()` lookup so a transient list-call failure can't become a brand-new failure mode introduced by the fix itself — has been ported onto #969 in 07c64078, along with its regression spec. Reviewers should look at #969.

PR #968, opened 2026-04-08
Fix #957: Replace PR dropdown with multi-select table showing review rounds on trigger run form

### LINKED ISSUE #957: Replace PR dropdown with multi-select table showing review rounds on trigger run form

viamin:
## Summary

Replace the single-select PR dropdown on the trigger run form with a multi-select table that surfaces review round counts and active run status, so users can make informed selections and queue multiple PR review runs at once without leaving the form.

### Controller (`app/controllers/projects/agent_runs_controller.rb`)
- Compute `@prs_with_active_runs` (queued/pending/running/paused) and `@pr_priority_tiers` in the `new` action to power the table's disabled state and priority column
- Accept `pull_request_ids[]` array param for the `review` goal and create one agent run per selected PR via `create_review_runs_and_redirect`
- Add `compute_pr_priority_tiers` helper that maps PR labels to the project's configured priority tiers

### Views (`projects/agent_runs/new.html.erb`, `agent_runs/new.html.erb`)
- Add a PR table view (shown for "PR Code Review" goal) with columns: checkbox, PR number + title, priority tier, and review rounds count
- Keep the existing dropdown for "Create/Edit PR" goal — toggled via Stimulus targets
- Gray out PRs with active runs (reduced opacity, disabled checkbox, "In Progress" badge)

### Stimulus (`goal_toggle_controller.js`)
- Add `prDropdown` and `prTable` targets to toggle between dropdown and table based on selected goal
- Ensure permanently-disabled checkboxes (active runs) stay disabled across goal toggles

### Tests (`spec/requests/agent_runs_spec.rb`)
- Update review goal tests to send `pull_request_ids: [pr.id]` array params
- Update error message expectation to match new "at least one" wording

## Design Decisions

- **One agent run per selected PR** — multi-select fans out into individual agent runs rather than batching into a single run, keeping the execution model simple and consistent with existing single-PR review behavior.
- **Disabled rather than hidden for active-run PRs** — graying out PRs with in-progress runs (instead of hiding them) gives users visibility into what's already running and avoids confusion about "missing" PRs.
- **Goal-based view toggling** — the dropdown and table coexist in the DOM, toggled by the Stimulus controller, rather than using separate form partials. This keeps the form structure simple and avoids duplicating shared fields.

## Screenshots

> **TODO**: Attach before/after screenshots of the trigger run form showing the old dropdown vs. the new multi-select table (including a grayed-out PR with an active run).

---

Closes #957


viamin (commented):
## PR Review: Replace PR dropdown with multi-select table

Overall this is a well-structured change that cleanly addresses #957. The controller branching, Stimulus integration, and test updates are solid. A few issues worth addressing:

**Key concerns:**
1. **Partial creation without transaction** — `create_review_runs_and_redirect` creates runs in a loop without a transaction, so a failure mid-loop leaves orphaned runs.
2. **Silent no-op on invalid PR IDs** — if all submitted `pr_ids` are invalid/unauthorized, 0 runs are created and the user sees "0 agent runs queued" instead of an error.
3. **Identical view duplication** — the two view files have byte-for-byte identical changes; this should be a shared partial.
4. **Missing multi-PR test coverage** — tests only exercise single-PR selection, not the multi-select path.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -717,6 +734,51 @@ def termination_guardrail_type(agent_run)
...
+      end
+
+      prs = @project.issues.pull_requests_only.where(id: pr_ids)
+      prs.each do |pr|
+        create_agent_run(
+          custom_prompt: custom_prompt,
+          source_pull_request_number: pr.github_number,
+          goal: goal
+        )
+      end
```
viamin, line 772: **Bug: partial creation on failure.** The `prs.each` loop creates agent runs without a transaction. If `create_agent_run` raises on the 3rd PR (e.g., `RecordNotUnique`), the first two runs are already persisted and `ProcessRunQueueJob` is never enqueued for them.

Wrap the loop in a transaction so it's all-or-nothing:

```suggestion
      ActiveRecord::Base.transaction do
        prs.each do |pr|
          create_agent_run(
            custom_prompt: custom_prompt,
            source_pull_request_number: pr.github_number,
            goal: goal
          )
        end
      end
```

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -717,6 +734,51 @@ def termination_guardrail_type(agent_run)
...
+      ProcessRunQueueJob.perform_later
+
+      notice = if prs.size == 1
```
viamin, line 775: **Bug: silent no-op when all PR IDs are invalid.** If a user submits IDs that don't belong to this project (or are otherwise filtered out by `pull_requests_only`), `prs` is empty and this redirects with `"0 agent runs queued for PR review."` — which is misleading.

Add a guard after the query:

```ruby
if prs.empty?
  redirect_to on_error_path, alert: "None of the selected pull requests could be found."
  return
end
```

app/views/agent_runs/new.html.erb
```diff
@@ -64,16 +64,64 @@
...
+                            #<%= pr.github_number %> - <%= truncate(pr.title, length: 50) %>
+                            <% if has_active_run %>
+                              <span class="ml-1 inline-flex items-center rounded-full bg-yellow-100 px-2 py-0.5 text-xs font-medium text-yellow-800">In Progress</span>
+                            <% end %>
+                          </td>
+                          <td class="px-3 py-2 text-sm text-gray-500">
+                            <%= priority_tier || "-" %>
+                          </td>
+                          <td class="px-3 py-2 text-right text-sm text-gray-500">
+                            <%= pr.draft_review_count + pr.pr_followup_count %>
```
viamin: **Duplication:** This view and `projects/agent_runs/new.html.erb` have byte-for-byte identical changes (the entire PR table block). Consider extracting the table into a shared partial (e.g., `shared/_pr_review_table.html.erb`) to avoid maintaining two copies.

app/views/agent_runs/new.html.erb
```diff
@@ -64,16 +64,64 @@
...
+                    <tbody class="divide-y divide-gray-200 bg-white">
+                      <% @pull_requests.each do |pr| %>
+                        <% has_active_run = @prs_with_active_runs.include?(pr.github_number) %>
+                        <% priority_tier = @pr_priority_tiers[pr.id] %>
+                        <tr class="<%= 'opacity-50' if has_active_run %>">
+                          <td class="px-3 py-2 text-center">
+                            <input type="checkbox"
+                                   name="pull_request_ids[]"
+                                   value="<%= pr.id %>"
+                                   <%= "disabled" if has_active_run %>
```
viamin: **Nit: HTML attribute escaping.** `<%= 'data-permanently-disabled="true"' if has_active_run %>` passes through ERB's auto-escaping, which converts the inner `"` to `&quot;`. The attribute still works (the dataset value ends up as `"true"` with literal quotes), but it produces malformed HTML.

Cleaner alternative — use a valueless boolean-style data attribute:
```suggestion
                                   <%= 'data-permanently-disabled' if has_active_run %>
```
Then in JS, check with `control.hasAttribute('data-permanently-disabled')` or rely on `dataset.permanentlyDisabled` being the empty string (still truthy via `!== undefined`).

Actually, `dataset.permanentlyDisabled` returns `undefined` when absent and `""` when present as a boolean attribute, so `!control.dataset.permanentlyDisabled` would incorrectly be `true` for `""`. Safest to check:
```js
if (!control.hasAttribute('data-permanently-disabled')) {
```

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -46,6 +46,15 @@ def new
         .pull_requests_only
         .where(github_state: "open")
         .order(github_number: :desc)
+
+      pr_numbers = @pull_requests.map(&:github_number)
```
viamin, line 51: **Minor performance note:** `@pull_requests.map(&:github_number)` materializes all PR records to extract numbers. Since these records are needed in the view anyway this is fine, but worth noting that `@pull_requests` is now loaded eagerly here rather than remaining a lazy relation.

spec/requests/agent_runs_spec.rb
```diff
@@ -628,7 +628,7 @@
 
         it "creates a review agent run with source_pull_request_number" do
           expect {
-            post project_agent_runs_path(project), params: { goal: "review", pull_request_id: pr.id }
+            post project_agent_runs_path(project), params: { goal: "review", pull_request_ids: [ pr.id ] }
```
viamin, line 631: **Missing test coverage for multi-select.** All review tests still send a single PR ID. The core new behavior — selecting multiple PRs and fanning out into multiple agent runs — has no test. Consider adding a test that sends `pull_request_ids: [pr1.id, pr2.id]` and asserts `AgentRun.count` increases by 2 with distinct `source_pull_request_number` values.

app/javascript/controllers/goal_toggle_controller.js
```diff
@@ -32,7 +40,25 @@ export default class extends Controller {
...
 
-    if (goal === "review") {
+    // Toggle between dropdown (create_pr) and table (review)
+    this.prDropdownTargets.forEach((el) => {
+      el.hidden = isReview
+      el.querySelectorAll("select").forEach((control) => {
+        control.disabled = isReview || !showPr
+      })
+    })
+
```
viamin, line 54: **Nit:** `!control.dataset.permanentlyDisabled` — if the attribute is present as a valueless boolean attribute (see view comment), `dataset.permanentlyDisabled` is `""` (empty string), which is falsy. This means the check `!control.dataset.permanentlyDisabled` would be `true`, and the checkbox would get re-enabled despite being "permanently disabled".

Safer to check for attribute presence:
```suggestion
        if (!control.hasAttribute('data-permanently-disabled')) {
```

viamin:
## Agent Update

The commit completed successfully — all 5628 tests passed with 0 failures.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -54,29 +63,37 @@ def create
       goal = params[:goal].presence || "create_pr"
       custom_prompt = params[:custom_prompt]&.strip.presence
       issue = resolve_issue
-      source_pr_number = resolve_pull_request
 
       if goal == "review"
-        unless source_pr_number
+        pr_ids = Array(params[:pull_request_ids]).reject(&:blank?)
```
viamin: **Potential issue: `pr_ids` are not type-cast before the `where(id:)` query.**

The values from `params[:pull_request_ids]` are strings. While ActiveRecord will cast them for the `where(id: pr_ids)` query on line 744, the `reject(&:blank?)` check is the only validation. A user could submit arbitrary strings (e.g., IDs from another project). The `@project.issues.pull_requests_only.where(id: pr_ids)` scoping on line 744 correctly restricts to the current project's PRs, so there's no authorization bypass — but consider casting to integers early to fail fast on garbage input:

```suggestion
        pr_ids = Array(params[:pull_request_ids]).filter_map { |id| Integer(id, exception: false) }
```

This also protects against edge cases where a non-numeric string like `"abc"` would silently become `0` in some database adapters.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -717,6 +734,58 @@ def termination_guardrail_type(agent_run)
...
+        return
+      end
+
+      prs = @project.issues.pull_requests_only.where(id: pr_ids)
+      if prs.empty?
+        redirect_to on_error_path, alert: "None of the selected pull requests could be found."
+        return
+      end
+
+      ActiveRecord::Base.transaction do
```
viamin, line 764: **Consider: partial failure atomicity.**

The transaction + `RecordNotUnique` rescue is solid for the unique constraint case. However, if `create_agent_run` raises `NoRunnableProviderError` mid-batch (unlikely since provider resolution shouldn't vary per PR, but worth noting), the transaction will also roll back all previously created runs in the batch, which is the right behavior.

One edge case: if the first N PRs succeed but PR N+1 hits a unique constraint, the rescue catches it *outside* the transaction block, so the transaction should already be rolled back. This looks correct as written — just confirming the intent is all-or-nothing semantics.

app/views/shared/_pr_review_table.html.erb
```diff
@@ -0,0 +1,44 @@
...
+              #<%= pr.github_number %> - <%= truncate(pr.title, length: 50) %>
+              <% if has_active_run %>
+                <span class="ml-1 inline-flex items-center rounded-full bg-yellow-100 px-2 py-0.5 text-xs font-medium text-yellow-800">In Progress</span>
+              <% end %>
+            </td>
+            <td class="px-3 py-2 text-sm text-gray-500">
+              <%= priority_tier || "-" %>
+            </td>
+            <td class="px-3 py-2 text-right text-sm text-gray-500">
+              <%= pr.draft_review_count + pr.pr_followup_count %>
```
viamin: **Minor: consider caching this computation.**

`pr.draft_review_count + pr.pr_followup_count` is fine since these are database columns, but if this table grows large, you might want to add a combined `review_rounds_count` column or a model method to encapsulate this logic:

```ruby
# On Issue model:
def review_rounds_count
  draft_review_count + pr_followup_count
end
```

This would make the intent clearer in the view: `pr.review_rounds_count` vs the raw addition.

app/views/shared/_pr_review_table.html.erb
```diff
@@ -0,0 +1,44 @@
...
+        <% @pull_requests.each do |pr| %>
+          <% has_active_run = @prs_with_active_runs.include?(pr.github_number) %>
+          <% priority_tier = @pr_priority_tiers[pr.id] %>
+          <tr class="<%= 'opacity-50' if has_active_run %>">
+            <td class="px-3 py-2 text-center">
+              <input type="checkbox"
+                     name="pull_request_ids[]"
+                     value="<%= pr.id %>"
+                     <%= "disabled" if has_active_run %>
+                     <%= "data-permanently-disabled" if has_active_run %>
```
viamin: **HTML nit: bare `data-permanently-disabled` attribute.**

This works in HTML5 and `hasAttribute()` handles it correctly in the JS, but for consistency with the HTML convention used elsewhere in the codebase, consider giving it an explicit value:

```suggestion
                     <%= 'data-permanently-disabled="true"' if has_active_run %>
```

This is purely stylistic — the current version is functionally correct.

app/views/agent_runs/new.html.erb
```diff
@@ -64,16 +64,21 @@
...
+                  <option value="">-- Select a pull request --</option>
+                  <% @pull_requests.each do |pr| %>
+                    <option value="<%= pr.id %>" <%= "selected" if params[:pull_request_id].to_s == pr.id.to_s %>>
+                      #<%= pr.github_number %> - <%= truncate(pr.title, length: 60) %>
+                    </option>
+                  <% end %>
+                </select>
+              </div>
+
+              <%= render "shared/pr_review_table" %>
```
viamin: **Note: `app/views/agent_runs/new.html.erb` is not routed to.**

The top-level `AgentRunsController` only defines `index`, and routes confirm `resources :agent_runs, only: [:index]`. This file appears to be a dead template. The changes here keep it in sync with the project-scoped version, which is fine, but this partial depends on `@prs_with_active_runs` and `@pr_priority_tiers` which would never be set if this view were rendered.

Consider either removing this orphaned view file in a follow-up or adding a comment noting it's kept in sync with `projects/agent_runs/new.html.erb`.

app/javascript/controllers/goal_toggle_controller.js
```diff
@@ -32,7 +40,25 @@ export default class extends Controller {
       )
     })
 
-    if (goal === "review") {
+    // Toggle between dropdown (create_pr) and table (review)
+    this.prDropdownTargets.forEach((el) => {
+      el.hidden = isReview
+      el.querySelectorAll("select").forEach((control) => {
+        control.disabled = isReview || !showPr
+      })
```
viamin, line 52: **Good pattern.** The `data-permanently-disabled` attribute check ensures active-run checkboxes stay disabled even when toggling between goals. This is a clean approach.

spec/requests/agent_runs_spec.rb
```diff
@@ -643,14 +643,34 @@
...
         it "enqueues ProcessRunQueueJob for review runs" do
           expect {
-            post project_agent_runs_path(project), params: { goal: "review", pull_request_id: pr.id }
+            post project_agent_runs_path(project), params: { goal: "review", pull_request_ids: [ pr.id ] }
           }.to have_enqueued_job(ProcessRunQueueJob)
         end
+
+        it "creates one agent run per selected PR when multiple are selected" do
+          pr2 = create(:issue, :pull_request, project: project, github_number: 56, title: "Second PR")
+
```
viamin, line 657: **Good test coverage.** The multi-PR test and the invalid-ID test cover the important new paths. Consider also adding a test for the case where some selected PRs have active runs (to verify the controller doesn't double-create runs for those), though this may be better handled as a separate issue since the current behavior relies on the UI disabling those checkboxes rather than server-side enforcement.

viamin (commented):
## Review Summary

Clean implementation that addresses #957 well. The multi-select table, disabled-state handling for active runs, and fan-out into individual agent runs all look correct. The transaction ensures atomicity, and the queries are efficient (no N+1 issues).

A few items worth addressing below — one potential bug, one minor code quality issue, and a couple of small suggestions.

viamin:
## Agent Update

The background setup agent finished — its findings are consistent with what I already resolved. All changes are committed and the commit passed pre-commit hooks (lint + style checks) and all 150 request specs + 128 issue model specs pass.

app/controllers/projects/agent_runs_controller.rb
viamin, line 51: **Performance (minor):** `@pull_requests.map(&:github_number)` loads all PR records into memory and then iterates them. Since `@pull_requests` is an ActiveRecord relation, this triggers the query. That's fine for now, but if the PR count grows large, consider using `.pluck(:github_number)` to avoid hydrating full Issue objects just for the numbers.

Not blocking — the relation is reused in the view anyway, so it would be loaded regardless.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -54,29 +63,37 @@ def create
       goal = params[:goal].presence || "create_pr"
       custom_prompt = params[:custom_prompt]&.strip.presence
       issue = resolve_issue
-      source_pr_number = resolve_pull_request
 
       if goal == "review"
-        unless source_pr_number
+        pr_ids = Array(params[:pull_request_ids]).filter_map { |id| Integer(id, exception: false) }
```
viamin: **Security note:** Good use of `Integer(id, exception: false)` to safely filter non-numeric values. This correctly prevents injection of arbitrary strings into the `where(id:)` clause.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -717,6 +734,58 @@ def termination_guardrail_type(agent_run)
...
     end
 
+    def create_review_runs_and_redirect(pr_ids:, on_error_path:, custom_prompt:, goal:)
+      budget_result = CostBudgets::Check.call(@project)
+      unless budget_result[:allowed]
+        redirect_to on_error_path, alert: "Your project's AI budget has been reached. Please adjust your budget settings or try again later."
+        return
+      end
+
+      prs = @project.issues.pull_requests_only.where(id: pr_ids)
```
viamin, line 747: **Potential issue:** The `prs` query uses `.pull_requests_only.where(id: pr_ids)` which is correct for scoping to the project. However, if a user submits IDs for PRs belonging to a *different* project, this query would silently exclude them (since `@project.issues` scopes to the project). That's actually the desired behavior — just calling it out as a positive security pattern.

One consideration: if the user selects 5 PRs but 2 have active runs (and are disabled client-side), a malicious user could still submit those IDs. The `ActiveRecord::RecordNotUnique` rescue on the unique constraint handles this, but it rolls back the *entire* transaction — so valid PRs also don't get runs created. Consider either:
1. Filtering out PRs with active runs server-side before the transaction, or
2. Using `create` (not `create!`) per-PR outside a transaction so partial success is possible.

This is the most impactful issue in the PR.

app/controllers/projects/agent_runs_controller.rb
viamin, line 764: **Robustness:** If one PR triggers `RecordNotUnique` (e.g. it already has an active run), the whole transaction rolls back and *no* runs are created — even for the valid PRs. Since the checkbox is only disabled client-side, a determined user (or a race condition between two users) could hit this.

Consider filtering out PRs with active runs server-side:

```suggestion
      active_pr_numbers = @project.agent_runs
        .where(source_pull_request_number: prs.map(&:github_number), status: AgentRun::UNFINISHED_STATUSES)
        .pluck(:source_pull_request_number)
      prs = prs.where.not(github_number: active_pr_numbers)
      if prs.empty?
        redirect_to on_error_path, alert: "All selected pull requests already have active runs."
        return
      end

      ActiveRecord::Base.transaction do
```

This way the unique constraint rescue becomes a true edge-case fallback rather than the primary guard.

app/views/shared/_pr_review_table.html.erb
```diff
@@ -0,0 +1,44 @@
...
+        <% @pull_requests.each do |pr| %>
+          <% has_active_run = @prs_with_active_runs.include?(pr.github_number) %>
+          <% priority_tier = @pr_priority_tiers[pr.id] %>
+          <tr class="<%= 'opacity-50' if has_active_run %>">
+            <td class="px-3 py-2 text-center">
+              <input type="checkbox"
+                     name="pull_request_ids[]"
+                     value="<%= pr.id %>"
+                     <%= "disabled" if has_active_run %>
+                     <%= 'data-permanently-disabled="true"' if has_active_run %>
```
viamin: **XSS note:** Using `<%= 'data-permanently-disabled="true"' if has_active_run %>` outputs a raw HTML attribute string. While `has_active_run` is a boolean derived from a set lookup (so it's safe here), this pattern bypasses Rails' attribute escaping. Prefer using Rails tag helpers or `content_tag` attributes for consistency. For ERB, the `tag.input` helper or explicitly using `data: { permanently_disabled: true }` would be safer by convention.

Not a real vulnerability in this case since the value is hardcoded, but worth noting as a pattern to avoid.

app/views/shared/_pr_review_table.html.erb
```diff
@@ -0,0 +1,44 @@
...
+              #<%= pr.github_number %> - <%= truncate(pr.title, length: 50) %>
+              <% if has_active_run %>
+                <span class="ml-1 inline-flex items-center rounded-full bg-yellow-100 px-2 py-0.5 text-xs font-medium text-yellow-800">In Progress</span>
+              <% end %>
+            </td>
+            <td class="px-3 py-2 text-sm text-gray-500">
+              <%= priority_tier || "-" %>
+            </td>
+            <td class="px-3 py-2 text-right text-sm text-gray-500">
+              <%= pr.review_rounds_count %>
```
viamin: Good: `review_rounds_count` uses `draft_review_count + pr_followup_count` which are both database columns, so no N+1 query here.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -717,6 +734,58 @@ def termination_guardrail_type(agent_run)
...
+    rescue ActiveRecord::RecordNotUnique => e
+      alert = if (e.cause&.message || e.message).include?("proxy_token")
+        "An unexpected error occurred. Please try again."
+      else
+        "An agent run is already queued or in progress for one of the selected PRs."
+      end
+      redirect_to on_error_path, alert: alert
+    end
+
+    def compute_pr_priority_tiers(pull_requests)
```
viamin, line 789: **Performance:** `compute_pr_priority_tiers` iterates all PRs × all priority tiers (3). Since `pr.labels` is a jsonb column loaded with the record, this is fine — no extra queries. Clean implementation.

viamin (commented):
## PR Review: Replace PR dropdown with multi-select table

Overall this is a well-structured PR that cleanly addresses #957. The controller split between review and non-review goals is sensible, the transaction wrapping is correct, and the test coverage for the new multi-select path is solid. A few items to address below.

app/javascript/controllers/goal_toggle_controller.js
```diff
@@ -32,7 +40,25 @@ export default class extends Controller {
...
+    })
+
+    this.prTableTargets.forEach((el) => {
```
viamin, line 55: **Best practice:** Using `data-permanently-disabled` as a custom HTML attribute works but is a non-standard pattern. Consider using a Stimulus value or a `data-*` attribute that's managed via the Stimulus controller's `connect()` lifecycle instead of a plain DOM attribute check. This is minor and works correctly as-is.

spec/requests/agent_runs_spec.rb
```diff
@@ -643,14 +643,41 @@
...
+          pr_numbers = AgentRun.last(2).map(&:source_pull_request_number)
+          expect(pr_numbers).to contain_exactly(55, 56)
+          expect(response).to redirect_to(project_path(project))
+          expect(flash[:notice]).to include("2 agent runs queued")
+        end
+
+        it "redirects with error when all PR IDs are invalid" do
+          post project_agent_runs_path(project), params: { goal: "review", pull_request_ids: [ -1 ] }
+
```
viamin, line 670: Good test coverage for the multi-select path. Consider also adding a test case for the race condition / server-side validation scenario: what happens when a user submits PR IDs that already have active runs? This would document the current behavior (full transaction rollback) and would need updating if the server-side filtering suggestion is adopted.

app/views/agent_runs/new.html.erb
```diff
@@ -1,3 +1,7 @@
+<%# This template mirrors projects/agent_runs/new.html.erb but is not currently
```
viamin: Good: the comment clarifies this template is unrouted and explains the `@prs_with_active_runs` / `@pr_priority_tiers` dependency. If this template is truly dead code, consider whether it should be removed entirely rather than maintained in parallel — keeping two copies in sync is a maintenance burden.

viamin:
## Agent Update

Commit succeeded. All pre-commit hooks passed. Here's a summary of what was addressed:

**Review feedback addressed:**

1. **Server-side active run filtering** (most impactful): Added filtering in `create_review_runs_and_redirect` to exclude PRs with active runs before the transaction. This prevents the entire transaction from rolling back when a bypassed checkbox or race condition submits duplicate PR IDs. The `RecordNotUnique` rescue is now a true edge-case fallback.

2. **`.pluck` over `.map`** (line 50): Changed `@pull_requests.map(&:github_number)` to `.pluck(:github_number)` to avoid hydrating full Issue objects.

3. **Raw HTML attribute pattern** (partial line 24): Replaced raw `<%= 'data-permanently-disabled="true"' %>` interpolation with Rails' `tag.input` helper and proper `data:` hash for safe attribute escaping.

4. **Dead template removal** (agent_runs/new.html.erb): Removed the unrouted template — the top-level `AgentRunsController` only defines `index`, so this was dead code adding maintenance burden.

5. **Test coverage**: Added two new tests covering the server-side filtering: one for partial success (some PRs have active runs) and one for the all-active-runs redirect.

**Positive feedback threads** (no action needed): Security note on `Integer()`, scoped query pattern, `review_rounds_count` N+1 check, `compute_pr_priority_tiers` perf, and `data-permanently-disabled` Stimulus pattern were all acknowledged as working correctly.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -46,6 +46,15 @@ def new
         .pull_requests_only
         .where(github_state: "open")
         .order(github_number: :desc)
+
+      pr_numbers = @pull_requests.pluck(:github_number)
```
viamin: **Performance**: `pluck` on an ActiveRecord relation that hasn't been loaded yet fires a `SELECT github_number FROM issues …` query. Then the view later iterates `@pull_requests`, firing a second `SELECT * FROM issues …` query.

Consider eagerly loading the records once and extracting the numbers from the loaded array:

```suggestion
      @pull_requests = @pull_requests.to_a
      pr_numbers = @pull_requests.map(&:github_number)
```

This avoids the extra SELECT and ensures the collection is loaded once for both the controller logic and the view iteration.

app/views/shared/_pr_review_table.html.erb
```diff
@@ -0,0 +1,44 @@
...
+      <thead class="bg-gray-50">
+        <tr>
+          <th scope="col" class="w-10 px-3 py-2"></th>
+          <th scope="col" class="px-3 py-2 text-left text-xs font-medium uppercase tracking-wide text-gray-500">PR</th>
+          <th scope="col" class="px-3 py-2 text-left text-xs font-medium uppercase tracking-wide text-gray-500">Priority</th>
+          <th scope="col" class="px-3 py-2 text-right text-xs font-medium uppercase tracking-wide text-gray-500">Review Rounds</th>
+        </tr>
+      </thead>
+      <tbody class="divide-y divide-gray-200 bg-white">
+        <% @pull_requests.each do |pr| %>
```
viamin: **Best practice**: This partial references three instance variables (`@pull_requests`, `@prs_with_active_runs`, `@pr_priority_tiers`) directly. Shared partials should receive data via locals to stay decoupled and reusable.

Consider rendering with locals:
```erb
<%= render "shared/pr_review_table", pull_requests: @pull_requests, prs_with_active_runs: @prs_with_active_runs, pr_priority_tiers: @pr_priority_tiers %>
```

And in the partial use `pull_requests`, `prs_with_active_runs`, `pr_priority_tiers` instead of the `@`-prefixed versions. This also makes it immediately clear from the render call what data the partial depends on.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -717,6 +734,69 @@ def termination_guardrail_type(agent_run)
...
+        redirect_to on_error_path, alert: "None of the selected pull requests could be found."
+        return
+      end
+
+      # Filter out PRs that already have active (unfinished) runs server-side,
+      # since the client-side disabled checkbox can be bypassed or hit by a race.
+      active_pr_numbers = @project.agent_runs
+        .where(source_pull_request_number: prs.map(&:github_number), status: AgentRun::UNFINISHED_STATUSES)
+        .pluck(:source_pull_request_number)
+      prs = prs.where.not(github_number: active_pr_numbers) if active_pr_numbers.any?
```
viamin, line 758: **Minor (TOCTOU)**: There's a small race window between the active-run check (lines 752-755) and the `create_agent_run` calls inside the transaction (lines 761-768). Two concurrent requests could both pass the filter and create duplicate runs for the same PR.

The `RecordNotUnique` rescue at line 780 catches this for unique-index violations, which is a reasonable safety net. Just flagging that if there's no unique index on `(project_id, source_pull_request_number, status)` for unfinished statuses, a partial unique index or advisory lock might be worth considering. If there is such an index, this is fine as-is.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -717,6 +734,69 @@ def termination_guardrail_type(agent_run)
         "unknown"
     end
 
+    def create_review_runs_and_redirect(pr_ids:, on_error_path:, custom_prompt:, goal:)
```
viamin, line 738: **Observation**: This method is ~50 lines with multiple early-return redirects, a transaction block, and two rescue clauses. It's well-organized, but if it grows further, extracting the PR-resolution and active-run filtering into a small private method (or even a service object like `AgentRuns::CreateReviewBatch`) would keep the controller leaner — especially since the controller is already 800+ lines.

app/javascript/controllers/goal_toggle_controller.js
viamin, line 54: Good use of the `data-permanently-disabled` attribute to prevent re-enabling checkboxes for PRs with active runs during goal toggles. Clean approach.

app/models/issue.rb
```diff
@@ -147,6 +147,10 @@ def has_associated_pull_requests?
     end
   end
 
+  def review_rounds_count
```
viamin, line 150: Clean convenience method. Since both `draft_review_count` and `pr_followup_count` are integer DB columns with `default: 0, null: false`, this is safe and efficient.

spec/requests/agent_runs_spec.rb
```diff
@@ -643,14 +643,66 @@
...
+          }.to change(AgentRun, :count).by(2)
+
+          pr_numbers = AgentRun.last(2).map(&:source_pull_request_number)
+          expect(pr_numbers).to contain_exactly(55, 56)
+          expect(response).to redirect_to(project_path(project))
+          expect(flash[:notice]).to include("2 agent runs queued")
+        end
+
+        it "redirects with error when all PR IDs are invalid" do
+          post project_agent_runs_path(project), params: { goal: "review", pull_request_ids: [ -1 ] }
```
viamin: Solid test coverage — the multi-select, invalid ID filtering, and active-run server-side filtering scenarios are all well covered. The test for non-numeric IDs (line 682) is a nice touch for input validation.

viamin (commented):
## PR Review: Replace PR dropdown with multi-select table (#957)

Overall this is a well-structured PR that cleanly implements the multi-select table for PR reviews. The server-side active-run filtering is a good security measure, the test coverage is thorough, and the deletion of the duplicate non-namespaced view is a nice cleanup.

A few issues to address — one performance concern, one partial coupling issue, and a minor TOCTOU race window.

viamin:
## Agent Update

Commit succeeded. All lint checks passed. Here's what was addressed:

1. **Performance fix (controller line 50)**: Replaced `.pluck(:github_number)` with `.to_a` + `.map(&:github_number)`. This eagerly loads `@pull_requests` once, so both the controller logic (extracting PR numbers) and the view iteration share the same loaded array — eliminating the extra `SELECT github_number FROM issues …` query.

2. **Shared partial decoupling**: The `_pr_review_table` partial now receives `pull_requests`, `prs_with_active_runs`, and `pr_priority_tiers` as locals instead of referencing `@`-prefixed instance variables directly. The render call in `new.html.erb` passes these explicitly, making the partial's data dependencies clear and keeping it reusable.

app/javascript/controllers/goal_toggle_controller.js
```diff
@@ -32,7 +40,25 @@ export default class extends Controller {
       )
     })
```
viamin, line 43: **Fragile disable/enable ordering** — The `prSectionTargets` loop (lines 34-41) enables *all* `input, select, textarea, button` controls when `showPr` is true. This re-enables:
1. The dropdown `<select>` (which the `prDropdownTargets` loop at line 44 then re-disables for review mode)
2. Checkboxes with `data-permanently-disabled` (which the `prTableTargets` loop at line 53 then re-disables)

This works only because the more-granular toggles run *after* the broad one. If someone reorders these blocks or adds early-returns, the permanently-disabled checkboxes will get re-enabled. Consider skipping controls with `data-permanently-disabled` in the broad loop too:

```suggestion
    this.prSectionTargets.forEach((el) => {
      el.hidden = !showPr
      el.querySelectorAll("input, select, textarea, button").forEach(
        (control) => {
          if (!control.hasAttribute('data-permanently-disabled')) {
            control.disabled = !showPr
          }
        }
      )
    })
```

viamin (commented):
## PR Review: Replace PR dropdown with multi-select table (#957)

Overall this is a well-structured PR that clearly addresses the requirements of #957. The controller logic is solid with good server-side validation, the Stimulus controller cleanly toggles between views, and the test coverage is thorough. A few items to address below — one potential bug and some minor improvements.

app/javascript/controllers/goal_toggle_controller.js
```diff
@@ -32,7 +40,25 @@ export default class extends Controller {
...
+        // Re-enable non-disabled-by-default checkboxes when review is shown
+        if (!control.hasAttribute('data-permanently-disabled')) {
+          control.disabled = !isReview
+        }
+      })
+    })
+
+    if (isReview) {
       this.prHeadingTargets.forEach((el) => {
         el.textContent = "Select PR to Review"
```
viamin: Minor: the heading still says "Select PR to Review" (singular), but the feature now supports multi-select. Consider updating to "Select PRs to Review" for consistency with the table label.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -54,29 +64,37 @@ def create
       goal = params[:goal].presence || "create_pr"
       custom_prompt = params[:custom_prompt]&.strip.presence
       issue = resolve_issue
-      source_pr_number = resolve_pull_request
 
       if goal == "review"
-        unless source_pr_number
+        pr_ids = Array(params[:pull_request_ids]).filter_map { |id| Integer(id, exception: false) }
```
viamin: Good defensive parsing with `Integer(id, exception: false)`. One edge case: `Integer("0", exception: false)` returns `0`, which would pass the `filter_map` (since `0` is truthy in Ruby) but would never match a real PR id. This is harmless since the subsequent `where(id: pr_ids)` query returns nothing for id=0, but if you want belt-and-suspenders:

```suggestion
        pr_ids = Array(params[:pull_request_ids]).filter_map { |id| Integer(id, exception: false) }.select(&:positive?)
```

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -46,6 +46,16 @@ def new
         .pull_requests_only
         .where(github_state: "open")
         .order(github_number: :desc)
+
+      @pull_requests = @pull_requests.to_a
```
viamin, line 50: Materializing the AR relation to an array here (`to_a`) means `@pull_requests` loses its AR relation interface. This is fine for the current usage (iteration in views and `map` in the controller), but be aware that anything downstream expecting a relation (e.g., chaining `.where` or `.includes`) would break.

Also, since `compute_pr_priority_tiers` calls `pr.labels` for each PR, and `labels` is a jsonb column on the same table, there's no N+1 here — good.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -717,6 +735,69 @@ def termination_guardrail_type(agent_run)
...
+      active_pr_numbers = @project.agent_runs
+        .where(source_pull_request_number: prs.map(&:github_number), status: AgentRun::UNFINISHED_STATUSES)
+        .pluck(:source_pull_request_number)
+      prs = prs.where.not(github_number: active_pr_numbers) if active_pr_numbers.any?
+      if prs.empty?
+        redirect_to on_error_path, alert: "All selected pull requests already have active runs."
+        return
+      end
+
+      ActiveRecord::Base.transaction do
```
viamin, line 764: If the transaction wraps many PRs and one `create_agent_run` raises `RecordNotUnique` (e.g., duplicate proxy_token), the entire batch rolls back. That means zero runs are created even if only one PR was the problem. Consider whether you'd prefer to create runs for the non-conflicting PRs and report partial success, or if the all-or-nothing approach is intentional. The current behavior is safe and predictable, just worth documenting the intent.

app/views/shared/_pr_review_table.html.erb
```diff
@@ -0,0 +1,44 @@
...
+              #<%= pr.github_number %> - <%= truncate(pr.title, length: 50) %>
+              <% if has_active_run %>
+                <span class="ml-1 inline-flex items-center rounded-full bg-yellow-100 px-2 py-0.5 text-xs font-medium text-yellow-800">In Progress</span>
+              <% end %>
+            </td>
+            <td class="px-3 py-2 text-sm text-gray-500">
+              <%= priority_tier || "-" %>
+            </td>
+            <td class="px-3 py-2 text-right text-sm text-gray-500">
+              <%= pr.review_rounds_count %>
```
viamin: Potential N+1 concern: `pr.review_rounds_count` calls `draft_review_count + pr_followup_count`, which are both columns on the issues table — so this is just attribute access, no extra queries. Good.

Minor UX thought: consider showing "0" styled differently or as "—" when review_rounds_count is 0, to distinguish "never reviewed" from "reviewed". Not blocking.

app/views/agent_runs/new.html.erb
```diff
@@ -1,121 +0,0 @@
-<% content_for(:title) { "Trigger Agent Run - #{@project.name} - Paid" } %>
```
viamin, line 1: Deleting the non-namespaced `agent_runs/new.html.erb` template — is there anything that still routes to `AgentRunsController#new` (without the `Projects::` namespace)? If the non-namespaced controller still exists and has a `new` action, this deletion will cause a `MissingTemplate` error. Worth a quick grep to confirm nothing references this path.

viamin:
## Agent Update

Commit succeeded with all pre-commit hooks passing. Here's a summary of what was addressed:

**Review feedback addressed:**

1. **Fragile disable/enable ordering** (goal_toggle_controller.js:34-41) — The broad `prSectionTargets` loop now skips controls with `data-permanently-disabled`, eliminating the ordering dependency where the `prTableTargets` loop had to run after to re-disable them.

2. **Singular heading** (goal_toggle_controller.js:63) — Changed "Select PR to Review" to "Select PRs to Review" for multi-select consistency.

3. **Non-positive PR IDs** (controller:69) — Added `.select(&:positive?)` to filter out zero and negative IDs that would pass `Integer()` parsing but never match real records.

4. **Review rounds UX** (shared partial:37) — Shows "—" (em dash) instead of "0" for never-reviewed PRs to visually distinguish them.

5. **Test updates** — Fixed the invalid-ID test to use `999_999_999` (a non-existent positive ID) instead of `-1` (which is now filtered earlier), and added a new test specifically for negative/zero ID rejection.

**Informational threads (no code change needed):**
- `.to_a` materialization (controller:50) — Acknowledged; current usage is fine
- Transaction rollback behavior (controller:762) — All-or-nothing is intentional and documented
- Dead template deletion (agent_runs/new.html.erb) — Already handled in prior commit

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -46,6 +46,16 @@ def new
...
         .where(github_state: "open")
         .order(github_number: :desc)
+
+      @pull_requests = @pull_requests.to_a
+      pr_numbers = @pull_requests.map(&:github_number)
+      @prs_with_active_runs = @project.agent_runs
+        .where(source_pull_request_number: pr_numbers, status: AgentRun::UNFINISHED_STATUSES)
+        .distinct
+        .pluck(:source_pull_request_number)
+        .to_set
```
viamin, line 56: **Performance (minor):** `@pull_requests` is converted `.to_a` at line 50 and `pr_numbers` is computed via `.map`. This is fine for a typical PR list, but note that the `@prs_with_active_runs` query and the `compute_pr_priority_tiers` loop both iterate over the full set. If projects ever accumulate a very large number of open PRs, consider adding a `.limit()` — but this is unlikely to matter in practice.

app/controllers/projects/agent_runs_controller.rb
viamin, line 764: **Race condition:** Between the `active_pr_numbers` check (line 753) and the `create_agent_run` call inside the transaction (line 764), another request could create a run for the same PR. The `RecordNotUnique` rescue at line 781 will catch a unique constraint violation on `proxy_token`, but there's no unique constraint on `(project_id, source_pull_request_number, status)` preventing duplicate queued runs for the same PR.

The current behavior is acceptable since duplicate review runs are a minor annoyance rather than a correctness issue, but if you want stricter enforcement, consider using a `SELECT ... FOR UPDATE` or an advisory lock within the transaction.

Alternatively, the `RecordNotUnique` rescue's `else` branch message ("An agent run is already queued or in progress for one of the selected PRs.") won't actually fire for this case since there's no unique index backing it — it would only fire for the `proxy_token` collision. Worth verifying whether the intended guard actually exists at the database level.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -717,6 +735,69 @@ def termination_guardrail_type(agent_run)
...
+
+      ActiveRecord::Base.transaction do
+        prs.each do |pr|
+          create_agent_run(
+            custom_prompt: custom_prompt,
+            source_pull_request_number: pr.github_number,
+            goal: goal
+          )
+        end
+      end
```
viamin, line 772: **Atomicity consideration:** If `create_agent_run` raises `NoRunnableProviderError` partway through the loop (e.g., after creating 2 of 5 runs), the transaction rolls back all of them and the user gets an error. This seems correct — good that the rescue is outside the transaction block.

app/controllers/projects/agent_runs_controller.rb
```diff
@@ -717,6 +735,69 @@ def termination_guardrail_type(agent_run)
         "unknown"
     end
 
+    def create_review_runs_and_redirect(pr_ids:, on_error_path:, custom_prompt:, goal:)
+      budget_result = CostBudgets::Check.call(@project)
```
viamin, line 741: **Budget check placement:** The budget check runs once before all runs are created. If the user selects, say, 10 PRs and the budget only allows 3 more runs, all 10 will be created. Consider either:
- Checking `budget_result` includes remaining capacity and comparing against `pr_ids.size`, or
- Accepting this as-is if the budget check is a soft gate (checked again at execution time).

Let me know if the budget is re-checked at run execution time — if so, this is fine.

app/views/shared/_pr_review_table.html.erb
```diff
@@ -0,0 +1,44 @@
...
+              <% if has_active_run %>
+                <span class="ml-1 inline-flex items-center rounded-full bg-yellow-100 px-2 py-0.5 text-xs font-medium text-yellow-800">In Progress</span>
+              <% end %>
+            </td>
+            <td class="px-3 py-2 text-sm text-gray-500">
+              <%= priority_tier || "-" %>
+            </td>
+            <td class="px-3 py-2 text-right text-sm text-gray-500">
+              <%= pr.review_rounds_count > 0 ? pr.review_rounds_count : "—" %>
+            </td>
```
viamin, line 38: **Minor:** `pr.review_rounds_count` calls `draft_review_count + pr_followup_count` which are both DB columns — no N+1 concern since `@pull_requests` is already loaded as an array. Nice.

One small nit: the em-dash `"—"` for zero rounds is a nice touch but it uses a different convention than the `"-"` for missing priority tier on line 36. Consider using the same character for both empty states for visual consistency.

```suggestion
              <%= pr.review_rounds_count > 0 ? pr.review_rounds_count : "-" %>
```

viamin (commented):
Good PR — clean separation between the dropdown and table views, solid server-side validation, and thorough test coverage for multi-select edge cases. A few items to consider below, mostly around a potential race condition and a minor N+1 risk.

app/javascript/controllers/goal_toggle_controller.js
```diff
@@ -27,14 +35,34 @@ export default class extends Controller {
...
+            control.disabled = !showPr
+          }
         }
       )
     })
 
-    if (goal === "review") {
+    // Toggle between dropdown (create_pr) and table (review)
+    this.prDropdownTargets.forEach((el) => {
+      el.hidden = isReview
```
viamin, line 49: **Correctness:** The `prSection` targets' `forEach` (line 35-42) will enable/disable controls within the section, and then `prDropdown` (line 45) and `prTable` (line 53) targets will toggle within that same section. Since `prDropdown` and `prTable` are nested inside `prSection`, there's a risk of the outer loop enabling a control that the inner loop should keep disabled (or vice versa).

It works because the inner loops run after the outer one, overriding the state — but this ordering dependency is fragile. If someone reorders the blocks, the behavior breaks silently. A brief comment noting the intentional ordering would help future maintainers.

app/views/agent_runs/new.html.erb
viamin, line 1: **Scope check:** This deletes the non-namespaced `agent_runs/new.html.erb` entirely. Just want to confirm there are no routes still pointing to `AgentRunsController#new` (without the `Projects::` namespace) that would now 404. If this view was already dead code, great cleanup.

app/models/issue.rb
```diff
@@ -147,6 +147,10 @@ def has_associated_pull_requests?
...
+    draft_review_count + pr_followup_count
```
viamin, line 151: Clean addition. Since both `draft_review_count` and `pr_followup_count` are `integer, default: 0, null: false` columns, this is always safe — no nil-handling needed.

spec/requests/agent_runs_spec.rb
```diff
@@ -643,14 +643,73 @@
...
+          expect(response).to redirect_to(new_project_agent_run_path(project, goal: "review"))
+          expect(flash[:alert]).to include("None of the selected pull requests could be found")
+        end
+
+        it "rejects negative and zero PR IDs" do
+          post project_agent_runs_path(project), params: { goal: "review", pull_request_ids: [ "-1", "0" ] }
+
+          expect(response).to redirect_to(new_project_agent_run_path(project, goal: "review"))
+          expect(flash[:alert]).to include("Please select at least one pull request to review.")
+        end
```
viamin, line 687: Good coverage of edge cases — invalid IDs, negative/zero IDs, non-numeric IDs, active run filtering, and the "all have active runs" scenario. This is thorough.

One thing not tested: what happens when a user submits `pull_request_ids` as an empty array `[]` vs. not providing the param at all? Both should hit the "at least one" error, but the `Array()` coercion handles them differently (`Array([])` → `[]`, `Array(nil)` → `[]`). Might be worth a quick test for the nil/missing param case.

viamin:
## Agent Update

All changes are committed. Here's a summary of what was addressed:

### Review feedback addressed:

1. **Race condition / misleading `RecordNotUnique` rescue** (controller:781) — Removed the `else` branch that would never fire (no unique index on `(project_id, source_pull_request_number, status)` exists). Simplified to a single generic error message with a comment explaining the constraint landscape.

2. **Budget check placement** (controller:739) — Added a comment clarifying that the budget is a soft gate, re-checked at execution time in `ProcessRunQueueJob#start_claimed_run`, so over-queuing is harmless.

3. **Inconsistent empty-state characters** (partial:34 vs :37) — Changed priority tier's `"-"` to `"—"` to match the em-dash used for zero review rounds.

4. **Fragile ordering dependency in JS** (goal_toggle_controller.js:47) — Added a comment explaining that the dropdown/table blocks intentionally run after `prSectionTargets` and override its enable/disable state.

5. **Missing test for empty array param** (spec:675) — Added test for `pull_request_ids: []` to complement the existing nil/missing param test.

### Informational threads (no code change needed):
- **Performance** (line 56) — Acknowledged as fine for typical PR counts
- **Atomicity** (line 770) — All-or-nothing rollback is correct behavior
- **Scope check** (deleted view) — Already confirmed dead code; top-level route only has `:index`
- **`review_rounds_count`** — Clean, no N+1 concern
- **Test coverage** — Acknowledged as thorough

viamin (commented):
Generated no new comments. The PR cleanly implements all requirements from #957: multi-select table with review round counts, disabled state for PRs with active runs, server-side validation of selected PR IDs (including filtering out active runs to prevent race conditions), and proper fan-out of one agent run per selected PR within a transaction. Test coverage is thorough, covering multi-select, invalid/negative IDs, and active-run filtering. The dead non-namespaced view removal is a reasonable cleanup.

PR #969, opened 2026-04-08
Fix #964: fix(workflow): CreatePullRequestActivity is not idempotent — retries leave orphan branches

### LINKED ISSUE #964: fix(workflow): CreatePullRequestActivity is not idempotent — retries leave orphan branches

viamin:
Based on the current code and the issue/agent output, I can write the PR description. The changes are clear from the issue and agent output.

## Summary

Make `CreatePullRequestActivity` idempotent so that Temporal retries after partial failures reuse an existing GitHub PR instead of hitting a 422 error, which previously left orphan branches and incorrectly failed agent runs.

## Changes

**Activity idempotency (`app/temporal/activities/create_pull_request_activity.rb`):**
- Add `find_existing_pr` method that checks for an open PR matching the branch via `client.pull_requests(repo, head: "owner:branch", state: 'open')` before attempting to create one
- On retry, if a PR already exists for the branch, reuse it instead of calling `create_pull_request` again — eliminates the GitHub 422 ("A pull request already exists") failure path
- Persist `agent_run.complete!` with PR URL/number immediately after PR creation/lookup, before any post-processing

**Best-effort post-processing:**
- Wrap `add_pr_labels` and `agent_run.log!` in a `best_effort` helper that rescues `StandardError` and logs a warning
- Failures in labeling or logging no longer cause the activity to raise after `complete!` has run, preventing Temporal from retrying into a non-idempotent state

**Tests (`spec/temporal/activities/create_pull_request_activity_spec.rb`):**
- PR creation checks for existing PR before creating a new one
- Existing open PR is reused without calling `create_pull_request`
- Agent run is marked completed even when post-processing raises
- Retry after partial failure (PR created but activity didn't return) reuses the existing PR
- No orphan branches produced when `log!` raises after completion

## Design Decisions

- **Check-then-create over rescue-and-retry**: Rather than catching the 422 and then looking up the PR, we proactively check for an existing PR first. This avoids relying on GitHub error response parsing and makes the idempotency explicit in the happy path.
- **Best-effort wrapper instead of separate activities**: Post-processing steps (labels, logging) are wrapped in-place rather than extracted into separate Temporal activities. This keeps the workflow definition simple while still isolating failures — a separate activity would add retry/timeout configuration overhead for operations that are purely cosmetic.
- **`complete!` as the commit point**: Once `complete!` succeeds, the activity treats everything after as best-effort. This ensures the agent run state is consistent even if the activity is never retried (e.g., workflow cancellation).

---

Closes #964

### DIFF
--- a/app/temporal/activities/create_pull_request_activity.rb
+++ b/app/temporal/activities/create_pull_request_activity.rb
@@ -12,7 +12,12 @@ module Activities
         issue = agent_run.issue
 
         client = project.github_token.client
-        pr = client.create_pull_request(
+
+        # Idempotency: reuse an existing PR for this branch if one was
+        # already created (e.g. by a prior attempt that failed after the
+        # GitHub API call but before the activity returned).
+        pr = find_existing_pr(client, project, agent_run.branch_name)
+        pr ||= client.create_pull_request(
           project.full_name,
           base: project.default_branch,
           head: agent_run.branch_name,
@@ -21,15 +26,19 @@ module Activities
           draft: true
         )
 
+        # Persist completion as the very first step after obtaining the PR,
+        # before any best-effort post-processing. This ensures a retry
+        # cannot overwrite status via MarkAgentRunFailedActivity.
         agent_run.complete!(
           result_commit: agent_run.result_commit_sha,
           pr_url: pr.html_url,
           pr_number: pr.number
         )
 
-        add_pr_labels(client, project, pr.number, agent_run_id, issue: issue)
-
-        agent_run.log!("system", "PR created: #{pr.html_url}")
+        # Best-effort post-processing — failures here must not cause the
+        # activity to be retried now that the run is already completed.
+        best_effort(agent_run_id) { add_pr_labels(client, project, pr.number, agent_run_id, issue: issue) }
+        best_effort(agent_run_id) { agent_run.log!("system", "PR created: #{pr.html_url}") }
 
         logger.info(
           message: "agent_execution.pull_request_created",
@@ -42,6 +51,26 @@ module Activities
     end
 
     private
+
+    def find_existing_pr(client, project, branch_name)
+      existing = client.pull_requests(
+        project.full_name,
+        head: "#{project.owner}:#{branch_name}",
+        state: "open"
+      )
+      existing.first
+    end
+
+    def best_effort(agent_run_id)
+      yield
+    rescue StandardError => e
+      logger.warn(
+        message: "agent_execution.post_processing_failed",
+        agent_run_id: agent_run_id,
+        error_class: e.class.name,
+        error: e.message
+      )
+    end
 
     def pr_title(issue)
       return "Agent changes" unless issue
--- a/app/temporal/activities/create_pull_request_activity.rb
+++ b/app/temporal/activities/create_pull_request_activity.rb
@@ -16,8 +16,8 @@ module Activities
         # Idempotency: reuse an existing PR for this branch if one was
         # already created (e.g. by a prior attempt that failed after the
         # GitHub API call but before the activity returned).
-        pr = find_existing_pr(client, project, agent_run.branch_name, agent_run_id: agent_run_id)
-        pr ||= client.create_pull_request(
+        existing_pr = find_existing_pr(client, project, agent_run.branch_name, agent_run_id: agent_run_id)
+        pr = existing_pr || client.create_pull_request(
           project.full_name,
           base: project.default_branch,
           head: agent_run.branch_name,
@@ -25,6 +25,7 @@ module Activities
           body: pr_body(issue, agent_run),
           draft: true
         )
+        pr_action = existing_pr ? "reused" : "created"
 
         # Persist completion as the very first step after obtaining the PR,
         # before any best-effort post-processing. This ensures a retry
@@ -38,10 +39,10 @@ module Activities
         # Best-effort post-processing — failures here must not cause the
         # activity to be retried now that the run is already completed.
         best_effort(agent_run_id) { add_pr_labels(client, project, pr.number, agent_run_id, issue: issue) }
-        best_effort(agent_run_id) { agent_run.log!("system", "PR created: #{pr.html_url}") }
+        best_effort(agent_run_id) { agent_run.log!("system", "PR #{pr_action}: #{pr.html_url}") }
 
         logger.info(
-          message: "agent_execution.pull_request_created",
+          message: "agent_execution.pull_request_#{pr_action}",
           agent_run_id: agent_run_id,
           pull_request_url: pr.html_url
         )
@@ -59,10 +60,11 @@ module Activities
         state: "open"
       )
       existing.first
-    rescue GithubClient::Error => e
-      # Lookup is best-effort: a transient failure here must not become a
-      # new failure mode introduced by the idempotency fix. Fall through
-      # to create_pull_request and let GitHub be the source of truth.
+    rescue StandardError => e
+      # Lookup is best-effort: a transient failure (including network-level
+      # errors like Faraday::TimeoutError) must not become a new failure
+      # mode introduced by the idempotency fix. Fall through to
+      # create_pull_request and let GitHub be the source of truth.
       logger.warn(
         message: "agent_execution.pull_request_lookup_failed",
         agent_run_id: agent_run_id,
@@ -169,13 +171,6 @@ module Activities
       return if labels.empty?
 
       client.add_labels_to_issue(project.full_name, pr_number, labels)
-    rescue GithubClient::Error => e
-      logger.warn(
-        message: "agent_execution.add_pr_labels_failed",
-        agent_run_id: agent_run_id,
-        pr_number: pr_number,
-        error: e.message
-      )
     end
   end
 end
--- a/app/temporal/activities/create_pull_request_activity.rb
+++ b/app/temporal/activities/create_pull_request_activity.rb
@@ -38,14 +38,16 @@ module Activities
 
         # Best-effort post-processing — failures here must not cause the
         # activity to be retried now that the run is already completed.
-        best_effort(agent_run_id) { add_pr_labels(client, project, pr.number, agent_run_id, issue: issue) }
-        best_effort(agent_run_id) { agent_run.log!("system", "PR #{pr_action}: #{pr.html_url}") }
+        best_effort(agent_run_id, context: "add_pr_labels") { add_pr_labels(client, project, pr.number, agent_run_id, issue: issue) }
+        best_effort(agent_run_id, context: "log_pr_action") { agent_run.log!("system", "PR #{pr_action}: #{pr.html_url}") }
 
-        logger.info(
-          message: "agent_execution.pull_request_#{pr_action}",
-          agent_run_id: agent_run_id,
-          pull_request_url: pr.html_url
-        )
+        best_effort(agent_run_id, context: "structured_log") do
+          logger.info(
+            message: "agent_execution.pull_request_#{pr_action}",
+            agent_run_id: agent_run_id,
+            pull_request_url: pr.html_url
+          )
+        end
 
         { agent_run_id: agent_run_id, pull_request_url: pr.html_url, pull_request_number: pr.number }
       end
@@ -74,12 +76,13 @@ module Activities
       nil
     end
 
-    def best_effort(agent_run_id)
+    def best_effort(agent_run_id, context: nil)
       yield
     rescue StandardError => e
       logger.warn(
         message: "agent_execution.post_processing_failed",
         agent_run_id: agent_run_id,
+        context: context,
         error_class: e.class.name,
         error: e.message
       )
--- a/app/temporal/activities/create_pull_request_activity.rb
+++ b/app/temporal/activities/create_pull_request_activity.rb
@@ -16,7 +16,7 @@ module Activities
         # Idempotency: reuse an existing PR for this branch if one was
         # already created (e.g. by a prior attempt that failed after the
         # GitHub API call but before the activity returned).
-        pr = find_existing_pr(client, project, agent_run.branch_name)
+        pr = find_existing_pr(client, project, agent_run.branch_name, agent_run_id: agent_run_id)
         pr ||= client.create_pull_request(
           project.full_name,
           base: project.default_branch,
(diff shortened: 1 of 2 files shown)


app/temporal/activities/create_pull_request_activity.rb
```diff
@@ -43,6 +52,37 @@ def execute(input)
...
     private
 
+    def find_existing_pr(client, project, branch_name, agent_run_id:)
+      existing = client.pull_requests(
+        project.full_name,
+        head: "#{project.owner}:#{branch_name}",
+        state: "open"
+      )
+      existing.first
+    rescue GithubClient::Error => e
```
viamin: **Nit: `GithubClient::Error` vs broader failures** — This rescue covers all `GithubClient::Error` subclasses, which is good. However, if the Octokit/Faraday layer raises something outside the `GithubClient::Error` hierarchy (e.g., `Faraday::TimeoutError`, `Net::OpenTimeout`), it would propagate and the activity would fail *before* even attempting `create_pull_request`. Consider whether rescuing `StandardError` here (like `best_effort` does) would be more defensive, since this lookup is also best-effort.

Not blocking — the current behavior is safe (it falls through to the create path which will fail with a clearer error), but worth considering.

app/temporal/activities/create_pull_request_activity.rb
```diff
@@ -43,6 +52,37 @@ def execute(input)
...
+        agent_run_id: agent_run_id,
+        branch: branch_name,
+        error: e.message
+      )
+      nil
+    end
+
+    def best_effort(agent_run_id)
+      yield
+    rescue StandardError => e
```
viamin, line 81: **Observation: `best_effort` catches `StandardError`** — This is intentionally broad, which makes sense for post-processing. Just noting that `add_pr_labels` at line 172 already has its own `rescue GithubClient::Error` internally. The double rescue is harmless (the inner one catches first for GitHub errors), but if you wanted to simplify, you could remove the inner rescue from `add_pr_labels` since `best_effort` now covers it with a broader net. Not blocking — just a potential future cleanup.

app/temporal/activities/create_pull_request_activity.rb
```diff
@@ -21,15 +26,19 @@ def execute(input)
...
           pr_url: pr.html_url,
           pr_number: pr.number
         )
 
-        add_pr_labels(client, project, pr.number, agent_run_id, issue: issue)
-
-        agent_run.log!("system", "PR created: #{pr.html_url}")
+        # Best-effort post-processing — failures here must not cause the
+        # activity to be retried now that the run is already completed.
+        best_effort(agent_run_id) { add_pr_labels(client, project, pr.number, agent_run_id, issue: issue) }
```
viamin: **Minor: consider logging when an existing PR is reused** — When `find_existing_pr` returns a hit, the activity silently reuses it. The `"PR created"` log message at line 41 and the `"agent_execution.pull_request_created"` structured log at line 44 are emitted in both cases, which is misleading.

Consider capturing whether the PR was found or created and adjusting the messages:
```ruby
existing_pr = find_existing_pr(client, project, agent_run.branch_name, agent_run_id: agent_run_id)
pr = existing_pr || client.create_pull_request(...)
pr_action = existing_pr ? "reused" : "created"
# ...
best_effort(agent_run_id) { agent_run.log!("system", "PR #{pr_action}: #{pr.html_url}") }
```
This would make production debugging much easier when a retry path is taken.

spec/temporal/activities/create_pull_request_activity_spec.rb
```diff
@@ -276,5 +286,64 @@
...
+        allow(github_client).to receive(:add_labels_to_issue)
+          .and_raise(StandardError.new("transient failure"))
+
+        result = activity.execute(agent_run_id: agent_run.id)
+
+        expect(agent_run.reload.status).to eq("completed")
+        expect(result[:pull_request_url]).to eq("https://github.com/owner/repo/pull/42")
+      end
+
+      it "reuses existing PR on retry after partial failure" do
```
viamin: **Test overlap** — This test ("reuses existing PR on retry after partial failure") has the same setup and assertions as "reuses an existing open PR instead of creating a new one" on line 293. The comment says it simulates a first attempt that created the PR but failed after `complete!`, but nothing in the test actually distinguishes this scenario (e.g., pre-setting `agent_run.status` to `completed` before the retry call). Consider either:
1. Making this test genuinely different by setting up the agent_run as already completed before the retry, or
2. Removing it to reduce test maintenance burden.

As-is, it exercises the same code path as the test above.

spec/temporal/activities/create_pull_request_activity_spec.rb
```diff
@@ -276,5 +286,64 @@
...
+
+        result = activity.execute(agent_run_id: agent_run.id)
+
+        expect(github_client).not_to have_received(:create_pull_request)
+        expect(agent_run.reload.status).to eq("completed")
+        expect(result[:pull_request_number]).to eq(99)
+      end
+
+      it "falls through to create_pull_request when the lookup itself errors" do
+        allow(github_client).to receive(:pull_requests).and_raise(GithubClient::ApiError.new("boom"))
```
viamin, line 343: Good coverage of the fallthrough path. Note this only tests `GithubClient::ApiError`, while the production rescue catches the broader `GithubClient::Error`. Low priority since this is a Ruby inheritance guarantee, but a second case with e.g. `GithubClient::RateLimitError` would add confidence.

viamin (commented):
Solid fix for a real idempotency gap. The check-then-create approach with best-effort post-processing is clean and well-tested. A few observations below, mostly minor.

viamin:
## Agent Update

All checks passed and committed. Here's a summary of the changes addressing each review thread:

1. **`find_existing_pr` rescue broadened** (line 63): Changed from `GithubClient::Error` to `StandardError` to also catch network-level errors like `Faraday::TimeoutError` and `Net::OpenTimeout`. Updated comment to explain the rationale.

2. **Removed redundant inner rescue from `add_pr_labels`**: Since `best_effort` already catches `StandardError`, the inner `rescue GithubClient::Error` was unnecessary. Removed to simplify.

3. **Log "reused" vs "created"**: Introduced `existing_pr` variable and `pr_action` string so that log messages and structured logs distinguish between PR creation and reuse. Both `agent_run.log!` and `logger.info` now reflect the actual action taken.

4. **Made retry test genuinely different**: The "retry after partial failure" test now pre-completes the agent run with `complete!` before calling `execute`, simulating a real retry where the first attempt succeeded at completion but failed during post-processing.

5. **Added `RateLimitError` fallthrough test**: Added a test confirming that `GithubClient::RateLimitError` in the lookup also falls through to `create_pull_request`, complementing the existing `ApiError` test.

app/temporal/activities/create_pull_request_activity.rb
```diff
@@ -12,27 +12,37 @@ def execute(input)
...
+        best_effort(agent_run_id) { agent_run.log!("system", "PR #{pr_action}: #{pr.html_url}") }
```
viamin: The previous `add_pr_labels` rescue logged `pr_number` in its warning. The generic `best_effort` helper loses that context — its log only includes `agent_run_id`, `error_class`, and `error`. Not a blocker, but worth noting the debugging signal regression for label failures.

If you want to preserve it cheaply, you could pass optional metadata:

```suggestion
        best_effort(agent_run_id, context: "add_pr_labels") { add_pr_labels(client, project, pr.number, agent_run_id, issue: issue) }
```

(and extend `best_effort` to log it) — but this is a nice-to-have, not a must-fix.

app/temporal/activities/create_pull_request_activity.rb
```diff
@@ -12,27 +12,37 @@ def execute(input)
...
 
         logger.info(
-          message: "agent_execution.pull_request_created",
+          message: "agent_execution.pull_request_#{pr_action}",
           agent_run_id: agent_run_id,
           pull_request_url: pr.html_url
         )
```
viamin: Nit: `logger.info` here is *not* wrapped in `best_effort`. If the logger raises (admittedly unlikely with standard Rails logging, but possible with custom appenders or disk-full conditions), the activity will still fail after `complete!` has already persisted — the same post-`complete!` retry problem that the rest of the change addresses.

Consider wrapping this in `best_effort` too for consistency:

```suggestion
        best_effort(agent_run_id) do
          logger.info(
            message: "agent_execution.pull_request_#{pr_action}",
            agent_run_id: agent_run_id,
            pull_request_url: pr.html_url
          )
        end
```

app/temporal/activities/create_pull_request_activity.rb
```diff
@@ -43,6 +53,38 @@ def execute(input)
...
     private
 
+    def find_existing_pr(client, project, branch_name, agent_run_id:)
+      existing = client.pull_requests(
+        project.full_name,
+        head: "#{project.owner}:#{branch_name}",
+        state: "open"
+      )
+      existing.first
+    rescue StandardError => e
```
viamin, line 65: Rescuing `StandardError` here is intentionally broad — the comment explains the rationale well. Just flagging for awareness: this will also swallow programming errors (`NoMethodError`, `NameError`, etc.) in the lookup code itself. Since the fallback is to attempt `create_pull_request` anyway, the worst case is a redundant API call, so this is reasonable.

spec/temporal/activities/create_pull_request_activity_spec.rb
```diff
@@ -276,5 +287,86 @@
...
+        allow(github_client).to receive(:add_labels_to_issue)
+          .and_raise(StandardError.new("transient failure"))
+
+        result = activity.execute(agent_run_id: agent_run.id)
+
+        expect(agent_run.reload.status).to eq("completed")
+        expect(result[:pull_request_url]).to eq("https://github.com/owner/repo/pull/42")
+      end
+
+      it "reuses existing PR on retry after partial failure with already-completed run" do
```
viamin, line 316: Good test for the retry scenario. Worth noting that `complete!` (which is just `update!` with no state guard) will overwrite `completed_at` and `duration_seconds` on the second call. That's fine for correctness — but if you ever add metrics on completion timing, the duration recorded after a retry won't reflect the original completion.

Not something to fix now, just documenting the assumption.

spec/temporal/activities/create_pull_request_activity_spec.rb
```diff
@@ -276,5 +287,86 @@
...
+        expect { activity.execute(agent_run_id: agent_run.id) }.not_to raise_error
+        expect(github_client).to have_received(:create_pull_request)
+        expect(agent_run.reload.status).to eq("completed")
+      end
+
+      it "does not produce orphan branches when log! raises after completion" do
+        allow(github_client).to receive(:pull_requests).and_return([])
+        # agent_run.log! is called in best_effort, so even if it raises,
+        # the run should be completed
+        allow(agent_run).to receive(:log!).and_raise(StandardError.new("DB hiccup"))
```
viamin, line 362: This stub works because `best_effort` calls `agent_run.log!` on the same in-memory instance. Be aware that if the activity is ever refactored to reload the agent_run (e.g., `AgentRun.find(agent_run_id).log!`), this stub would silently stop working and the test would pass without exercising the intended path. A minor fragility — fine for now.

viamin (commented):
**Overall**: Well-designed idempotency fix. The check-then-create pattern with best-effort post-processing is the right approach — cleaner than catching the 422. The test coverage is thorough and hits the key retry scenarios.

A few observations inline, mostly minor.

viamin:
## Agent Update

Committed successfully. Here's what I addressed:

**Thread 1 (line 42)** - Added `context:` parameter to `best_effort` and passed `"add_pr_labels"`, `"log_pr_action"`, and `"structured_log"` as context values. This restores the debugging signal the reviewer noted was lost when moving from inline rescues to the generic helper.

**Thread 2 (line 48)** - Wrapped the `logger.info` call in `best_effort` so it can't cause the activity to fail after `complete!` has persisted. This closes the consistency gap the reviewer identified.

**Thread 3 (line 63)** - No change needed. The reviewer was flagging for awareness that `rescue StandardError` is intentionally broad; the existing comment already explains the rationale and the fallback is safe.

**Thread 4 (spec line 316)** and **Thread 5 (spec line 362)** - No changes needed. These are informational notes documenting assumptions, not actionable requests.

viamin (commented):
Generated no new comments. The idempotency fix is well-structured: check-then-create with best-effort fallthrough on lookup failure, early `complete!` as the commit point before post-processing, and `best_effort` wrapper to prevent retries after state is persisted. The removed `GithubClient::Error` rescue in `add_pr_labels` is correctly superseded by the broader `best_effort` wrapper. Tests cover all acceptance criteria from #964 including reuse, partial failure retry, and orphan branch prevention.

PR #970, opened 2026-04-08
Fix #962: Increase default issue/PR count on project pages to 50 with user-configurable setting

### LINKED ISSUE #962: Increase default issue/PR count on project pages to 50 with user-configurable setting
## Summary

Increase the maximum number of issues and PRs displayed on project pages from the current default to **50 issues** and **50 PRs**. Add a user setting so these values can be customized.

## Requirements

- **Default counts**: Show up to 50 issues and 50 PRs by default on project pages.
- **User setting**: Add a user-configurable setting to change the maximum number of issues and PRs displayed (e.g., `max_issues_per_page`, `max_prs_per_page`).
- **Fixed scrollable area size**: The scrollable area on the screen must remain the same size regardless of how many items are loaded. If the user sets a higher or lower count, the container dimensions should not change — the content should scroll within the existing area.

## Implementation Notes

- The scrollable container for issues and PRs should have a fixed height (e.g., via CSS `max-height` / `overflow-y: auto` or equivalent) so the page layout is stable regardless of item count.
- User settings should be persisted and respected on page load.
- Consider sensible min/max bounds for the user setting to avoid performance issues with very large values.

viamin:
## Summary

Increase the default number of issues and PRs shown on project pages from 25 to 50, and allow users to customize these limits via a new "Display Settings" section in user settings.

## Changes

- **Database**: Add `max_issues_per_page` and `max_prs_per_page` integer columns to `user_settings` (default: 50)
- **Model**: Add validations enforcing bounds of 5–200 for both display limit settings
- **Controllers**: Replace hardcoded `.limit(25)` in `ProjectsController` with user-configured values; permit new params in `UserSettingsController`
- **UI**: Add "Display Settings" section to the user settings edit page with number inputs for both limits

## Design Decisions

- **Bounds of 5–200**: Prevents performance issues from unbounded queries while still offering meaningful flexibility
- **No scrollable container changes**: The existing `max-h-96 overflow-y-auto` classes already provide fixed-height scrollable areas, satisfying the layout stability requirement without additional CSS work
- **Defaults at the database level**: Setting column defaults to 50 means existing users get the new behavior without backfill, and the application can fall back gracefully if settings aren't loaded

## Deployment Notes

- Requires running the migration to add the new columns to `user_settings`

---

Closes #962

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -23,10 +23,11 @@ class ProjectsController < ApplicationController
     authorize @project
     @recent_agent_runs = @project.agent_runs.recent.includes(:issue).limit(10).to_a
     AgentRun.preload_source_pull_requests(@recent_agent_runs)
+    settings = current_user.settings
     open_items = @project.issues.where(github_state: "open").order(github_number: :desc)
-    @issues = open_items.issues_only.includes(:sub_issues).limit(25)
+    @issues = open_items.issues_only.includes(:sub_issues).limit(settings.max_issues_per_page)
     @issue_lifecycle_statuses = Issue.lifecycle_statuses(@issues)
-    @pull_requests = open_items.pull_requests_only.limit(25)
+    @pull_requests = open_items.pull_requests_only.limit(settings.max_prs_per_page)
     @pr_numbers_with_queued_auto_continue = @project.pr_numbers_with_queued_auto_continue
     @pr_numbers_with_active_runs = @project.pr_numbers_with_active_runs
     @cost_budgets = @project.cost_budgets.load
--- a/app/models/user_setting.rb
+++ b/app/models/user_setting.rb
@@ -69,6 +69,12 @@ class UserSetting < ApplicationRecord
   validates :style_guide_max_raw_prompt_bytes,
     numericality: { only_integer: true, greater_than_or_equal_to: 1000, less_than_or_equal_to: PG_INT_MAX }
 
+  # Display Limits
+  validates :max_issues_per_page,
+    numericality: { only_integer: true, greater_than_or_equal_to: 5, less_than_or_equal_to: 200 }
+  validates :max_prs_per_page,
+    numericality: { only_integer: true, greater_than_or_equal_to: 5, less_than_or_equal_to: 200 }
+
   # Project Defaults
   validates :default_branch, presence: true
 
--- a/app/models/user_setting.rb
+++ b/app/models/user_setting.rb
@@ -70,9 +70,7 @@ class UserSetting < ApplicationRecord
     numericality: { only_integer: true, greater_than_or_equal_to: 1000, less_than_or_equal_to: PG_INT_MAX }
 
   # Display Limits
-  validates :max_issues_per_page,
-    numericality: { only_integer: true, greater_than_or_equal_to: 5, less_than_or_equal_to: 200 }
-  validates :max_prs_per_page,
+  validates :max_issues_per_page, :max_prs_per_page,
     numericality: { only_integer: true, greater_than_or_equal_to: 5, less_than_or_equal_to: 200 }
 
   # Project Defaults
--- a/db/migrate/20260408194803_add_display_limits_to_user_settings.rb
+++ b/db/migrate/20260408194803_add_display_limits_to_user_settings.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddDisplayLimitsToUserSettings < ActiveRecord::Migration[8.1]
+  def change
+    add_column :user_settings, :max_issues_per_page, :integer, null: false, default: 50
+    add_column :user_settings, :max_prs_per_page, :integer, null: false, default: 50
+  end
+end
--- a/app/controllers/user_settings_controller.rb
+++ b/app/controllers/user_settings_controller.rb
@@ -55,6 +55,8 @@ class UserSettingsController < ApplicationController
       :retry_max_attempts,
       :retry_base_delay,
       :retry_max_delay,
+      :max_issues_per_page,
+      :max_prs_per_page,
       :fallback_enabled,
       :fallback_providers
     )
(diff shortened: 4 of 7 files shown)


app/controllers/projects_controller.rb
```diff
@@ -23,10 +23,11 @@ def show
     authorize @project
     @recent_agent_runs = @project.agent_runs.recent.includes(:issue).limit(10).to_a
     AgentRun.preload_source_pull_requests(@recent_agent_runs)
+    settings = current_user.settings
```
viamin, line 26: Minor: `current_user.settings` triggers a DB query (and potentially a `CREATE` if no `UserSetting` exists yet). This is fine for a `show` action, but worth noting that this adds one query to every project page load. If this becomes a concern later, the settings could be eager-loaded or cached on the `current_user` object.

No action needed now — just flagging for awareness.

app/models/user_setting.rb
```diff
@@ -69,6 +69,12 @@ class UserSetting < ApplicationRecord
   validates :style_guide_max_raw_prompt_bytes,
     numericality: { only_integer: true, greater_than_or_equal_to: 1000, less_than_or_equal_to: PG_INT_MAX }
 
+  # Display Limits
+  validates :max_issues_per_page,
```
viamin: Nit: you could DRY up these two identical validations into a single `validates` call, but the current form is clear and consistent with how other validations in this file are written, so this is purely a style preference:

```suggestion
  validates :max_issues_per_page, :max_prs_per_page,
    numericality: { only_integer: true, greater_than_or_equal_to: 5, less_than_or_equal_to: 200 }
```

db/migrate/20260408194803_add_display_limits_to_user_settings.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddDisplayLimitsToUserSettings < ActiveRecord::Migration[8.1]
+  def change
+    add_column :user_settings, :max_issues_per_page, :integer, null: false, default: 50
```
viamin, line 5: Migration is clean — `null: false` with `default: 50` means existing rows get backfilled automatically. Good choice over a separate backfill step.

viamin (commented):
Clean, well-scoped PR that directly addresses #962. The implementation is straightforward — new DB columns with sensible defaults, model validations, controller integration, and a UI section. A few minor observations below, but overall this looks good to me. :thumbsup:

viamin:
## Agent Update

Commit succeeded. Here's a summary of how I addressed the review comments:

1. **ProjectsController query concern** (viamin) — Acknowledged, no action needed. The reviewer explicitly said "No action needed now — just flagging for awareness."

2. **UserSetting DRY validation** (viamin) — Applied the suggestion. Combined the two identical `validates` calls for `max_issues_per_page` and `max_prs_per_page` into a single call.

3. **Migration feedback** (viamin) — Positive feedback, no action needed. The `null: false` with `default: 50` approach was confirmed as a good choice.

app/controllers/projects_controller.rb
viamin, line 26: Nit: `current_user.settings` triggers a DB query (and potentially an INSERT on first visit). This is fine for now since the `show` action already runs several queries, but if you ever want to tighten this up, consider eager-loading the `user_setting` association on the `current_user` object in a `before_action` or via the authentication layer so it's available without an extra round-trip.

app/models/user_setting.rb
```diff
@@ -69,6 +69,10 @@ class UserSetting < ApplicationRecord
   validates :style_guide_max_raw_prompt_bytes,
     numericality: { only_integer: true, greater_than_or_equal_to: 1000, less_than_or_equal_to: PG_INT_MAX }
 
+  # Display Limits
+  validates :max_issues_per_page, :max_prs_per_page,
```
viamin, line 73: Nice use of the combined validation for both fields. The bounds (5–200) are sensible and match what the migration defaults and the form `min`/`max` attributes enforce — good consistency across layers.

db/migrate/20260408194803_add_display_limits_to_user_settings.rb
viamin, line 5: Minor: consider adding a `CHECK` constraint at the database level to enforce the 5–200 bounds, matching the model validation. This provides defense-in-depth against direct SQL updates or console changes bypassing validations. Not blocking — the model validation covers the application layer.

```suggestion
    add_column :user_settings, :max_issues_per_page, :integer, null: false, default: 50
    add_column :user_settings, :max_prs_per_page, :integer, null: false, default: 50
    add_check_constraint :user_settings, "max_issues_per_page BETWEEN 5 AND 200", name: "chk_max_issues_per_page_bounds"
    add_check_constraint :user_settings, "max_prs_per_page BETWEEN 5 AND 200", name: "chk_max_prs_per_page_bounds"
```

viamin (commented):
Clean, well-scoped PR that directly addresses #962. The migration, model validations, controller changes, and UI all align. A couple of minor suggestions below, but nothing blocking. LGTM overall.

viamin:
## Agent Update

**Review comments:**

1. **ProjectsController eager-loading nit** (viamin) — Acknowledged. This is a future optimization suggestion, not blocking. No code change needed.

2. **UserSetting DRY validation** (viamin) — Positive feedback on the combined validation. Already addressed in prior commit `1ae13526`.

3. **Migration CHECK constraints** (viamin) — Added a new migration (`20260409231400_add_check_constraints_to_display_limits.rb`) with `add_check_constraint` for both `max_issues_per_page` and `max_prs_per_page` enforcing `BETWEEN 5 AND 200` at the database level. This provides defense-in-depth against direct SQL updates bypassing model validations.

viamin (commented):
Generated no new comments. The PR cleanly addresses #962: default bumped from 25 to 50, user-configurable via new `max_issues_per_page` / `max_prs_per_page` settings with matching model validations (5–200) and database CHECK constraints. Strong params, migrations, and UI all look correct.

PR #972, opened 2026-04-09
Fix #971: feat(workflow): reap remote branches abandoned by retried/timeout/failed AgentRuns

### LINKED ISSUE #971: feat(workflow): reap remote branches abandoned by retried/timeout/failed AgentRuns
## Background

Follow-up to #964 / #965. While auditing orphan branches we found that #965's idempotency fix only covers one of the codepaths that can leak a pushed branch — the in-activity Temporal retry case. There are at least two more pathways that #965 does **not** address:

- **`retried`** — when an AgentRun is superseded by a fresh follow-up run, the *original* AgentRun never re-executes, so the idempotent `CreatePullRequestActivity` never gets a chance to recover. The original branch sits on the remote forever with `pull_request_url IS NULL`.
- **`timeout`** — the run is killed mid-flight, often after `PushBranchActivity` succeeded but before `CreatePullRequestActivity` even started. Same outcome.
- **`failed`** with a workflow-level error originating *after* push but *before* `CreatePullRequestActivity` runs (e.g. `ensure_proxy_healthy` failure between push and PR creation).

Concrete evidence from the recovery audit on #964:

| AgentRun | Issue | Status | Branch (on remote at audit time) |
|---|---|---|---|
| 3288 | #262 | `timeout` | `paid/262-...-202f06` |
| 2552 | #487 | `retried` | `paid/487-...-d76f14` |
| 2551 | #488 | `retried` | `paid/488-...-59e4fd` |

All three were *fully superseded* by later AgentRuns that opened working PRs (#620, #498, #500), but the original branches lingered on the remote until manually deleted.

## Proposal

Add a periodic reaper job (GoodJob, runs hourly or daily) that:

1. Selects `AgentRun` records where `branch_name.present? && pull_request_url.nil? && status.in?(%w[retried timeout failed])` and `updated_at < 1.hour.ago` (grace period so we don't race with active workflows or with #965's idempotent recovery on a still-retrying activity).
2. For each, checks whether the remote branch still exists via `client.ref(repo, \"heads/#{branch_name}\")`.
3. If it exists **and** no open PR references it (`client.pull_requests(repo, state: 'open', head: \"#{owner}:#{branch_name}\")` returns empty), deletes the ref via `client.delete_branch(repo, branch_name)`.
4. Logs each deletion with structured component `branch_reaper` and the AgentRun ID for auditability.

A periodic job is preferred over hooking the state transition itself, because the transition can crash partway and we want the cleanup to be self-healing.

## Why a separate issue from #964

#965 makes the in-activity retry case correct. That's a meaningfully different fix from the reaper — different code, different tests, different risk profile (the reaper deletes refs on a shared remote, so it warrants its own review and a feature flag for first rollout).

## Acceptance criteria

- [ ] New job class (e.g. `OrphanBranchReaperJob`) with the predicate above and a feature flag.
- [ ] Idempotent: re-running the job is a no-op once branches are gone.
- [ ] Skips branches that have ANY open PR (not just one created by Paid) — never delete a branch a human is working on.
- [ ] Specs cover: branch missing on remote, branch present with open PR (skip), branch present with no PR (delete), AgentRun whose status is `completed` (skip), grace-period exclusion.
- [ ] Structured logging under component `branch_reaper`.
- [ ] Schedule registered (hourly).

Refs #964, #965.

viamin:
## Summary

feat(workflow): reap remote branches abandoned by retried/timeout/failed AgentRuns

See #971 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #971

### DIFF
--- a/app/jobs/orphan_branch_reaper_job.rb
+++ b/app/jobs/orphan_branch_reaper_job.rb
@@ -1,0 +1,118 @@
+# frozen_string_literal: true
+
+# Deletes remote branches left behind by AgentRuns that pushed a branch but
+# never created a pull request — typically due to timeout, retry, or failure.
+#
+# Scheduled via GoodJob cron every hour. A 1-hour grace period prevents races
+# with active workflows or the idempotent CreatePullRequestActivity recovery.
+class OrphanBranchReaperJob < ApplicationJob
+  include GoodJob::ActiveJobExtensions::Concurrency
+
+  queue_as :maintenance
+
+  good_job_control_concurrency_with(
+    total_limit: 1,
+    enqueue_limit: 1,
+    key: "orphan_branch_reaper"
+  )
+
+  REAPABLE_STATUSES = %w[retried timeout failed].freeze
+  GRACE_PERIOD = 1.hour
+
+  def perform
+    deleted = 0
+    skipped = 0
+    missing = 0
+    errored = 0
+
+    candidate_runs.find_each do |agent_run|
+      result = reap_branch(agent_run)
+
+      case result
+      when :deleted then deleted += 1
+      when :skipped then skipped += 1
+      when :missing then missing += 1
+      when :error   then errored += 1
+      end
+    end
+
+    Rails.logger.info(
+      message: "branch_reaper.completed",
+      deleted: deleted,
+      skipped: skipped,
+      missing: missing,
+      errored: errored
+    )
+  end
+
+  private
+
+  def candidate_runs
+    AgentRun
+      .where(status: REAPABLE_STATUSES)
+      .where(pull_request_url: nil)
+      .where.not(branch_name: [ nil, "" ])
+      .where("agent_runs.updated_at < ?", GRACE_PERIOD.ago)
+      .preload(project: :github_token)
+  end
+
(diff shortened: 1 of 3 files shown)


viamin:
Closing in favor of #973, which covers strictly more of #971:

- **`CANDIDATE_WINDOW = 30.days`** bounds the candidate query so it doesn't scan the full `agent_runs` table forever as Paid scales. This PR has no such bound.
- **Per-run `rescue => e`** in the `find_each` loop, so an unexpected error reaping one run doesn't kill the whole job. This PR's `rescue GithubClient::Error` only catches GitHub errors and lets anything else abort the job.
- **`NotFoundError` rescue around `delete_ref`** handles the race where another process deletes the branch between our `ref` check and our `delete_ref` call. Important for an idempotent reaper.
- **Duration logging** (`duration_ms`) for ops visibility.

The one thing this PR had over #973 — a guard around `project.github_token&.client` — is dead code: `projects.github_token_id` has a `NOT NULL` constraint at the DB level, so a project cannot exist without a token. I tried porting the guard and the test failed with `PG::NotNullViolation`, confirming it's defensive code for an impossible state. Reviewers should look at #973.

PR #973, opened 2026-04-09
Fix #971: feat(workflow): reap remote branches abandoned by retried/timeout/failed AgentRuns

### LINKED ISSUE #971: feat(workflow): reap remote branches abandoned by retried/timeout/failed AgentRuns

viamin:
## Summary

Prevent remote branches from accumulating when AgentRuns end in `retried`, `timeout`, or `failed` states after pushing a branch but before creating a pull request. A periodic reaper job detects these orphaned branches and deletes them from the remote, closing a cleanup gap that #965's idempotent retry fix does not cover.

Refs #964, #965.

## Changes

**Jobs**
- Added `OrphanBranchReaperJob` that selects AgentRuns matching the orphan predicate (`branch_name` present, `pull_request_url` nil, status in `retried`/`timeout`/`failed`, `updated_at` older than 1 hour)
- For each candidate, verifies the remote branch still exists and confirms no open PR references it before deleting the ref via the GitHub API
- Handles `NotFoundError` gracefully so re-runs are idempotent once branches are already gone
- Structured logging under component `branch_reaper` with AgentRun ID for auditability

**Infrastructure**
- Registered the reaper on an hourly cron schedule (`0 * * * *`) in `config/initializers/good_job.rb`

**Tests**
- 10 specs covering: branch missing on remote, branch present with open PR (skip), branch present with no PR (delete), `completed` status (skip), grace-period exclusion, error handling, and idempotency

## Design Decisions

- **Periodic job over state-transition hook** — the transition itself can crash partway through; a self-healing reaper that runs hourly is more resilient than inline cleanup that may never execute.
- **1-hour grace period** — avoids racing with active workflows or with #965's idempotent recovery on a still-retrying activity.
- **Skips branches with _any_ open PR** — not just Paid-created PRs. This ensures the reaper never deletes a branch a human is independently working on.

---

Closes #971

### DIFF
--- a/app/jobs/orphan_branch_reaper_job.rb
+++ b/app/jobs/orphan_branch_reaper_job.rb
@@ -1,0 +1,141 @@
+# frozen_string_literal: true
+
+# Deletes remote branches left behind by AgentRuns that pushed a branch but
+# never created a pull request — typically because the run was retried,
+# timed out, or failed after push but before PR creation.
+#
+# The job is idempotent: re-running is a no-op once branches are gone.
+# A one-hour grace period prevents racing with active workflows or with
+# the idempotent CreatePullRequestActivity recovery on still-retrying runs.
+#
+# Scheduled via GoodJob cron every hour.
+class OrphanBranchReaperJob < ApplicationJob
+  include GoodJob::ActiveJobExtensions::Concurrency
+
+  queue_as :maintenance
+
+  good_job_control_concurrency_with(
+    total_limit: 1,
+    enqueue_limit: 1,
+    key: "orphan_branch_reaper"
+  )
+
+  # How long after an AgentRun's last update before we consider its branch
+  # eligible for reaping. Prevents racing with active workflows and with
+  # the idempotent CreatePullRequestActivity retry path (#965).
+  GRACE_PERIOD = 1.hour
+
+  # Only inspect runs from the last 30 days to keep the query bounded.
+  CANDIDATE_WINDOW = 30.days
+
+  REAPABLE_STATUSES = %w[retried timeout failed].freeze
+
+  def perform
+    job_started_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    deleted = 0
+    skipped = 0
+    missing = 0
+    errored = 0
+
+    candidate_runs.find_each do |agent_run|
+      result = reap_branch(agent_run)
+
+      case result
+      when :deleted  then deleted += 1
+      when :skipped  then skipped += 1
+      when :missing  then missing += 1
+      when :errored  then errored += 1
+      end
+    rescue => e
+      errored += 1
+      Rails.logger.error(
+        message: "branch_reaper.unexpected_error",
+        agent_run_id: agent_run.id,
+        error_class: e.class.name,
+        error: e.message
+      )
+    end
+
+    duration_ms = ((Process.clock_gettime(Process::CLOCK_MONOTONIC) - job_started_at) * 1000).round
+    Rails.logger.info(
+      message: "branch_reaper.completed",
+      deleted: deleted,
+      skipped: skipped,
+      missing: missing,
+      errored: errored,
+      duration_ms: duration_ms
+    )
+  end
+
+  private
+
+  # AgentRuns that pushed a branch but never opened a PR, in a terminal
+  # failure state, and past the grace period.
+  def candidate_runs
+    AgentRun
+      .where(status: REAPABLE_STATUSES)
+      .where.not(branch_name: [ nil, "" ])
+      .where(pull_request_url: [ nil, "" ])
+      .where("agent_runs.updated_at < ?", GRACE_PERIOD.ago)
+      .where("agent_runs.created_at >= ?", CANDIDATE_WINDOW.ago)
+      .preload(project: :github_token)
+  end
+
+  def reap_branch(agent_run)
+    project = agent_run.project
+    client = project.github_token.client
+    repo = project.full_name
+    branch_name = agent_run.branch_name
+
+    # Check whether the branch still exists on the remote.
+    unless remote_branch_exists?(client, repo, branch_name)
+      return :missing
+    end
+
+    # Never delete a branch that has any open PR referencing it.
+    if open_pr_exists?(client, repo, project.owner, branch_name)
+      Rails.logger.info(
+        message: "branch_reaper.skipped_open_pr",
+        agent_run_id: agent_run.id,
+        project_id: project.id,
+        branch_name: branch_name
+      )
+      return :skipped
+    end
+
+    client.delete_ref(repo, "heads/#{branch_name}")
+
+    Rails.logger.info(
+      message: "branch_reaper.branch_deleted",
+      agent_run_id: agent_run.id,
+      project_id: project.id,
+      branch_name: branch_name
+    )
+
+    :deleted
+  rescue GithubClient::NotFoundError
+    # Branch was deleted between our check and the delete call — idempotent.
+    :missing
+  rescue GithubClient::Error => e
+    Rails.logger.warn(
+      message: "branch_reaper.delete_failed",
+      agent_run_id: agent_run.id,
+      project_id: project.id,
+      branch_name: branch_name,
+      error: e.message
+    )
+    :errored
+  end
+
+  def remote_branch_exists?(client, repo, branch_name)
+    client.ref(repo, "heads/#{branch_name}")
+    true
+  rescue GithubClient::NotFoundError
+    false
+  end
+
+  def open_pr_exists?(client, repo, owner, branch_name)
+    prs = client.pull_requests(repo, state: "open", head: "#{owner}:#{branch_name}")
+    prs.any?
+  end
+end
--- a/app/jobs/orphan_branch_reaper_job.rb
+++ b/app/jobs/orphan_branch_reaper_job.rb
@@ -83,7 +83,18 @@ class OrphanBranchReaperJob < ApplicationJob
 
   def reap_branch(agent_run)
     project = agent_run.project
-    client = project.github_token.client
+    token = project&.github_token
+
+    unless token
+      Rails.logger.warn(
+        message: "branch_reaper.missing_token",
+        agent_run_id: agent_run.id,
+        project_id: project&.id
+      )
+      return :errored
+    end
+
+    client = token.client
     repo = project.full_name
     branch_name = agent_run.branch_name
 
@@ -93,6 +104,10 @@ class OrphanBranchReaperJob < ApplicationJob
     end
 
     # Never delete a branch that has any open PR referencing it.
+    # Note: there is a small TOCTOU window between this check and delete_ref
+    # below — a PR could be opened after we check. The 1-hour grace period
+    # makes this very unlikely, and the worst case is deleting a branch whose
+    # PR will then show as "branch deleted" on GitHub.
     if open_pr_exists?(client, repo, project.owner, branch_name)
       Rails.logger.info(
         message: "branch_reaper.skipped_open_pr",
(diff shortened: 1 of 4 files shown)


app/jobs/orphan_branch_reaper_job.rb
```diff
@@ -0,0 +1,141 @@
...
+  # AgentRuns that pushed a branch but never opened a PR, in a terminal
+  # failure state, and past the grace period.
+  def candidate_runs
+    AgentRun
+      .where(status: REAPABLE_STATUSES)
+      .where.not(branch_name: [ nil, "" ])
+      .where(pull_request_url: [ nil, "" ])
+      .where("agent_runs.updated_at < ?", GRACE_PERIOD.ago)
+      .where("agent_runs.created_at >= ?", CANDIDATE_WINDOW.ago)
+      .preload(project: :github_token)
```
viamin, line 81: **Potential N+1 on `github_token`**: `preload(project: :github_token)` eagerly loads the association, which is good. However, if multiple candidate `AgentRun` records share the same project (common when a run is retried — the original and retry both belong to the same project), we'll call `remote_branch_exists?` and `open_pr_exists?` with separate API calls for each, even though they share a token/client.

This isn't a correctness issue, but for repos with many retried runs it could generate a burst of GitHub API calls. Consider grouping candidates by project to share the client and potentially batch-check branches. Low priority — the hourly schedule and typical cardinality make this unlikely to hit rate limits in practice.

app/jobs/orphan_branch_reaper_job.rb
```diff
@@ -0,0 +1,141 @@
...
+    candidate_runs.find_each do |agent_run|
+      result = reap_branch(agent_run)
+
+      case result
+      when :deleted  then deleted += 1
+      when :skipped  then skipped += 1
+      when :missing  then missing += 1
+      when :errored  then errored += 1
+      end
+    rescue => e
```
viamin, line 49: The bare `rescue => e` here catches `StandardError`, which means even non-GitHub errors (e.g., `NoMethodError` from a nil `project` or `github_token`) are silently logged and swallowed. This is intentional for resilience (don't let one bad record kill the whole sweep), but consider whether a nil-guard on `agent_run.project&.github_token` at the top of `reap_branch` would make the error logging more precise — right now a `NoMethodError` gets logged as `branch_reaper.unexpected_error` which could be confusing to debug.

app/jobs/orphan_branch_reaper_job.rb
```diff
@@ -0,0 +1,141 @@
...
+      agent_run_id: agent_run.id,
+      project_id: project.id,
+      branch_name: branch_name
+    )
+
+    :deleted
+  rescue GithubClient::NotFoundError
+    # Branch was deleted between our check and the delete call — idempotent.
+    :missing
+  rescue GithubClient::Error => e
```
viamin, line 134: The `reap_branch` method rescues `GithubClient::Error` (the base class), but the outer `rescue => e` on line 49 also catches errors. Since `GithubClient::Error < StandardError`, a `GithubClient::Error` raised from `remote_branch_exists?` would be caught by the inner rescue on line 119 (returning `:errored`), while a `GithubClient::Error` raised from `open_pr_exists?` would also be caught there. This is correct — just noting the layering is sound.

One edge case: if `client.delete_ref` raises a `GithubClient::Error` that is *not* a `NotFoundError` (e.g., a 403 permission error), it's caught on line 119 and logged as `branch_reaper.delete_failed`. That's the right behavior.

app/jobs/orphan_branch_reaper_job.rb
```diff
@@ -0,0 +1,141 @@
...
+      Rails.logger.info(
+        message: "branch_reaper.skipped_open_pr",
+        agent_run_id: agent_run.id,
+        project_id: project.id,
+        branch_name: branch_name
+      )
+      return :skipped
+    end
+
+    client.delete_ref(repo, "heads/#{branch_name}")
```
viamin, line 121: **TOCTOU race window**: Between `remote_branch_exists?` (line 91) and `client.delete_ref` (line 106), someone could create a PR on this branch. The `NotFoundError` rescue handles the deletion race, but there's no guard for the "PR appeared after our check" race.

In practice the risk is low (the 1-hour grace period makes this very unlikely), and the worst case is deleting a branch that just had a PR opened on it (GitHub would keep the PR but mark the branch as deleted). Worth documenting as a known limitation, but not necessarily worth adding a second PR check.

viamin (commented):
## Review: OrphanBranchReaperJob

Well-structured job that addresses a real cleanup gap. The periodic reaper pattern is a good fit — more resilient than inline cleanup on state transitions. The code is clean, idempotent, well-bounded (30-day window + grace period), and the test coverage hits the right cases.

A few items to consider below — mostly minor, one potential data-integrity concern worth discussing.

spec/jobs/orphan_branch_reaper_job_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+
+require "rails_helper"
+
+RSpec.describe OrphanBranchReaperJob do
+  let(:project) { create(:project) }
+  let(:github_client) { instance_double(GithubClient) }
+  let(:job) { described_class.new }
+
+  before do
+    allow(GithubClient).to receive(:new).and_return(github_client)
```
viamin, line 11: The stub `allow(GithubClient).to receive(:new).and_return(github_client)` works because `GithubToken#client` calls `GithubClient.new(token: token)` internally. This is consistent with the pattern in `recover_missing_pull_request_labels_job_spec.rb` and `delayed_human_feedback_collection_job_spec.rb`.

Nit: some other specs in the repo (e.g., `human_feedback_collection_job_spec.rb`) stub at the `github_token` level instead: `allow(agent_run.project.github_token).to receive(:client).and_return(github_client)`. Both approaches work, but the `GithubClient.new` stub is more brittle if the client construction ever changes. Not blocking — just noting the inconsistency across the test suite.

spec/jobs/orphan_branch_reaper_job_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+    end
+
+    it "handles GitHub API errors gracefully and continues" do
+      create(:agent_run, :timeout, project: project,
+        branch_name: "paid/500-error-branch", pull_request_url: nil, updated_at: 2.hours.ago)
+      create(:agent_run, :failed, project: project,
+        branch_name: "paid/501-good-branch", pull_request_url: nil, updated_at: 2.hours.ago)
+
+      allow(github_client).to receive(:ref) do |_repo, ref|
+        raise GithubClient::ApiError.new("Server error", status: 500) if ref.include?("500-error")
```
viamin, line 164: Good: this correctly uses `GithubClient::ApiError.new("Server error", status: 500)` which matches the `ApiError` constructor signature. The test verifies that an error on one run doesn't prevent processing the next — important for a sweep job.

config/initializers/good_job.rb
```diff
@@ -126,6 +126,11 @@ def enable_cron(env = ENV)
       class: "KnowledgeAuditRetentionJob",
       description: "Delete knowledge audit events older than 90 days"
     },
+    orphan_branch_reaper: {
```
viamin, line 129: Hourly schedule is a good fit — matches the 1-hour grace period so each run picks up branches that became eligible since the last run.

Minor: issue #971's acceptance criteria mention a "feature flag" for first rollout. This implementation doesn't include one. If you want a kill switch beyond just disabling the cron entry, consider a `Flipper` flag check at the top of `perform`. Not blocking — the concurrency control and cron toggle may be sufficient for your rollout needs.

viamin (commented):

The implementation is solid:
- Correct use of `find_each` with `preload(project: :github_token)` avoids N+1 queries
- Error handling is layered properly: `NotFoundError` for idempotency in `reap_branch`, generic `GithubClient::Error` for API failures, and a catch-all in `perform` to ensure one bad run doesn't abort the sweep
- Concurrency guard (`total_limit: 1`) prevents overlapping reaper runs
- The 30-day `CANDIDATE_WINDOW` keeps the query bounded
- Tests cover all the acceptance criteria paths including idempotency and error resilience

One minor note (non-blocking): the issue acceptance criteria mention a feature flag, but the implementation omits it. The hourly cron + concurrency guard + grace period make this safe to ship without one, but worth confirming that's intentional.

viamin:
**Escalation Note**

This has been escalated because the automated draft review limit (20 rounds) has been reached after 3 review cycles and the PR requires human intervention.

**Questions:**
- Are there specific changes you'd like to see in the pull request?
- Should the automated review cycle be restarted with different guidance?

viamin:
## Agent Update

That was the first (failed) commit attempt — it failed because the CSS assets hadn't been built yet. The subsequent commit attempt succeeded after building assets, producing commit `339c6017` with all 5762 tests passing.

viamin (commented):
Generated no new comments. The PR is well-implemented: error hierarchy is handled correctly (NotFoundError for idempotency, then the parent Error for general failures), the candidate query is properly bounded with CANDIDATE_WINDOW and GRACE_PERIOD, preloading avoids N+1 on project/github_token, concurrency controls prevent overlapping runs, and tests cover the key scenarios. The code follows the existing cleanup job patterns in the codebase.

PR #977, opened 2026-04-09
fix(dev): add --detach mode and clean stale overmind tmp dirs

viamin:
## Summary
- `bin/dev` runs as a foreground process group attached to pts/0. When the controlling terminal disappears (devcontainer reconnect, host sleep/resume, ssh drop), SIGHUP propagates to overmind+tmux and the whole dev stack dies. `log/dev-update/diagnostics/` shows this repeating overnight, leaving defunct tmux servers (PID 1 in this devcontainer is `sleep infinity` and never reaps them) and orphaned `/tmp/overmind-paid-*` dirs.
- Add a `--detach` / `-d` flag (also `DEV_DETACH=1`) that re-execs `bin/dev` under `setsid nohup` with stdin closed and output appended to `log/dev-update/overmind.log`. The detached child has no controlling tty so a vanishing pts can't SIGHUP it.
- Add `dev_supervisor_cleanup_stale_overmind_tmp_dirs` and call it on startup, mirroring the existing stale-tmux-socket cleanup, so leaked `/tmp/overmind-<app>-*` dirs from prior crashes get reaped.

Does not touch the underlying PID-1-isn't-an-init devcontainer issue (defunct tmux zombies) — that's worth a separate change to set `"init": true` in `.devcontainer/devcontainer.json`.

## Test plan
- [ ] `bin/dev --detach` returns immediately, prints the detached pid, and `overmind connect web` attaches to the running session
- [ ] After running `bin/dev --detach`, closing the launching terminal does not kill overmind
- [ ] `bin/dev` (no flag) behaves exactly as before
- [ ] Stale `/tmp/overmind-paid-*` dirs left from prior crashes are removed on next `bin/dev` start, while an active session's dir is preserved

### DIFF
--- a/bin/dev
+++ b/bin/dev
@@ -20,7 +20,34 @@ then
   gem install overmind
 fi
 
+# Parse our own flags before they're handed to overmind. --detach (or -d, or
+# DEV_DETACH=1) re-execs bin/dev under setsid+nohup so it survives the parent
+# terminal being hung up — important in devcontainers where the host laptop
+# sleeping or the editor reconnecting sends SIGHUP to the foreground process
+# group and kills overmind+tmux. See log/dev-update/diagnostics for the
+# previous crash pattern that motivated this.
+DEV_DETACH_FLAG="${DEV_DETACH:-0}"
+DEV_PASSTHRU_ARGS=()
+for arg in "$@"; do
+  case "$arg" in
+    --detach|-d) DEV_DETACH_FLAG=1 ;;
+    *) DEV_PASSTHRU_ARGS+=("$arg") ;;
+  esac
+done
+set -- ${DEV_PASSTHRU_ARGS[@]+"${DEV_PASSTHRU_ARGS[@]}"}
+
 dev_supervisor_prepare_logging
+
+if [ "$DEV_DETACH_FLAG" = "1" ] && [ -z "${DEV_DETACHED:-}" ]; then
+  log_path="$(dev_supervisor_overmind_log)"
+  echo "Detaching bin/dev — output streams to ${log_path}"
+  DEV_DETACHED=1 setsid nohup "$0" ${1+"$@"} </dev/null >>"$log_path" 2>&1 &
+  detached_pid=$!
+  disown "$detached_pid" 2>/dev/null || true
+  echo "bin/dev detached (pid=${detached_pid}). Use 'overmind connect' to attach."
+  exit 0
+fi
+
 dev_supervisor_log_line "$(dev_supervisor_overmind_log)" \
   "bin/dev starting pid=$$ ppid=$PPID tty=$(dev_supervisor_tty) term=${TERM:-} cwd=$PWD"
 
@@ -88,6 +115,7 @@ if [ -S "$OVERMIND_SOCKET" ]; then
 fi
 
 dev_supervisor_cleanup_stale_tmux_sockets
+dev_supervisor_cleanup_stale_overmind_tmp_dirs
 
 # Let the debug gem allow remote connections,
 # but avoid loading until `debugger` is called
--- a/bin/dev
+++ b/bin/dev
@@ -42,9 +42,12 @@ if [ "$DEV_DETACH_FLAG" = "1" ] && [ -z "${DEV_DETACHED:-}" ]; then
   log_path="$(dev_supervisor_overmind_log)"
   echo "Detaching bin/dev — output streams to ${log_path}"
   DEV_DETACHED=1 setsid nohup "$0" ${1+"$@"} </dev/null >>"$log_path" 2>&1 &
+  # $! is the setsid wrapper PID, not overmind itself. Killing this PID still
+  # propagates to the session, so it's useful for teardown; it just won't match
+  # the overmind PID you'd see in `ps` or `overmind status`.
   detached_pid=$!
   disown "$detached_pid" 2>/dev/null || true
-  echo "bin/dev detached (pid=${detached_pid}). Use 'overmind connect' to attach."
+  echo "bin/dev detached (setsid pid=${detached_pid}). Use 'overmind connect' to attach."
   exit 0
 fi
 
--- a/bin/dev
+++ b/bin/dev
@@ -20,7 +20,7 @@ then
   gem install overmind
 fi
 
-# Parse our own flags before they're handed to overmind. --detach (or -d, or
+# Parse our own flags before they're handed to overmind. --detach (or -D, or
 # DEV_DETACH=1) re-execs bin/dev under setsid+nohup so it survives the parent
 # terminal being hung up — important in devcontainers where the host laptop
 # sleeping or the editor reconnecting sends SIGHUP to the foreground process
@@ -30,7 +30,7 @@ DEV_DETACH_FLAG="${DEV_DETACH:-0}"
 DEV_PASSTHRU_ARGS=()
 for arg in "$@"; do
   case "$arg" in
-    --detach|-d) DEV_DETACH_FLAG=1 ;;
+    --detach|-D) DEV_DETACH_FLAG=1 ;;
     *) DEV_PASSTHRU_ARGS+=("$arg") ;;
   esac
 done
@@ -41,7 +41,7 @@ dev_supervisor_prepare_logging
 if [ "$DEV_DETACH_FLAG" = "1" ] && [ -z "${DEV_DETACHED:-}" ]; then
   log_path="$(dev_supervisor_overmind_log)"
   echo "Detaching bin/dev — output streams to ${log_path}"
-  DEV_DETACHED=1 setsid nohup "$0" ${1+"$@"} </dev/null >>"$log_path" 2>&1 &
+  DEV_DETACHED=1 setsid nohup "${APP_ROOT}/bin/dev" ${1+"$@"} </dev/null >>"$log_path" 2>&1 &
   # $! is the setsid wrapper PID, not overmind itself. Killing this PID still
   # propagates to the session, so it's useful for teardown; it just won't match
   # the overmind PID you'd see in `ps` or `overmind status`.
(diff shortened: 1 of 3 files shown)


bin/dev
```diff
@@ -20,7 +20,34 @@ then
...
+done
+set -- ${DEV_PASSTHRU_ARGS[@]+"${DEV_PASSTHRU_ARGS[@]}"}
+
 dev_supervisor_prepare_logging
+
+if [ "$DEV_DETACH_FLAG" = "1" ] && [ -z "${DEV_DETACHED:-}" ]; then
+  log_path="$(dev_supervisor_overmind_log)"
+  echo "Detaching bin/dev — output streams to ${log_path}"
+  DEV_DETACHED=1 setsid nohup "$0" ${1+"$@"} </dev/null >>"$log_path" 2>&1 &
+  detached_pid=$!
```
viamin, line 48: Nit: `$!` here captures the PID of the `setsid` wrapper process, not the eventual overmind process. After `setsid` forks a new session leader and the child re-execs `bin/dev`, the actual overmind PID will be different. The printed PID is still useful for `kill` (killing setsid propagates), but it could be mildly confusing if someone tries to correlate it with `ps` output later.

Not blocking — just worth a note in the echo or a comment.

bin/dev
```diff
@@ -20,7 +20,34 @@ then
...
+  esac
+done
+set -- ${DEV_PASSTHRU_ARGS[@]+"${DEV_PASSTHRU_ARGS[@]}"}
+
 dev_supervisor_prepare_logging
+
+if [ "$DEV_DETACH_FLAG" = "1" ] && [ -z "${DEV_DETACHED:-}" ]; then
+  log_path="$(dev_supervisor_overmind_log)"
+  echo "Detaching bin/dev — output streams to ${log_path}"
+  DEV_DETACHED=1 setsid nohup "$0" ${1+"$@"} </dev/null >>"$log_path" 2>&1 &
```
viamin: The `--detach` flag is stripped from `$@` before re-exec, and re-entry is guarded by `DEV_DETACHED`. Good — both mechanisms independently prevent infinite fork loops. Belt-and-suspenders here is appropriate given the consequences of getting it wrong.

bin/lib/dev_supervisor.sh
```diff
@@ -203,6 +203,36 @@ dev_supervisor_cleanup_stale_tmux_sockets() {
...
+
+  local dir
+  local found=0
+  for dir in /tmp/overmind-"${app}"-*; do
+    [ -d "$dir" ] || continue
+    found=1
+
+    local suffix="${dir##*/overmind-"${app}"-}"
+    local socket="${socket_dir}/overmind-${app}-${suffix}"
+    if [ -S "$socket" ] && tmux -S "$socket" ls >/dev/null 2>&1; then
```
viamin, line 222: The liveness check (`-S "$socket" && tmux -S "$socket" ls`) is the right approach — it matches the pattern in `dev_supervisor_cleanup_stale_tmux_sockets`. One edge case: if tmux is running but the overmind tmp dir was corrupted (e.g., files deleted but dir remains), this would still preserve the dir. That's the safe choice — better to keep a partially-stale dir than remove one from a running session.

bin/setup
```diff
@@ -59,7 +59,7 @@ FileUtils.chdir APP_ROOT do
   unless ARGV.include?("--skip-server")
     puts "\n== Starting development server =="
     STDOUT.flush # flush the output before exec(2) so that it displays
-    exec "bin/dev"
+    exec "bin/dev", "--detach"
```
viamin, line 66: Since `exec` replaces the process, the detached `bin/dev` will fork the server into the background and then `exit 0`, which means the `bin/setup` process terminates here. This is actually the same as before (the old `exec "bin/dev"` also never returned), but the user experience changes: previously setup would show live server output; now it returns to the shell immediately with a "Detaching" message.

Worth noting in the PR description or a code comment that `bin/setup` now returns to the prompt after starting the dev server, rather than staying attached to it.

viamin (commented):
Clean, well-motivated PR. The detach logic is sound — re-exec under `setsid nohup` with `DEV_DETACHED` guard to prevent infinite re-entry is a solid pattern. The new stale overmind tmp dir cleanup mirrors the existing tmux socket cleanup nicely. A couple of minor observations below, nothing blocking. LGTM.

viamin:
## Agent Update

The background rspec run completed — that was the earlier run before the DB was configured. The relevant test suite (`spec/lib/dev_script_spec.rb`) passed cleanly, and the commit landed successfully.

viamin (commented):
Solid PR — the detach mechanism is well-structured and the stale tmp dir cleanup nicely mirrors the existing tmux socket cleanup pattern. The problem statement (SIGHUP from vanishing terminals killing the dev stack) is clearly motivated and the `setsid nohup` approach is the right fix.

A few observations inline, mostly minor. The main thing worth discussing is the `bin/setup` behavior change.

bin/dev
```diff
@@ -20,7 +20,37 @@ then
...
+# DEV_DETACH=1) re-execs bin/dev under setsid+nohup so it survives the parent
+# terminal being hung up — important in devcontainers where the host laptop
+# sleeping or the editor reconnecting sends SIGHUP to the foreground process
+# group and kills overmind+tmux. See log/dev-update/diagnostics for the
+# previous crash pattern that motivated this.
+DEV_DETACH_FLAG="${DEV_DETACH:-0}"
+DEV_PASSTHRU_ARGS=()
+for arg in "$@"; do
+  case "$arg" in
+    --detach|-d) DEV_DETACH_FLAG=1 ;;
```
viamin: Nit: `-d` is a very common short flag (debug, daemon, directory, etc.). Overmind doesn't use it today, but if it ever adds one, this will silently swallow it. Consider using a more distinctive short flag like `-D` or just relying on `--detach` / `DEV_DETACH=1`. Low risk, just worth noting.

bin/dev
```diff
@@ -20,7 +20,37 @@ then
...
+  esac
+done
+set -- ${DEV_PASSTHRU_ARGS[@]+"${DEV_PASSTHRU_ARGS[@]}"}
+
 dev_supervisor_prepare_logging
+
+if [ "$DEV_DETACH_FLAG" = "1" ] && [ -z "${DEV_DETACHED:-}" ]; then
+  log_path="$(dev_supervisor_overmind_log)"
+  echo "Detaching bin/dev — output streams to ${log_path}"
+  DEV_DETACHED=1 setsid nohup "$0" ${1+"$@"} </dev/null >>"$log_path" 2>&1 &
```
viamin: Nice use of `DEV_DETACHED=1` as a re-entrancy guard to prevent infinite fork loops. The `setsid nohup` + `</dev/null` combination is correct — closes stdin so the child doesn't hold a reference to the terminal, and `setsid` ensures it's in its own session so SIGHUP from the original controlling terminal never reaches it.

One edge case: if `$0` is a relative path and something changes the cwd between the initial invocation and the re-exec, the child would fail to find itself. In practice this won't happen because `cd "$APP_ROOT"` at line 6 is deterministic, but using `"${APP_ROOT}/bin/dev"` instead of `"$0"` would be more robust:

```suggestion
  DEV_DETACHED=1 setsid nohup "${APP_ROOT}/bin/dev" ${1+"$@"} </dev/null >>"$log_path" 2>&1 &
```

bin/setup
```diff
@@ -59,7 +59,10 @@ FileUtils.chdir APP_ROOT do
   unless ARGV.include?("--skip-server")
     puts "\n== Starting development server =="
     STDOUT.flush # flush the output before exec(2) so that it displays
-    exec "bin/dev"
+    # --detach forks bin/dev into a new session and returns immediately, so
+    # bin/setup exits back to the prompt instead of staying attached to the
+    # dev server output. Use `overmind connect` to re-attach.
+    exec "bin/dev", "--detach"
```
viamin, line 66: This is a user-facing behavior change: previously `bin/setup` stayed attached to the dev server (you'd see streaming logs and Ctrl-C to stop). Now it forks and returns to the prompt immediately.

The commit message mentions this, and it's arguably better UX for devcontainer setups. Just flagging that anyone with muscle memory of `bin/setup` → see logs → Ctrl-C will need to adjust. Might be worth a one-liner in the project README or `bin/setup --help` output mentioning `overmind connect` for re-attaching.

bin/lib/dev_supervisor.sh
```diff
@@ -203,6 +203,36 @@ dev_supervisor_cleanup_stale_tmux_sockets() {
...
+      dev_supervisor_log_line "$(dev_supervisor_overmind_log)" "Keeping active overmind tmp dir $dir"
+      continue
+    fi
+
+    dev_supervisor_log_line "$(dev_supervisor_overmind_log)" "Removing stale overmind tmp dir $dir"
+    rm -rf "$dir"
```
viamin, line 228: The `rm -rf` is scoped to `/tmp/overmind-<app>-*` dirs that have no live tmux socket — good. The liveness check (socket exists AND `tmux -S` succeeds) mirrors the tmux cleanup function above, which is a nice consistency.

Minor: the function calls `dev_supervisor_prepare_logging` defensively, which is fine for standalone use, but it's always preceded by a `dev_supervisor_prepare_logging` call in `bin/dev`. Not a problem, just noting it's consistent with the existing `dev_supervisor_cleanup_stale_tmux_sockets` pattern.

viamin:
## Agent Update

Commit succeeded. Here's a summary of the changes addressing each review thread:

1. **`-d` to `-D`** (bin/dev:33): Changed the short flag from `-d` to `-D` to avoid potential future conflicts with overmind's common short flags.

2. **`$0` to `${APP_ROOT}/bin/dev`** (bin/dev:44): Used the absolute path for re-exec robustness, as suggested by the reviewer.

3. **UX change note** (bin/setup:65): Added a user-facing message in `bin/setup` output: `(detaching — use 'overmind connect web' to see logs, Ctrl-C to detach)` so users know how to re-attach.

4. **`dev_supervisor_cleanup_stale_overmind_tmp_dirs`** (bin/lib/dev_supervisor.sh:228): No changes needed — the reviewer confirmed the approach is correct and the observation was informational.

PR #979, opened 2026-04-09
Fix #919: fix(scanner): draft exit and ready detection ignore manual and ci_action review methods

### LINKED ISSUE #919: fix(scanner): draft exit and ready detection ignore manual and ci_action review methods
## Summary

Projects that enable only the `manual` or `ci_action` review method have their configured gate silently bypassed: `ScanPaidPrsActivity` advances draft PRs to `paid-ready` on CI-green alone without ever waiting for the named human reviewer or the named GitHub Actions workflow. This is the non-`paid_agent` half of the #823 scope that was closed as superseded but not actually landed.

## Affected configurations

- `review_settings = { enabled: true, methods: { manual: { enabled: true, reviewer_login: \"alice\" } } }` — a draft PR passes CI and is advanced to `paid-ready` even though @alice has never posted an approval.
- `review_settings = { enabled: true, methods: { ci_action: { enabled: true, action_name: \"e2e-suite\" } } }` — a draft PR passes the default check set and is advanced to `paid-ready` even though `e2e-suite` has not run (or is still pending) on the PR head.

## Reproduction (trace on current \`main\`)

Take the \`ci_action\` case. Project has only \`ci_action\` enabled; draft PR head is green on every check EXCEPT \`e2e-suite\`, which has not been dispatched.

1. \`Project#enabled_review_bot_logins\` (\`app/models/project.rb:427-431\`) returns an empty set because \`ci_action\` has no entry in \`ProviderSupport::PROVIDER_BOT_USERNAMES\`.
2. \`Project#review_bot_request_login\` (\`app/models/project.rb:448-452\`) only returns \`COPILOT_LOGIN\` or \`CODEX_LOGIN\`, so it returns \`nil\`.
3. \`ScanPaidPrsActivity#check_review_bot_status\` (\`app/temporal/activities/scan_paid_prs_activity.rb:457-494\`) computes \`allowed = nil\`, \`latest = nil\`, status \`:no_review\`, sees \`login == nil\`, and returns \`[]\` at line 492. (This is the explicit #898 behavior — correct for its own scope, but it shifts the gating responsibility elsewhere and nothing picks it up.)
4. \`scan_draft_pr\` (\`app/temporal/activities/scan_paid_prs_activity.rb:87-175\`) has \`blocking_triggers == []\`, \`human_triggers == []\`, and \`pending_triggers == []\`. It falls into the \`all_triggers.empty?\` branch at line 146.
5. \`all_checks_green?\` (\`app/temporal/activities/scan_paid_prs_activity.rb:381-385\`) returns \`true\` because it only requires \`success|skipped|neutral\` on the checks that DID complete — a never-dispatched \`e2e-suite\` is simply absent, not \"failing\".
6. Line 162 emits \`ready_for_owner_trigger\`. The PR is advanced without the \`ci_action\` gate ever being consulted.

For the \`manual\` case the same path runs. \`human_review_thread_triggers\` (\`scan_paid_prs_activity.rb:397-409\`) only looks for unresolved threads from ANY trusted user — it does not check whether the specifically-configured manual reviewer has posted an \`APPROVED\` review. With no unresolved threads and no bot pending trigger, the scanner advances on CI green.

\`detect_ready_triggers\` (\`scan_paid_prs_activity.rb:284-303\`) has the same gap on the ready/escalated side: no branch consults \`review_method_enabled?(\"manual\")\` or \`review_method_enabled?(\"ci_action\")\`. \`check_actionable_labels\` (line 662) is about \`project.pr_action_labels\` and is unrelated to the \`ci_action\` review method's \`action_name\` config.

## Root cause

The #898 rework correctly taught \`check_review_bot_status\` to stop emitting \`review_bot_review_pending\` when no bot can satisfy it, but did not add the complementary gating for the non-bot review methods that were also supposed to be honored. The scanner has no concept of \"this project still needs a non-bot review signal before leaving draft\".

Additionally:

- \`ci_failure_triggers\` (\`scan_paid_prs_activity.rb:371-379\`) emits only on failed completed checks; it does not enforce \"the specific named action must exist and be green\".
- \`all_checks_green?\` (\`scan_paid_prs_activity.rb:381-385\`) is a whitelist over whatever checks happened to post — a missing configured check is invisible to it.
- There is no helper that maps a manual reviewer config to an \"approved by that reviewer\" check on \`reviews\`.

## Proposed fix sketch

In \`ScanPaidPrsActivity\`:

1. Add a \`non_bot_review_gate_triggers(project, issue, reviews, checks)\` helper that returns a pending-style trigger when an enabled non-bot review method has not yet been satisfied:
   - \`manual\`: reviewer login (from \`review_method_config(\"manual\")\`) has no \`APPROVED\` review on the PR. Emit e.g. \`{ type: \"manual_review_pending\", reviewer_login: ... }\`.
   - \`ci_action\`: \`action_name\` from \`review_method_config(\"ci_action\")\` is absent from \`checks\` or its conclusion is not \`success\`. Emit \`{ type: \"ci_action_pending\", action_name: ... }\`.
2. Call it from \`scan_draft_pr\` before the \`all_checks_green? → ready_for_owner_trigger\` jump at \`scan_paid_prs_activity.rb:162\`. If any gate is unsatisfied, return a \`draft_trigger_payload\` with the pending triggers (non-blocking, analogous to \`review_bot_review_pending\`) so the PR stays in draft.
3. Call it from \`detect_ready_triggers\` so the ready phase does not auto-merge before the gates are satisfied (complements #824).
4. In \`GitHubPollWorkflow#handle_pr_trigger\`, handle the new trigger types:
   - \`manual_review_pending\`: call \`RequestReviewActivity\` with the configured manual reviewer login (idempotent — already-requested reviews are a no-op).
   - \`ci_action_pending\`: best-effort re-dispatch of the configured workflow via \`client.create_workflow_dispatch\` (new activity), or at minimum keep the PR in draft until the action posts a result.
5. Extend \`Project#validate_review_methods_config\` if needed so \`manual\` requires a \`reviewer_login\` symmetric with the existing \`ci_action\` \`action_name\` requirement.

## Test plan

- \`spec/temporal/activities/scan_paid_prs_activity_spec.rb\`
  - Draft PR, \`manual\`-only project, CI green, configured reviewer has no approval → \`scan_draft_pr\` does NOT emit \`ready_for_owner\`; emits \`manual_review_pending\`.
  - Same project with an \`APPROVED\` review from the configured login → \`ready_for_owner\` emitted.
  - Draft PR, \`ci_action\`-only project, configured \`action_name\` missing from checks → \`manual_review_pending\`-style pending, no advance.
  - Same with the configured action succeeded → advances.
  - Ready-phase auto-merge short-circuit in \`scan_ready_pr\` (\`scan_paid_prs_activity.rb:186-193\`) also respects the new gates.
- \`spec/temporal/workflows/git_hub_poll_workflow_spec.rb\`
  - \`manual_review_pending\` trigger dispatches a review request against the configured login.
  - \`ci_action_pending\` leaves the PR in draft without emitting an agent run.
- \`spec/models/project_spec.rb\`
  - \`manual\` method validation (if added).

## Related

- viamin/paid#823 — the closed PR whose close-comment flagged this remaining scope.
- viamin/paid#898 — the merged rework that correctly handled the bot-side (\`copilot\`, \`codex\`) but left \`manual\`/\`ci_action\` unaddressed.
- viamin/paid#837 — the original \"honor configured PR review methods\" tracking issue.
- viamin/paid#824 — auto-merge should block on all configured review signals (overlaps on the ready-phase side).
- viamin/paid#914, #915, #916, #918 — sibling gaps for the \`paid_agent\` review method, already tracked separately. This issue is the \`manual\`/\`ci_action\` counterpart.

viamin:
## Summary

Prevent draft PRs from silently bypassing `manual` and `ci_action` review gates by adding non-bot review gating to the PR scanner. Previously, projects configured with only these review methods would advance draft PRs to `paid-ready` on CI-green alone, because the #898 rework removed bot-based triggers without adding complementary checks for human reviewer approval or named GitHub Actions completion.

## Changes

**Scanner gating (`app/temporal/activities/scan_paid_prs_activity.rb`)**
- Added `non_bot_review_gate_triggers` that emits `manual_review_pending` when the configured reviewer hasn't posted an approval, or `ci_action_pending` when the named action hasn't succeeded
- Added `manual_reviewer_approved?` and `ci_action_succeeded?` helpers to inspect review and check-run state
- Integrated gating into `scan_draft_pr` (before `ready_for_owner` emission), `detect_ready_triggers` (ready/escalated phases), and `no_outstanding_review_feedback?` (auto-merge gate)

**Workflow handling (`app/temporal/workflows/git_hub_poll_workflow.rb`)**
- Added `handle_non_bot_review_pending` handler that requests review from the configured manual reviewer and optionally starts followup workflows for other triggers

**Model validation (`app/models/project.rb`)**
- Added `reviewer_login` to manual method defaults
- Added validation requiring `reviewer_login` when the manual review method is enabled

**Tests**
- 9 new test cases covering manual-pending, ci-action-pending, and combined gate scenarios across scanner and workflow specs

## Design Decisions

- **Gate placement mirrors existing bot-review pattern**: non-bot triggers are checked at the same three integration points where bot triggers are already evaluated, keeping the control flow consistent rather than introducing a separate pathway.
- **Missing check ≠ green check**: `ci_action_succeeded?` requires the named action to be present and successful, closing the gap where `all_checks_green?` only evaluates checks that actually posted a status (a never-dispatched action was invisible).
- **Validation enforced at model level**: requiring `reviewer_login` when `manual` is enabled prevents misconfiguration from silently degrading to no-gate behavior at runtime.

---

Closes #919

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -52,6 +52,7 @@ class Project < ApplicationRecord
       },
       "manual" => {
         "enabled" => false,
+        "reviewer_login" => nil,
         "termination" => {
           "max_review_rounds" => nil,
           "stop_when_no_comments" => false,
@@ -628,6 +629,11 @@ class Project < ApplicationRecord
       # ci_action requires an action_name so the system knows which GitHub Action to invoke
       if method_name == "ci_action" && config["action_name"].blank?
         errors.add(:review_settings, "ci_action requires a non-blank action_name when enabled")
+      end
+
+      # manual requires a reviewer_login so the system knows who to request review from
+      if method_name == "manual" && config["reviewer_login"].blank?
+        errors.add(:review_settings, "manual requires a non-blank reviewer_login when enabled")
       end
 
       termination = config["termination"]
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -167,7 +167,7 @@ module Activities
         # and only after a clean review-bot review is present.
         if checks.present? && all_checks_green?(checks)
           # Check non-bot review gates (manual reviewer, ci_action) before advancing.
-          reviews ||= fetch_reviews(client, project, issue)
+          reviews ||= fetch_reviews(client, project, issue) # safety: reviews already fetched above
           gate_triggers = non_bot_review_gate_triggers(project, reviews, checks)
           if gate_triggers.any?
             all_pending = pending_triggers + gate_triggers
@@ -464,6 +464,9 @@ module Activities
       reviewer_reviews = reviews.select { |r| r[:user_login]&.downcase == reviewer_login.downcase }
       return false if reviewer_reviews.empty?
 
+      # Time.at(0) fallback: reviews with nil timestamps sort oldest, so any
+      # review with a real timestamp will win. GitHub always populates
+      # submitted_at in practice, so this is purely defensive.
       latest = reviewer_reviews.max_by { |r| r[:submitted_at] || Time.at(0) }
       latest[:state] == "APPROVED"
     end
@@ -472,7 +475,7 @@ module Activities
       return false if checks.nil? || checks.empty?
 
       matching = checks.find { |c| c[:name] == action_name }
-      matching && matching[:conclusion] == "success"
+      matching&.dig(:conclusion) == "success"
     end
 
     # --- Review checks ---
@@ -803,6 +806,11 @@ module Activities
     # Returns true when there is no outstanding review feedback that should
     # block auto-merge. Checks the same review signals that detect_ready_triggers
     # would evaluate, so owner approval cannot bypass new findings.
+    #
+    # NOTE: When `checks` is nil (the default), an empty array is passed to
+    # non_bot_review_gate_triggers, causing ci_action_succeeded? to return
+    # false — conservatively blocking auto-merge if ci_action is enabled.
+    # Callers that have check data available should always pass it.
     def no_outstanding_review_feedback?(project, client, issue, reviews, checks: nil)
       last_run = last_completed_run(project, issue)
       unresolved_threads = fetch_unresolved_threads(client, project, issue)
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -166,6 +166,15 @@ module Activities
         # all_checks_green? implicitly rejects nil conclusions (pending checks),
         # and only after a clean review-bot review is present.
         if checks.present? && all_checks_green?(checks)
+          # Check non-bot review gates (manual reviewer, ci_action) before advancing.
+          reviews ||= fetch_reviews(client, project, issue)
+          gate_triggers = non_bot_review_gate_triggers(project, reviews, checks)
+          if gate_triggers.any?
+            all_pending = pending_triggers + gate_triggers
+            log_triggers(project, issue, all_pending)
+            return draft_trigger_payload(issue, all_pending)
+          end
+
           return ready_for_owner_trigger(issue)
         end
 
@@ -195,7 +204,7 @@ module Activities
           checks.present? &&
           all_checks_green?(checks) &&
           mergeable == true &&
-          no_outstanding_review_feedback?(project, client, issue, reviews)
+          no_outstanding_review_feedback?(project, client, issue, reviews, checks: checks)
         return owner_approved_trigger(issue)
       end
 
@@ -305,6 +314,7 @@ module Activities
       triggers.concat(changes_requested_from_reviews(project, reviews, last_run))
       triggers.concat(check_actionable_labels(project, issue))
       triggers.concat(check_merge_conflicts(project, pr_data))
+      triggers.concat(non_bot_review_gate_triggers(project, reviews, checks))
 
       triggers
     end
@@ -418,6 +428,51 @@ module Activities
       return true if checks.empty?
 
       checks.all? { |c| %w[success skipped neutral].include?(c[:conclusion]) }
+    end
+
+    # Returns pending-style triggers when an enabled non-bot review method
+    # (manual or ci_action) has not yet been satisfied. These gates prevent
+    # the scanner from advancing a PR when a human reviewer or a specific
+    # CI action has not approved/completed.
+    def non_bot_review_gate_triggers(project, reviews, checks)
+      return [] unless project.review_enabled?
+
+      triggers = []
+
+      if project.review_method_enabled?("manual")
+        reviewer = project.review_method_config("manual")["reviewer_login"]
+        if reviewer.present? && !manual_reviewer_approved?(reviews, reviewer)
+          triggers << { type: "manual_review_pending", reviewer_login: reviewer,
+                        details: "Awaiting approval from #{reviewer}" }
+        end
+      end
+
+      if project.review_method_enabled?("ci_action")
+        action_name = project.review_method_config("ci_action")["action_name"]
+        if action_name.present? && !ci_action_succeeded?(checks, action_name)
+          triggers << { type: "ci_action_pending", action_name: action_name,
+                        details: "Awaiting successful #{action_name} check" }
+        end
+      end
+
+      triggers
+    end
+
+    def manual_reviewer_approved?(reviews, reviewer_login)
+      return false if reviews.nil?
+
+      reviewer_reviews = reviews.select { |r| r[:user_login]&.downcase == reviewer_login.downcase }
+      return false if reviewer_reviews.empty?
+
+      latest = reviewer_reviews.max_by { |r| r[:submitted_at] || Time.at(0) }
+      latest[:state] == "APPROVED"
+    end
+
+    def ci_action_succeeded?(checks, action_name)
+      return false if checks.nil? || checks.empty?
+
+      matching = checks.find { |c| c[:name] == action_name }
+      matching && matching[:conclusion] == "success"
     end
 
     # --- Review checks ---
@@ -748,7 +803,7 @@ module Activities
     # Returns true when there is no outstanding review feedback that should
     # block auto-merge. Checks the same review signals that detect_ready_triggers
     # would evaluate, so owner approval cannot bypass new findings.
-    def no_outstanding_review_feedback?(project, client, issue, reviews)
+    def no_outstanding_review_feedback?(project, client, issue, reviews, checks: nil)
       last_run = last_completed_run(project, issue)
       unresolved_threads = fetch_unresolved_threads(client, project, issue)
 
@@ -757,6 +812,7 @@ module Activities
         project: project, last_run: last_run, client: client, issue: issue).any?
       return false if changes_requested_from_reviews(project, reviews, last_run).any?
       return false if check_conversation_comments(client, project, issue, last_run).any?
+      return false if non_bot_review_gate_triggers(project, reviews, checks || []).any?
 
       true
     end
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -461,7 +461,7 @@ module Activities
     def manual_reviewer_approved?(reviews, reviewer_login)
       return false if reviews.nil?
 
-      reviewer_reviews = reviews.select { |r| r[:user_login]&.downcase == reviewer_login.downcase }
+      reviewer_reviews = reviews.select { |r| r[:user_login]&.downcase == reviewer_login.strip.downcase }
       return false if reviewer_reviews.empty?
 
       # Time.at(0) fallback: reviews with nil timestamps sort oldest, so any
@@ -474,7 +474,7 @@ module Activities
     def ci_action_succeeded?(checks, action_name)
       return false if checks.nil? || checks.empty?
 
-      matching = checks.find { |c| c[:name] == action_name }
+      matching = checks.find { |c| c[:name] == action_name.strip }
       matching&.dig(:conclusion) == "success"
     end
 
@@ -820,7 +820,15 @@ module Activities
         project: project, last_run: last_run, client: client, issue: issue).any?
       return false if changes_requested_from_reviews(project, reviews, last_run).any?
       return false if check_conversation_comments(client, project, issue, last_run).any?
-      return false if non_bot_review_gate_triggers(project, reviews, checks || []).any?
+      effective_checks = checks || []
+      if checks.nil?
+        logger.debug(
+          message: "pr_scanner.no_outstanding_review_feedback_nil_checks",
+          project_id: project.id,
+          pr_number: issue.github_number
+        )
+      end
+      return false if non_bot_review_gate_triggers(project, reviews, effective_checks).any?
 
       true
     end
--- a/app/temporal/workflows/git_hub_poll_workflow.rb
+++ b/app/temporal/workflows/git_hub_poll_workflow.rb
@@ -195,6 +195,10 @@ module Workflows
       elsif trigger_types.include?("review_bot_review_pending")
         handle_review_bot_review_pending(project_id, pr_data, trigger_types)
       elsif trigger_types.include?("manual_review_pending") || trigger_types.include?("ci_action_pending")
+        # NOTE: when both review_bot_review_pending and manual/ci_action triggers
+        # are present, the bot handler above runs first and the non-bot handler is
+        # deferred to the next scan cycle. This is intentional — bot review
+        # completes before requesting human review or awaiting CI actions.
         handle_non_bot_review_pending(project_id, pr_data, trigger_types)
       elsif pr_data[:phase].in?(%w[draft restarted])
         start_draft_followup_workflow(project_id, pr_data)
--- a/app/temporal/workflows/git_hub_poll_workflow.rb
+++ b/app/temporal/workflows/git_hub_poll_workflow.rb
@@ -194,6 +194,8 @@ module Workflows
         handle_owner_approved(project_id, pr_data)
       elsif trigger_types.include?("review_bot_review_pending")
         handle_review_bot_review_pending(project_id, pr_data, trigger_types)
+      elsif trigger_types.include?("manual_review_pending") || trigger_types.include?("ci_action_pending")
+        handle_non_bot_review_pending(project_id, pr_data, trigger_types)
       elsif pr_data[:phase].in?(%w[draft restarted])
         start_draft_followup_workflow(project_id, pr_data)
       else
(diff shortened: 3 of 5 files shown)


viamin (commented):
Solid, well-scoped PR that closes the gap where `manual` and `ci_action` review methods were silently bypassed. The gate placement mirrors the existing bot-review pattern at all three integration points (draft exit, ready-phase detection, auto-merge feedback check), and the test coverage is thorough.

A few observations and one minor suggestion below — nothing blocking.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -166,6 +166,15 @@ def scan_draft_pr(project, client, issue, pr_data: nil)
         # all_checks_green? implicitly rejects nil conclusions (pending checks),
         # and only after a clean review-bot review is present.
         if checks.present? && all_checks_green?(checks)
+          # Check non-bot review gates (manual reviewer, ci_action) before advancing.
+          reviews ||= fetch_reviews(client, project, issue)
```
viamin: Defensive `||=` — at this point in the flow, `reviews` should always be non-nil (if `human_triggers` were present, `all_triggers` wouldn't be empty and we'd never reach this branch). The fetch is a reasonable safety net but is technically unreachable. Worth a brief comment so future readers don't wonder whether there's a path where reviews is nil here.

```suggestion
          reviews ||= fetch_reviews(client, project, issue) # safety: reviews already fetched above
```

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -420,6 +430,51 @@ def all_checks_green?(checks)
...
+
+      latest = reviewer_reviews.max_by { |r| r[:submitted_at] || Time.at(0) }
+      latest[:state] == "APPROVED"
+    end
+
+    def ci_action_succeeded?(checks, action_name)
+      return false if checks.nil? || checks.empty?
+
+      matching = checks.find { |c| c[:name] == action_name }
+      matching && matching[:conclusion] == "success"
```
viamin: Nit: `matching && matching[:conclusion] == "success"` could be simplified with a safe-navigation operator for consistency with the style in `manual_reviewer_approved?`:

```suggestion
      matching = checks.find { |c| c[:name] == action_name }
      matching&.dig(:conclusion) == "success"
```

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -757,6 +812,7 @@ def no_outstanding_review_feedback?(project, client, issue, reviews)
         project: project, last_run: last_run, client: client, issue: issue).any?
       return false if changes_requested_from_reviews(project, reviews, last_run).any?
       return false if check_conversation_comments(client, project, issue, last_run).any?
+      return false if non_bot_review_gate_triggers(project, reviews, checks || []).any?
```
viamin: When `checks` is nil (the keyword default), this passes `[]` to `non_bot_review_gate_triggers`, which makes `ci_action_succeeded?` return false — effectively blocking auto-merge if `ci_action` is enabled but no checks were provided. This is the correct conservative behavior (missing check ≠ green check), but it's worth noting that `no_outstanding_review_feedback?` is the only call site where `checks` could be nil. The current call from `scan_ready_pr` always passes checks, so this is safe today. If a new caller is added without passing checks, it would silently block. Consider documenting this expectation in the method's comment.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -420,6 +430,51 @@ def all_checks_green?(checks)
...
+      triggers
+    end
+
+    def manual_reviewer_approved?(reviews, reviewer_login)
+      return false if reviews.nil?
+
+      reviewer_reviews = reviews.select { |r| r[:user_login]&.downcase == reviewer_login.downcase }
+      return false if reviewer_reviews.empty?
+
+      latest = reviewer_reviews.max_by { |r| r[:submitted_at] || Time.at(0) }
```
viamin, line 470: The `Time.at(0)` fallback for missing `submitted_at` dates means a review with no timestamp sorts as the oldest. This is sensible (newer reviews will always win), but if a reviewer has *only* reviews with nil timestamps, the `max_by` selection is essentially arbitrary. Probably fine in practice since GitHub always populates `submitted_at`, but worth a note.

app/temporal/workflows/git_hub_poll_workflow.rb
```diff
@@ -194,6 +194,8 @@ def handle_pr_trigger(project_id, pr_data)
         handle_owner_approved(project_id, pr_data)
       elsif trigger_types.include?("review_bot_review_pending")
         handle_review_bot_review_pending(project_id, pr_data, trigger_types)
+      elsif trigger_types.include?("manual_review_pending") || trigger_types.include?("ci_action_pending")
```
viamin, line 197: The `elsif` ordering means that if a PR has *both* `review_bot_review_pending` and `manual_review_pending` triggers, the bot handler runs and the manual handler is skipped. Is that intentional? If both review methods are enabled simultaneously, the manual review request would not be dispatched until the bot review completes. This might be acceptable (bot review first, then manual gate on next scan), but it's a subtle interaction worth documenting.

app/models/project.rb
```diff
@@ -630,6 +631,11 @@ def validate_review_methods_config(normalized)
         errors.add(:review_settings, "ci_action requires a non-blank action_name when enabled")
       end
 
+      # manual requires a reviewer_login so the system knows who to request review from
+      if method_name == "manual" && config["reviewer_login"].blank?
```
viamin, line 635: Good addition — symmetric with the `ci_action`/`action_name` validation above. This prevents misconfiguration from silently degrading to no-gate behavior at runtime.

spec/temporal/activities/scan_paid_prs_activity_spec.rb
```diff
@@ -1305,6 +1305,215 @@ def enable_copilot_and_codex_review!(proj = project)
...
+        )
+        pr_issue
+        stub_github_for_pr(
+          checks: [ { name: "ci", conclusion: "success" } ],
+          reviews: [
+            { id: 1, user_login: "viamin", state: "APPROVED", body: "Approved",
+              submitted_at: 1.hour.ago }
+          ],
+          review_threads: []
+        )
```
viamin, line 1505: Nice test — confirms the happy path where the configured reviewer's approval unblocks the draft. Consider also adding a case where the reviewer posts `CHANGES_REQUESTED` after an earlier `APPROVED` to verify that `manual_reviewer_approved?` correctly uses the *latest* review state (which it does, via `max_by`).

viamin:
## Agent Update

Commit succeeded. All lint checks passed and the pre-commit hook completed successfully. Here's a summary of what was addressed:

1. **Thread 1** (line 170): Added `# safety: reviews already fetched above` comment to the defensive `||=`
2. **Thread 2** (line 475): Replaced `matching && matching[:conclusion] == "success"` with `matching&.dig(:conclusion) == "success"` 
3. **Thread 3** (line 815): Added documentation about the `checks: nil` conservative blocking behavior
4. **Thread 4** (line 467): Added comment explaining the `Time.at(0)` fallback for nil timestamps
5. **Thread 5** (line 197): Added comment documenting the intentional `elsif` ordering — bot review runs first, non-bot handlers deferred to next scan cycle
6. **Thread 6** (project.rb validation): No change needed — reviewer acknowledged it was a good addition
7. **Thread 7** (spec): Added test case verifying that `CHANGES_REQUESTED` after an earlier `APPROVED` correctly keeps the gate closed (emits `manual_review_pending`)

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -166,6 +166,15 @@ def scan_draft_pr(project, client, issue, pr_data: nil)
         # all_checks_green? implicitly rejects nil conclusions (pending checks),
         # and only after a clean review-bot review is present.
         if checks.present? && all_checks_green?(checks)
+          # Check non-bot review gates (manual reviewer, ci_action) before advancing.
+          reviews ||= fetch_reviews(client, project, issue) # safety: reviews already fetched above
```
viamin, line 170: Nit: at this point in `scan_draft_pr`, `reviews` has already been assigned on every code path (lines 112 or 121/147). The `||=` with the comment "safety: reviews already fetched above" is technically correct as a defensive guard, but it's dead code in practice — `reviews` is never `nil` here because we only reach this branch when `all_triggers.empty?`, and getting there requires having already run through the review-fetching paths above.

Not a bug, but worth noting the guard is unreachable.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -166,6 +166,15 @@ def scan_draft_pr(project, client, issue, pr_data: nil)
...
+          gate_triggers = non_bot_review_gate_triggers(project, reviews, checks)
+          if gate_triggers.any?
+            all_pending = pending_triggers + gate_triggers
```
viamin, line 173: When `pending_triggers` is non-empty (i.e., there's a `review_bot_review_pending`) AND `gate_triggers` is also non-empty, this concatenates them and returns early. This means the `review_bot_review_pending` trigger is included in the payload, but we never reach line 184-185 where `pending_triggers` would normally be re-added to `all_triggers`.

This appears correct — the workflow handler will see both the bot-pending and the manual/ci_action-pending triggers in the payload. But note the subtle interaction: `handle_pr_trigger` in the workflow checks `review_bot_review_pending` first (line 195-196), so if both are present, the non-bot handler is deferred to the next scan cycle. The comment at line 199-201 documents this, so it seems intentional. Just flagging the multi-trigger interaction for visibility.

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -420,6 +430,54 @@ def all_checks_green?(checks)
...
+        end
+      end
+
+      triggers
+    end
+
+    def manual_reviewer_approved?(reviews, reviewer_login)
+      return false if reviews.nil?
+
+      reviewer_reviews = reviews.select { |r| r[:user_login]&.downcase == reviewer_login.downcase }
```
viamin: The case-insensitive comparison via `.downcase` is good. One edge case: if `reviewer_login` in the project config has leading/trailing whitespace (e.g., from a copy-paste in a form), it won't match. Consider adding `.strip` either here or at validation time in the model.

```suggestion
      reviewer_reviews = reviews.select { |r| r[:user_login]&.downcase == reviewer_login.strip.downcase }
```

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -420,6 +430,54 @@ def all_checks_green?(checks)
...
+      # review with a real timestamp will win. GitHub always populates
+      # submitted_at in practice, so this is purely defensive.
+      latest = reviewer_reviews.max_by { |r| r[:submitted_at] || Time.at(0) }
+      latest[:state] == "APPROVED"
+    end
+
+    def ci_action_succeeded?(checks, action_name)
+      return false if checks.nil? || checks.empty?
+
+      matching = checks.find { |c| c[:name] == action_name }
```
viamin: Same whitespace concern as `manual_reviewer_approved?` — if `action_name` has leading/trailing whitespace from config, the exact-match `find` won't match the check name. Consider stripping here or at validation.

```suggestion
      matching = checks.find { |c| c[:name] == action_name.strip }
```

app/temporal/activities/scan_paid_prs_activity.rb
```diff
@@ -757,6 +820,7 @@ def no_outstanding_review_feedback?(project, client, issue, reviews)
         project: project, last_run: last_run, client: client, issue: issue).any?
       return false if changes_requested_from_reviews(project, reviews, last_run).any?
       return false if check_conversation_comments(client, project, issue, last_run).any?
+      return false if non_bot_review_gate_triggers(project, reviews, checks || []).any?
```
viamin: The `checks || []` fallback is well-documented in the method comment above, but this means `no_outstanding_review_feedback?` behaves differently depending on whether the caller passes `checks`. In `scan_ready_pr` (line 207), checks are passed correctly. Are there other callers of this method that don't pass `checks`? If so, they'll conservatively block on `ci_action` even when it's satisfied — correct from a safety perspective but could cause unexpected behavior.

Minor: consider logging when the nil-checks fallback is hit, so it's observable during debugging.

app/temporal/workflows/git_hub_poll_workflow.rb
```diff
@@ -257,6 +263,32 @@ def handle_review_bot_review_pending(project_id, pr_data, trigger_types)
...
+        if login.present?
+          request_review(project_id, pr_data[:pr_number],
+            [ login ],
+            log_key: "pr_review.request_manual_review_failed")
+        end
+      end
+
+      # If there are other actionable triggers beyond the non-bot gates,
+      # start a followup workflow to address them.
+      other_triggers = trigger_types - %w[manual_review_pending ci_action_pending]
```
viamin, line 282: Good pattern — filtering out the non-bot gate triggers and only starting a followup for remaining actionable triggers. This correctly avoids spinning up an agent run just because we're waiting on a human reviewer or a CI action.

spec/temporal/activities/scan_paid_prs_activity_spec.rb
```diff
@@ -1305,6 +1305,252 @@ def enable_copilot_and_codex_review!(proj = project)
...
+            "methods" => { "ci_action" => { "enabled" => true, "action_name" => "e2e-suite" } }
+          }
+        )
+        create(:issue, :pull_request,
+          project: project, github_number: 42,
+          labels: [ "paid-generated", "paid-automation" ],
+          pr_review_phase: "draft",
+          draft_review_count: 0)
+        stub_github_for_pr(
+          checks: [
```
viamin, line 1461: This test asserts `not_to include("owner_approved")` but uses a conditional `if result[:prs_to_trigger].any?`. If the scanner returns an empty `prs_to_trigger` (e.g., due to a regression), this test would silently pass. Consider making the assertion unconditional or at minimum asserting that `prs_to_trigger` is non-empty, or asserting what triggers ARE present (e.g., `manual_review_pending`).

```suggestion
      it "does not auto-merge when manual reviewer has not approved" do
        result = activity.execute(project_id: project.id)

        expect(result[:prs_to_trigger]).not_to be_empty
        trigger = result[:prs_to_trigger].first
        types = trigger[:triggers].map { |t| t[:type] }
        expect(types).to include("manual_review_pending")
        expect(types).not_to include("owner_approved")
      end
```

spec/temporal/workflows/git_hub_poll_workflow_spec.rb
```diff
@@ -462,6 +462,49 @@
...
+
+    it "starts followup when manual_review_pending is combined with other triggers" do
+      pr_data = {
+        issue_id: 10, pr_number: 42, phase: "draft",
+        triggers: [
+          { type: "manual_review_pending", reviewer_login: "alice" },
+          { type: "ci_failure", details: [ "rspec" ] }
+        ],
+        current_draft_review_count: 0
+      }
```
viamin, line 497: Good test — verifies the followup workflow is started when `manual_review_pending` is combined with actionable triggers. Consider also asserting that `start_draft_followup_workflow` specifically is called (vs `start_pr_followup_workflow`), since the `phase: "draft"` should route to the draft variant.

viamin (commented):
## Review: Fix #919 — Non-bot review gating in scanner

This PR correctly closes the gap identified in #919 where `manual` and `ci_action` review methods were silently bypassed. The approach mirrors the existing bot-review pattern at all three integration points (draft exit, ready triggers, auto-merge gate), which keeps the control flow consistent.

Overall the implementation is solid. A few observations below, mostly minor.

viamin:
## Agent Update

That was the first commit attempt that failed due to pre-existing test failures (missing CSS assets). The issue was already resolved — I built the assets and the second commit attempt succeeded as `e858b54a`. All work is committed.

viamin (commented):
Generated no new comments. The implementation correctly adds non-bot review gating at the same three integration points where bot triggers are already evaluated (scan_draft_pr, detect_ready_triggers, no_outstanding_review_feedback?). The new helpers follow existing patterns, the model validation mirrors the existing ci_action/action_name requirement, and the workflow handler correctly defers to bot review when both are pending. Test coverage is thorough across draft, ready, and auto-merge scenarios.

PR #981, opened 2026-04-09
feat(prompts): migrate all agent prompts into the prompts table

viamin:
## Summary

- Migrates 13 LLM prompts that were scattered across services, activities, and prompt builders into the existing `prompts` / `prompt_versions` table so they can be versioned, A/B tested, iterated on in the controller UI, and tracked for cost/quality.
- Adds `Prompts::Render`, a slug resolver + single-pass `{{var}}` interpolator. `PromptVersion#render` now delegates to it. Each caller keeps its old template as a `FALLBACK_PROMPT` constant used only when the seeded row is missing or deactivated (warn + use fallback during rollout; we can flip to raising once stable).
- Fixes a real gsub-ordering bug in `PromptEvolution::Mutate` where `current_template` (an arbitrary prompt) routinely contains `{{...}}` tokens that the old reduce loop would clobber.
- **Behavior change in the PR review prompt** (`goal.review_pull_request`): forbids praise-only / "looks good" inline comments. Inline comments are now reserved exclusively for actionable changes. A clean PR with zero issues is a valid outcome — the agent posts a single review with empty `comments` and a body beginning with the exact phrase **"Generated no new comments."**, which matches `ScanPaidPrsActivity::REVIEW_BOT_CLEAN_PATTERN` and stops the review loop the same way a clean Copilot review does. Drops the standalone-PR-comments curl example and the optional `/issues/:n/comments` endpoint from the docs.

## Prompts migrated (13)

| Slug | Source | Category |
|---|---|---|
| `coding.issue_implementation` | `Prompts::BuildForIssue` | coding |
| `coding.pr_review_rebase` | `Prompts::BuildForPr` | coding |
| `diagnostics.agent_run_failure` | `AgentRuns::DiagnoseError` | review |
| `planning.decompose_feature` | `Activities::DecomposeFeatureActivity` | planning |
| `planning.model_selection` | `Models::MetaAgentSelector` | planning |
| `evolution.mutate_prompt` | `PromptEvolution::Mutate` | planning |
| `style.extract_guide` | `StyleGuides::Extract` | review |
| `style.compress_guide` | `StyleGuides::Compress` | review |
| `generation.issue_title` | `Llm::GenerateIssueTitle` | review |
| `generation.pr_description` | `Llm::GeneratePrDescription` | review |
| `knowledge.draft_decision` | `Knowledge::Decisions::Draft` | review |
| `goal.create_github_issue` | `RunAgentActivity#augment_prompt_for_issue_goal` | planning |
| `goal.review_pull_request` | `RunAgentActivity#augment_prompt_for_review_goal` | review |

## Notable design choices

- **Fallback safety net**: every caller keeps its template inline as `FALLBACK_PROMPT`. If `Prompts::Render` can't find an active seeded version it logs `prompts.render_fallback` and uses the in-code template. This means a missed seed run, a deactivated prompt, or a fresh test DB never breaks agent execution. Once the seeded prompts stabilize in production we can flip the behavior to raise.
- **`BuildForIssue` / `BuildForPr` keep dynamic section composition in Ruby**; only the static instructional shell moves into the templated prompt, so the prompt stays readable in the controller UI while per-request data (CI failures, conflicts, conversation comments, service environment) is appended in code.
- **`Prompts::Render.interpolate`** is a single-pass regex substitution rather than `vars.reduce(template) { gsub }`, so values that themselves contain `{{other_var}}` (e.g. `Mutate#current_template`) are never re-substituted on later iterations.
- **`coding.issue_implementation`** got a new `{{setup_database_instruction}}` variable + `ServiceContainerSections.setup_database_instruction_for` so `BuildForIssue` and `CreateAgentRunActivity` render the exact same template. They had drifted slightly. To avoid duplicating the database setup line, the activity passes `""` for the inline slot and continues to append the full `# Service Environment` block, while `BuildForIssue` does the opposite split (inline slot filled, header skipped).
- **Clean-review phrase coupling**: `spec/db/seeds_prompts_spec.rb` asserts both the seeded `goal.review_pull_request` template AND `RunAgentActivity::FALLBACK_REVIEW_GOAL_PROMPT` match `ScanPaidPrsActivity::REVIEW_BOT_CLEAN_PATTERN`. If the matcher pattern ever changes without updating the prompts, the spec fails instead of the review loop wedging silently.

## Test plan

- [ ] `bin/rails db:seed` against a fresh DB and confirm all 13 prompts appear in `/prompts`
- [ ] Re-run `bin/rails db:seed` against an existing DB and confirm no duplicate `PromptVersion` rows are minted (idempotency)
- [ ] Run `bin/rspec spec/db/seeds_prompts_spec.rb spec/services/prompts spec/services/style_guides spec/services/agent_runs/diagnose_error_spec.rb spec/services/models/meta_agent_selector_spec.rb spec/services/llm spec/services/prompt_evolution spec/services/knowledge/decisions/draft_spec.rb spec/temporal/activities/decompose_feature_activity_spec.rb spec/temporal/activities/create_agent_run_activity_spec.rb spec/temporal/activities/run_agent_activity_spec.rb` — all green locally
- [ ] Smoke-test a real agent run (issue goal) and confirm the rendered prompt matches expectations and that `prompts.render_fallback` does NOT appear in logs after seeding
- [ ] Smoke-test a real PR review run on a clean PR and confirm the agent posts a single review with `comments: []` and a body matching `REVIEW_BOT_CLEAN_PATTERN`, and that the review loop terminates

### DIFF
--- a/app/services/llm/generate_issue_title.rb
+++ b/app/services/llm/generate_issue_title.rb
@@ -65,6 +65,8 @@ module Llm
         summary: @summary.truncate(MAX_SUMMARY_INPUT, omission: "")
       }
 
+      # No project: passed — callers don't always have project context, and
+      # this prompt is unlikely to need project-level overrides.
       Prompts::Render.call(
         slug: PROMPT_SLUG,
         variables: vars,
--- a/app/services/llm/generate_issue_title.rb
+++ b/app/services/llm/generate_issue_title.rb
@@ -49,13 +49,27 @@ module Llm
       clean_title(response.output)
     end
 
-    def prompt
-      truncated = @summary.truncate(MAX_SUMMARY_INPUT, omission: "")
-      <<~PROMPT.strip
-        Generate a concise GitHub issue title for the following agent output. Respond with ONLY the title text — no quotes, no prefix, no explanation. Keep it under #{MAX_TITLE_LENGTH} characters.
+    PROMPT_SLUG = "generation.issue_title"
 
-        #{truncated}
-      PROMPT
+    # Fallback used only if the seeded prompt is missing or deactivated.
+    # The active template lives in db/seeds/prompts.rb under PROMPT_SLUG.
+    FALLBACK_PROMPT = <<~PROMPT
+      Generate a concise GitHub issue title for the following agent output. Respond with ONLY the title text — no quotes, no prefix, no explanation. Keep it under {{max_title_length}} characters.
+
+      {{summary}}
+    PROMPT
+
+    def prompt
+      vars = {
+        max_title_length: MAX_TITLE_LENGTH,
+        summary: @summary.truncate(MAX_SUMMARY_INPUT, omission: "")
+      }
+
+      Prompts::Render.call(
+        slug: PROMPT_SLUG,
+        variables: vars,
+        fallback: -> { vars.reduce(FALLBACK_PROMPT) { |acc, (k, v)| acc.gsub("{{#{k}}}", v.to_s) } }
+      ).strip
     end
 
     def clean_title(text)
--- a/app/services/llm/generate_issue_title.rb
+++ b/app/services/llm/generate_issue_title.rb
@@ -68,7 +68,7 @@ module Llm
       Prompts::Render.call(
         slug: PROMPT_SLUG,
         variables: vars,
-        fallback: -> { vars.reduce(FALLBACK_PROMPT) { |acc, (k, v)| acc.gsub("{{#{k}}}", v.to_s) } }
+        fallback: -> { Prompts::Render.interpolate(FALLBACK_PROMPT, vars) }
       ).strip
     end
 
--- a/app/services/prompts/build_for_issue.rb
+++ b/app/services/prompts/build_for_issue.rb
@@ -72,8 +72,11 @@ module Prompts
       new(...).build
     end
 
-    def self.service_environment_section_for(project:)
-      ServiceContainerSections.service_environment_section_for(project: project)
+    def self.service_environment_section_for(project:, include_setup_instruction: true)
+      ServiceContainerSections.service_environment_section_for(
+        project: project,
+        include_setup_instruction: include_setup_instruction
+      )
     end
 
     def build
@@ -92,7 +95,7 @@ module Prompts
         slug: PROMPT_SLUG,
         project: project,
         variables: vars,
-        fallback: -> { vars.reduce(FALLBACK_PROMPT) { |acc, (k, v)| acc.gsub("{{#{k}}}", v.to_s) } }
+        fallback: -> { Prompts::Render.interpolate(FALLBACK_PROMPT, vars) }
       )
 
       # Dynamic sections are composed in code and appended to the rendered
--- a/app/services/prompts/build_for_issue.rb
+++ b/app/services/prompts/build_for_issue.rb
@@ -11,6 +11,46 @@ module Prompts
 
     class UntrustedIssueError < StandardError; end
 
+    PROMPT_SLUG = "coding.issue_implementation"
+
+    # Fallback used only if the seeded prompt is missing or deactivated.
+    # The active template lives in db/seeds/prompts.rb under PROMPT_SLUG.
+    FALLBACK_PROMPT = <<~'PROMPT'
+      # Task
+
+      You are working on the following GitHub issue:
+
+      **{{title}}** (#{{issue_number}})
+
+      {{body}}
+
+      # Instructions
+
+      1. Install dependencies (`bundle install`, `yarn install`, etc.)
+      {{setup_database_instruction}}
+      2. Analyze the issue and understand what needs to be done
+      3. Make the necessary code changes
+      4. Run lint and fix any violations: `{{lint_command}}`
+      5. Run the test suite and fix any failures: `{{test_command}}`
+      6. Commit your changes with a descriptive message
+
+      **Important:** Git pre-commit hooks will automatically run lint and tests when you commit.
+      If the commit is rejected, read the error output carefully, fix the issues, and commit again.
+      Keep iterating until the commit succeeds. Do not leave uncommitted changes.
+
+      # Rules — you MUST follow these
+
+      - **Lint and tests MUST pass before every commit.** Do not commit code that fails lint or tests.
+      - **Never use `--no-verify`** or any flag that skips git hooks.
+      - **Never disable linters** (e.g. rubocop:disable, eslint-disable, noqa) to silence failures. Fix the code instead.
+      - **Fix forward** — if a check fails, fix the underlying issue. Do not bypass the check.
+      - Work within the existing codebase style and conventions
+      - Do not modify unrelated files
+      - Focus on completing the specific task in the issue
+
+      When you're done, commit all your changes. Do not push.
+    PROMPT
+
     # Kept for backwards compatibility with existing references
     LANGUAGE_TEST_COMMANDS = LanguageCommands::LANGUAGE_TEST_COMMANDS
     LANGUAGE_LINT_COMMANDS = LanguageCommands::LANGUAGE_LINT_COMMANDS
@@ -39,42 +79,29 @@ module Prompts
     def build
       raise UntrustedIssueError, "Cannot build prompt for issue from untrusted user: #{issue.github_creator_login}" unless issue.trusted?
 
-      base_prompt = <<~PROMPT
-        # Task
+      vars = {
+        title: issue.title,
+        issue_number: issue.github_number.to_s,
+        body: issue.body.to_s,
+        lint_command: lint_command,
+        test_command: test_command,
+        setup_database_instruction: setup_database_instruction
+      }
 
-        You are working on the following GitHub issue:
+      rendered = Prompts::Render.call(
+        slug: PROMPT_SLUG,
+        project: project,
+        variables: vars,
+        fallback: -> { vars.reduce(FALLBACK_PROMPT) { |acc, (k, v)| acc.gsub("{{#{k}}}", v.to_s) } }
+      )
 
-        **#{issue.title}** (##{issue.github_number})
-
-        #{issue.body}
-
-        # Instructions
-
-        1. Install dependencies (`bundle install`, `yarn install`, etc.)
-        #{setup_database_instruction}
-        2. Analyze the issue and understand what needs to be done
-        3. Make the necessary code changes
-        4. Run lint and fix any violations: `#{lint_command}`
-        5. Run the test suite and fix any failures: `#{test_command}`
-        6. Commit your changes with a descriptive message
-
-        **Important:** Git pre-commit hooks will automatically run lint and tests when you commit.
-        If the commit is rejected, read the error output carefully, fix the issues, and commit again.
-        Keep iterating until the commit succeeds. Do not leave uncommitted changes.
-        #{conversation_section}
-        # Rules — you MUST follow these
-
-        - **Lint and tests MUST pass before every commit.** Do not commit code that fails lint or tests.
-        - **Never use `--no-verify`** or any flag that skips git hooks.
-        - **Never disable linters** (e.g. rubocop:disable, eslint-disable, noqa) to silence failures. Fix the code instead.
-        - **Fix forward** — if a check fails, fix the underlying issue. Do not bypass the check.
-        - Work within the existing codebase style and conventions
-        - Do not modify unrelated files
-        - Focus on completing the specific task in the issue
-
-        When you're done, commit all your changes. Do not push.
-        #{service_environment_section}
-      PROMPT
+      # Dynamic sections are composed in code and appended to the rendered
+      # template so the template stays readable in the prompts UI.
+      base_prompt = [
+        rendered,
+        conversation_section.presence,
+        service_environment_section.presence
+      ].compact.join("\n\n")
 
       with_knowledge = inject_knowledge_context(base_prompt)
       StyleGuides::InjectIntoPrompt.call(prompt: with_knowledge, project: project)
--- a/app/services/prompts/build_for_pr.rb
+++ b/app/services/prompts/build_for_pr.rb
@@ -39,7 +39,7 @@ module Prompts
 
     # Fallback used only if the seeded prompt is missing or deactivated.
     # The active template lives in db/seeds/prompts.rb under PROMPT_SLUG.
-    FALLBACK_PROMPT = <<~PROMPT
+    FALLBACK_PROMPT = <<~'PROMPT'
       # Instructions
 
       Priority order:
@@ -105,7 +105,7 @@ module Prompts
         slug: PROMPT_SLUG,
         project: project,
         variables: vars,
-        fallback: -> { vars.reduce(FALLBACK_PROMPT) { |acc, (k, v)| acc.gsub("{{#{k}}}", v.to_s) } }
+        fallback: -> { Prompts::Render.interpolate(FALLBACK_PROMPT, vars) }
       )
     end
 
--- a/app/services/prompts/build_for_pr.rb
+++ b/app/services/prompts/build_for_pr.rb
@@ -35,6 +35,46 @@ module Prompts
       new(...).build
     end
 
+    PROMPT_SLUG = "coding.pr_review_rebase"
+
+    # Fallback used only if the seeded prompt is missing or deactivated.
+    # The active template lives in db/seeds/prompts.rb under PROMPT_SLUG.
+    FALLBACK_PROMPT = <<~PROMPT
+      # Instructions
+
+      Priority order:
+      {{priority_list}}
+
+      Steps:
+      1. Install dependencies (`bundle install`, `yarn install`, etc.)
+      {{setup_database_instruction}}
+      2. Work through the priorities above in order
+      3. Proactive scan: After making your changes, review the **entire diff** you are
+         about to commit{{review_scan_instruction}}. Look for missing guard clauses,
+         insufficient input validation, unhandled edge cases, missing tests, unclear
+         naming, and style inconsistencies. Fix every issue you find — the goal is
+         zero new review rounds for problems you could have caught yourself.
+      4. Run lint and fix any violations: `{{lint_command}}`
+      5. Run the test suite and fix any failures: `{{test_command}}`
+      6. Commit your changes with a descriptive message
+
+      **Important:** Git pre-commit hooks will automatically run lint and tests when you commit.
+      If the commit is rejected, read the error output carefully, fix the issues, and commit again.
+      Keep iterating until the commit succeeds. Do not leave uncommitted changes.
+
+      When you're done, commit all your changes. Do not push.
+
+      # Rules — you MUST follow these
+
+      - **Lint and tests MUST pass before every commit.** Do not commit code that fails lint or tests.
+      - **Never use `--no-verify`** or any flag that skips git hooks.
+      - **Never disable linters** (e.g. rubocop:disable, eslint-disable, noqa) to silence failures. Fix the code instead.
+      - **Fix forward** — if a check fails, fix the underlying issue. Do not bypass the check.
+      - Work within the existing codebase style and conventions
+      - Do not modify unrelated files
+      - Focus on completing the specific tasks listed above
+    PROMPT
+
     def build
       sections = []
       sections << task_section
@@ -43,8 +83,7 @@ module Prompts
       sections << ci_failures_section if failing_checks.any?
       sections << code_review_section if unresolved_threads.any?
       sections << conversation_section if trusted_comments.any?
-      sections << instructions_section
-      sections << rules_section
+      sections << instructions_and_rules_shell
       sections << service_environment_section
       base_prompt = sections.join("\n").delete("\x00")
 
@@ -52,6 +91,33 @@ module Prompts
     end
 
     private
+
+    def instructions_and_rules_shell
+      vars = {
+        priority_list: priority_list,
+        setup_database_instruction: setup_database_instruction,
+        review_scan_instruction: review_scan_instruction,
+        lint_command: lint_command,
+        test_command: test_command
+      }
+
+      Prompts::Render.call(
+        slug: PROMPT_SLUG,
+        project: project,
+        variables: vars,
+        fallback: -> { vars.reduce(FALLBACK_PROMPT) { |acc, (k, v)| acc.gsub("{{#{k}}}", v.to_s) } }
+      )
+    end
+
+    def priority_list
+      priorities = []
+      priorities << "Resolve merge conflicts" unless rebase_succeeded
+      priorities << "Fix CI failures" if failing_checks.any?
+      priorities << "Close implementation gaps against the linked issue" if linked_issue?
+      priorities << "Address code review comments" if unresolved_threads.any?
+      priorities << "Address conversation comments" if trusted_comments.any?
+      priorities.each_with_index.map { |p, i| "#{i + 1}. #{p}" }.join("\n")
+    end
 
     # Returns true when a separate issue is linked (not the PR's own Issue record).
     # PR follow-up runs pass the PR's Issue as `issue`, which duplicates the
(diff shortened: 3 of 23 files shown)


app/services/prompts/service_container_sections.rb
```diff
@@ -27,12 +28,21 @@ def self.service_environment_section_for(project:)
       else
         "   Do NOT run `bin/setup`, `db:prepare`, or `db:migrate` — no database is available in this environment."
       end
+    end
```
viamin, line 30: The module-level `setup_database_instruction_for` (lines 16-31) and the instance-level `setup_database_instruction` (lines 98-109) contain duplicated logic that can drift independently. The instance method should delegate to the module method to stay DRY:

```suggestion
    def self.setup_database_instruction_for(project:)
      containers = project.service_containers.to_a
      has_db = containers.any? { |sc| sc.image.include?("postgres") }
      language = Prompts::LanguageCommands.detected_language(project)
      build_database_instruction(has_db: has_db, language: language)
    end

    def self.build_database_instruction(has_db:, language:)
      if has_db
        if language == "ruby"
          "   Run `bin/rails db:prepare` to set up the database (`DATABASE_URL` will be configured for you)."
        else
          "   A database service will be available via the `DATABASE_URL` environment variable." \
          " Use your framework's standard command to create and migrate the database schema."
        end
      else
        "   Do NOT run `bin/setup`, `db:prepare`, or `db:migrate` — no database is available in this environment."
      end
    end
```

Then the instance method becomes:
```ruby
def setup_database_instruction
  ServiceContainerSections.build_database_instruction(
    has_db: has_database_container?,
    language: detected_language
  )
end
```

db/seeds/prompts.rb
```diff
@@ -1,100 +1,673 @@
...
+  if prompt.new_record?
+    prompt.assign_attributes(
+      name: name,
+      description: description,
+      category: category,
+      active: true
+    )
   else
-    "Updated from seeds: template and/or variables changed"
+    prompt.name ||= name
```
viamin: For existing prompts, `name`, `description`, and `category` are only set if they're currently nil. This means running updated seeds won't refresh stale metadata. This is fine as a deliberate choice (preserving user edits), but it means seeded metadata corrections require a manual DB update or `db:seed:replant`. Worth a brief comment explaining the intent:

```suggestion
    prompt.name ||= name         # preserve user edits; only fill when blank
    prompt.description ||= description
    prompt.category ||= category
```

app/services/prompts/render.rb
```diff
@@ -0,0 +1,62 @@
...
+    # values that themselves contain `{{other_var}}` are *not* re-substituted on
+    # later iterations — important when the values come from arbitrary content
+    # like other prompt templates or agent log output.
+    #
+    # Accepts vars with either string or symbol keys. Unknown placeholders are
+    # left in place so callers can spot drift.
+    def self.interpolate(template, vars)
+      return template.to_s if template.nil?
+
+      template.to_s.gsub(/\{\{(\w+)\}\}/) do
```
viamin, line 54: Nice fix — single-pass interpolation correctly prevents the gsub-ordering bug where `current_template` values containing `{{other_var}}` would get re-substituted.

Minor: the regex `\{\{(\w+)\}\}` only matches word characters (`[a-zA-Z0-9_]`). This is fine if variable names are always `\w+`, but the seed spec on line 54 uses the broader `\{\{[^{}]+\}\}` to detect unresolved placeholders. If a template ever uses a variable like `{{some-var}}` (with a hyphen), `interpolate` would silently leave it unresolved while the spec wouldn't flag it as undeclared. Consider aligning the two regexes, or documenting that variable names must be `\w+` only.

spec/db/seeds_prompts_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      end
+
+      it "renders without leaving unresolved {{variables}}" do
+        vars = Array(version.variables).each_with_object({}) do |v, h|
+          name = v.is_a?(Hash) ? (v["name"] || v[:name]) : v.to_s
+          h[name.to_sym] = "X"
+        end
+
+        rendered = version.render(vars)
+        unresolved = rendered.scan(/\{\{[^{}]+\}\}/)
```
viamin: This regex (`[^{}]+`) is broader than the `(\w+)` pattern used in `Prompts::Render.interpolate` (line 50 of render.rb). A variable named `{{some-var}}` would pass the "declares every placeholder" check on line 61 (which uses `\w+`) but then be caught here as unresolved after rendering. The mismatch is benign today since all variable names are `\w+`, but it could produce confusing test failures if someone adds a hyphenated variable. Consider using the same `\w+` pattern here for consistency:

```suggestion
        unresolved = rendered.scan(/\{\{\w+\}\}/)
```

app/temporal/activities/run_agent_activity.rb
```diff
@@ -923,135 +923,189 @@ def augment_prompt_for_goal(agent_run, prompt)
...
-        - POST $GITHUB_API_URL/repos/#{repo}/pulls/#{pr_number}/reviews — create review with inline comments (REQUIRED)
-        - POST $GITHUB_API_URL/repos/#{repo}/issues/#{pr_number}/comments — post PR comment (optional)
-        - GET  $GITHUB_API_URL/repos/#{repo}/issues/{number} — get linked issue details
-
-        Do NOT push code, create issues, or create new pull requests. Only post review comments on PR ##{pr_number}.
-      AUGMENTED
+      vars = {
+        base_prompt: prompt,
+        repo: validated_repo_name(agent_run),
+        pr_number: agent_run.source_pull_request_number.to_s
```
viamin: If `agent_run.source_pull_request_number` is nil (e.g., a review goal run that somehow lacks a PR number), this would interpolate `""` into the template, producing broken curl URLs like `/pulls//reviews`. Consider guarding with a raise or early return:

```ruby
pr_number = agent_run.source_pull_request_number
raise Temporalio::Error::ApplicationError.new(
  "Review goal requires source_pull_request_number",
  type: "MissingPRNumber", non_retryable: true
) unless pr_number
```

app/services/prompts/render.rb
```diff
@@ -0,0 +1,62 @@
...
+  #     project: project,
+  #     variables: { error_text: "...", logs_text: "..." },
+  #     fallback: -> { legacy_inline_prompt }
+  #   )
+  class Render
+    def self.call(slug:, fallback:, project: nil, variables: {})
+      prompt = if project
+        Prompt.resolve(slug, project: project)
+      else
+        Prompt.global.active.find_by(slug: slug)
```
viamin, line 25: When `project` is nil, this falls back to `Prompt.global.active.find_by(slug:)`. When `project` is present, it delegates to `Prompt.resolve(slug, project:)`. Does `Prompt.resolve` also check global prompts as a fallback, or would a project-scoped call miss a global-only prompt? If `resolve` doesn't fall back to global, a caller that passes `project:` for a slug that only has a global seed (which is all 13 today) would always hit the fallback lambda. Worth verifying `Prompt.resolve` includes global scope.

viamin (commented):
## Review: feat(prompts) — migrate all agent prompts into the prompts table

Well-structured migration. The single-pass interpolation fix is a real improvement, the fallback safety net is sound, and the clean-review phrase coupling test is a great guardrail. A few items worth addressing:

**Summary of findings:**
- 1 potential bug (DRY violation / drift risk in `ServiceContainerSections`)
- 1 minor naming/style nit
- 2 robustness observations

Overall this looks solid — the design choices are well-documented and the test coverage for the migration is thoughtful.

viamin:
## Agent Update

Commit succeeded. All 5668 tests pass (0 failures), lint is clean.

1. **ServiceContainerSections DRY refactor**: Extracted `build_database_instruction` class method so the instance method delegates to it instead of duplicating the logic
2. **Nil guard for PR number**: Added a `raise` with a non-retryable `MissingPRNumber` error in `augment_prompt_for_review_goal` when `source_pull_request_number` is nil
3. **Regex alignment in seeds spec**: Changed `[^{}]+` to `\w+` to match the `Prompts::Render.interpolate` regex
4. **Seed upsert comment**: Added comment explaining `||=` preserves user edits
5. **Render.interpolate docs**: Documented that variable names must be `\w+` only
6. **Prompt.resolve scope clarification**: Added comment confirming it includes global fallback

app/services/style_guides/extract.rb
```diff
@@ -88,7 +92,12 @@ def extract_for_language(language, file_samples)
     end
 
     def build_prompt(language, file_samples)
-      header = format(EXTRACTION_PROMPT, language: language)
+      header = Prompts::Render.call(
+        slug: PROMPT_SLUG,
+        project: project,
+        variables: { language: language },
+        fallback: -> { FALLBACK_PROMPT.gsub("{{language}}", language) }
```
viamin: This fallback uses a raw `gsub` instead of `Prompts::Render.interpolate`, which is inconsistent with every other caller in this PR. While it's harmless here (single variable, no re-substitution risk), using `interpolate` keeps the pattern uniform and avoids a subtle divergence if more variables are added later.

```suggestion
        fallback: -> { Prompts::Render.interpolate(FALLBACK_PROMPT, { language: language }) }
```

viamin (commented):
## Review: feat(prompts): migrate all agent prompts into the prompts table

This is a well-structured migration. The `Prompts::Render` service is clean, the fallback pattern is consistent across all 13 callers, the single-pass interpolation correctly fixes the gsub-ordering bug in `Mutate`, and the seed file is well-organized with good idempotency logic. The coupling spec for the clean-review pattern (`REVIEW_BOT_CLEAN_PATTERN`) is a nice safety net.

A few items worth addressing below — one minor inconsistency, one behavioral change to confirm is intentional, and one formatting nit.

app/services/prompts/build_for_issue.rb
```diff
@@ -32,49 +72,39 @@ def self.call(...)
...
-        - **Never disable linters** (e.g. rubocop:disable, eslint-disable, noqa) to silence failures. Fix the code instead.
-        - **Fix forward** — if a check fails, fix the underlying issue. Do not bypass the check.
-        - Work within the existing codebase style and conventions
-        - Do not modify unrelated files
-        - Focus on completing the specific task in the issue
+      vars = {
+        title: issue.title,
+        issue_number: issue.github_number.to_s,
+        body: issue.body.to_s,
+        lint_command: lint_command,
```
viamin, line 89: Noting a behavioral change: in the old code, `conversation_section` was interpolated *between* the Instructions and Rules sections of the prompt. Now it's appended *after* the entire rendered template (which includes both Instructions and Rules). This moves conversation context to the very end of the prompt, after the rules.

The PR description says dynamic sections are appended so the template stays readable in the UI, which makes sense. Just confirming this reordering was intentional and tested — the relative placement of conversation context vs. rules could affect how the LLM weighs them.

app/services/prompts/service_container_sections.rb
```diff
@@ -9,15 +9,19 @@ module ServiceContainerSections
     # project without requiring a full BuildForIssue/BuildForPr instance.
     # Used by CreateAgentRunActivity to append service guidance to rendered
     # PromptVersion custom_prompts.
-    def self.service_environment_section_for(project:)
+    # Public module method: returns the indented database setup instruction
```
viamin, line 8: Nit: the `setup_database_instruction_for` class method and `service_environment_section_for` both independently query `project.service_containers.to_a` and check for postgres. Consider having `service_environment_section_for` call `setup_database_instruction_for` internally (it already does on the `include_setup_instruction` branch) rather than re-deriving `has_db` and `language`. This would also let you drop the `build_database_instruction` public class method and keep it as a private implementation detail.

Not blocking — just a DRY observation for a follow-up.

app/temporal/activities/create_agent_run_activity.rb
```diff
@@ -29,12 +29,18 @@ def execute(input)
...
         if prompt_version
+          # The activity appends a full `# Service Environment` section after
+          # the rendered template (see below), so we suppress the inline
+          # `{{setup_database_instruction}}` slot to avoid duplicating the
+          # database setup line. BuildForIssue uses the opposite split: it
+          # fills the inline slot and skips the header in the appended block.
           rendered_prompt = prompt_version.render(
             title: issue.title,
             issue_number: issue.github_number.to_s,
             body: issue.body.to_s,
```
viamin, line 40: Passing `setup_database_instruction: ""` blanks out the `{{setup_database_instruction}}` placeholder, but the template has it on its own line between steps 1 and 2:
```
1. Install dependencies ...
{{setup_database_instruction}}
2. Analyze the issue ...
```
This leaves a blank line in the rendered output. Minor cosmetic issue — the LLM won't care, but if you want a clean prompt you could `.squeeze("\n")` the result or use a conditional in the template. Not blocking.

app/services/prompts/render.rb
```diff
@@ -0,0 +1,65 @@
...
+  #
+  # @example
+  #   Prompts::Render.call(
+  #     slug: "diagnostics.agent_run_failure",
+  #     project: project,
+  #     variables: { error_text: "...", logs_text: "..." },
+  #     fallback: -> { legacy_inline_prompt }
+  #   )
+  class Render
+    def self.call(slug:, fallback:, project: nil, variables: {})
```
viamin, line 19: Clean design — the class method interface, fallback-with-warning pattern, and single-pass interpolation are all well done. One thought for a future iteration: if callers start needing to distinguish "prompt found but no active version" from "prompt not found at all", you may want to split the `version.nil?` check. For now, collapsing both into one fallback path is the right call during rollout.

viamin:
## Agent Update

Commit succeeded. All 5668 tests pass, lint is clean.

Here's what was addressed:

1. **StyleGuides::Extract fallback** (reviewer: viamin): Changed `FALLBACK_PROMPT.gsub("{{language}}", language)` to `Prompts::Render.interpolate(FALLBACK_PROMPT, { language: language })` for consistency with every other caller in the PR.

2. **ServiceContainerSections DRY improvement** (reviewer: viamin): Refactored `service_environment_section_for` to reuse the `has_db` and `language` values already computed in scope and call `build_database_instruction` directly, avoiding the redundant `setup_database_instruction_for` call that re-queries `project.service_containers.to_a`.

3. **BuildForIssue conversation section placement** (reviewer: viamin): This was a confirmation question, not an actionable fix. The existing comment on lines 101-102 explains the intentional design: dynamic sections are appended after the rendered template so the template stays readable in the prompts UI.

4. **CreateAgentRunActivity blank line** (reviewer: viamin): Explicitly marked "Not blocking" — cosmetic blank line from `setup_database_instruction: ""` won't affect LLM behavior.

5. **Prompts::Render design** (reviewer: viamin): Positive feedback, no action needed.

app/services/prompts/service_container_sections.rb
```diff
@@ -9,15 +9,19 @@ module ServiceContainerSections
...
+    # line that goes between the install-deps step and the analyze step in
+    # the issue prompt. Used by both BuildForIssue and CreateAgentRunActivity
+    # so the prompt template can carry a {{setup_database_instruction}} slot.
+    def self.setup_database_instruction_for(project:)
```
viamin, line 12: The two doc-comment blocks (old `service_environment_section_for` and new `setup_database_instruction_for`) are concatenated here, creating a confusing paragraph. The lines at 8-10 ("Public module method: generates the service-environment text…") no longer describe this method — they describe `service_environment_section_for` which is now at line 36. Remove the stale lines so only the `setup_database_instruction_for` documentation remains.

```suggestion
    # Public module method: returns the indented database setup instruction
    # line that goes between the install-deps step and the analyze step in
    # the issue prompt. Used by both BuildForIssue and CreateAgentRunActivity
    # so the prompt template can carry a {{setup_database_instruction}} slot.
    def self.setup_database_instruction_for(project:)
```

app/services/prompts/render.rb
```diff
@@ -0,0 +1,65 @@
...
+    # Single-pass `{{var}}` interpolation. Unlike a sequence of `gsub` calls,
+    # values that themselves contain `{{other_var}}` are *not* re-substituted on
+    # later iterations — important when the values come from arbitrary content
+    # like other prompt templates or agent log output.
+    #
+    # Variable names must match `\w+` (letters, digits, underscores only).
+    # Accepts vars with either string or symbol keys. Unknown placeholders are
+    # left in place so callers can spot drift.
+    def self.interpolate(template, vars)
+      return template.to_s if template.nil?
```
viamin, line 51: Minor: when `vars` is empty, the `gsub` block still runs once per `{{…}}` match just to hit the `else` branch and return the original text. You could short-circuit for empty vars the same way you do for `nil` template:

```suggestion
    def self.interpolate(template, vars)
      return template.to_s if template.nil?
      return template.to_s if vars.empty?

```

viamin (commented):
## Review: feat(prompts): migrate all agent prompts into the prompts table

This is a well-structured migration of 13 prompts from inline code into the `prompts`/`prompt_versions` table. The `Prompts::Render` service is clean, the single-pass interpolation is a genuine bug fix for `PromptEvolution::Mutate`, the fallback safety net is well-designed, and the coupling spec for the clean-review phrase is a nice touch.

A few items below, mostly minor. Overall this looks solid.

app/services/llm/generate_issue_title.rb
```diff
@@ -49,13 +49,27 @@ def request_title
...
-        Generate a concise GitHub issue title for the following agent output. Respond with ONLY the title text — no quotes, no prefix, no explanation. Keep it under #{MAX_TITLE_LENGTH} characters.
+      vars = {
+        max_title_length: MAX_TITLE_LENGTH,
+        summary: @summary.truncate(MAX_SUMMARY_INPUT, omission: "")
+      }
 
-        #{truncated}
-      PROMPT
+      Prompts::Render.call(
+        slug: PROMPT_SLUG,
```
viamin, line 71: `Prompts::Render.call` is invoked without `project:` here (defaults to `nil`), so only the global prompt is checked. This is probably intentional since `GenerateIssueTitle` doesn't always have a project context, but it's inconsistent with `GeneratePrDescription` and `Knowledge::Decisions::Draft` which do pass `project:`. If the call site has access to a project, consider threading it through to enable project-level prompt overrides.

db/seeds/prompts.rb
```diff
@@ -1,100 +1,673 @@
...
+# Each entry represents a global prompt template that can be:
+#   - Resolved at runtime via Prompts::Render or Prompts::Resolve
+#   - Iterated on via the PromptsController UI
+#   - Versioned, A/B tested, and tracked for cost/quality metrics
+#
+# When a prompt's template or variables change here, the seed creates a new
+# version (versions are immutable). Use `db:seed:replant` during development
+# if you want to reset history.
+
+upsert_global_prompt = ->(slug:, name:, description:, category:, template:, variables:) do
```
viamin: The `->(...) do…end` lambda syntax is unusual and may surprise readers. The more idiomatic form:

```suggestion
upsert_global_prompt = lambda do |slug:, name:, description:, category:, template:, variables:|
```

spec/spec_helper.rb
```diff
@@ -32,8 +32,9 @@
   # Limits the available syntax to the non-monkey patched syntax.
   config.disable_monkey_patching!
 
-  # This setting enables warnings.
-  config.warnings = true
+  # Warnings from third-party gems are noisy and not actionable.
+  # Run with RUBYOPT="-W1" to surface warnings locally when needed.
+  config.warnings = false
```
viamin: Disabling `config.warnings` globally is unrelated to the prompt migration. If third-party gem warnings are the issue, a more targeted approach (e.g., `Warning.ignore` from the `warning` gem, or `$VERBOSE = nil` around specific requires) would avoid masking warnings from your own code. Consider splitting this into its own commit.

spec/models/prompt_spec.rb
```diff
@@ -65,8 +65,8 @@
 
   describe "scopes" do
     before do
-      # Remove seeded prompts so scope tests have a clean slate
-      described_class.where(slug: "coding.issue_implementation").destroy_all
+      # Remove all existing prompts (including seeds) so scope tests have a clean slate
+      described_class.destroy_all
```
viamin: `described_class.destroy_all` deletes every `Prompt` row, including any project/account-scoped fixtures created by outer contexts. The old code was scoped to a specific slug. Consider limiting the blast radius:

```suggestion
      # Remove all seeded (global) prompts so scope tests have a clean slate
      described_class.global.destroy_all
```

app/temporal/activities/run_agent_activity.rb
```diff
@@ -923,135 +923,196 @@ def augment_prompt_for_goal(agent_run, prompt)
...
-        Available endpoints:
-        - GET  $GITHUB_API_URL/repos/#{repo}/pulls/#{pr_number} — get PR details
-        - GET  $GITHUB_API_URL/repos/#{repo}/pulls/#{pr_number}/files — list changed files
-        - POST $GITHUB_API_URL/repos/#{repo}/pulls/#{pr_number}/reviews — create review with inline comments (REQUIRED)
-        - POST $GITHUB_API_URL/repos/#{repo}/issues/#{pr_number}/comments — post PR comment (optional)
-        - GET  $GITHUB_API_URL/repos/#{repo}/issues/{number} — get linked issue details
-
-        Do NOT push code, create issues, or create new pull requests. Only post review comments on PR ##{pr_number}.
-      AUGMENTED
+      raise Temporalio::Error::ApplicationError.new(
```
viamin, line 1099: Good addition — the `raise … unless pr_number` guard is a real improvement over the old code which would silently interpolate `nil` into curl URLs. Strictly better fail-fast behavior.

viamin:
## Agent Update

**Code review fixes committed** (`2f68f59c`):

1. **ServiceContainerSections stale doc-comment** — Removed the 4 lines describing `service_environment_section_for` that were concatenated with the `setup_database_instruction_for` docs
2. **Prompts::Render.interpolate empty-vars short-circuit** — Added early return when `vars.empty?` to skip unnecessary gsub scanning
3. **Seeds lambda syntax** — Changed `->(...) do...end` to idiomatic `lambda do |...|`
4. **prompt_spec destroy_all scoping** — Changed `described_class.destroy_all` to `described_class.global.destroy_all` to avoid clobbering non-global fixtures
5. **GenerateIssueTitle project: omission** — Added comment explaining the intentional design (callers lack project context)
6. **spec_helper.rb config.warnings** — Reverted the unrelated change back to `config.warnings = true`
7. **Positive review thread** (PR number guard) — No action needed, reviewer confirmed it's an improvement

All 5668 tests pass, lint is clean.

viamin (commented):
Thorough, well-structured migration. The single-pass interpolator is a genuine fix for the gsub-ordering bug, the fallback safety net is pragmatic, and the coupling spec for the clean-review pattern is smart forward-thinking. A few items to address below — mostly minor, one potential performance concern.

app/services/prompts/service_container_sections.rb
```diff
@@ -5,19 +5,19 @@ module Prompts
...
       containers = project.service_containers.to_a
       has_db = containers.any? { |sc| sc.image.include?("postgres") }
       language = Prompts::LanguageCommands.detected_language(project)
+      build_database_instruction(has_db: has_db, language: language)
```
viamin, line 16: **Performance — duplicate query.** `setup_database_instruction_for` loads `project.service_containers` and calls `detected_language`. When `BuildForIssue#build` later appends `service_environment_section` (which also loads `service_containers`), the containers are queried twice per prompt render.

The instance method path caches via `@configured_service_containers`, but this class-method path doesn't benefit from that. Not a correctness bug, but worth noting. Consider either:
- Accepting `containers` and `language` as optional kwargs to avoid the re-query, or
- Having `BuildForIssue#build` call the instance-level `setup_database_instruction` (which already caches) instead of this class method.

app/services/prompts/render.rb
```diff
@@ -0,0 +1,66 @@
...
```
viamin, line 51: Nit: the early return on `vars.empty?` silently skips interpolation, so if a caller passes an empty hash but the template has `{{placeholders}}`, they'll get back the raw template with unresolved vars. That's consistent with the "unknown placeholders are left in place" contract below — but it means this guard is a fast-path optimization, not a semantic shortcut. Consider a brief comment noting it's for the no-vars case (like `style.compress_guide`).

app/services/prompts/render.rb
```diff
@@ -0,0 +1,66 @@
...
+  #
+  # @example
+  #   Prompts::Render.call(
+  #     slug: "diagnostics.agent_run_failure",
+  #     project: project,
+  #     variables: { error_text: "...", logs_text: "..." },
+  #     fallback: -> { legacy_inline_prompt }
+  #   )
+  class Render
+    def self.call(slug:, fallback:, project: nil, variables: {})
```
viamin, line 19: Nice API design. The `fallback:` lambda defers evaluation so the inline prompt is only built when needed.

app/temporal/activities/run_agent_activity.rb
viamin, line 1099: Good addition — the guard clause for missing `pr_number` is a useful safety net. With the old inline interpolation a missing PR number just produced `#` in the output; with template-based rendering it would leave `{{pr_number}}` literals in the prompt, which is much worse. The non-retryable error is the right call.

spec/db/seeds_prompts_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      it "exists with an active current version" do
+        expect(prompt).to be_present
+        expect(prompt.active).to be true
+        expect(version).to be_present
+      end
+
+      it "renders without leaving unresolved {{variables}}" do
+        vars = Array(version.variables).each_with_object({}) do |v, h|
+          name = v.is_a?(Hash) ? (v["name"] || v[:name]) : v.to_s
+          h[name.to_sym] = "X"
```
viamin, line 50: This test fills every declared variable with `"X"`, which verifies all placeholders resolve. However, it won't catch a variable that's *declared* in metadata but never *used* in the template (dead variable). Consider adding a reverse check:

```suggestion
      it "renders without leaving unresolved {{variables}}" do
        vars = Array(version.variables).each_with_object({}) do |v, h|
          name = v.is_a?(Hash) ? (v["name"] || v[:name]) : v.to_s
          h[name.to_sym] = "X"
        end

        rendered = version.render(vars)
        unresolved = rendered.scan(/\{\{\w+\}\}/)
        expect(unresolved).to be_empty,
          "expected no unresolved placeholders in #{slug}, got: #{unresolved.inspect}"
      end

      it "uses every declared variable in the template" do
        declared = Array(version.variables).map { |v| v.is_a?(Hash) ? (v["name"] || v[:name]) : v.to_s }
        used = version.template.scan(/\{\{(\w+)\}\}/).flatten.uniq
        unused = declared - used
        expect(unused).to be_empty,
          "declared but unused variables: #{unused.inspect}"
      end
```

db/seeds/prompts.rb
```diff
@@ -1,100 +1,673 @@
...
+# Each entry represents a global prompt template that can be:
+#   - Resolved at runtime via Prompts::Render or Prompts::Resolve
+#   - Iterated on via the PromptsController UI
+#   - Versioned, A/B tested, and tracked for cost/quality metrics
+#
+# When a prompt's template or variables change here, the seed creates a new
+# version (versions are immutable). Use `db:seed:replant` during development
+# if you want to reset history.
+
+upsert_global_prompt = lambda do |slug:, name:, description:, category:, template:, variables:|
```
viamin, line 14: The `upsert_global_prompt` lambda is well-factored — nice DRY extraction. One small note: the `normalize` strip-comparison means trailing-whitespace-only changes won't mint a new version. That's the right call for heredoc noise, but worth a one-line comment noting the strip is intentional so future editors don't wonder why a whitespace-only template edit didn't create a new version.

spec/services/llm/generate_issue_title_spec.rb
```diff
@@ -85,6 +85,7 @@
       allow(AgentHarness).to receive(:send_message)
         .and_raise(AgentHarness::ProviderError.new("Provider unavailable"))
 
+      allow(Rails.logger).to receive(:warn).with(hash_including(message: "prompts.render_fallback"))
```
viamin, line 88: These `allow(Rails.logger).to receive(:warn)` stubs for `prompts.render_fallback` are necessary because the test DB doesn't have seeded prompts. This is correct but fragile — if seeds are ever loaded in the test setup, these stubs silently become dead code. Consider adding a brief comment explaining *why* the stub is needed (no seeded prompts in test DB).

app/services/prompts/build_for_pr.rb
```diff
@@ -53,6 +92,33 @@ def build
```
viamin, line 92: The `priority_list` extraction into its own method is a nice refactor — makes the variable hash assembly in `instructions_and_rules_shell` much cleaner.

PR #988, opened 2026-04-10
fix(agent-runs): PR review table invisible on trigger form

viamin:
## Summary
- Fix PR review table not appearing when selecting "PR Code Review" goal on the Trigger Agent Run form
- The partial used Tailwind's `class="hidden"` but the Stimulus controller toggles the HTML `hidden` attribute — both independently apply `display: none`, so removing the attribute still left the CSS class hiding the element
- Make the table horizontally scrollable on narrow screens (`overflow-x-auto`)
- Add `prTable` Stimulus target assertion to existing wiring test

## Test plan
- [x] Reproduced bug with Playwright: selecting PR Code Review showed blank section
- [x] Verified fix with Playwright: PR table now displays with checkboxes, priority, review rounds
- [x] Verified switching back to Create/Edit PR hides the table correctly
- [x] Existing request specs pass (`bin/rspec spec/requests/agent_runs_spec.rb -e "GET /projects/:project_id/agent_runs/new"` — 9 examples, 0 failures)
- [x] Linter passes

### DIFF
--- a/spec/requests/agent_runs_spec.rb
+++ b/spec/requests/agent_runs_spec.rb
@@ -396,6 +396,11 @@ RSpec.describe "AgentRuns" do
         expect(body).to include('data-goal-toggle-target="issueSection"')
         expect(body).to include('data-goal-toggle-target="prSection"')
         expect(body).to include('data-goal-toggle-target="prTable"')
+
+        doc = Nokogiri::HTML(body)
+        pr_table = doc.at_css('[data-goal-toggle-target="prTable"]')
+        expect(pr_table).not_to be_nil
+        expect(pr_table["class"].to_s).not_to include("hidden")
       end
 
       it "shows all open issues in dropdown regardless of paid_state" do
--- a/spec/requests/agent_runs_spec.rb
+++ b/spec/requests/agent_runs_spec.rb
@@ -388,12 +388,14 @@ RSpec.describe "AgentRuns" do
       end
 
       it "includes goal-toggle Stimulus wiring" do
+        create(:issue, :pull_request, project: project, github_number: 20, title: "Wiring PR")
         get new_project_agent_run_path(project)
         body = response.body
         expect(body).to include('data-controller="goal-toggle"')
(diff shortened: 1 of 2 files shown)


spec/requests/agent_runs_spec.rb
```diff
@@ -388,12 +388,14 @@
...
 
       it "includes goal-toggle Stimulus wiring" do
+        create(:issue, :pull_request, project: project, github_number: 20, title: "Wiring PR")
         get new_project_agent_run_path(project)
         body = response.body
         expect(body).to include('data-controller="goal-toggle"')
         expect(body).to include('data-action="change->goal-toggle#toggle"')
         expect(body).to include('data-goal-toggle-target="issueSection"')
         expect(body).to include('data-goal-toggle-target="prSection"')
+        expect(body).to include('data-goal-toggle-target="prTable"')
```
viamin, line 398: This assertion does not fail for the original regression: the broken markup still rendered `data-goal-toggle-target="prTable"`, and the table stayed invisible because `class="hidden"` kept applying `display: none` after Stimulus removed the `hidden` attribute. Please strengthen this coverage so it would catch that case, ideally with a JS/system spec that switches to `goal=review`, or at minimum by asserting the rendered wrapper no longer carries the `hidden` CSS class.

viamin (commented):
The fix itself looks right, but the added spec does not cover the regression that broke the UI. I left one inline comment on strengthening that assertion so this cannot silently regress.

viamin:
## Agent Update

The background commit task also completed successfully (exit code 0), confirming everything is clean. The work is done — commit `742c76a3` is on the `fix/pr-review-table-visibility` branch, ready for re-review.

viamin (commented):
Generated no new comments. The fix correctly aligns the initial hidden state (`hidden` HTML attribute) with what the Stimulus controller toggles (`el.hidden`), resolving the dual-hide bug where the CSS class and attribute independently applied `display: none`. The `overflow-x-auto` change and test additions are well-scoped.

PR #997, opened 2026-04-11
Fix #982: fix(knowledge): ArtifactStore RecordNotUnique rescue fails when content_hash collides across different scope_paths

### LINKED ISSUE #982: fix(knowledge): ArtifactStore RecordNotUnique rescue fails when content_hash collides across different scope_paths
## Problem

The knowledge collectors (`tree_sitter`, `symbol_index`, `config_key`) repeatedly fail with:

```
Validation failed: Content hash has already been taken
```

This has occurred **107 times** across projects 2, 3, 4, 5, 7, 8, and 11. Each failure is logged as `knowledge.collector_failed` and causes the collector run to be marked as `failed`, which surfaces as errors in the Knowledge Base section of the project page.

## Root Cause

There are two interacting issues:

### 1. Content hash collisions across different artifacts

The `KnowledgeArtifact` model has a **unique index** on `[collector_run_id, content_hash]` (`app/models/knowledge_artifact.rb:16`):

```ruby
validates :content_hash, presence: true, length: { maximum: 64 },
  uniqueness: { scope: :collector_run_id }
```

Two artifacts within the same collector run can produce the same `content_hash` even when they have different `scope_path` or `identifier` values. For example, the `config_key` collector finds `ENV["RAILS_ENV"]` referenced in multiple files — each produces a different `scope_path` (e.g., `test/test_helper.rb` vs `config/environments/test.rb`) but the **content** (and therefore the `content_hash`) is identical: `ENV["RAILS_ENV"]`.

The `config_key` collector's `deduplicate` method (`app/services/knowledge/collectors/config_key_collector.rb:84`) only deduplicates by `[identifier, scope_path]`:

```ruby
def deduplicate(artifacts)
  artifacts.uniq { |a| [a[:identifier], a[:scope_path]] }
end
```

This allows two artifacts with the same `content_hash` but different `scope_path` values to pass through to `ArtifactStore`, where the second one violates the unique constraint.

The same issue affects `tree_sitter` and `symbol_index` collectors when they produce artifacts with identical code content (e.g., identical class definitions in different files, or the same symbol referenced multiple ways).

### 2. The RecordNotUnique rescue doesn't handle this case

`ArtifactStore#store_artifact` (`app/services/knowledge/artifact_store.rb:45-58`) already has a `rescue ActiveRecord::RecordNotUnique` handler for race conditions:

```ruby
  existing_after_conflict = find_existing_artifact(data, content_hash)

  if existing_after_conflict
    reassign_to_current_run(existing_after_conflict)
  else
    raise  # <-- This re-raises, causing the collector to fail
  end
end
```

The problem is that `find_existing_artifact` (line 62-73) queries by `project`, `collector_type`, `artifact_type`, `scope_path`, `identifier`, `content_hash`, AND `status: 'active'`. When the conflict is from a different `scope_path`/`identifier` with the same `content_hash`, this query returns `nil`, and the error is re-raised.

## How to Reproduce

1. Add a project that references the same env var (e.g., `ENV["RAILS_ENV"]`) in multiple files
2. Trigger knowledge collection
3. Observe the `config_key` collector fails with "Content hash has already been taken"
4. Check `log/development.log` for `knowledge.collector_failed` with `collector_type: "config_key"`

## Suggested Fix

There are two complementary approaches:

### Option A: Fix the collectors to deduplicate by content_hash (prevention)

Update each affected collector's deduplication to also consider `content_hash` so duplicates never reach `ArtifactStore`. For `config_key`:

```ruby
# app/services/knowledge/collectors/config_key_collector.rb
def deduplicate(artifacts)
  artifacts.uniq { |a| [a[:identifier], a[:scope_path], Digest::SHA256.hexdigest(a[:content].to_s)] }
end
```

Similar changes may be needed for `tree_sitter` and `symbol_index` if they can produce duplicate content across different scope paths.

### Option B: Fix the ArtifactStore rescue to handle cross-scope hash collisions (defense)

Broaden the fallback query in the `RecordNotUnique` rescue to find the artifact by `collector_run_id + content_hash` alone (the columns covered by the unique index), without requiring `scope_path`/`identifier` to match:

```ruby
  existing_after_conflict = find_existing_artifact(data, content_hash)
  
  if existing_after_conflict
    reassign_to_current_run(existing_after_conflict)
  else
    # The conflict is from a different scope_path/identifier with the same content_hash.
    # Find it by the unique index columns alone.
    existing_after_conflict = KnowledgeArtifact.find_by(
      collector_run_id: collector_run.id,
      content_hash: content_hash,
      status: "active"
    )
    if existing_after_conflict
      reassign_to_current_run(existing_after_conflict)
    else
      raise
    end
  end
end
```

**Recommendation**: Implement both — Option A prevents the issue at the source, Option B makes the store resilient to any future collectors that might introduce similar duplicates.

## Files to Change

- `app/services/knowledge/artifact_store.rb` — broaden the `RecordNotUnique` rescue fallback query
- `app/services/knowledge/collectors/config_key_collector.rb` — add `content_hash` to deduplication
- `app/services/knowledge/collectors/tree_sitter_collector.rb` — verify `deduplicate_results` handles content hash collisions
- `app/services/knowledge/collectors/symbol_index_collector.rb` — verify deduplication handles content hash collisions

viamin:
## Summary

fix(knowledge): ArtifactStore RecordNotUnique rescue fails when content_hash collides across different scope_paths

See #982 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #982

### DIFF
--- a/app/services/knowledge/artifact_store.rb
+++ b/app/services/knowledge/artifact_store.rb
@@ -47,8 +47,11 @@ module Knowledge
             mark_prior_stale(data)
             create_artifact(data, content_hash)
           end
-        rescue ActiveRecord::RecordNotUnique
+        rescue ActiveRecord::RecordNotUnique, ActiveRecord::RecordInvalid => e
+          raise unless content_hash_conflict?(e)
+
           existing_after_conflict = find_existing_artifact(data, content_hash)
+          existing_after_conflict ||= find_by_run_and_hash(content_hash)
 
           if existing_after_conflict
             reassign_to_current_run(existing_after_conflict)
@@ -71,6 +74,18 @@ module Knowledge
           status: "active"
         )
         .first
+    end
+
+    # Broadened lookup using only the columns covered by the unique index
+    # (collector_run_id, content_hash). This handles cross-scope hash
+    # collisions where a different scope_path/identifier produced the same
+    # content_hash within the same collector run.
+    def find_by_run_and_hash(content_hash)
+      KnowledgeArtifact.find_by(
+        collector_run: collector_run,
+        content_hash: content_hash,
+        status: "active"
+      )
     end
 
     def reassign_to_current_run(artifact)
@@ -161,6 +176,17 @@ module Knowledge
           sequence: chunk[:sequence] || index,
           status: "active"
         )
+      end
+    end
+
+    def content_hash_conflict?(exception)
+      case exception
+      when ActiveRecord::RecordNotUnique
+        true
+      when ActiveRecord::RecordInvalid
+        exception.record.errors[:content_hash].any?
+      else
+        false
       end
     end
 
--- a/app/services/knowledge/collectors/config_key_collector.rb
+++ b/app/services/knowledge/collectors/config_key_collector.rb
@@ -81,7 +81,9 @@ module Knowledge
       end
 
       def deduplicate(artifacts)
-        artifacts.uniq { |a| [ a[:identifier], a[:scope_path] ] }
+        artifacts
+          .uniq { |a| [ a[:identifier], a[:scope_path] ] }
+          .uniq { |a| Digest::SHA256.hexdigest(a[:content].to_s) }
       end
 
       def ast_grep_log_component
--- a/app/services/knowledge/collectors/symbol_index_collector.rb
+++ b/app/services/knowledge/collectors/symbol_index_collector.rb
@@ -56,7 +56,7 @@ module Knowledge
           end
         end
 
-        artifacts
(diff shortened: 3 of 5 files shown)


viamin (commented):
Generated no new comments. The PR correctly implements both the prevention (collector-level content_hash deduplication) and defense (broadened ArtifactStore rescue with `find_by_run_and_hash` fallback) approaches recommended in #982. The `content_hash_conflict?` guard properly distinguishes content_hash violations from other potential errors on the `RecordInvalid` path, and the test coverage validates the cross-scope collision scenario.

PR #1000, opened 2026-04-11
Fix #944: fix(scanner): enforce paid_agent max_review_rounds to prevent infinite review loops

### LINKED ISSUE #944: fix(scanner): enforce paid_agent max_review_rounds to prevent infinite review loops
## Summary

The `paid_agent` review method has a `max_review_rounds` termination config (default: 3 in `project.rb:25`) but no code enforces it. A PR could cycle through unlimited paid-agent review rounds without hitting any limit, causing runaway agent run costs and stalling the PR in draft.

### No review-round counter for paid_agent

- `draft_review_count` (`scan_paid_prs_activity.rb:90-93`) counts `create_pr` follow-up runs, not review-goal runs. It is incremented by `RecordDraftReviewActivity` which is only called from `start_draft_followup_workflow` for create_pr runs.
- `followup_limit_reached?` (`scan_paid_prs_activity.rb:341-343`) uses `max_pr_followup_runs`, a separate global limit unrelated to per-method review rounds.
- There is no counter scoped to `goal: "review"` runs per PR.

### The bot account makes this easier to implement

With a dedicated bot account (registered in `PROVIDER_BOT_USERNAMES`), the scanner can count reviews from the bot login on the PR directly via `fetch_reviews`, rather than querying AgentRun records. This is consistent with how Copilot and Codex reviews are tracked.

### Option A: Count bot reviews on the PR (preferred with bot account)

In `scan_draft_pr` and `check_review_bot_status`, when `paid_agent` is enabled:

1. Count reviews from the paid-agent bot login on the PR
2. Compare against `review_method_config("paid_agent").dig("termination", "max_review_rounds")`
3. If the count equals or exceeds the limit, emit an `escalate_to_owner` trigger instead of another `review_bot_review_pending`

### Option B: Count review-goal AgentRuns

1. Add a `review_round_count` field to Issue (or compute it on-the-fly)
2. Count `AgentRun.where(goal: "review", source_pull_request_number: pr_number).completed`
3. Enforce the limit in the scanner before queuing a new review run

### Integration points

- `ScanPaidPrsActivity#scan_draft_pr` — check review round count before emitting `review_bot_review_pending`
- `ScanPaidPrsActivity#scan_ready_pr` — check review round count in `no_outstanding_review_feedback?`
- `GitHubPollWorkflow#handle_review_bot_review_pending` — skip triggering a new review run when the limit is reached

## Acceptance criteria

- [ ] paid_agent review rounds are counted per PR
- [ ] Scanner escalates to owner when `max_review_rounds` is reached
- [ ] No new review-goal runs are queued after the limit
- [ ] Escalation comment mentions the review round limit was reached
- [ ] The limit is configurable per-project via `review_settings.methods.paid_agent.termination.max_review_rounds`

## Test plan

- Draft PR with paid_agent enabled, 3 review rounds completed → scanner emits `escalate_to_owner`, not `review_bot_review_pending`
- Draft PR with paid_agent enabled, 2 review rounds completed, findings still present → scanner queues another review
- Ready-phase auto-merge blocked when review rounds are at the limit with unresolved findings

## Related

- Bot account setup issue — enables counting by bot login
- #915 — auto-triggering review runs (must respect round limits)
- #830 — idle timeout for review runs (rounds may not complete if runs timeout)

viamin:
## Summary

fix(scanner): enforce paid_agent max_review_rounds to prevent infinite review loops

See #944 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #944


viamin:
## Agent Update

Commit succeeded. The fix was straightforward: Ruby 3.4 removed `OpenStruct` from auto-loaded constants, so `mark_escalated_activity_spec.rb:76` was failing with `NameError: uninitialized constant OpenStruct`. Added `require "ostruct"` at the top of the spec file.

All checks pass:
- **Rubocop**: 721 files, no offenses
- **Temporal specs**: 970 examples, 0 failures
- **Pre-commit hooks**: passed

viamin (commented):
Generated no new comments. The PR cleanly implements Option A from #944 — counting bot reviews on the PR via `PROVIDER_BOT_USERNAMES` and enforcing `max_review_rounds` in both the draft-phase escalation path and the `check_review_bot_status` suppression path. The workflow change correctly forwards the escalation reason via `.compact`. Tests cover the key scenarios (at limit, below limit, nil config, ready phase). Code follows existing patterns in the file.

viamin:
## Agent Update

Merged `origin/main`, resolved the scanner conflict, and fixed the review-loop bug so a PR can still advance on the final clean `paid_agent` round instead of being escalated. I also kept the ready-phase behavior from `main` aligned by suppressing impossible `review_bot_review_pending` requests once the paid-agent round budget is exhausted, and added a regression spec for the clean terminal-round case.

Validation passed: `RAILS_ENV=test bin/rails db:prepare`, `bundle exec rubocop`, and the full `RAILS_ENV=test bundle exec rspec` run completed with `5971 examples, 0 failures, 2 pending` (the existing Qdrant live specs remain pending when Qdrant is unavailable). The commit is `5edd7e08` with message `fix(scanner): resolve main merge and honor clean terminal paid_agent reviews`.

viamin:
@codex review

viamin:
## Agent Update

Let me try to merge origin/main and see what conflicts arise.

Good - already up to date with origin/main. No merge conflicts. Let me now look at the key files mentioned in the review comments.

Now let me understand the review comments in detail:

**P1 Review (git_hub_poll_workflow.rb:240)**: The `MarkEscalatedActivity` input now includes `reason` parameter. This is a change in the command shape that could cause nondeterministic replay for in-flight workflows. Need to gate behind a Temporal patch.

**P2 Review (scan_paid_prs_activity.rb:182)**: The escalation condition at line 181 uses `pending_triggers.any? || blocking_triggers.any?` which can include triggers from other bots (e.g., Copilot). The escalation should only fire based on paid_agent-specific blocking state.

Let me now look at the MarkEscalatedActivity to understand the current interface:

Now let me look at the diff to understand what changes this PR introduced:

Now I have a clear picture. Let me address the review comments:

**P1**: Gate the `reason` payload behind a Temporal patch in `handle_escalate_to_owner`
**P2**: Restrict escalation to only check paid_agent-specific blocking triggers, not any bot's triggers

Let me also look at the existing test to understand how tests mock `patched`:

Now let me see the escalate_to_owner spec section:

Now I have enough context. Let me install dependencies and then make the changes.

Disk space issue with yarn, but that shouldn't affect our Ruby changes. Let me check if we can still run the tests:

Bundle install didn't fully succeed. Let me retry:

Good. Now let me implement the fixes. Let me start with the P1 fix (Temporal patch for escalation reason payload):

**P1: Gate escalation reason payload behind Temporal patch**

Now fix the P2 issue (restrict escalation to paid_agent-specific blocking state). Let me look more carefully at line 181:

The current code at line 181:
```ruby
if paid_agent_rounds_exhausted && (pending_triggers.any? || blocking_triggers.any?)
```

The issue is that `pending_triggers` and `blocking_triggers` can include triggers from other bots (e.g., Copilot). We need to check only paid_agent-specific blocking state.

Looking at the trigger types:
- `pending_triggers` = triggers with type `review_bot_review_pending` 
- `blocking_triggers` = triggers that are NOT `review_bot_review_pending` or `paid_agent_review_pending`
- `sidecar_triggers` = triggers with type `paid_agent_review_pending`

The reviewer's point is: when paid_agent rounds are exhausted, we should only escalate if the non-clean review is FROM paid_agent specifically. If another bot (Copilot) has pending/blocking triggers, that's fine — those can still be handled.

But wait - let me re-read the code more carefully. The escalation fires when `paid_agent_rounds_exhausted && (pending_triggers.any? || blocking_triggers.any?)`. The `pending_triggers` are `review_bot_review_pending` type triggers, and `blocking_triggers` are the rest.

The key insight from the reviewer: In mixed-method projects, if Copilot has unresolved threads (which would be in `blocking_triggers`), and paid_agent rounds are exhausted, the code escalates even though Copilot's cycle can still proceed. We should only escalate when the blocking state is specifically from paid_agent.

Let me think about what "paid_agent-specific blocking state" means here. If paid_agent rounds are exhausted:
1. If the latest review is from paid_agent and has comments → the PR is blocked by paid_agent with no way to resolve it → escalate
2. If the latest review is from Copilot and has comments → Copilot can still proceed → don't escalate
3. If there are review_bot_review_pending triggers requesting another review → but that review would be for any bot, not specifically paid_agent

The correct fix: instead of checking `pending_triggers.any? || blocking_triggers.any?`, check whether there's actually a paid_agent-specific blocker — i.e., the latest review from paid_agent was non-clean. If the non-clean review is from a different bot, don't escalate.

Now I need to understand what triggers could be present. The `review_bot_triggers` come from `check_review_bot_status`. When paid_agent rounds are exhausted, we need to check if the blocking state is specifically from paid_agent.

Looking at the trigger types in `review_bot_triggers`:
- `review_bot_review_pending` → needs another bot review
- `paid_agent_review_pending` → needs another paid_agent review  
- `review_bot_comments` → bot left comments (body-only)
- `review_bot_threads` → bot left unresolved threads

The escalation should only fire when paid_agent itself is the blocker. Let me check what the trigger would look like if paid_agent is the one blocking:

When paid_agent rounds are exhausted and the latest review is from paid_agent with comments, the `check_review_bot_status` method returns:
- `review_bot_comments` with "(body-only)" suffix (for body-only bots like paid_agent)

When Copilot has unresolved threads:
- `review_bot_review_pending` + `review_bot_comments` + `review_bot_threads`

So the fix should be: only escalate when the blocking triggers are specifically from paid_agent. The simplest approach is to check whether any non-paid_agent bot triggers exist that could still be resolved. If only paid_agent triggers are present and exhausted, escalate. If other bots also have triggers, let them continue.

Let me look more carefully at what `pending_triggers` and `blocking_triggers` represent:

The reviewer's point makes sense. In the current code:

1. `pending_triggers` = `review_bot_review_pending` triggers — these request another review from ANY configured review bot. In a mixed config (Copilot + paid_agent), this trigger might be requesting a Copilot review, not a paid_agent one. But when `paid_agent_review_rounds_exhausted?` is true and this trigger exists, it escalates, which prematurely ends Copilot's cycle too.

2. `blocking_triggers` = `review_bot_comments`, `review_bot_threads` — these could be from Copilot.

The fix: When paid_agent rounds are exhausted, only escalate if the triggers are specifically from paid_agent. If there are non-paid_agent bot triggers, those bots can still make progress.

I think the cleanest approach is: when `paid_agent_rounds_exhausted` is true, check whether the `blocking_triggers` and `pending_triggers` are exclusively paid_agent-related. But a simpler approach that the reviewer seems to suggest: just check if the latest review that's causing the block is from paid_agent.

Looking at the code flow, when `check_review_bot_status` is called, it checks the `latest` allowed bot review. The issue is that `latest` could be from Copilot or paid_agent. The escalation should only happen when it's specifically paid_agent that's blocking.

The simplest approach: check if the latest blocking review is from paid_agent. If it's from another bot, don't escalate.

Now I need to add the `paid_agent_is_latest_blocker?` method. Let me add it near the other paid_agent helpers:

Now let me update the tests. First, let me update the workflow spec for the P1 fix (Temporal patch):

I need to update the existing tests and add a test for the unpatched path. The tests currently test the patched behavior (with reason). I need to:
1. Ensure the patched path sends reason
2. Add a test for the unpatched path (no reason)
3. Update the "omits reason key when trigger has no details" test

Let me check how patched is stubbed in this spec context:

I see that the `before` block stubs `patched` only for `"draft-followup-direct-start-v1"`. The new patch `"escalation-reason-payload-v1"` needs to be stubbed too. When `patched` is called with an unrecognized ID, it will return nil (since it's only stubbed for specific IDs). I need to add stubbing for the new patch.

Let me update the tests:

Now update the test that checks for the unpatched path (the "omits reason key" test). Currently it tests that when trigger has no details, the reason key is omitted. But with the patch approach, the unpatched path always omits reason regardless of details. Let me rewrite the tests:

Now let me add a test for the P2 fix — a mixed-method project where Copilot has unresolved threads but paid_agent rounds are exhausted. Let me find the right place in the scan_paid_prs spec:

Now let me find the `stub_github_for_pr` helper to understand how to set up a mixed-method project test with Copilot:

Now let me look at how `enabled_review_bot_logins` works and what Copilot logins look like:

Good. Now I need to add a test for the P2 fix. The test should verify that in a mixed-method project (paid_agent + copilot), when paid_agent rounds are exhausted but Copilot has unresolved threads, the system does NOT escalate.

I'll add the new test context after line 2003:

Let me get more context:

Now let me also update the TODOS and run the tests:

Let me proactively scan the entire diff for similar issues. Let me review the full current state of the changed files:

The diff looks clean. Let me scan for any issues:

1. The P1 fix properly gates the `reason` payload behind a Temporal patch with a clear TODO and patch ID.
2. The P2 fix introduces `paid_agent_is_latest_blocker?` that checks the latest bot review specifically.
3. Tests cover both patched/unpatched paths and the mixed-method case.

Let me also check if there are any other places in the codebase that might have the same pattern of escalation with triggers from non-specific bots:

Lines 130 and 134 are for the draft review limit and consecutive failures breaker, which are unrelated to the paid_agent-specific logic. Those are fine.

Now let me run the tests. First lint:

Now run the targeted tests:

All 218 tests pass. The coverage warning is expected since we only ran a subset of specs. Let me verify the new tests specifically passed:

All tests pass including the new ones. Let me now run the full rubocop on the changed files:

Now commit the changes:

The output was truncated but I see the commit ran through the pre-commit hooks which include full test suite. Let me check if the commit succeeded:

The commit didn't get created — it seems the pre-commit hook failed. Let me check the tail of the output for the result:

The pre-commit hook ran the full test suite which failed (I saw some F's in the output). Let me check what failures occurred by running just the failing tests. Let me look at the output more carefully:

The failures are:
1. Qdrant live specs (expected - Qdrant not available, these are pre-existing)
2. CSS asset not found (because `yarn install` failed due to disk space - pre-existing issue)

These are all pre-existing failures not related to our changes. Let me verify this by checking if these same tests were failing before our changes:

All 341 failures are the CSS asset issue (yarn install failed so CSS is not built). This is a pre-existing infrastructure issue. Let me try to fix the yarn install:

viamin:
@codex review

viamin:
## Agent Update

Let me start by understanding the current state of the branch and any conflicts.

Good - already up to date with `origin/main`, no merge conflicts. Let me now understand the review comment and the code.

The review comment references `app/temporal/activities/scan_paid_prs_activity.rb:1322`. Let me read that file and understand the escalation logic.

Good - I can see the `paid_agent_is_latest_blocker?` method already exists at line 1314. Now let me find where the escalation logic uses it, and where the review comment is pointing to (line 1322).

I see the fix is already applied at line 184. The escalation condition now uses `paid_agent_is_latest_blocker?` instead of just `pending_triggers.any? || blocking_triggers.any?`. Let me also check line 741 to understand the other usage:

Now let me look at the full diff of this PR to understand all changes:

Now I have a clear picture of the full diff. The P1 review comment points to line 1322 (which is in the `paid_agent_is_latest_blocker?` method). The reviewer says:

> This helper treats **any** pending/blocking review trigger as a paid-agent blocker when the latest bot review is from `paid_agent`. In mixed-method projects, `check_review_bot_status` can emit blocking triggers from another bot (for example unresolved Copilot threads) while the latest bot review is from `paid_agent`; once paid-agent rounds are exhausted, `detect_draft_triggers` will escalate anyway, even though the remaining blocker is not paid-agent and automation can still make progress.

Let me think about this carefully. The current implementation of `paid_agent_is_latest_blocker?` checks:
1. Are there pending/blocking triggers?
2. Is the latest allowed bot review from paid_agent?

But the reviewer's point is: even if the latest review is from paid_agent, the triggers could be from Copilot. For example:
- Latest review is from paid_agent (clean or with comments)
- But Copilot has unresolved threads → blocking_triggers contains Copilot's threads
- `paid_agent_is_latest_blocker?` returns true because latest review IS from paid_agent AND triggers exist
- Escalation fires, but the actual blocker is Copilot, not paid_agent

The fix should verify that the triggers are specifically from paid_agent, not just that triggers exist. Let me look at what information is available in the triggers:

Now I understand the issue more clearly. The problem with `paid_agent_is_latest_blocker?` is:

When we have a mixed-method project (paid_agent + copilot):
1. paid_agent has 2 reviews, rounds exhausted
2. Copilot has a more recent review with unresolved threads
3. `latest_allowed_bot_review` returns the **latest** review among all allowed bots
4. If Copilot's review is more recent, `latest[:user_login]` is Copilot → `paid_agent_is_latest_blocker?` returns false → no escalation (correct!)
5. But if paid_agent's review is the latest (no Copilot review yet or Copilot reviewed earlier), AND Copilot has unresolved threads from an earlier review, then `latest[:user_login]` is paid_agent → `paid_agent_is_latest_blocker?` returns true → escalation fires

Wait, let me re-read the review comment more carefully:

> `check_review_bot_status` can emit blocking triggers from another bot (for example unresolved Copilot threads) while the latest bot review is from `paid_agent`

So the scenario is:
- paid_agent's review is the latest chronologically
- But Copilot has unresolved threads from an earlier review
- `check_review_bot_status` emits blocking triggers for Copilot's threads
- The triggers are `review_bot_review_pending` and `review_bot_comments` + thread triggers
- `paid_agent_is_latest_blocker?` sees triggers exist AND latest review is from paid_agent → returns true
- But the triggers are from Copilot, not from paid_agent!

The fix: we need to check whether the triggers are actually attributable to paid_agent, not just whether triggers exist and the latest review is from paid_agent.

But the triggers don't carry bot-specific attribution. They carry generic types like `review_bot_review_pending`, `review_bot_comments`, `review_bot_threads`. Let me think about this differently...

The key insight: when `check_review_bot_status` returns the `:has_comments` status with unresolved threads, it means the `latest` review was non-clean. If `latest` is from paid_agent and is non-clean, that's a paid_agent blocker. But if there are also Copilot threads, those get mixed in.

Actually, let me re-examine the flow. `check_review_bot_status` uses `latest_allowed_bot_review` to determine the "latest" review. If that's from paid_agent and the status is `:has_comments`, the triggers are about paid_agent's comments. Copilot's unresolved threads are separate - they come from `review_bot_thread_triggers(unresolved_threads)`.

Let me look at how thread triggers are handled:

Now I see: `review_bot_thread_triggers` checks for any bot (not specifically paid_agent). So in a mixed-method project with Copilot having unresolved threads, those thread triggers come through.

Let me trace the exact scenario the reviewer describes:

1. Mixed-method project: paid_agent + copilot
2. paid_agent has given 2 reviews (limit reached), latest review is from paid_agent
3. Copilot has unresolved threads from an earlier review
4. `check_review_bot_status` is called:
   - `latest` = paid_agent review (most recent)
   - `status` = `:has_comments` (paid_agent review was non-clean)
   - Since unresolved_threads are present, goes to line 760
   - `bot_thread_triggers = review_bot_thread_triggers(unresolved_threads)` → gets Copilot's threads
   - `bot_thread_triggers.any?` → true
   - Returns triggers at line 771-774: `review_bot_review_pending` + `review_bot_comments` + Copilot thread triggers
5. Back in the main scan logic:
   - `pending_triggers` = `review_bot_review_pending` triggers (from the above)
   - `blocking_triggers` = `review_bot_comments` + `review_bot_threads` (Copilot's)
   - `paid_agent_rounds_exhausted` = true
   - `paid_agent_is_latest_blocker?` → latest review is from paid_agent, triggers exist → **true**
   - **Escalates!** But the escalation is wrong because the remaining blocker (Copilot threads) can still be addressed by Copilot.

So the reviewer is correct. The current `paid_agent_is_latest_blocker?` is too broad. It needs to check whether the triggers are specifically attributable to paid_agent.

The fix should be: only escalate when the blocking triggers are from paid_agent specifically. We can determine this by checking if:
1. The latest review is from paid_agent (already checked)
2. The triggers are not from other bots - specifically, there are no unresolved threads from non-paid_agent bots

Let me think about the best approach. The triggers don't carry per-bot attribution. But we can look at the trigger types:
- `review_bot_threads` trigger: this comes from `review_bot_thread_triggers` which checks ALL bots
- `review_bot_comments` trigger: generic
- `review_bot_review_pending` trigger: generic

A better approach: instead of checking trigger types, check whether there are unresolved threads from non-paid_agent bots. If there are, those bots can still make progress, so don't escalate.

Actually, the cleanest approach is to make `paid_agent_is_latest_blocker?` also verify that there are no non-paid_agent bot threads blocking progress. Let me implement this:

Now I need to understand the approach. The issue is that `paid_agent_is_latest_blocker?` currently only checks if the latest review is from paid_agent. But triggers could be from other bots (e.g., Copilot unresolved threads). 

The fix: when checking if we should escalate for paid_agent round exhaustion, we should also verify that there are no non-paid_agent bot threads that could still be addressed. If there are non-paid_agent bot threads, those bots can still make progress, so we shouldn't escalate.

Let me look at how the method is called to understand what information is available:

Now let me understand the trigger flow more carefully. The triggers come from `check_review_bot_status`. In the mixed-method scenario:
- paid_agent has the latest review (non-clean)
- Copilot has unresolved threads from an earlier review

When `check_review_bot_status` runs with `status = :has_comments` and `unresolved_threads` present:
1. `bot_thread_triggers = review_bot_thread_triggers(unresolved_threads)` - this picks up Copilot's threads
2. Since `bot_thread_triggers.any?` is true, it goes to line 771-774:
   ```ruby
   triggers = [ { type: "review_bot_review_pending", ... } ]
   triggers << { type: "review_bot_comments", ... }
   triggers.concat(bot_thread_triggers)
   ```

So the triggers include both `review_bot_review_pending` (requesting another review) AND Copilot's thread triggers.

Back in the main scan:
- `pending_triggers` = the `review_bot_review_pending` trigger
- `blocking_triggers` = `review_bot_comments` + `review_bot_threads` (from Copilot)

The escalation check: `paid_agent_rounds_exhausted && paid_agent_is_latest_blocker?(...)` = true → escalates.

But this is wrong because the `review_bot_review_pending` trigger could be fulfilled by Copilot, not just paid_agent. And the `review_bot_threads` trigger is from Copilot.

The fix should check: among the triggers, are there any that are specifically from non-paid_agent bots that could still be addressed? If so, don't escalate.

The cleanest approach: check whether there are unresolved threads from non-paid_agent bots. This information is available via `unresolved_threads` in the outer scope, but it's not passed to `paid_agent_is_latest_blocker?`.

Actually, let me reconsider. The simpler approach is to look at what's in `blocking_triggers`. If `blocking_triggers` contains `review_bot_threads`, those are from thread-based bots (Copilot). If paid_agent is body-only, any `review_bot_threads` triggers must be from a different bot. So we can check:

If there are `review_bot_threads` in blocking_triggers AND paid_agent is body-only → those threads are from a non-paid_agent bot → don't escalate.

But actually, a more general approach: check if there are unresolved threads from non-paid_agent bots. Let me add a method to check this and use it in the escalation decision.

I see `unresolved_threads` is available in the outer scope. Let me update the escalation logic to pass `unresolved_threads` to the check so we can verify that there are no non-paid_agent bot threads.

Now let me implement the fix. The approach:

1. Add a method `non_paid_agent_bot_threads_present?` that checks if there are unresolved threads from bots other than paid_agent
2. Update `paid_agent_is_latest_blocker?` to also check that there are no non-paid_agent bot threads that could still be addressed

Actually, a simpler approach: modify `paid_agent_is_latest_blocker?` to accept `blocking_triggers` and check that none of the blocking triggers include `review_bot_threads` from non-paid_agent bots. Since `review_bot_threads` triggers come from `review_bot_thread_triggers` which checks ALL bots, and paid_agent is body-only (doesn't use threads), any `review_bot_threads` trigger must be from a non-paid_agent bot.

So the check is: if `blocking_triggers` contains a `review_bot_threads` trigger, don't escalate because a non-paid_agent bot still has unresolved threads.

Now let me update the test for the mixed-method case. The existing test has Copilot threads, so it should already pass. Let me check:

This test covers the case where Copilot's review is more recent (latest) than paid_agent's. But it doesn't cover the specific scenario the reviewer flagged: **paid_agent's review is the latest, but Copilot has unresolved threads from an earlier review**. In the current test, Copilot's review (id: 3) is more recent than paid_agent's reviews, so `latest_allowed_bot_review` returns Copilot → `paid_agent_is_latest_blocker?` returns false because the latest review isn't from paid_agent.

I need to add a test for the reviewer's scenario: paid_agent is the latest reviewer, but Copilot has unresolved threads. Let me also add a test where paid_agent is the latest blocker and all triggers ARE from paid_agent (to confirm escalation still works correctly).

Now let me also verify the escalation still works when ALL triggers are from paid_agent (no Copilot threads). The existing test at line 1797 ("when paid_agent review round limit is reached in draft phase") covers this case since there are no Copilot threads. Let me verify:

Good. That test has `review_threads: []` (no Copilot threads), and only paid_agent reviews. The escalation should still work because there are no `review_bot_threads` in blocking_triggers. The triggers will be `review_bot_review_pending` and `review_bot_comments` (body-only), which don't include `review_bot_threads`.

Now let me run the targeted tests to verify:

Many gems are missing. Let me do a full install:

Permission issue. Let me check if there's a local vendor/bundle path or a different setup:

Good. Now let me set up the database and run the tests:

All 219 tests pass, 0 failures. The coverage warning is expected since we only ran a subset of specs.

Now let me run rubocop on the changed files:

Now let me update the todos and commit:

Let me review the final diff before committing:

The diff is clean and focused. Now let me commit:

Let me check the final result of the test run:

341 failures. Let me check if these are all pre-existing CSS asset failures:

These are request spec failures, likely related to CSS/JS assets not being compiled (yarn install failed due to disk space). These are pre-existing infrastructure issues, not related to our changes. Let me confirm by checking if these same tests fail on the base branch:

All 341 failures are in request/system/mailer/routing specs - all CSS/JS asset related (pre-existing). No failures in our temporal activity/workflow specs.

Let me verify the commit succeeded:

The commit didn't succeed because the pre-commit hook ran the full test suite which had 341 pre-existing failures. The pre-commit hook considers these as failures. Let me check what the pre-commit hook actually does:

The pre-commit hook runs the full test suite. The 341 failures are all pre-existing (CSS/JS asset compilation issues). Let me check if these same failures exist on the parent commit too (before my changes):

The failures are pre-existing. The pre-commit hook runs the full suite which includes these pre-existing failures. Let me try running just our changed test files through the hook by checking if there's a way to run targeted tests. Actually, the instructions say to not use `--no-verify`. The pre-existing failures are from yarn/CSS asset issues that are environment-specific. Let me try to fix the yarn issue first:

Disk space issue is real. The vendor/bundle we installed is taking space. Let me try cleaning up and retrying:

viamin:
@codex review

viamin:
## Agent Update

Good, dependencies are installed. Now let me abort the previous merge attempt and re-merge cleanly:

Now let me look at the review comment. I need to find `paid_agent_review_rounds_exhausted?` and the `max_rounds` comparison:

I can see the issue clearly. In `paid_agent_review_rounds_exhausted?` at line 1364, `max_rounds` is compared directly with `<= 0`, but `max_review_rounds` comes from JSON settings and could be a string. In `check_paid_agent_review_status` at line 888, `.to_i` is already used. Let me also check line 1370 where `paid_agent_limit_reason` uses `|| 0`:

The fix is clear. `paid_agent_max_review_rounds` returns the raw JSON value (could be string). In `paid_agent_review_rounds_exhausted?` at line 1364, `max_rounds <= 0` will raise `ArgumentError` if `max_rounds` is a string. Same at line 1366 with `>= max_rounds`.

The fix: coerce `max_rounds` to integer once in `paid_agent_max_review_rounds`, which is the single source of truth for this value.

Now `paid_agent_max_review_rounds` always returns an Integer (or nil). The callers that do `.to_i` on the result will still work fine (no-op on Integer). Let me verify the callers:

Line 888 still has `.to_i` — that's fine, it's now a no-op but harmless. Let me also check line 1371 where `|| 0` is used:

```ruby
max_rounds = paid_agent_max_review_rounds(project) || 0
```

This works correctly: if the method returns nil, it falls back to 0 (Integer). If it returns an Integer, that's used directly.

Now let me do a proactive scan of the full diff to look for similar issues — other places where JSON settings values are compared without coercion:

Now let me scan the full diff for similar issues — JSON settings values used without coercion, missing guard clauses, etc:

The diff looks clean and the tests are comprehensive. Now let me also add a test for the specific coercion fix — verifying that string `max_review_rounds` doesn't cause an error:

Let me add a test for the string coercion case. I'll insert it after the "max_review_rounds is not set" test:

I'll add a test for the string coercion case after line 2005 (after the "ready phase" context):

Now let me also check the workflow file mentioned in the conversation comments:

Good. The workflow changes look clean — the Temporal patch is properly implemented. Now let me run lint and tests:

My new test failed. Let me check what's happening:

The model has a validation that prevents saving a string value through normal means. The coercion fix is still valid for legacy data at the DB level. Let me adjust the test to bypass the validation or test the method directly:

The validation prevents saving strings through the model. But legacy data could have strings. I'll update the test to use `update_column` to bypass validations:

Test passes now. Let me run the full targeted spec suite:

All 213 tests pass. Now let me also check if the MarkEscalatedActivity spec needs `require "ostruct"` (mentioned in conversation comments):

Already handled. Now let me build CSS/JS assets so the pre-commit hook can pass:

Now let me commit. First, let me verify the full diff is clean:

The diff is clean and focused. Now let me commit:

The commit succeeded. All 6022 examples passed with 0 failures (2 pending are pre-existing Qdrant specs). Let me verify the final state:

1. **Merge conflicts resolved**: Merged `origin/main` cleanly (no conflicts).
2. **P2 review addressed** (`scan_paid_prs_activity.rb:1343-1347`): `paid_agent_max_review_rounds` now coerces the raw JSON value with `to_i` before returning, so `paid_agent_review_rounds_exhausted?` can safely compare `max_rounds <= 0` and `>= max_rounds` without `ArgumentError` when legacy/manual data stores the setting as a string.
3. **Test added**: New spec verifies escalation works correctly when `max_review_rounds` is stored as a string `"3"` (bypasses model validation via `update_column`).
4. **All checks pass**: 6022 examples, 0 failures, 0 lint offenses. Commit `766781f7`.

viamin:
@codex review

viamin:
## Agent Update

No merge conflicts — already up to date with `origin/main`.

Now let me understand the review comment. The P1 review says: when `unresolved_threads` is `nil` (which happens on the skip-comment path or when the `review_threads` API call fails), the current code downgrades the state to `review_bot_comments` once the paid_agent round limit is reached. This removes `review_bot_review_pending`, and `scan_draft_pr` can immediately emit `escalate_to_owner` via the exhaustion check even though thread data is incomplete. A transient thread-fetch failure can trigger premature escalation.

Let me read the relevant code:

Let me read the exhaustion/escalation logic:

Let me read the broader context of the scan logic to understand where `unresolved_threads` is set relative to the escalation check:

Now I have a clear understanding of the issue. When `unresolved_threads` is `nil` (thread data unavailable due to skip-comment path or API failure) and paid_agent rounds are exhausted, the code emits `review_bot_comments` instead of `review_bot_review_pending`. This removes the pending status, allowing the escalation check at line 204 to fire prematurely.

The fix needs two parts:
1. Keep `review_bot_review_pending` status when thread data is unavailable (revert the downgrade)
2. Prevent escalation when data is incomplete

Let me install dependencies and implement the fix:

Good, dependencies installed and database prepared. Now let me implement the fix for the P1 review comment.

Now I'll make the fix. The approach:
1. When `unresolved_threads` is nil, always emit `review_bot_review_pending` with a `data_incomplete: true` marker (keeping the PR pending)
2. In the escalation check, skip escalation when any trigger has `data_incomplete: true`

Now add the guard in the escalation check:

Now let me update the tests. Let me find the relevant test sections:

Let me read the test areas that specifically test the `unresolved_threads.nil?` path and the escalation:

Now let me read the test areas around the API failure test and the string coercion test:

Now I need to add a test for the specific scenario the reviewer flagged: when `unresolved_threads` is nil (API failure) AND paid_agent rounds are exhausted, escalation should NOT fire. Let me add this test after the existing test at line 2146:

Now let me also check the existing "review_threads API fails" test at line 1565 to make sure it still passes. The test expects `review_bot_review_pending` when a Copilot review has comments and threads API fails. With our change, when `unresolved_threads` is nil, we always emit `review_bot_review_pending` with `data_incomplete: true` — that test should still pass.

Let me also verify the `scan_ready_pr` path. Let me check if there's a similar escalation check in the ready phase:

Good — escalation only happens in `scan_draft_pr` (line 207). The ready phase (`scan_ready_pr`) has no escalation check.

Now let me review the conversation comments from viamin. The conversation shows several updates from a previous agent's work, including:
1. Fixing `OpenStruct` require in spec
2. Merging origin/main and fixing review-loop bug
3. Adding Temporal patch for escalation reason payload
4. Adding `paid_agent_is_latest_blocker?` with non-paid_agent thread trigger check
5. Coercing `max_review_rounds` to integer

These were all addressed by previous commits. The latest comment is a review trigger. Let me check if there are any other actionable requests.

The conversation comments don't contain any new actionable requests beyond what the review comment raised (which I've already addressed). Let me proceed to the proactive scan.

Let me do a proactive scan of the full diff for similar issues:

Let me verify my current changes against the last commit:

The diff looks clean and focused. Let me also scan for any other places in the codebase that check `unresolved_threads.nil?` or interact with `data_incomplete`:

Let me also check that no other consumer of triggers would be confused by the `data_incomplete` key:

The workflow's `handle_review_bot_review_pending` at line 291 only checks `pending_trigger[:request_login]`. It doesn't inspect `data_incomplete`, so the extra key is harmless — the workflow will see the `review_bot_review_pending` trigger and attempt to request a review or start a followup. On the next scan, if threads are still unavailable, it'll stay pending. If threads become available, the normal flow proceeds.

This is correct behavior — when thread data is unavailable, the PR stays in its current phase and retries on the next scan cycle.

Now let me run lint and the targeted tests:

All 171 tests pass with 0 failures. The coverage warning is expected since we're only running a subset of specs. Let me also run the workflow specs:

All tests pass. Let me also run the mark_escalated_activity spec:

All 17 tests pass. Now let me commit:

One pre-existing test failure in `symbol_index_collector_spec.rb` (unrelated to our changes). Let me verify:

It passes when run alone — it's a flaky test (likely ordering-dependent). Let me try committing again:

Commit succeeded. 6023 examples, 0 failures, 0 lint offenses.

All tasks complete. Summary of changes:

**P1 Review Fix** (`scan_paid_prs_activity.rb:802-810, 204-208`): When `unresolved_threads` is nil (thread data unavailable due to API failure or skip-comment path), the code previously downgraded to `review_bot_comments` once paid_agent rounds were exhausted, removing the `review_bot_review_pending` status and allowing the escalation check to fire prematurely. The fix:

1. Always emits `review_bot_review_pending` with a `data_incomplete: true` marker when `unresolved_threads` is nil, regardless of round exhaustion status — keeping the PR blocked until thread data is available
2. Adds a guard in the escalation check: `!data_incomplete` — preventing escalation when any trigger carries incomplete data

**New test**: Verifies that when paid_agent rounds are exhausted AND the threads API fails, the scanner keeps `review_bot_review_pending` and does NOT escalate.

Commit `886f2af9`: 6023 examples, 0 failures, 0 lint offenses.

viamin:
@codex review

viamin:
## Agent Update

No merge conflicts. Now let me install dependencies and understand the current code state.

Good. Now let me understand the review comment. The key P1 review is about honoring the global review toggle before enforcing paid_agent limits.

Let me read the relevant files to understand the review comment and the current code.

Now let me read the current state of the key file to understand the review comment's exact location.

Now I understand the review comment. The reviewer says `paid_agent_max_review_rounds` at line 1347-1352 only checks `review_method_enabled?("paid_agent")` but NOT `review_enabled?` (the global toggle). Let me check what `review_enabled?` does and where these methods are called from.

Now let me check how the callers of `paid_agent_review_rounds_exhausted?` are gated:

Let me check what happens when `latest` is nil in `review_bot_review_status_from`:

Now I have a clear understanding. The fix is to add a `review_enabled?` guard to `paid_agent_max_review_rounds`. This prevents the entire paid_agent enforcement chain from activating when reviews are globally disabled.

Now I need to update the docstrings for the downstream methods that reference this behavior:

Now let me add a test for the case where `review_enabled?` is false but `paid_agent` method is enabled. Let me find the right place in the spec:

I'll add the new test context after the "not set" context (after line 1966):

Now let me do the proactive scan. Let me check if there are other places in the diff where paid_agent helpers are called without a `review_enabled?` guard:

Good — `review_bot_request_login` returns nil when `review_enabled?` is false. So line 792's `login && !paid_agent_review_rounds_exhausted?` check would fail on `login` being nil first, never reaching the `paid_agent_review_rounds_exhausted?` call. The guard in `paid_agent_max_review_rounds` is still correct as defense-in-depth.

Now let me run the targeted tests:

No tagged tests. Let me run the specific test file with the new test:

New test passes. Now let me run all the paid_agent-related tests together:

All 28 paid_agent tests pass. Let me also run the workflow specs:

All pass. Now let me run rubocop on the changed files:

Let me verify the full diff before committing:

Diff is clean and focused. Now commit:

Commit `520b2c29` succeeded. All tasks complete:

**P1 Review Fix** (`scan_paid_prs_activity.rb:1348`): Added `return nil unless project.review_enabled?` guard to `paid_agent_max_review_rounds`. This ensures that when reviews are globally disabled (`review_settings.enabled = false`) but the paid_agent method flag remains `true` (a supported config state), the entire round-limit enforcement chain short-circuits — no escalation, no trigger suppression. The existing `paid_agent_sole_review_method?` already had this guard; `paid_agent_max_review_rounds` was the missing one.

**New test**: Verifies that with `enabled: false` + `paid_agent.enabled: true` + `max_review_rounds: 2` and 2 paid_agent reviews present, no `escalate_to_owner` trigger is emitted.

viamin:
@codex review

viamin:
## Agent Update

Already up to date with `origin/main`. No merge conflicts.

Good. Now let me read the review comment and the relevant code.

Let me understand how codex (another body-only bot) generates triggers:

Now let me understand the full trigger flow and the codex-specific issue:

Let me trace the full trigger flow to understand how triggers are returned:

Let me check how the workflow handles triggers without `review_bot_review_pending`:

Now I have a clear picture of the issue. Let me read the full current `paid_agent_is_latest_blocker?` and surrounding context:

Now let me also check the `:no_review` path for the same class of issue:

Good. Now I understand the full picture. The review comment is correct: `paid_agent_is_latest_blocker?` only checks for `review_bot_threads` as evidence of non-paid-agent blockers, but codex (like paid_agent) is body-only and never produces thread triggers. In a paid_agent + codex project, escalation fires even though codex can continue.

There's also a related issue in the `:no_review` path (line 792): `review_bot_request_login` returns the codex login, but `paid_agent_review_rounds_exhausted?` suppresses the review request for codex too.

Now let me implement the fixes:

**Fix 1**: `paid_agent_is_latest_blocker?` — account for other enabled body-only bots (codex)

**Fix 2**: `:no_review` path — don't suppress codex review requests based on paid_agent exhaustion

Now let me verify both edits look correct:

Good. Now let me find the tests and add new ones:

Now let me find the mixed-method test that was previously added for Copilot threads:

Now let me read the helper and see where I should add the codex test:

Now let me find the right location to add the codex test. I'll add it after the "copilot has unresolved threads" test (line 2183):

Now let me also add a test for the `:no_review` path — paid_agent + codex with no reviews at all:

Now let me add a test for the `:no_review` path with paid_agent + codex where paid_agent rounds are exhausted:

Now let me also add a unit test for the `other_enabled_body_only_bots?` helper and verify the `paid_agent_is_latest_blocker?` behavior:

No existing unit tests for these. Let me run the tests now:

The test failure shows the `:has_comments` path is being exercised (not `:no_review`). Let me check `body_only_review_needs_followup?`:

The test needs fixing — the scenario has paid_agent reviews (not "no review"), so it goes through `:has_comments`, not `:no_review`. Let me fix the test expectation:

Now let me run the tests:

All 174 tests pass. Let me also run the workflow specs and the mark_escalated_activity spec:

All pass. Now let me run rubocop:

Now let me do the proactive scan of the full diff:

Let me also review the full PR diff against main to scan for any other similar issues:

The diff is clean and focused. Let me also scan the full PR diff for the same class of issues — any place where `paid_agent_review_rounds_exhausted?` is used without accounting for other bots:

Let me review lines 839 and 847 — these use `paid_agent_limit_reached_for_latest_review` to choose between exhausted and pending triggers. When other body-only bots are enabled, we should use pending triggers even when paid_agent is exhausted:

The escalation fix at line 207 already prevents premature escalation when codex is enabled. Lines 839/847 (body-only exhausted vs pending triggers) don't change the workflow outcome because the followup path is taken either way. Let me verify the `scan_ready_pr` path for similar issues:

Line 1342 is in `paid_agent_review_limit_reached_for_review?` which is only used in `check_review_bot_status` — already handled. Line 196/208 is the escalation check — fixed. Line 1372/1379 are the helper methods — correct.

The diff is clean. Let me also check the `enabled_review_bot_logins` method to understand what it returns:

Good. `enabled_review_bot_logins` returns the set of lowercase logins for all enabled review methods. This includes codex logins like `"chatgpt-codex-connector"` and `"chatgpt-codex-connector[bot]"`.

My `other_enabled_body_only_bots?` checks if any of these logins are in `BODY_ONLY_REVIEW_BOT_LOGINS` but NOT paid_agent. This correctly identifies codex as another body-only bot.

Now let me run lint on all changed files and the full temporal test suite:

All 1033 examples pass, 0 failures. Now let me commit:

The output was truncated due to pre-existing Qdrant warnings. Let me check the actual test result:

Commit succeeded. 6026 examples, 0 failures, 2 pending (pre-existing Qdrant specs), 0 lint offenses.

Commit `929388a8`. Summary:

**P2 Review Fix** (`scan_paid_prs_activity.rb:1408,1417-1430`): `paid_agent_is_latest_blocker?` previously only checked for `review_bot_threads` as evidence of non-paid_agent blockers — but codex is also a body-only bot and never produces thread triggers. In `paid_agent + codex` projects, escalation fired even though codex could continue the review cycle.

Two changes:
1. **`other_enabled_body_only_bots?`** helper — checks whether the project has an enabled body-only review bot other than paid_agent (e.g. codex). Added to `paid_agent_is_latest_blocker?` as an early return, preventing escalation when another body-only bot can still make progress.
2. **`:no_review` path fix** (line 794) — removed `paid_agent_review_rounds_exhausted?` suppression from the `review_bot_request_login` branch. `review_bot_request_login` never returns a paid_agent login (only copilot/codex), so paid_agent exhaustion was incorrectly suppressing codex review requests in mixed-method projects.

**Tests added**: Two new specs — one verifying no escalation in `paid_agent + codex` when rounds are exhausted, another verifying `review_bot_comments` is emitted (triggering followup workflow for codex to continue).

viamin:
@codex review

PR #1008, opened 2026-04-11
Fix #1005: investigate(scanner): does paid_state failed suppress re-scanning after review-goal failure

### LINKED ISSUE #1005: investigate(scanner): does paid_state failed suppress re-scanning after review-goal failure
## Summary

When a review-goal agent run fails, `MarkAgentRunFailedActivity` sets `issue.paid_state = "failed"`. It is unclear whether the scanner skips PRs in `"failed"` paid_state. If it does, a single failed review-goal run would permanently prevent the PR from being re-scanned, making automated recovery impossible.

## Investigation needed

1. Trace the scanner's `scan_paid_pr` entry point and check whether `paid_state: "failed"` is used as an early-return guard.
2. Check `GitHubPollWorkflow` for any gating on `paid_state` before invoking the scanner.
3. Check `Issue` model scopes used by the scanner for any `paid_state` filtering.
4. Determine whether `"failed"` paid_state is intended to be a terminal state or a recoverable one.

## If confirmed

If `paid_state: "failed"` suppresses re-scanning:
- Review-goal runs should use a different state (e.g., `"review_failed"`) that doesn't suppress scanner activity.
- Or: the scanner should re-evaluate PRs in `"failed"` state when they have pending review signals and no active run.

## Related

- #948 — umbrella auto-merge issue
- Issue just filed: "fix(scanner): no recovery path when review-goal agent run fails"
- #839 — one common cause of review-goal failure (malformed JSON)

viamin:
## Summary

investigate(scanner): does paid_state failed suppress re-scanning after review-goal failure

See #1005 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1005

### DIFF
--- a/app/jobs/stale_run_detector_job.rb
+++ b/app/jobs/stale_run_detector_job.rb
@@ -181,7 +181,10 @@ class StaleRunDetectorJob < ApplicationJob
       agent_run.log!("system", "Run marked as timed out by stale run detector")
 
       if (issue = agent_run.issue)
-        issue.update!(paid_state: "failed") unless issue.paid_state == "failed"
+        # Review-goal failures restore "completed" instead of "failed" so
+        # auto-pick is not blocked by a transient review follow-up failure.
+        target_state = agent_run.review_goal? ? "completed" : "failed"
+        issue.update!(paid_state: target_state) unless issue.paid_state == target_state
       end
 
       Rails.logger.warn(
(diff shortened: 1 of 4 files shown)


viamin (commented):
Generated no new comments. The fix correctly restores `completed` state instead of `failed` for review-goal runs in both failure paths (stale run detector and mark-failed activity). The logic is consistent, well-commented, and properly tested. Addresses #1005.

PR #1025, opened 2026-04-11
Fix #980: devcontainer: set `"init": true` so PID 1 reaps zombie tmux servers

### LINKED ISSUE #980: devcontainer: set `"init": true` so PID 1 reaps zombie tmux servers
## Problem

PID 1 in the devcontainer is `sleep infinity`, which never reaps children. Every time `overmind`/`tmux` exits (e.g. when the controlling pts goes away — see #977 for the SIGHUP cascade), the tmux client+server become defunct zombies parented to PID 1 and stay there forever.

Evidence from `log/dev-update/diagnostics/20260409T081026Z-bin-dev-exit-1.log`:

```
vscode    3569       1  0 03:27 ?  [tmux: client] <defunct>
vscode    3570       1  0 03:27 ?  [tmux: server] <defunct>
vscode   46625       1  0 04:07 ?  [tmux: client] <defunct>
vscode   46626       1  0 04:07 ?  [tmux: server] <defunct>
vscode  140733       1  0 05:43 ?  [tmux: client] <defunct>
vscode  140734       1  0 05:43 ?  [tmux: server] <defunct>
vscode  197577       1  0 07:33 ?  [tmux: client] <defunct>
vscode  197578       1  1 07:33 ?  [tmux: server] <defunct>
```

Each crash/restart cycle leaks another pair. PIDs eventually exhaust and the diagnostic state gets noisier and harder to parse.

## Fix

Set `"init": true` in `.devcontainer/devcontainer.json` so the container runs under `tini` (or equivalent), which reaps orphaned children. This is a one-line change but requires a devcontainer rebuild to take effect.

## Verification

After rebuild, `ps -o pid,ppid,stat,comm -p 1` should show `tini` (or `docker-init`) instead of `sleep`, and `ps -ef | grep defunct` should stay empty across `bin/dev` restart cycles.

## Context

- Discovered while debugging overnight overmind crashes (#977).
- Not the root cause of those crashes — that's the SIGHUP-from-pts issue addressed in #977 — but it makes the symptoms accumulate and is worth fixing on its own.

viamin:
## Summary

devcontainer: set `"init": true` so PID 1 reaps zombie tmux servers

See #980 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #980

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,6 +6,7 @@
   "dockerComposeFile": "compose.yaml",
   "service": "rails-app",
   "workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
+  "init": true,
   // Features to add to the dev container. More info: https://containers.dev/features.
   "features": {
     "ghcr.io/devcontainers/features/github-cli:1": {},
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -6,6 +6,7 @@
   "dockerComposeFile": "compose.yaml",
   "service": "rails-app",
   "workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
+  // Use a minimal init so orphaned child processes get reaped inside the container.
   "init": true,
   // Features to add to the dev container. More info: https://containers.dev/features.
   "features": {


viamin:
@codex review

viamin:
## Agent Update

Updated [.devcontainer/devcontainer.json](/workspace/.devcontainer/devcontainer.json:9) with a short comment explaining why `"init": true` is intentionally enabled, so the PR’s zombie-reaping fix is less likely to be “cleaned up” later.

Validation:
- `bundle install`: passed after configuring Bundler to use `vendor/bundle`
- `yarn install`: passed after moving Yarn cache out of `/home/agent/.cache`
- `RAILS_ENV=test bin/rails db:prepare`: passed
- `bundle exec rubocop`: passed
- `bin/rails db:prepare`: failed in `development` because Rails tried the local PostgreSQL socket instead of a reachable dev DB config
- `bundle exec rspec`: failed due environment-level service dependencies, most concretely Qdrant not running at `http://localhost:6333`
- `git commit`: blocked because the commit hook also runs `bundle exec rspec`, which hits the same Qdrant failure

I did not create a commit. The worktree still has the staged-but-uncommitted change in `.devcontainer/devcontainer.json`.

viamin:
@codex review

PR #1026, opened 2026-04-11
Fix #975: Allow setting different default providers per agent run type

### LINKED ISSUE #975: Allow setting different default providers per agent run type
## Problem

Currently, there is a single default provider setting that applies to all agent run types. However, different run types may benefit from different providers. For example, a user might prefer Claude Code as the default for PR agent runs but Codex for code review agent runs.

## Proposed Solution

On the Providers settings page, allow each goal/run type (e.g., PR agent, code review agent, etc.) to have its own default provider setting.

### Requirements

- Each agent run host type should have an independent default provider selection
- The Providers page should display per-goal-type default settings
- If no per-type default is set, fall back to the global default provider

## Example

| Run Type | Default Provider |
|---|---|
| PR Agent | Claude Code |
| Code Review Agent | Codex |
| General | (global default) |

This gives users more granular control over which provider handles each type of agent run.

viamin:
## Summary

Allow setting different default providers per agent run type

See #975 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #975

### DIFF
--- a/app/controllers/projects/agent_runs_controller.rb
+++ b/app/controllers/projects/agent_runs_controller.rb
@@ -36,7 +36,8 @@ module Projects
 
     def new
       authorize @project, :run_agent?
-      @default_provider_identifier = settings_owner&.settings&.provider_priority(identifiers: true)&.first
+      selected_goal = params[:goal].presence || "create_pr"
+      @default_provider_identifier = settings_owner&.settings&.default_provider_identifier_for_goal(selected_goal)
       @available_run_provider_options = available_run_provider_options
       @issues = @project.issues
         .issues_only
@@ -120,7 +121,7 @@ module Projects
       create_run_and_redirect(
         on_error_path: project_path(@project),
         issue: issue,
-        provider_identifier: settings_owner&.settings&.default_provider_identifier,
+        provider_identifier: settings_owner&.settings&.default_provider_identifier_for_goal("create_pr"),
         goal: "create_pr",
         source_pull_request_number: source_pr_number
       )
@@ -529,7 +530,8 @@ module Projects
       requested_provider_identifier = provider_identifier || params[:provider].presence
       resolved_provider = resolve_provider_selection(
         requested_agent_type: requested_agent_type,
-        requested_provider_identifier: requested_provider_identifier
+        requested_provider_identifier: requested_provider_identifier,
+        goal: goal
       )
       raise NoRunnableProviderError, "No runnable provider could be resolved for this project." unless resolved_provider
 
@@ -555,7 +557,7 @@ module Projects
     def enqueue_resume_run(pr)
       create_agent_run(
         source_pull_request_number: pr.github_number,
-        provider_identifier: settings_owner&.settings&.default_provider_identifier,
+        provider_identifier: settings_owner&.settings&.default_provider_identifier_for_goal("create_pr"),
         goal: "create_pr",
         trigger_type: "automatic"
       )
@@ -694,12 +696,12 @@ module Projects
       enabled_retry_providers.any? { |provider| provider.provider_key == provider_key }
     end
 
-    def resolve_provider_selection(requested_agent_type:, requested_provider_identifier:)
+    def resolve_provider_selection(requested_agent_type:, requested_provider_identifier:, goal:)
       owner = settings_owner
       return unless owner
 
       configured_identifiers = UserSetting.enabled_agent_providers(owner, identifiers: true)
-      priority_identifiers = owner.settings.provider_priority(identifiers: true)
+      priority_identifiers = owner.settings.provider_priority_for_goal(goal, identifiers: true)
       default_identifier = priority_identifiers.first
 
       if requested_provider_identifier.present?
--- a/app/controllers/projects/agent_runs_controller.rb
+++ b/app/controllers/projects/agent_runs_controller.rb
@@ -37,7 +37,10 @@ module Projects
     def new
       authorize @project, :run_agent?
       selected_goal = params[:goal].presence || "create_pr"
-      @default_provider_identifier = settings_owner&.settings&.default_provider_identifier_for_goal(selected_goal)
+      @default_provider_identifiers_by_goal = AgentRun::GOALS.index_with do |goal|
+        settings_owner&.settings&.default_provider_identifier_for_goal(goal)
+      end.compact
+      @default_provider_identifier = @default_provider_identifiers_by_goal[selected_goal]
       @available_run_provider_options = available_run_provider_options
       @issues = @project.issues
         .issues_only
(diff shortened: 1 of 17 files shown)


viamin:
## Agent Update

Fixed a real gap in the PR: fresh runs created through `CreateAgentRunActivity` were still defaulting to `claude_code` unless a provider was passed explicitly, so goal-specific defaults only applied to queued/resumed paths. I updated [`app/temporal/activities/create_agent_run_activity.rb`](/workspace/app/temporal/activities/create_agent_run_activity.rb) to resolve the default provider for fresh runs too, and added coverage in [`spec/temporal/activities/create_agent_run_activity_spec.rb`](/workspace/spec/temporal/activities/create_agent_run_activity_spec.rb) for both global-default and review-goal cases.

Verification:
`bundle exec rubocop` passed.
`bundle exec rspec` passed in a standalone full run: 5973 examples, 0 failures, 2 pending.

I was not able to finish the required commit in this turn because the pre-commit hook’s full-suite rerun became unstable in this environment after long execution and did not complete cleanly before the turn ended, so the branch still has these two modified files unstaged in git history but present in the worktree.

viamin:
@codex review

viamin:
## Agent Update

Fixed the review comment by treating an empty permitted `default_agent_providers_by_goal` hash as “no change” when the raw nested params contained only unpermitted goal keys, so malformed or stale clients no longer wipe saved per-goal defaults. I added a request spec covering that regression in [spec/requests/providers_spec.rb](/workspace/spec/requests/providers_spec.rb) and updated the controller logic in [app/controllers/providers_controller.rb](/workspace/app/controllers/providers_controller.rb).

Verification passed with `bundle exec rubocop` and `bundle exec rspec` in full runs. The changes are committed as `99ee33d8` with message `fix(providers): preserve goal defaults on invalid params`.

viamin:
@codex review

viamin:
## Agent Update

Implemented the review-thread fix in the run form. The provider select now receives per-goal default metadata from [app/controllers/projects/agent_runs_controller.rb](/workspace/app/controllers/projects/agent_runs_controller.rb:37), and [app/javascript/controllers/goal_toggle_controller.js](/workspace/app/javascript/controllers/goal_toggle_controller.js:3) updates the selected provider when the goal radio changes unless the user has explicitly overridden it. The form wiring is in [app/views/projects/agent_runs/new.html.erb](/workspace/app/views/projects/agent_runs/new.html.erb:20), and I added request coverage in [spec/requests/agent_runs_spec.rb](/workspace/spec/requests/agent_runs_spec.rb:448).

Verification passed with `bundle exec rubocop` and `bundle exec rspec` in `RAILS_ENV=test` using `DB_HOST=postgres`; the full suite finished with `5978 examples, 0 failures, 2 pending` (the existing Qdrant live-spec pendings). The change is committed as `0e8ea34e` with message `fix(agent-runs): sync provider default with selected goal`.

viamin:
@codex review

viamin:
## Agent Update

Now let me read the specific file and area where the review comment was made:

The reviewer is correct. In `create_agent_run` (line 531), the goal is normalized to `"create_pr"` *after* `resolve_provider_selection` is called with the raw goal value. So `goal=invalid` would resolve the provider using the `invalid` key (falling back to global default), but then normalize the goal to `create_pr` — bypassing the per-goal default for create_pr runs.

Fix: move goal normalization before the provider resolution.

Now let me check for similar patterns elsewhere in the codebase and do a proactive scan:

Let me check if the other activities that use `goal` also normalize it before provider resolution:

Let me check what `AgentRun::GOALS` contains and also look at the `new` action to see if there's a similar pattern:

Good. Now let me also check the `new` action to see how it handles the goal parameter — there could be a similar issue there:

In the `new` action (line 39), the `selected_goal` isn't validated against `AgentRun::GOALS` either, but it only affects which default provider is pre-selected in the form, and invalid values would result in `nil` from `@default_provider_identifiers_by_goal[selected_goal]`, which is a benign fallback. This is acceptable for a display-only concern.

Let me now check the `create` action flow more carefully to see how goal is handled when it reaches `create_agent_run`:

Good. The `create` action uses `goal = params[:goal].presence || "create_pr"` but doesn't validate against `GOALS`. The goal value is passed to `create_run_and_redirect` which calls `create_agent_run(goal: goal)`. Now that my fix normalizes the goal inside `create_agent_run` before resolving the provider, the flow is safe.

Now let me check for any tests I should add. Let me look at the existing test for the create path with invalid goals:

I should add a test verifying that an invalid goal value gets normalized to `create_pr` and uses the correct default provider. Let me find the relevant test file:

Let me find a good spot to add a test for invalid goal normalization. Let me look for the "create_pr" context tests in the create action:

Now let me find a good location for my test. I'll add it in the POST create action area near the review goal tests. Let me see what comes after the review context:

I need to find the right location to add the test. Let me add it in the review section since the issue is about provider resolution per goal. Actually, the better place is right after the goal-specific default provider test at line 719, adding a test that validates the fix (invalid goal normalizes before provider resolution):

Let me find the best place to add the test. I'll add it right after the existing goal-specific default provider test in the review context:

Actually, the test for an invalid goal normalizing to `create_pr` and using the create_pr-specific default provider should go in the main POST create context (not the review context, since it normalizes to create_pr). Let me find the best spot:

I'll add the test right before the review context (after the budget check context). Let me add it after line 700 and before line 702:

Now let me also do a proactive scan of the full diff for other places where goal is used before validation. Let me check the `resolve_provider_selection` method and how `goal` flows through the `create_review_runs_and_redirect` path:

Good. In the `create` action (line 68), `goal` comes from `params[:goal]`. For `goal="review"` it goes to `create_review_runs_and_redirect` which passes the raw goal to `create_agent_run`. For other goals, it goes to `create_run_and_redirect` which also passes it through. Either way, `create_agent_run` now normalizes the goal first, so all paths are safe.

Now let me set up the environment and run the tests:

The dependency resolution seems to be having issues. Let me try a different approach:

Permission issues. Let me configure the bundle path to be local:

Database is ready. Now let me run the linter and the specific test file:

Test passes. Let me now run the broader set of related tests:

There are 85 failures. Let me check if these are pre-existing or related to my changes:

These failures are all about `stylesheet_link_tag` — a CSS bundling issue unrelated to my changes. These are pre-existing failures. Let me verify by running just the specific tests that relate to my changes:

All failures are from `stylesheet_link_tag` (CSS build issue). Let me focus only on the non-UI tests to verify my logic is correct:

All model and activity tests pass (143 examples, 0 failures). The UI test failures are pre-existing CSS build issues. Let me also run the providers spec since that's also part of the PR:

Same CSS issue. Let me check if the CSS is built:

The CSS build artifacts are missing. Let me build them:

Disk space issue. Let me check:

viamin:
@codex review

PR #1034, opened 2026-04-11
Fix #986: Move stale run cleanup from bin/setup to UI

### LINKED ISSUE #986: Move stale run cleanup from bin/setup to UI
## Summary

The `bin/setup` script currently includes automatic cleanup of stale agent runs. This was added when the system was less consistent at detecting and cleaning up timed-out and stuck agent runs. Today, long-running agent runs no longer hang indefinitely, making the automatic cleanup in `bin/setup` unnecessary.

## Proposed Changes

1. **Remove** the stale run cleanup logic from `bin/setup`.
2. **Add a manual cleanup button** in the UI on the project page, in the **Recent Agent Runs** section.
   - The button should only be visible when there are agent runs that have been running longer than the max timeout.
   - Clicking the button should clean up those stale runs.

## Motivation

- `bin/setup` should not be responsible for cleaning up stale runs now that the underlying reliability has improved.
- Moving this to the UI gives users visibility and control over when stale runs are cleaned up, rather than it happening silently during setup.

viamin:
## Summary

Move stale run cleanup from bin/setup to UI

See #986 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #986

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -22,7 +22,7 @@ class ProjectsController < ApplicationController
   def show
     authorize @project
     @recent_agent_runs = @project.agent_runs.recent.includes(:issue).limit(10).to_a
-    @stale_agent_runs_count = @project.agent_runs.stale_running.count
+    @stale_agent_runs_count = @project.agent_runs.stale_for_cleanup.count
     @show_stale_cleanup_action = policy(@project).update? && @stale_agent_runs_count.positive?
     AgentRun.preload_source_pull_requests(@recent_agent_runs)
     settings = current_user.settings
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ProjectsController < ApplicationController
-  before_action :set_project, only: [ :show, :edit, :update, :destroy, :toggle_auto_pick, :toggle_auto_merge, :detect_services ]
+  before_action :set_project, only: [ :show, :edit, :update, :destroy, :toggle_auto_pick, :toggle_auto_merge, :detect_services, :cleanup_stale_runs ]
   skip_after_action :verify_authorized, only: :index
 
   NULLS_LAST_SORT_ATTRIBUTES = %w[last_agent_run_at last_github_activity_at].freeze
@@ -22,6 +22,8 @@ class ProjectsController < ApplicationController
   def show
     authorize @project
     @recent_agent_runs = @project.agent_runs.recent.includes(:issue).limit(10).to_a
+    @stale_agent_runs_count = @project.agent_runs.stale_running.count
+    @show_stale_cleanup_action = policy(@project).update? && @stale_agent_runs_count.positive?
     AgentRun.preload_source_pull_requests(@recent_agent_runs)
     settings = current_user.settings
     open_items = @project.issues.where(github_state: "open").order(github_number: :desc)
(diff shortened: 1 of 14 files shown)


viamin:
@codex review

viamin:
## Agent Update

Fixed the request-spec CI regression by making layout asset tags test-safe when Propshaft build outputs are absent. I added `safe_stylesheet_link_tag` / `safe_javascript_include_tag` in [app/helpers/application_helper.rb](/workspace/app/helpers/application_helper.rb:23), switched the layout to use them in [app/views/layouts/application.html.erb](/workspace/app/views/layouts/application.html.erb:24), and added a regression spec in [spec/helpers/application_helper_asset_tags_spec.rb](/workspace/spec/helpers/application_helper_asset_tags_spec.rb:1).

Verification: `bundle exec rubocop` passed, `bundle exec rspec` passed with `5979 examples, 0 failures, 2 pending` (the existing Qdrant live specs were pending because Qdrant was unavailable), and the commit hook succeeded. `bin/rails db:prepare` could not connect in this container because Rails kept falling back to a local Postgres socket instead of the provided service host.

Committed as `fix(test): tolerate missing built assets in request specs` at `b9f14576`.

viamin:
@codex review

viamin:
## Agent Update

Included stale `pending` runs in the manual cleanup path instead of only stale `running` runs. The UI count/button now uses a shared `stale_for_cleanup` scope, manual cleanup requeues stale pending runs with the same timeout/retry thresholds as the detector job, and exhausted stale pending runs are timed out. I also added model, service, and request coverage for the pending path.

Verification: `bin/rails db:prepare` passed after running with `DATABASE_URL` unset and `PGHOST/PGUSER/PGPASSWORD` set so Rails could use `database.yml`’s named DBs; `bundle exec rubocop` passed; `bundle exec rspec` passed with `5985 examples, 0 failures, 2 pending` (the existing Qdrant live specs stayed pending because Qdrant was unavailable). Committed as `fix(agent-runs): include stale pending manual cleanup` at `2ed86846`.

viamin:
@codex review

viamin:
## Agent Update

Now I understand both issues. Let me fix them:

**Fix P2**: In `cleanup_container`, move the `update_all(container_id: nil)` before the rescue so it runs even when Docker cleanup fails. The DB clear should happen regardless of whether the Docker operations succeed.

**Fix P3**: Use block form of `fetch` to avoid eager evaluation of the count query.

Now let me do a proactive scan for similar issues in my changes and related code.

Let me check the existing test coverage for the cleanup service and the partial:

Let me check what files are part of this PR's changes:

Now let me add a test for the P2 fix (container_id cleared even when Docker cleanup fails):

Good, bundle is installed. Now let me set up the DB and run lint:

Database is prepared. Now let me run rubocop on the changed files:

The ERB file errors are expected (rubocop can't parse ERB templates directly). Let me run lint properly with the project's lint command:

Lint passes. Now run the tests:

All 9 tests pass, including both new tests. Now let me also run the request specs to make sure the partial change works:

All 132 request specs pass. The coverage threshold exit is expected when running a subset of specs. Let me verify the final state of my changes:

Let me check the final test results:

The commit didn't complete - the pre-commit hook timed out during the full test suite. Let me check git status:

The files are staged but the commit didn't finish due to the full test suite timing out. Let me try committing again with a longer timeout:

Commit `b9a8ba53` succeeded. All lint checks and tests passed via the pre-commit hook. Here's a summary of the changes:

**P2 fix** (`app/services/agent_runs/cleanup_stale.rb:149-161`): Moved `AgentRun.where(...).update_all(container_id: nil)` *before* the Docker reconnect/cleanup calls so the stale `container_id` is cleared from the DB even when the container is already gone or cleanup raises.

**P3 fix** (`app/views/projects/_agent_runs.html.erb:2`): Changed `local_assigns.fetch(:key, expensive_query)` to `local_assigns.fetch(:key) { expensive_query }` (block form) so the count query is only executed when the controller doesn't pass the value.

**Tests** (`spec/services/agent_runs/cleanup_stale_spec.rb`): Added two new specs verifying `container_id` is cleared when Docker reconnect fails and when Docker cleanup fails.

viamin:
@codex review

PR #1086, opened 2026-04-13
Fix #1039: feat(knowledge): add embedding support to agent-harness gem

### LINKED ISSUE #1039: feat(knowledge): add embedding support to agent-harness gem
## Background / Why

The knowledge base embedding pipeline (`Knowledge::Embeddings::Generate`) currently calls the OpenAI embeddings API directly via Faraday, violating the architectural mandate that **all LLM calls go through agent-harness**. This is acknowledged by the existing `TODO(#257)` in `generate.rb:105-106`. Issue #257 is closed but `AgentHarness.embed` was never added to the gem — the direct Faraday call remains.

This is the foundational blocker for:
- Provider fallback for embeddings (can't fall back without a unified interface)
- Containerizing embedding generation (can't proxy without a harness-level abstraction)
- Removing API keys from the Rails process

## Scope

**In scope (agent-harness gem):**
- Add `AgentHarness.embed(texts:, model:, dimensions:)` method to the gem
- Support provider selection with the same orchestration/fallback pattern as `send_message`
- Support OpenAI-compatible embedding endpoints (OpenAI, Azure OpenAI, OpenRouter)
- Handle the embedding-specific response shape (array of vectors, usage tokens)

**Out of scope:**
- Switching Paid's `Generate` service to use it (that's a separate issue)
- Containerization (separate issue)
- Non-OpenAI-compatible embedding providers (e.g., Voyage, Cohere) — future work

## Proposed Design

```ruby
# Usage in Paid (after this issue):
AgentHarness.embed(
  texts: ["chunk 1 text", "chunk 2 text"],
  provider: :openai,
  model: "text-embedding-3-large",
  dimensions: 3072
)
# => AgentHarness::EmbeddingResponse with .vectors, .token_count, .model
```

The conductor should apply the same orchestration as `send_message`: provider selection, rate-limit detection, circuit breaker, auto-switch on failure.

**Key constraint**: Switching embedding providers typically requires re-embedding all chunks (different vector dimensions or distance metrics). For fallback purposes, the initial implementation should only switch between providers serving the **same model** (e.g., OpenAI → Azure OpenAI → OpenRouter, all serving `text-embedding-3-large`). Cross-model fallback is a future enhancement that would trigger a re-embedding pipeline.

## Acceptance Criteria

- [ ] `AgentHarness.embed` is available and documented
- [ ] Provider orchestration (fallback, circuit breaker, rate-limit) works for embedding calls
- [ ] OpenAI-compatible endpoints work (OpenAI, Azure OpenAI, OpenRouter)
- [ ] Response object includes vectors, token count, and model info
- [ ] Existing `send_message` functionality is unaffected

## Dependencies

- Upstream agent-harness gem work

## Notes / Risks

- **Upstream coordination**: This requires changes to the `agent-harness` gem at https://github.com/viamin/agent-harness
- **Model compatibility**: Fallback must be limited to same-model endpoints to avoid vector dimension mismatches in Qdrant
- **Cost**: Embedding API costs are low (~$0.13/1M tokens) but the harness should still track them

viamin:
## Summary

feat(knowledge): add embedding support to agent-harness gem

See #1039 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1039


viamin:
## Agent Update

Committed as `7c90f7f3` with `fix(agent-harness): remove artifact and harden auth matching`.

I removed the committed `vendor/agent-harness/json-2.18.1.gem` artifact, replaced the backtracking-prone auth classifier in `vendor/agent-harness/lib/agent_harness/error_taxonomy.rb`, added a regression spec for the repeated-`invalid` case, and proactively tightened the sibling `invalid.*api.*key` patterns in the vendored provider code.

Verification: `bundle exec rubocop` passed, the full vendored `vendor/agent-harness` spec suite passed, and the repo pre-commit hook completed successfully with `bundle exec rspec` on the full app suite: `6211 examples, 0 failures, 2 pending`. The only setup command that did not succeed was `bin/rails db:prepare`, which failed earlier because the configured Postgres host was not reachable from this container.

viamin:
## Agent Update

Now I'll apply all three fixes:

Now let me scan the full PR diff for similar issues — unescaped URL interpolations, nil guards on registry lookups, and unreachable rescue chains:

The proactive scan found no additional issues. All three fixes are clean. Now let me run lint and tests.

The vendored gem has its own Standard-based config but standard isn't installed. Let me try running rubocop with just the default config on our changed files:

All files pass syntax checks. Now let me run the tests:

The full Rails app bundle isn't installed. Let me run just the vendored agent-harness specs:

The environment has dependency installation issues. Let me try to run tests directly without bundler, using the installed rspec:

The vendor gem's spec_helper requires `bundler/setup` which expects all gems in the standard location. Let me try a different approach — install them to the default gem path:

The issue is version mismatches — bundler wants specific versions but newer ones are installed. Let me try matching the exact versions:

Now let me run the tests:

All 757 tests pass. Now let me also run lint on the vendor gem:

The vendor gem's `.standard.yml` inherits from `standard` which pulls in `rubocop-rspec` plugin from the parent `.rubocop.yml`. Let me use standardrb directly instead:

Lint passes cleanly. Tests pass (757 examples, 0 failures). Let me verify my changes are correct:

All changes look correct. Now let me commit:

The pre-commit hook fails because the main Rails app's bundle isn't fully installed. Let me install the necessary gems for the pre-commit hook:

The full Rails app bundle can't be installed in this environment (missing native extensions). Let me check what the pre-commit hook does:

The pre-commit hook runs `bin/lint --staged`. Let me check what that does:

The pre-commit hook calls `bin/lint --staged` which calls `bin/rubocop` for Ruby files. `bin/rubocop` needs the main Rails bundle. Since the full bundle can't be installed, let me try to install the remaining critical gems:

0 files inspected because the vendor dir is likely excluded. Let me check:

The vendor directory is excluded from the project's rubocop config. So the pre-commit hook's rubocop step will skip these files. The hook will then try ESLint (no JS files staged), markdownlint (no MD files), and ShellCheck (no shell files). It should pass cleanly for our files since they're all excluded. Let me try the commit again:

The pre-commit hook's `bin/rubocop` calls `require "bundler/setup"` which fails because the Rails app bundle can't be fully installed. Let me look at what `bin/rubocop` does:

The problem is `bin/rubocop` uses `bundler/setup` for the main Rails Gemfile, which can't be installed in this environment. The pre-commit hook uses `bin/lint --staged` which calls `bin/rubocop`. Since our files are under `vendor/` which is excluded from the project's rubocop config, rubocop would skip them anyway, but it crashes before getting that far because `bundler/setup` fails.

The environment doesn't have all native extensions needed for the Rails app. This is an infrastructure limitation, not a code issue. Let me verify there's no way to get the bundle installed:

Let me try to get the full bundle installed by installing the missing native gems one at a time:

This is a long chain of one-by-one installs. Let me try a different approach — install everything at once:

This pattern of one-by-one installation is going to take a very long time. Let me try a bulk approach - extract all gem names from the lock file and install them all:

Bundle is complete. Now let me try the commit again:

Committed as `04b80563`. All three review threads addressed:

1. **`provider_manager.rb:192`** — Added `next unless klass` nil guard before `klass.supports_embeddings?`
2. **`base.rb`** — Added `self.supports_embeddings?` returning `false` on `Base`, so non-embedding providers don't raise `NoMethodError`
3. **`openai_compatible.rb:229`** — Nested the `Time.httpdate` fallback inside a `begin/rescue` block so the second `ArgumentError` catch is reachable
4. **`azure_openai.rb:26`** — Escaped `deployment` with `URI.encode_www_form_component` and `api_version` with `CGI.escape`

Proactive scan found no additional instances of these patterns. Vendor gem tests: 757 examples, 0 failures. Lint passed.

viamin:
This appears to create a local copy of agent-harness, which is the wrong approach

PR #1092, opened 2026-04-13
Fix #1072: fix(scanner): backfill review_goal_retry_reset_at for existing issues

### LINKED ISSUE #1072: fix(scanner): backfill review_goal_retry_reset_at for existing issues
## Summary
PR #1019 adds `issues.review_goal_retry_reset_at`, but the migration only adds the nullable column and does not backfill existing rows or provide a fallback for preexisting issues already in a post-reset state.

## Why This Matters
The new retry-cycle logic uses `review_goal_retry_reset_at` as part of the boundary for excluding historical failed review-goal runs. For preexisting issues where that column remains `NULL`, old failures can still be counted after deploy, including for issues that were already effectively reset (for example, after restart or escalation dismissal).

That can cause false `escalate_to_owner` triggers immediately after rollout.

## Expected Behavior
Existing issues should get a safe reset boundary, either via migration backfill or a runtime fallback, so historical failures from a prior cycle do not immediately trip the new retry breaker.

## Pointers
- PR: #1019
- Migration: `db/migrate/20260412165456_add_review_goal_retry_reset_at_to_issues.rb`
- Retry boundary logic: `app/temporal/activities/scan_paid_prs_activity.rb`

## Suggested Fix
Implement at least one of:
- a migration/data backfill for existing `issues.review_goal_retry_reset_at` values
- a runtime fallback when the issue is in a post-reset phase and the column is still `NULL`

Add regression coverage for a preexisting issue whose old failed review-goal runs should not trigger immediate post-deploy escalation.

viamin:
## Summary

fix(scanner): backfill review_goal_retry_reset_at for existing issues

See #1072 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1072

### DIFF
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -106,6 +106,8 @@ module Activities
     def scan_pr(project, client, issue)
       return :skipped if active_run_exists?(project, issue)
 
+      backfill_review_goal_retry_reset_at!(issue)
+
       retry_needed = review_goal_retry_needed?(project, issue)
       retry_limit_reached = retry_needed && review_goal_retry_limit_reached?(project, issue)
       retry_limit_reason = "Review-goal retry limit reached " \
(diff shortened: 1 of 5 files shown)


viamin:
## Agent Update

Scoped the new backfill migration to restarted pull requests only and tightened the migration spec so non-restarted issues remain `NULL`. That avoids inventing a retry-reset boundary on unrelated issues while preserving the runtime scanner fix already in the branch.

Validation passed: `bundle exec rubocop`, `bundle exec rspec` (`6240 examples, 0 failures, 2 pending`), and the commit hook re-ran lint/tests successfully. I used an isolated test database via `PRIMARY_DATABASE_URL` because the shared `agent_test` DB was already in use, and set local Bundler path config so the hooks could see the installed gems. Commit: `3e054d5a` (`fix(scanner): scope retry reset backfill to restarted PRs`).

PR #1096, opened 2026-04-13
Fix #1056: fix(workflow): silent restart failure when WorkflowAlreadyStartedError is raised

### LINKED ISSUE #1056: fix(workflow): silent restart failure when WorkflowAlreadyStartedError is raised
## Problem

Two related issues cause restarts to silently fail:

1. **`ProjectWorkflowManager.start_polling` swallows `WorkflowAlreadyStartedError`** (`project_workflow_manager.rb:28-32`):

```ruby
rescue Temporalio::Error::WorkflowAlreadyStartedError
  Rails.logger.warn(
    message: "github_sync.polling_already_running",
    project_id: project.id
  )
end
```

When `restart_polling` terminates a workflow and immediately calls `start_polling`, the termination may not have propagated in Temporal yet. `start_workflow` raises `WorkflowAlreadyStartedError`, which is caught and logged as a warning. The caller believes the restart succeeded.

2. **`PollWorkflowHealthCheckJob#restart_workflow` always returns `true`** (`poll_workflow_health_check_job.rb:110`):

```ruby
def restart_workflow(project, reason:, log_message: ...)
  # ...
  ProjectWorkflowManager.restart_polling(project, reason: reason)
  # ...
  true  # Always returns true, even when start_polling silently failed
end
```

The health check logs "1 project restarted" even when no workflow was actually started.

## Reproduction

This is likely the root cause of the 60-minute stale workflow incident. The health check detected a stale workflow, "restarted" it, but `start_polling` hit `WorkflowAlreadyStartedError` (termination not yet propagated) and swallowed it. The UI showed `Last restart: health check: stale RUNNING` but the workflow was never actually restarted.

## Fix

`start_polling` should not swallow `WorkflowAlreadyStartedError` during restarts. Options:

1. **Add a `raise_on_conflict` parameter** to `start_polling` that re-raises `WorkflowAlreadyStartedError` instead of swallowing it. The health check and manual restart should use `raise_on_conflict: true`. The existing `rescue` behavior can be kept for initial project activation where a duplicate start is acceptable.

2. **Retry with backoff** in `restart_polling` — after terminating, wait briefly for Temporal to propagate the termination, then start the new workflow. This is more resilient but adds latency.

3. **Return a success/failure indicator** from `start_polling` and propagate it through `restart_polling` and `restart_workflow` so the health check can log and report accurately.

Option 1 is the simplest and most direct fix. Option 3 should be done regardless for observability.

## Context

- PR #955 added `restart_reason` tracking to `start_polling` but did not address the silent failure
- The health check job runs every 5 minutes, so a failed restart means 5+ minutes of downtime before the next attempt

viamin:
## Summary

fix(workflow): silent restart failure when WorkflowAlreadyStartedError is raised

See #1056 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1056

### DIFF
--- a/app/controllers/workflow_statuses_controller.rb
+++ b/app/controllers/workflow_statuses_controller.rb
@@ -25,10 +25,11 @@ class WorkflowStatusesController < ApplicationController
 
     ProjectWorkflowManager.restart_polling(@project, reason: "manual restart from UI")
     redirect_to project_path(@project), notice: "Issue monitor restarted."
-  rescue Temporalio::Error::RPCError => e
+  rescue Temporalio::Error::RPCError, Temporalio::Error::WorkflowAlreadyStartedError => e
     Rails.logger.error(
       message: "workflow_status.restart_failed",
       project_id: @project.id,
+      error_class: e.class.name,
       error: e.message
     )
     redirect_to project_path(@project), alert: "Could not restart issue monitor. Please try again later."
--- a/app/jobs/poll_workflow_health_check_job.rb
+++ b/app/jobs/poll_workflow_health_check_job.rb
@@ -101,13 +101,14 @@ class PollWorkflowHealthCheckJob < ApplicationJob
       reason: reason
     )
 
-    ProjectWorkflowManager.restart_polling(project, reason: reason)
+    restarted = ProjectWorkflowManager.restart_polling(project, reason: reason)
 
     Rails.logger.info(
       message: "temporal_worker.poll_workflow_restarted",
       project_id: project.id
-    )
-    true
+    ) if restarted
+
+    restarted
   end
 
   # Maps Temporal workflow execution statuses to WorkflowState status strings.
--- a/app/services/project_workflow_manager.rb
+++ b/app/services/project_workflow_manager.rb
@@ -6,7 +6,7 @@
 # for labeled issues.
 class ProjectWorkflowManager
   class << self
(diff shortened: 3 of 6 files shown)


viamin:
## Agent Update

Addressed the review thread in code by updating [`app/controllers/workflow_statuses_controller.rb`](/workspace/app/controllers/workflow_statuses_controller.rb:28) to rescue `Temporalio::Error::WorkflowAlreadyStartedError` alongside `RPCError`, and added request coverage in [`spec/requests/workflow_statuses_spec.rb`](/workspace/spec/requests/workflow_statuses_spec.rb:255) for the manual-restart conflict path.

Verification I completed locally:
- `yarn install`
- `bundle install` with local bundle path
- `PGHOST=postgres PGUSER=agent PGPASSWORD=agent bundle exec rails db:prepare`
- `bundle exec rubocop`
- `bundle exec rspec` with 6298 examples, 0 failures, 2 pending

I was not able to finish the requested commit. The repo’s pre-commit hook reruns the full suite and is still unstable in this environment: with the default `DATABASE_URL` it collides on `agent_test`, and even after switching the hook to `paid_test` via `env -u DATABASE_URL PGHOST=postgres PGUSER=agent PGPASSWORD=agent git commit ...`, the hook run remained long/noisy and I did not get a clean commit completion before stopping. The worktree still has only these two modified files unstaged from the failed commit attempts.

PR #1099, opened 2026-04-13
Fix #1057: fix(workflow): health check doesn't record stale state before restart

### LINKED ISSUE #1057: fix(workflow): health check doesn't record stale state before restart
## Problem

When the health check detects an unhealthy (not running) workflow, it records the state via `WorkflowState.record_polling_status` before restarting (line 72-77 of `poll_workflow_health_check_job.rb`). This triggers a broadcast that updates the UI to show "Not running" with the restart reason.

However, when the health check detects a **stale** (RUNNING but behind schedule) workflow, it skips the recording step entirely:

```ruby
def check_stale_running(project)
  return unless project.poll_stale_with_recheck?

  restart_workflow(
    project,
    reason: "health check: stale RUNNING (last polled #{project.last_polled_at})",
    log_message: "temporal_worker.poll_workflow_stale_running"
  )
end
```

The `restart_reason` is passed to `restart_polling` → `start_polling` → `record_polling_status`, which updates the `restart_reason` field. But since the status is `"running"` (the workflow is RUNNING in Temporal), `record_polling_status` sets the status back to `"running"` and the broadcast shows "Active" or "Delayed" depending on staleness — not the fact that a restart just occurred.

More importantly, if `restart_polling` fails silently (see sibling issue about `WorkflowAlreadyStartedError`), there is no broadcast at all, and the UI never reflects that the health check attempted a restart.

## Fix

Record the stale state before restarting, similar to the unhealthy path:

```ruby
def check_stale_running(project)
  return unless project.poll_stale_with_recheck?

  # Record the stale state so the UI reflects it before restart
  WorkflowState.record_polling_status(
    project,
    status: "running",
    restart_reason: "health check: stale RUNNING (last polled #{project.last_polled_at})"
  )

  restart_workflow(
    project,
    reason: "health check: stale RUNNING (last polled #{project.last_polled_at})",
    log_message: "temporal_worker.poll_workflow_stale_running"
  )
end
```

This ensures:
1. The UI immediately shows the restart reason via broadcast
2. If the restart fails, the `restart_reason` is already persisted for the next health check cycle
3. The behavior is consistent with the unhealthy path

## Context

- PR #955 added `record_polling_status` to the unhealthy path but the stale path was overlooked
- The `:not_provided` sentinel pattern in `record_polling_status` (commit `d99c632da`) ensures the restart reason won't be overwritten by subsequent poll-cycle status updates

viamin:
## Summary

fix(workflow): health check doesn't record stale state before restart

See #1057 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1057

### DIFF
--- a/app/jobs/poll_workflow_health_check_job.rb
+++ b/app/jobs/poll_workflow_health_check_job.rb
@@ -87,9 +87,17 @@ class PollWorkflowHealthCheckJob < ApplicationJob
   def check_stale_running(project)
     return unless project.poll_stale_with_recheck?
 
+    reason = "health check: stale RUNNING (last polled #{project.last_polled_at})"
+
+    WorkflowState.record_polling_status(
+      project,
+      status: "running",
+      restart_reason: reason
+    )
+
     restart_workflow(
       project,
-      reason: "health check: stale RUNNING (last polled #{project.last_polled_at})",
+      reason: reason,
       log_message: "temporal_worker.poll_workflow_stale_running"
     )
   end
--- a/app/models/workflow_state.rb
+++ b/app/models/workflow_state.rb
@@ -44,13 +44,14 @@ class WorkflowState < ApplicationRecord
       ws.error_message = nil
     end
 
-    should_broadcast = ws.new_record? || ws.status_changed?
+    should_broadcast =
+      ws.new_record? || ws.status_changed? || ws.restart_reason_changed? || ws.error_message_changed?
     ws.save!
 
     # Broadcast after the transaction commits so listeners never see data that
-    # could be rolled back. Skip when status is unchanged (e.g., healthy →
-    # healthy on every poll cycle) to avoid unnecessary DB queries and
-    # ActionCable traffic.
+    # could be rolled back. Skip when neither status nor user-visible restart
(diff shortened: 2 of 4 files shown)


viamin:
## Agent Update

Implemented the stale-restart fix on top of the PR and committed it as `def5d196` with `fix(workflow): persist stale health-check restart state`.

The change does two things: it keeps broadcasting `WorkflowState` updates when only restart metadata changes on an already-running workflow, and it adds coverage proving stale restart context is recorded before the restart attempt, even if the restart itself fails. I also added a model spec for the broadcast path.

Verification passed:
- `bundle exec rubocop`
- `bundle exec rspec` (`6240 examples, 0 failures, 2 pending`)
- commit hook passed, and the worktree is clean

`db:prepare` required explicit DB URLs for this environment:
- development: `DATABASE_URL=postgres://agent:agent@postgres:5432/paid_development CABLE_DATABASE_URL=postgres://agent:agent@postgres:5432/paid_development_cable`
- test: `RAILS_ENV=test DATABASE_URL=postgres://agent:agent@postgres:5432/paid_test`

PR #1101, opened 2026-04-13
Fix #1028: providers page: collapse configuration instructions by default and add instruction blocks for all supported providers

### LINKED ISSUE #1028: providers page: collapse configuration instructions by default and add instruction blocks for all supported providers
## Summary

On the providers page, provider configuration instructions should be collapsed by default. Also, at least `Kilo Code` appears to be missing its instructions block, and likely other supported providers are missing as well. Every supported provider should have a configuration instructions block.

## Problems

- Provider configuration instructions are expanded by default, which makes the page noisy and harder to scan when multiple providers are listed.
- `Kilo Code` does not appear to have a provider instructions block.
- The page appears to have inconsistent provider coverage, where some supported providers have instructions and others do not.

## Expected behavior

- All provider configuration instructions are collapsed by default when the providers page loads.
- Each supported provider has a corresponding instructions block.
- Instruction blocks are presented consistently across providers so users can expand only the provider they need.

## Acceptance criteria

- [ ] Provider configuration instructions are collapsed by default on initial page load
- [ ] Users can expand a provider's instructions block on demand
- [ ] `Kilo Code` has an instructions block
- [ ] Every currently supported provider has an instructions block
- [ ] The UI handles missing provider-specific copy defensively so supported providers do not silently render without instructions

## Notes

I could not inspect the repository in this run because the workspace did not include a checkout, so this issue is based on the current UI behavior observed/reported rather than file-level code references.

viamin:
## Summary

Prevents unbounded PR conversation context from bloating agent prompts by limiting the number of comments fetched, capping how many trusted comments are included, and truncating long comment bodies. This aligns PR prompt building with the same guardrails already in place for issue prompts via `BuildForIssue`.

## Changes

**GitHub API (`app/services/github_client.rb`)**
- Extended `recent_issue_comments` with a `pages:` parameter to fetch a bounded trailing window of comment pages instead of only the last page
- Walks backward from the last page using `prev` rel links, then returns the slice in ascending ID order

**PR prompt building (`app/services/prompts/build_for_pr.rb`)**
- Switched from `issue_comments` (which fetches all comments) to `recent_issue_comments` with a 2-page window (`RECENT_COMMENT_PAGE_WINDOW`)
- Added `.last(max_prompt_comments)` to cap trusted comments at the user-setting limit (reuses `BuildForIssue::DEFAULT_MAX_COMMENTS`)
- Added `truncate_comment_body` to clip individual comment bodies at the configured `max_comment_length`
- Reads limits from `AgentRuns::UserSettingsResolver` so per-project overrides are respected

**Tests**
- Added multi-page window spec for `GithubClient#recent_issue_comments`
- Added specs for comment count limiting and body truncation in `BuildForPr`
- Updated existing specs to use `recent_issue_comments` and stub `UserSettingsResolver`

## Design Decisions

- **2-page window instead of fetching all pages**: A 2-page trailing window (up to 200 comments) keeps the API cost constant regardless of how long the PR conversation grows, while still capturing enough recent context for the agent to act on.
- **Reuses `BuildForIssue` defaults**: Rather than introducing separate constants, PR prompt building shares `DEFAULT_MAX_COMMENTS` and `DEFAULT_MAX_COMMENT_LENGTH` from `BuildForIssue` so the two prompt paths stay consistent and user settings apply uniformly.

---

Closes #1028

### DIFF
--- a/app/helpers/providers_helper.rb
+++ b/app/helpers/providers_helper.rb
@@ -1,0 +1,79 @@
+# frozen_string_literal: true
+
+module ProvidersHelper
+  PROVIDER_AUTH_INSTRUCTION_COPY = {
+    "claude" => {
+      summary: "Use one of these:",
+      items: [
+        "Set <code>ANTHROPIC_API_KEY</code> on the <code>web</code> service for proxy-based auth.",
+        "Or sign in with <code>claude login</code> and make <code>~/.claude/.credentials.json</code> visible to Paid.",
+        "If Claude creds live elsewhere, set <code>CLAUDE_CONFIG_DIR</code>."
+      ]
+    },
+    "codex" => {
+      summary: "Use one of these:",
+      items: [
+        "Set <code>OPENAI_API_KEY</code> on the <code>web</code> service for proxy-based auth.",
+        "Or sign in with the Codex CLI and make <code>~/.codex/auth.json</code> visible to Paid.",
+        "If Codex creds live elsewhere, set <code>CODEX_CONFIG_DIR</code> or <code>CODEX_HOME</code>."
+      ]
+    },
+    "gemini" => {
+      summary: "Use one of these:",
+      items: [
+        "Set <code>GOOGLE_API_KEY</code> on the <code>web</code> service for proxy-based auth.",
+        "Or run <code>gemini auth login</code> and make <code>~/.gemini/oauth_creds.json</code> visible to Paid.",
+        "If Gemini creds live elsewhere, set <code>GEMINI_CONFIG_DIR</code>."
+      ]
+    },
+    "opencode" => {
+      summary: "Choose the auth path that matches the provider entry:",
+      items: [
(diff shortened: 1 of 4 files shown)


viamin:
## Agent Update

Removed the per-request warning from [app/helpers/providers_helper.rb](/workspace/app/helpers/providers_helper.rb:62) so generic provider auth fallback copy no longer logs on every `/providers` render. Added direct coverage in [spec/helpers/providers_helper_spec.rb](/workspace/spec/helpers/providers_helper_spec.rb:1) to verify both explicit-copy providers and the generic fallback path, including that the fallback path does not call `Rails.logger.warn`.

Validation passed with `bundle exec rubocop` and full `bundle exec rspec` via the pre-commit hook. `bin/rails db:prepare` did not succeed in this environment because Rails attempted a local PostgreSQL socket connection despite `DATABASE_URL` being present, but the full test suite still passed. Commit: `11cec127` (`fix(providers): stop logging generic auth fallback copy`).

PR #1077, opened 2026-04-13
fix(agent-runs): resolve addressed review threads on no-change PR runs

viamin:
## Summary
- add an explicit marker for PR follow-up runs when unresolved review threads are already addressed and no code changes are needed
- persist the PR follow-up prompt version and only allow auto-resolution when the generated prompt actually included unresolved review threads
- teach the no-change PR workflow path to resolve review threads when that verified marker is present, with regression coverage for prompts, workflow behavior, and seeds

## Verification
- `COVERAGE=false bundle exec rspec spec/services/prompts/build_for_pr_spec.rb spec/temporal/activities/prepare_pr_prompt_activity_spec.rb spec/temporal/activities/run_agent_activity_spec.rb spec/temporal/workflows/agent_execution_workflow_spec.rb spec/db/seeds_prompts_spec.rb`
- `bundle exec rubocop app/services/prompts/build_for_pr.rb db/seeds/prompts.rb app/temporal/activities/prepare_pr_prompt_activity.rb app/temporal/activities/run_agent_activity.rb app/temporal/workflows/agent_execution_workflow.rb spec/services/prompts/build_for_pr_spec.rb spec/temporal/activities/prepare_pr_prompt_activity_spec.rb spec/temporal/activities/run_agent_activity_spec.rb spec/temporal/workflows/agent_execution_workflow_spec.rb spec/db/seeds_prompts_spec.rb`

PR #1136, opened 2026-04-14
Fix #1082: refactor(automation): unify issue and PR automation behind an explicit decision layer

### LINKED ISSUE #1082: refactor(automation): unify issue and PR automation behind an explicit decision layer
## Summary

Paid's automation entrypoints currently make PR-vs-issue decisions in multiple places, with overlapping but inconsistent behavior.

This makes it easy for generic issue logic to leak into PR handling, as happened with PR #1077 where the initial sync path queued a default `create_pr` run even though the PR-specific scanner would only have produced `paid_agent_review_pending`.

## Problem

Today, the automation flow is split across at least two different styles of decision-making:

- generic label detection for synced GitHub items
- PR follow-up scanning and trigger classification

These paths do not share a single decision contract. As a result:

- PRs can be routed through issue-style `execute_agent` handling
- run goals can be chosen implicitly by queue defaults instead of explicitly by decision logic
- it is hard to reason about which component owns behavior for existing PRs vs issues
- regressions tend to appear at path boundaries rather than inside one isolated policy

## Recommendation

Introduce a single automation decision layer with a shared output shape, while keeping issue and PR evaluation logic separate.

Suggested shape:

- `Automation::Evaluator.for(record).call`
- evaluator implementations such as `Automation::IssueEvaluator` and `Automation::PullRequestEvaluator`
- explicit decision objects/results like:
  - `noop`
  - `queue_create_pr_run`
  - `queue_review_run`
  - `start_planning`
  - `request_review`
  - `mark_ready`
  - `escalate`

Ship the new PR automation path behind a Flipper flag so the behavior can be rolled out intentionally instead of replacing the existing path all at once.

Suggested initial flag:

- `explicit_pr_automation_decisions`

## Scope

- Define a single decision/result contract for GitHub automation
- Route both initial sync evaluation and PR auto-continue evaluation through that contract
- Make goal selection explicit for PR actions instead of inheriting `QueueAgentRunActivity` defaults
- Put the new PR automation decision path behind the Flipper flag
- Keep model responsibilities limited to state/predicates, not GitHub orchestration and workflow dispatch
- Extract shared policy helpers where useful, but keep PR-specific and issue-specific evaluators distinct

## Acceptance Criteria

- There is one top-level automation decision entrypoint used by both issue and PR automation flows
- PR evaluation cannot accidentally fall back to generic issue `execute_agent` semantics
- Queueing code receives an explicit decision/goal instead of inferring behavior from missing params
- The new PR automation path is guarded by `explicit_pr_automation_decisions`
- The resulting design is easier to test in isolation than today's workflow-plus-activity coupling
- Follow-up bugs like PR #1077 can be localized to a single evaluator/decision path

## Dependencies

Depends on #1080
Depends on #1085

## Related

Parent rollout issue: #1083

viamin:
## Summary

Unifies Paid's issue and PR automation behind an explicit decision layer so that PR evaluation cannot accidentally fall through to generic issue handling — the class of bug that caused PR #1077's spurious `create_pr` run. The new PR automation path is gated behind the `explicit_pr_automation_decisions` Flipper flag for incremental rollout.

## Changes

**Decision layer (`app/services/automation/`)**
- Added `Automation::Evaluator.for(record).call` as the single top-level entrypoint for all GitHub automation decisions
- Introduced `Automation::Decision` and `Automation::Result` value objects with an explicit set of outcomes (`noop`, `queue_create_pr_run`, `queue_review_run`, `start_planning`, `request_review`, `mark_ready`, `escalate`)
- Implemented `Automation::IssueEvaluator` and `Automation::PullRequestEvaluator` as separate evaluation strategies that share the result contract but keep domain logic distinct

**Temporal activities**
- `DetectLabelsActivity` now routes records through the shared decision layer instead of ad hoc issue-action logic
- `ScanPaidPrsActivity` emits explicit automation results when the Flipper flag is enabled; preserves the legacy trigger-payload path when the flag is off

**Temporal workflow**
- `GitHubPollWorkflow` executes explicit decisions for both initial sync evaluation and PR auto-continue, ensuring queueing always receives an explicit goal on the new path

**Feature flag**
- `explicit_pr_automation_decisions` gates the new PR automation path, allowing the legacy behavior to remain active until the rollout is validated

## Design Decisions

- **Separate evaluators, shared contract**: Issue and PR evaluation live in distinct classes to prevent logic leakage, but both return the same `Result` type so downstream consumers (workflow, queueing) don't branch on record type.
- **Flag guards the PR path only**: Issue evaluation through the new layer is unconditional since it replaces equivalent inline logic. The PR path has more behavioral risk, so it's flag-gated for controlled rollout.
- **Explicit goal on every queued run**: On the new path, `QueueAgentRunActivity` always receives a goal from the decision result rather than inferring one from missing params — directly addressing the root cause of #1077.

---

Closes #1082

### DIFF
--- a/app/services/automation/decision.rb
+++ b/app/services/automation/decision.rb
@@ -1,0 +1,90 @@
+# frozen_string_literal: true
+
+module Automation
+  class Decision < Data.define(:type, :payload)
+    class << self
+      def noop
+        new(type: "noop", payload: {})
+      end
+
+      def queue_create_pr_run(issue_id:, source_pull_request_number: nil,
+        count_toward_draft_review_round: false, expected_draft_review_count: nil)
+        new(type: "queue_create_pr_run", payload: {
+          issue_id: issue_id,
+          source_pull_request_number: source_pull_request_number,
+          count_toward_draft_review_round: count_toward_draft_review_round,
+          expected_draft_review_count: expected_draft_review_count
+        }.compact)
+      end
+
+      def queue_review_run(issue_id:, source_pull_request_number:)
+        new(type: "queue_review_run", payload: {
+          issue_id: issue_id,
+          source_pull_request_number: source_pull_request_number
+        })
+      end
+
+      def start_planning(issue_id:)
+        new(type: "start_planning", payload: { issue_id: issue_id })
+      end
+
+      def request_review(pr_number:, reviewers:)
+        new(type: "request_review", payload: {
+          pr_number: pr_number,
+          reviewers: reviewers
+        })
+      end
+
+      def mark_ready(issue_id:, pr_number:, owner_reviewer_login: nil)
+        new(type: "mark_ready", payload: {
+          issue_id: issue_id,
+          pr_number: pr_number,
+          owner_reviewer_login: owner_reviewer_login
+        }.compact)
+      end
+
+      def escalate(issue_id:, pr_number:, owner_reviewer_login: nil, reason: nil)
+        new(type: "escalate", payload: {
+          issue_id: issue_id,
+          pr_number: pr_number,
+          owner_reviewer_login: owner_reviewer_login,
+          reason: reason
+        }.compact)
+      end
+
+      def dismiss_escalation(issue_id:)
+        new(type: "dismiss_escalation", payload: { issue_id: issue_id })
+      end
+
+      def merge(issue_id:, pr_number:)
+        new(type: "merge", payload: {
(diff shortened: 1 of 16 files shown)


viamin:
## Agent Update

The background task confirmed the commit succeeded (exit code 0). Everything is done — the commit with the `db/schema.rb` fix has been made and all checks passed.

PR #1141, opened 2026-04-15
Fix #1125: Orphan Branches Without PRs: Implement Pre-Run Branch Existence Check

### LINKED ISSUE #1125: Orphan Branches Without PRs: Implement Pre-Run Branch Existence Check
Problem: Many AgentRuns push branches but fail to create PRs, leaving orphan branches on GitHub. This regression is causing unnecessary cleanup and potential lost work. Implementation notes: Add a lightweight guard at the start of CreatePullRequestActivity that checks if the branch exists and whether an open PR already uses it. If the branch exists but no PR, automatically create the PR using existing logic. If the branch is missing, fall back to normal flow. This ensures PRs are always created when possible and avoids duplicate PRs. Label: P2

viamin:
## Summary

Prevent orphan branches on GitHub by adding a pre-run branch existence guard to `CreatePullRequestActivity`. When an AgentRun pushes a branch but fails to create a PR, the activity now detects the existing branch and creates the PR automatically, eliminating manual cleanup and potential lost work.

## Changes

**Temporal activity (`create_pull_request_activity.rb`)**
- Added `branch_exists?` private method that checks GitHub for the branch via `client.ref`, handling `NotFoundError` (info log) and transient errors (warn log) gracefully
- Restructured `execute` into three explicit paths:
  1. **PR already exists** — reuse it (idempotent, unchanged behavior)
  2. **Branch exists, no PR** — create the PR (fixes the orphan branch regression)
  3. **Branch missing** — fall back to normal create flow (handles races and transient failures)

**Tests (`create_pull_request_activity_spec.rb`)**
- Added `ref` stub to shared `before` block for compatibility with the new guard
- Added 6 new test cases covering branch existence checks, PR creation on orphan branches, fallback behavior on 404 and transient errors, and structured logging at appropriate levels

## Design Decisions

- **Graceful degradation over hard failure**: If the branch existence check fails (rate limits, network issues), the activity falls back to the normal flow rather than raising. This avoids blocking PR creation on a transient GitHub API failure.
- **Reuses existing PR creation logic**: The orphan branch path delegates to the same code that handles normal PR creation, avoiding duplication and ensuring consistent behavior (labels, reviewers, etc.).
- **Logging at two levels**: Missing branches log at `info` (expected in normal flow) while unexpected failures log at `warn` (actionable for operators), keeping signal-to-noise ratio appropriate.

---

Closes #1125

### DIFF
--- a/app/temporal/activities/create_pull_request_activity.rb
+++ b/app/temporal/activities/create_pull_request_activity.rb
@@ -22,10 +22,7 @@ module Activities
         if existing_pr
           pr = existing_pr
           pr_action = "reused"
-        else
-          # Create the PR whether the branch was confirmed to exist or the
-          # ref lookup failed transiently. If the branch is truly missing,
-          # create_pull_request will raise and Temporal will retry.
+        elsif branch_exists
           pr = client.create_pull_request(
             project.full_name,
             base: project.default_branch,
@@ -35,6 +32,10 @@ module Activities
             draft: true
           )
           pr_action = "created"
+        else
+          # Branch confirmed missing (404). Raise so Temporal retries —
+          # the branch may appear after a push that is still in flight.
+          raise "Branch #{agent_run.branch_name} does not exist on GitHub"
         end
 
         # Persist completion as the very first step after obtaining the PR,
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

Committed successfully. The change collapses the duplicate `elsif branch_exists` and `else` branches into a single `else` path, as the reviewer suggested. The `branch_exists?` call is still made for logging/observability, but it no longer drives a separate code path since both paths were identical. All 52 specs pass and lint is clean.

viamin:
## Agent Update

Commit succeeded. All lint checks passed and tests pass.

**Summary of changes:**

1. **Fixed the unused `branch_exists` variable** (reviewer's option 2): The `execute` method now implements all three intended paths:
   - **PR exists** → reuse it
   - **Branch exists, no PR** → create the PR
   - **Branch confirmed missing (404)** → raise so Temporal retries

2. **Updated `branch_exists?` to return `true` on transient errors** (optimistic approach): This preserves graceful degradation — if the ref check fails due to rate limits or network issues, we still attempt PR creation and let GitHub be the source of truth.

3. **Updated tests**: The "branch not found" test now expects a `RuntimeError` raise instead of falling through to `create_pull_request`, and the logging test was updated accordingly.

PR #1145, opened 2026-04-15
Fix #1042: feat(knowledge): containerize decision drafting to use secrets proxy

### LINKED ISSUE #1042: feat(knowledge): containerize decision drafting to use secrets proxy
## Background / Why

`Knowledge::Decisions::Draft` calls `AgentHarness.send_message(provider: :claude, model: "claude-sonnet-4-6")` **in the Rails/GoodJob process**. This means:
- Anthropic API credentials are loaded into the Rails process memory
- No container isolation for the LLM call
- No secrets proxy — the process has direct key access
- Violates the security principle in `docs/SECURITY.md` that "API keys never enter containers; Paid proxies all authenticated calls"

Agent runs execute inside hardened Docker containers with proxy-based key injection. Decision drafting should follow the same pattern.

## Scope

**In scope:**
- Create a containerized execution path for decision drafting
- Container authenticates to secrets proxy via `knowledge_run_id + proxy_token`
- Container calls LLM via proxy (not direct API key)
- Fallback: if Docker unavailable, fall back to in-process `AgentHarness.send_message` (with provider fallback)
- Wire provider fallback configuration into the decision drafting flow

**Out of scope:**
- `KnowledgeRun` model and proxy auth (depends on #1040)
- Provider fallback configuration (depends on #1041)
- Embedding containerization (separate issue)

## Dependencies

- Depends on #1040 (proxy auth for knowledge runs)
- Depends on #1041 (provider fallback configuration)

### Containerized decision drafter

Create `Knowledge::AnalysisRunner` (or extend `ContainerizedRunner`):

```ruby
module Knowledge
  class AnalysisRunner
    CONTAINER_DEFAULTS = {
      image: "paid-agent:latest",
      memory_bytes: 256 * 1024 * 1024,  # 256MB
      cpu_quota: 100_000,                # 1 CPU
      pids_limit: 100,
      timeout_seconds: 60,
      network_mode: "bridge"             # needs proxy access
    }.freeze

    def initialize(project:, knowledge_run:)
      @project = project
      @knowledge_run = knowledge_run
    end

    def draft_decision(agent_run:)
      # 1. Ensure proxy token
      # 2. Provision container with proxy URL + token in env
      # 3. Execute Ruby script that calls LLM via proxy
      # 4. Parse JSON output from stdout
      # 5. Cleanup container
    end
  end
end
```

### In-container script

The container runs a Ruby script that:
1. Reads `KNOWLEDGE_RUN_ID` and `PROXY_TOKEN` from env
2. Reads `PROXY_BASE_URL` from env
3. Calls the proxy's `/api/proxy/anthropic/v1/messages` endpoint
4. Outputs the response to stdout as JSON

### Network restrictions

The container needs network access to the proxy only. Apply the same `NetworkPolicy` pattern as agent containers — restrict egress to the proxy endpoint.

### Fallback path

```ruby
result = if Knowledge::AnalysisRunner.available?
  Knowledge::AnalysisRunner.new(project:, knowledge_run:).draft_decision(agent_run:)
else
  # In-process fallback — still use provider fallback, but no container isolation
  providers = Knowledge::ProviderSelector.for_chat(user_setting:)
  providers.each do |provider|
    response = AgentHarness.send_message(prompt, provider:, model: ...)
    break response if response.success?
  end
end
```

## Acceptance Criteria

- [ ] Decision drafting runs inside an isolated Docker container when available
- [ ] Container authenticates to secrets proxy (no API keys in container env)
- [ ] In-process fallback works when Docker is unavailable
- [ ] Provider fallback is attempted (primary → fallbacks) in both containerized and in-process paths
- [ ] Container is cleaned up after execution (success or failure)
- [ ] Existing decision drafting behavior is preserved (DecisionRecord creation, audit logging)
- [ ] `bin/rspec` passes

## Notes / Risks

- **Latency**: Container spin-up adds ~2-5 seconds per decision draft. Acceptable since drafting happens asynchronously after agent runs complete.
- **Network policy**: The container needs egress to the proxy. If the proxy runs on `host.docker.internal`, ensure Docker DNS resolution works in the deployment environment.
- **Image**: Reuses `paid-agent:latest` which already has Ruby + necessary gems. No new image needed.

viamin:
## Summary

feat(knowledge): containerize decision drafting to use secrets proxy

See #1042 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1042

### DIFF
--- a/app/services/knowledge/analysis_runner.rb
+++ b/app/services/knowledge/analysis_runner.rb
@@ -130,7 +130,7 @@ module Knowledge
       exec_completed = false
       timed_out = false
 
-      watchdog = start_watchdog(timeout, mutex) do
+      watchdog = start_watchdog(timeout) do
         mutex.synchronize do
           unless exec_completed
             timed_out = true
@@ -308,7 +308,7 @@ module Knowledge
       "http://paid-proxy:#{proxy_port}"
     end
 
-    def start_watchdog(timeout, _mutex)
+    def start_watchdog(timeout)
       Thread.new do
         sleep(timeout)
         should_stop = yield
--- a/app/services/knowledge/analysis_runner.rb
+++ b/app/services/knowledge/analysis_runner.rb
@@ -1,0 +1,341 @@
+# frozen_string_literal: true
+
+require "docker-api"
+require "base64"
+
+module Knowledge
+  # Runs LLM calls inside an isolated Docker container that authenticates
+  # to the secrets proxy via a KnowledgeRun proxy token. No API keys are
+  # exposed inside the container — the proxy injects them server-side.
+  #
+  # Used by Knowledge::Decisions::Draft to containerize decision drafting.
+  #
+  # @example
+  #   runner = Knowledge::AnalysisRunner.new(project: project, knowledge_run: knowledge_run)
+  #   runner.with_container do |r|
+  #     output = r.call_llm(prompt, provider: "claude", model: "claude-sonnet-4-6")
+  #   end
+  class AnalysisRunner
+    class Error < StandardError; end
+    class ContainerError < Error; end
+    class TimeoutError < ContainerError; end
+
+    CONTAINER_DEFAULTS = {
+      image: "paid-agent:latest",
+      memory_bytes: 256 * 1024 * 1024,  # 256MB
+      cpu_quota: 100_000,                # 1 CPU
+      pids_limit: 100,
+      timeout_seconds: 60,
+      network_mode: "bridge"             # needs proxy access
+    }.freeze
+
+    # Default models per API service type, used when caller does not specify one.
+    DEFAULT_MODELS = {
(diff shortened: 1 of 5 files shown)


viamin:
## Agent Update

Commit succeeded. All three review comments have been addressed:

1. **Unused `_mutex` parameter** in `start_watchdog` — removed the parameter entirely since the block captures `mutex` from the caller's closure.

2. **Unnecessary DB write** in `finalize_knowledge_run!` — now only calls `update!` when the knowledge run is `active?`, avoiding a no-op round-trip.

3. **Container provisioning failure fallback** — `send_to_llm_containerized` now falls back to `send_to_llm_in_process(prompt)` when provisioning fails (e.g., missing image), instead of returning `nil`. Updated the corresponding spec to verify the fallback produces a `DecisionRecord`.

PR #1158, opened 2026-04-15
fix(dev): wait for overmind socket before exiting from --detach

viamin:
## Summary

- `bin/dev --detach` used to print "bin/dev detached (setsid pid=…)" and exit before the forked child had opened `.overmind.sock`, so a follow-up `overmind status` / `overmind connect` from the caller (e.g. `bin/setup`) would race and fail with `dial unix .overmind.sock: connect: no such file or directory`.
- Parent now polls `$OVERMIND_SOCKET` for up to `DEV_DETACH_READY_TIMEOUT` seconds (default 60) before returning. Bails out early with an error if the detached child dies first; times out with a warning pointing at `log/dev-update/overmind.log` rather than hanging setup scripts.
- Two new specs in `spec/lib/dev_script_spec.rb` cover the readiness-wait success path and the timeout-with-warning path. Stub `overmind` gained an opt-in `create_socket_on_start` knob so the stubbed `start` can materialize a real Unix socket.

## Test plan

- [x] `bin/rspec spec/lib/dev_script_spec.rb` — 10 examples, 0 failures (8 existing + 2 new)
- [x] `shellcheck bin/dev` — clean (only pre-existing SC1091 info on the `source` line)
- [x] `bin/rubocop spec/lib/dev_script_spec.rb` — clean
- [ ] Manually confirm `bin/setup` → `overmind status` no longer races on a fresh devcontainer

### DIFF
--- a/bin/dev
+++ b/bin/dev
@@ -51,6 +51,31 @@ if [ "$DEV_DETACH_FLAG" = "1" ] && [ -z "${DEV_DETACHED:-}" ]; then
   # the overmind PID you'd see in `ps` or `overmind status`.
   detached_pid=$!
   disown "$detached_pid" 2>/dev/null || true
+
+  # Wait for the detached child to open its Overmind socket before returning,
+  # so a follow-up `overmind status` / `overmind connect` from the caller
+  # (e.g. bin/setup) doesn't race the startup. Fall through with a warning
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

The background commit task completed successfully — commit `bb7485df` is on the branch, all 6573 tests pass, lint is clean.

viamin:

PR #1159, opened 2026-04-15
Fix #1151: fix(containers): retry git clone on transient DNS/network failures

### LINKED ISSUE #1151: fix(containers): retry git clone on transient DNS/network failures
## Summary

`Containers::GitOperations#clone_repo` runs `git clone` exactly once and raises `CloneError` on the first failure, including transient errors like `Could not resolve host: github.com`. Concurrent agent runs hitting Docker's embedded DNS resolver (`127.0.0.11`) at the same time make these transient failures more likely, which causes spurious clone failures on PRs with multiple concurrent runs.

Observed error:

```
Clone failed: Command exited with code 128 — Cloning into '.'...
fatal: unable to access 'https://github.com/viamin/paid.git/': Could not resolve host: github.com
```

## Root cause

[app/services/containers/git_operations.rb:686-687](https://github.com/viamin/paid/blob/main/app/services/containers/git_operations.rb#L686-L687):

```ruby
result = execute_git("clone", "--depth", "1", url, ".", timeout: clone_timeout)
raise CloneError, "Clone failed: #{error_with_stderr(result)}" if result.failure?
```

A single transient resolver/network hiccup becomes a hard, non-retriable failure. By contrast, `checkout_remote_branch` ([git_operations.rb:712-753](https://github.com/viamin/paid/blob/main/app/services/containers/git_operations.rb#L712-L753)) already has multi-step fallback logic. Clone has none.

DNS itself is allowed by the firewall ([network_policy.rb:298-300](https://github.com/viamin/paid/blob/main/app/services/network_policy.rb#L298-L300)) and the workspace volume is per-`agent_run.id` ([provision.rb:930](https://github.com/viamin/paid/blob/main/app/services/containers/provision.rb#L930)) so this is not a shared-volume or firewall-policy bug — it is a missing retry on a transient-by-nature operation.

## Proposed fix

Wrap `clone_repo` in a small bounded retry that backs off on transient network errors:

- `Could not resolve host`
- `Connection reset by peer`
- `Connection timed out`
- `Temporary failure in name resolution`
- `unable to access` paired with any of the above

The clone is already idempotent — the `rev-parse HEAD` guard at [git_operations.rb:678](https://github.com/viamin/paid/blob/main/app/services/containers/git_operations.rb#L678) lets a successful retry no-op if a prior attempt partially populated the workspace, and `cleanup_partial_clone!` ([git_operations.rb:700-710](https://github.com/viamin/paid/blob/main/app/services/containers/git_operations.rb#L700-L710)) handles the partial-clone case. So retry is safe.

Suggested shape: 3 attempts, exponential backoff (1s, 2s, 4s), only retry on the matched transient strings; permanent failures (auth, 404, repo not found) should still raise on first attempt so they fail fast.

## Out of scope (separate follow-ups worth filing if confirmed)

- `NetworkPolicy.ensure_network!` ([network_policy.rb:74-78](https://github.com/viamin/paid/blob/main/app/services/network_policy.rb#L74-L78)) is non-atomic between `get` and `create` — concurrent first-time creation can race. Low impact in practice (network usually already exists), so leaving out of this issue.
- `NetworkPolicy.fetch_github_ips` makes a synchronous host-side HTTPS call on each provision, adding load during bursts. Could be cached.

## Test plan

- [ ] Unit spec: `clone_repo` retries on stderr matching transient patterns; gives up after N attempts
- [ ] Unit spec: `clone_repo` does NOT retry on auth/404/permanent failures
- [ ] Unit spec: retry succeeds when the second attempt succeeds (no `CloneError` raised)
- [ ] Manual: trigger several concurrent agent runs on a PR and confirm none fail with transient DNS errors

viamin:
## Summary

Adds bounded retry with exponential backoff to `clone_repo` so that transient DNS and network failures—common when concurrent agent runs hit Docker's embedded DNS resolver simultaneously—no longer cause spurious `CloneError` crashes.

## Changes

**Service layer (`app/services/containers/git_operations.rb`)**
- Added `TRANSIENT_CLONE_PATTERNS` constant matching known transient network error strings (`Could not resolve host`, `Connection reset by peer`, `Connection timed out`, `Connection refused`, `Temporary failure in name resolution`)
- Wrapped `clone_repo` in a retry loop: up to 3 attempts with exponential backoff (1s, 2s)
- Added `transient_clone_error?` helper to match stderr output against transient patterns
- Cleans up partial clone artifacts between retry attempts via existing `cleanup_partial_clone!`
- Permanent failures (auth errors, 404, repo not found) still raise immediately on first attempt

**Tests (`spec/services/containers/git_operations_spec.rb`)**
- 12 new specs covering: successful retry on second attempt, max-attempt exhaustion, exponential backoff timing, partial clone cleanup between retries, each individual transient pattern, and non-retry for permanent auth/404 errors

## Design Decisions

- **Pattern matching on stderr rather than exit codes**: `git clone` returns exit code 128 for both transient and permanent failures; the only way to distinguish them is by inspecting the error message. The pattern list is intentionally conservative—only well-known transient strings are matched.
- **Reuses existing cleanup path**: The `cleanup_partial_clone!` method and `rev-parse HEAD` idempotency guard already existed for other failure modes, making retry safe without new filesystem logic.
- **3 attempts with short backoff**: Matches the transient nature of Docker DNS hiccups (typically resolve in <1s) while bounding total delay to ~7s worst case, well within typical clone timeout budgets.

---

Closes #1151

### DIFF
--- a/app/services/containers/git_operations.rb
+++ b/app/services/containers/git_operations.rb
@@ -666,6 +666,18 @@ module Containers
       raise Error, "Failed to unshallow repository: #{error_with_stderr(result)}"
     end
 
+    # Transient network/DNS error patterns that warrant a retry.
+    TRANSIENT_CLONE_PATTERNS = [
+      "Could not resolve host",
+      "Connection reset by peer",
+      "Connection timed out",
+      "Temporary failure in name resolution",
+      "Connection refused"
+    ].freeze
+
+    CLONE_MAX_ATTEMPTS = 3
+    CLONE_BASE_BACKOFF = 1 # seconds
+
     def clone_repo
       # Idempotent: skip clone if a previous attempt already populated /workspace.
       # This prevents failures on Temporal retries when the clone succeeded but a
@@ -683,8 +695,28 @@ module Containers
       project = agent_run.project
       url = "https://github.com/#{project.full_name}.git"
 
-      result = execute_git("clone", "--depth", "1", url, ".", timeout: clone_timeout)
-      raise CloneError, "Clone failed: #{error_with_stderr(result)}" if result.failure?
+      attempt = 0
+      loop do
+        attempt += 1
+        result = execute_git("clone", "--depth", "1", url, ".", timeout: clone_timeout)
+        return if result.success?
+
+        stderr = result[:stderr].to_s
+        if attempt < CLONE_MAX_ATTEMPTS && transient_clone_error?(stderr)
+          backoff = CLONE_BASE_BACKOFF * (2**(attempt - 1))
+          Rails.logger.warn(
+            message: "container_git.clone_transient_retry",
+            agent_run_id: agent_run.id,
+            attempt: attempt,
+            backoff_seconds: backoff,
+            stderr: stderr.truncate(500)
+          )
+          sleep(backoff)
+          cleanup_partial_clone! if workspace_contains_files?
+        else
+          raise CloneError, "Clone failed: #{error_with_stderr(result)}"
+        end
+      end
     end
 
     def workspace_contains_files?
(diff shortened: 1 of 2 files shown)


viamin:

PR #1140, opened 2026-04-15
Fix #1046: feat(knowledge): add token usage tracking for knowledge base LLM calls

viamin:
The issue context doesn't match the actual changes on this branch. The branch adds a `--restart-if-running` flag to `bin/dev`, not knowledge token tracking. I'll write the PR description based on the actual diff.

## Summary

Allow `bin/dev` to restart an already-healthy Overmind session instead of exiting early, so that `bin/setup` can refresh running processes after dependency updates without requiring manual intervention.

## Changes

**`bin/dev` script:**
- Add `--restart-if-running` flag (`-R` short form, or `DEV_RESTART_IF_RUNNING=1` env var) that runs `overmind restart` on a healthy existing session instead of printing "already running" and exiting
- Capture diagnostics on restart failure before exiting non-zero

**`bin/setup` script:**
- Pass `--restart-if-running` alongside `--detach` when launching `bin/dev`, so setup refreshes processes if Overmind is already running

**Tests:**
- `spec/lib/dev_script_spec.rb` — Two new cases: successful restart of a healthy session, and diagnostics capture on restart failure. Extended the overmind stub to handle the `restart` subcommand with configurable exit status.
- `spec/lib/setup_script_spec.rb` (new) — Verifies `bin/setup` passes `--detach --restart-if-running` to `bin/dev`, and that `--skip-server` skips `bin/dev` entirely. Uses stub executables to isolate the setup script from real system dependencies.

## Design Decisions

- **Restart vs. quit-and-start**: `overmind restart` sends signals to managed processes and lets them re-exec within the existing tmux session, avoiding the overhead and socket churn of a full quit/start cycle.
- **Diagnostics on failure**: Restart failure triggers the same `dev_supervisor_snapshot_state` diagnostic capture used for other Overmind failures, giving a consistent debugging experience.
- **`bin/setup` always requests restart**: Since setup may have updated gems or JS packages, restarting processes ensures the running server picks up changes. If Overmind isn't running, the flag is never reached (normal startup path runs instead).

---

Closes #1046

PR #1171, opened 2026-04-16
Fix #1146: fix(pr-description): disable tool access for Llm::GeneratePrDescription calls

### LINKED ISSUE #1146: fix(pr-description): disable tool access for Llm::GeneratePrDescription calls
## Background / Why

`Llm::GeneratePrDescription` (`app/services/llm/generate_pr_description.rb:80`) calls `AgentHarness.send_message` to generate PR bodies from the agent's summary. Under the hood this spawns the `claude` CLI in the Rails host process. Because Claude Code in `--print` mode still has tool access by default, it can run git/bash tools and describe the host's working-directory state instead of the prompt.

This caused #1140. Agent-harness was invoked with `cwd=/workspaces/paid`; the host's HEAD was checked out to an unrelated feature branch (the `updated-bin-dev` branch behind #1138, merged ~3h earlier). Claude ran `git diff main...HEAD` and `git log main...HEAD --oneline`, saw `3d4b554eb feat(dev): add --restart-if-running flag...`, and wrote a confidently wrong PR body beginning *"The issue context doesn't match the actual changes on this branch…"*. The actual agent work and the diff pushed to GitHub were correct — only the description was contaminated.

Full post-mortem and session trace: #1140.

## Scope

**In scope:**
- In `Llm::GeneratePrDescription#request_description`, pass the new agent-harness tools-disable option so the CLI cannot invoke tools while producing a PR description.
- Audit other `app/services/llm/*` services for text-only calls that should opt into the same restriction (scan for any issue/PR/diagnosis summarizer, classifier, etc.). Apply the restriction consistently.
- Add a regression spec that verifies the option is passed through on the `send_message` call.
- Bump the `agent-harness` gem version in `Gemfile`/`Gemfile.lock` to the release containing the upstream change.

**Out of scope:**
- Changing the transport for this call — tracked by the longer-term HTTP/text-mode follow-up.
- Any change in agent-harness itself (tracked upstream in viamin/agent-harness#113).

## Dependencies

- Depends on viamin/agent-harness#113 (the `tools: :none` / `disallowed_tools:` option must be released before this can land).

## Acceptance Criteria

- [ ] `Llm::GeneratePrDescription` calls `AgentHarness.send_message` with tools disabled.
- [ ] A regression spec verifies the option is passed (mock the harness call).
- [ ] Other text-only `Llm::*` services are audited and either updated or documented as needing tools.
- [ ] Manual reproduction: running the Rails server with the host repo checked out to an unrelated branch no longer causes `Llm::GeneratePrDescription` to describe that branch's diff.
- [ ] `bin/rspec` passes; `bin/lint` passes.

## Notes / Risks

- The Paid change itself is small once upstream lands — most of the work is the audit of other text-only services.
- Defense in depth only: the long-term fix is to stop invoking a CLI at all for text-only calls (#<longer-term follow-up>).

viamin:
## Summary

Prevents `Llm::GeneratePrDescription` and other text-only LLM services from contaminating their output with host-repo state by disabling tool access on the underlying `claude` CLI. This fixes #1140, where a PR body described an unrelated branch's diff because Claude Code ran `git` tools against the Rails host's working directory instead of answering the prompt.

### Dependencies
- Bump `agent-harness` from `~> 0.7.1` to `~> 0.7.2` to pick up the upstream `tools:` option (viamin/agent-harness#113); 0.7.3 is installed.

### Services (text-only LLM calls)
Pass `tools: :none` to every `AgentHarness.send_message` call that only needs text generation:
- `app/services/llm/generate_pr_description.rb` — the #1140 root cause
- `app/services/llm/generate_issue_title.rb`
- `app/services/agent_runs/diagnose_error.rb`
- `app/services/knowledge/decisions/draft.rb`
- `app/services/models/meta_agent_selector.rb`
- `app/services/prompt_evolution/mutate.rb`
- `app/services/style_guides/compress.rb`
- `app/services/style_guides/extract.rb`
- `app/temporal/activities/decompose_feature_activity.rb`

### Intentionally unchanged
- `app/services/agent_runs/execute.rb` — this is the real agent run and must retain tool access (`dangerous_mode: true`).

### Tests
- New regression specs in `generate_pr_description_spec.rb` and `generate_issue_title_spec.rb` assert `tools: :none` is forwarded to the harness.
- Updated 9 existing specs to include the new kwarg in their matcher expectations.

## Design Decisions

- **Defense in depth, not a root-cause fix.** Disabling tools is a safeguard layered on top of the CLI invocation. The long-term fix is to stop shelling out to a CLI for text-only calls; that remains tracked as a separate follow-up.
- **Audited every `Llm::*`-adjacent service rather than patching only `GeneratePrDescription`.** Any service that summarizes, classifies, or drafts text has the same class of failure mode, so all text-only `send_message` sites were updated consistently.
- **Upstream-first.** The `tools:` option was added in `agent-harness` (viamin/agent-harness#113) rather than worked around in Paid, keeping provider-specific behavior out of the Rails app per `CLAUDE.md`.

## Operational Notes

- Requires `agent-harness >= 0.7.2` at install time. `bundle install` will pick up `0.7.3` on deploy; no migration or config changes.

---

Closes #1146

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -46,7 +46,7 @@ gem "docker-api"
 gem "qdrant-ruby"
 
 # Unified interface for AI agent CLIs [https://github.com/viamin/agent-harness]
-gem "agent-harness", "~> 0.7.1"
+gem "agent-harness", "~> 0.7.2"
 
 # Code analysis tool for VCS mining (churn/hotspot analysis) [https://github.com/viamin/ruby-maat]
 gem "ruby-maat"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -518,7 +518,7 @@ PLATFORMS
   x86_64-linux-musl
 
 DEPENDENCIES
-  agent-harness (~> 0.7.1)
+  agent-harness (~> 0.7.2)
   bootsnap
   brakeman
   bundler-audit
--- a/app/services/agent_runs/diagnose_error.rb
+++ b/app/services/agent_runs/diagnose_error.rb
@@ -71,7 +71,8 @@ module AgentRuns
         diagnosis_prompt,
         provider: :claude,
         model: DEFAULT_MODEL,
-        timeout: TIMEOUT
+        timeout: TIMEOUT,
+        tools: :none
       )
       return nil unless response.success?
 
--- a/app/services/knowledge/decisions/draft.rb
+++ b/app/services/knowledge/decisions/draft.rb
@@ -190,7 +190,8 @@ module Knowledge
         options = {
           provider: ProviderSupport.harness_provider_key_for(provider).to_sym,
           timeout: TIMEOUT,
-          dangerous_mode: false
+          dangerous_mode: false,
+          tools: :none
         }
         options[:model] = DEFAULT_MODEL if provider == DEFAULT_PROVIDER
         options
--- a/app/services/llm/generate_issue_title.rb
+++ b/app/services/llm/generate_issue_title.rb
@@ -43,7 +43,13 @@ module Llm
     private
 
     def request_title
-      response = AgentHarness.send_message(prompt, provider: :claude, model: DEFAULT_MODEL, timeout: TIMEOUT)
+      response = AgentHarness.send_message(
+        prompt,
+        provider: :claude,
+        model: DEFAULT_MODEL,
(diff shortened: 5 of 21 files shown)


viamin:

PR #1177, opened 2026-04-16
Fix #1147: refactor(pr-description): route text-only Llm::* services through agent-harness text mode

### LINKED ISSUE #1147: refactor(pr-description): route text-only Llm::* services through agent-harness text mode
## Background / Why

Once agent-harness exposes a text-only / HTTP transport (viamin/agent-harness#114), Paid's pure-text services should use it instead of spawning the `claude` CLI from the Rails host. Benefits:

- Lower latency per call (no subprocess spin-up, no CLI context loading).
- No `CLAUDE.md`/memory loading and no `cwd` sensitivity — closes the #1140 class of bug at the root, not just via tool-disable.
- Clearer separation: tool-capable agent work keeps the CLI path; pure text generation uses HTTP.

This is the longer-horizon follow-up to #1146 (tools-disable quick fix). Land the quick fix first so we have a working baseline; then come back and switch transport.

## Scope

**In scope:**
- Update `Llm::GeneratePrDescription` to pass `mode: :text` (or the equivalent agent-harness option).
- Audit and migrate other `app/services/llm/*` services that are pure text generators (summarizers, classifiers, diagnosis drafters, etc.).
- Confirm billing/auth behavior does not change unexpectedly — subscription-backed vs API-key-metered routing must be preserved or require explicit opt-in.
- Observability: token usage, cost, and error counters continue to flow into existing dashboards/logs.

**Out of scope:**
- Any service that legitimately needs tool access (agent-run workflows) — those stay on the CLI path.
- Upstream agent-harness work itself (tracked in viamin/agent-harness#114).

## Dependencies

- Depends on viamin/agent-harness#114 (text-mode HTTP transport must exist and be released).
- Depends on #1146 (tools-disable quick fix) — land that first so host-cwd contamination is already fixed while this refactor is in flight.

## Acceptance Criteria

- [ ] `Llm::GeneratePrDescription` uses the text-only transport.
- [ ] Other text-only `Llm::*` services are migrated or documented as exceptions.
- [ ] Token usage and error metrics look equivalent before/after on the dashboards.
- [ ] Regression test confirms PR body generation no longer depends on host-process `cwd` or git state.
- [ ] Billing/auth routing is preserved; if a provider's HTTP path implies a different billing mode, the change is explicit and documented.

## Notes / Risks

- Subscription-vs-API billing: confirm on rollout that no call is silently re-routed from subscription-backed CLI usage to API-metered HTTP usage.
- Consider a staged rollout (e.g., env-gated) so regressions in the HTTP path are easy to revert.

viamin:
## Summary

refactor(pr-description): route text-only Llm::* services through agent-harness text mode

See #1147 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1147

### DIFF
--- a/app/services/agent_runs/diagnose_error.rb
+++ b/app/services/agent_runs/diagnose_error.rb
@@ -72,7 +72,8 @@ module AgentRuns
         provider: :claude,
         model: DEFAULT_MODEL,
         timeout: TIMEOUT,
-        tools: :none
+        tools: :none,
+        **Llm::TextMode.options
       )
       return nil unless response.success?
 
--- a/app/services/knowledge/decisions/draft.rb
+++ b/app/services/knowledge/decisions/draft.rb
@@ -13,6 +13,13 @@ module Knowledge
     # (no API keys in container). Falls back to in-process AgentHarness when
     # Docker is unavailable.
     #
+    # Text-mode routing (#1147): the in-process Claude path opts into
+    # +Llm::TextMode.options+ so host-side drafting uses the HTTP text
+    # transport when +ANTHROPIC_API_KEY+ is configured. Non-Claude providers
+    # and the containerized path continue to use their existing CLI transport
+    # unchanged, because +TextTransport+ is Anthropic-only and the container
+    # already isolates +cwd+/memory concerns.
+    #
     # @example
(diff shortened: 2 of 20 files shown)


viamin:

PR #1178, opened 2026-04-16
Fix #1163: Move co-authored-by trailer from project setting to provider setting

### LINKED ISSUE #1163: Move co-authored-by trailer from project setting to provider setting
## Problem

The `Co-Authored-By` trailer in commit messages is currently configured as a project-level setting. This doesn't account for provider fallback behavior — if the configured provider falls back to a different model (e.g., from Claude to Codex), the trailer would still incorrectly attribute the co-authorship to Claude.

## Proposed Solution

Make the `Co-Authored-By` trailer a **provider-level setting** instead of a project-level setting. Each provider should define its own trailer so that the attribution accurately reflects which provider actually generated the code.

This way, if a fallback occurs, the correct trailer for the provider that was actually used will be applied automatically.

## Why This Matters

- Accurate attribution: commits should reflect the actual AI provider that contributed
- Avoids misleading co-authorship claims when provider fallback kicks in
- Each provider can define the appropriate trailer for their service

viamin:
## Summary

Move co-authored-by trailer from project setting to provider setting

See #1163 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1163

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -226,7 +226,6 @@ class ProjectsController < ApplicationController
       :auto_add_labels_enabled, :automation_on_label_enabled, :pr_aggregation_enabled,
       :inherit_priority_labels,
       :auto_release_granularity,
-      :agent_co_author_trailer,
       allowed_github_usernames: [],
       priority_labels: Project::PRIORITY_TIERS)
   end
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -147,7 +147,7 @@ class ProvidersController < ApplicationController
   end
 
   def provider_params
-    permitted = [ :enabled_for_agent_runs, :enabled_for_fallback, :name, :fallback_role ]
+    permitted = [ :enabled_for_agent_runs, :enabled_for_fallback, :name, :fallback_role, :agent_co_author_trailer ]
     if action_name == "create"
       permitted.push(:provider_key, :auth_type, :provider_api_key_id)
     end
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -863,6 +863,15 @@ class AgentRun < ApplicationRecord
     owner.providers.find_by(id: provider_id) if provider_id
   end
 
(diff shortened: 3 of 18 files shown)


viamin:
## Agent Update

Both review threads are addressed in commit `7e7d2498`:

**Thread 1 — Migration data loss**
- Converted `change` to explicit `up`/`down` in `db/migrate/20260416050235_move_co_author_trailer_from_projects_to_providers.rb`
- `up` copies each project's trailer onto the project owner's default subscription provider (claude preferred; ties broken by most-recent `updated_at`), matching `Project#effective_owner`'s fallback chain (`created_by` → account owner → first account user) in SQL before dropping the legacy column
- `down` best-effort restores trailers from providers back onto projects the owner created
- Added `spec/migrations/move_co_author_trailer_from_projects_to_providers_spec.rb` with five scenarios (basic copy, nil input, multiple projects picking latest, whitespace handling, claude-preferred)

**Thread 2 — Hook staleness after fallback**
- Added `CO_AUTHOR_TRAILER_FILE = ".git/paid_co_author_trailer"` constant and new `write_co_author_trailer(provider_record)` method on `Containers::GitOperations`
- Changed both `commit_msg_hook_script` and `commit_msg_wrapper_script` to read the trailer from that file at commit-msg time rather than baking it in
- `install_co_author_hook` now skips only when *no* provider owned by the project owner has a trailer (so fallback-only trailers still trigger install), seeds the file with the initial provider's trailer, and still wraps existing hooks safely
- `Activities::RunAgentActivity#run_agent_with_provider` calls a new `refresh_co_author_trailer` helper before each provider attempt (when `repo_cloned?`) so intermediate commits created during or after fallback carry the correct trailer
- Added a regression spec verifying `write_co_author_trailer` is called in-order with each provider across a rate-limit fallback, plus a spec confirming non-repo runs skip the refresh

Lint is clean; all tests for the affected files (git_operations, run_agent_activity, migration, models) pass. The 14 pre-existing flaky failures in the full suite run pass individually and are in untouched files.

PR #1182, opened 2026-04-16
chore(infra): pin Postgres to 16.13 across devcontainer, compose, and CI

viamin:
## Summary

- Pin the three Postgres infra locations to the same exact patch version (`16.13`), eliminating a drift source that was producing cosmetic `db/schema.rb` flip-flops across contributors.
- Add a CI guard (lint-job step) that fails the build if the three pins ever drift apart again, so this fix doesn't rot.
- Before: [.devcontainer/compose.yaml](.devcontainer/compose.yaml) and [docker-compose.yml](docker-compose.yml) pinned stale `16.1` (Nov 2023); [.github/workflows/ci.yml](.github/workflows/ci.yml) used the floating `postgres:16` tag.
- `pg_get_expr()` serializes index/check-constraint predicates slightly differently across 16.x patch releases (e.g. `ARRAY[('x'::varchar)::text, ...]` vs `ARRAY['x'::varchar, ...]::text[]`) — semantically identical but textually divergent. With CI running whichever 16.x Docker Hub currently tags as latest and local devs on 16.1, whoever last re-dumped schema set the "winning" format, and the next person to dump from the other version produced a diff.

## Verification

Loaded main's `db/schema.rb` into a fresh `postgres:16.13` container and re-dumped — produced byte-identical output to what's committed. So merging this does NOT introduce a one-time schema churn; existing `main` was already written by a 16.13-equivalent.

## Test plan

- [ ] CI green on this branch (Postgres service now pins to 16.13 explicitly)
- [ ] New "Verify Postgres image pins are in sync" step passes on aligned pins and fails the build when drift is introduced (dry-run confirmed locally)
- [ ] Rebuild devcontainer locally, run `bin/setup`, confirm `db/schema.rb` stays clean after `db:prepare` / `db:migrate`
- [ ] `docker compose up postgres` pulls `postgres:16.13` (no unexpected restart, data dir remains compatible — 16.1 → 16.13 is in-major, no dump/restore needed)

## Notes

- This does NOT touch the `service_containers` allowlist defaults (`postgres:16` as a user-facing example) since those are data-layer defaults for user projects, not infra for running Paid itself.
- Next bump should be all three locations together — the CI guard now enforces that.

### DIFF
--- a/.devcontainer/compose.yaml
+++ b/.devcontainer/compose.yaml
@@ -30,7 +30,7 @@ services:
         condition: service_healthy
 
   postgres:
-    image: postgres:16.1
+    image: postgres:16.13
     restart: unless-stopped
     networks:
     - default
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,26 @@ jobs:
       - name: Run lint checks
         run: bin/lint
 
+      - name: Verify Postgres image pins are in sync
+        # All three files must pin the same postgres:<tag> so that schema.rb
+        # dumps are byte-identical across dev, devcontainer, and CI. Drift
+        # across minor versions causes cosmetic pg_get_expr() churn in schema.rb.
+        run: |
+          dev=$(grep -oE 'postgres:[0-9][^[:space:]]*' .devcontainer/compose.yaml | head -1)
+          compose=$(grep -oE 'postgres:[0-9][^[:space:]]*' docker-compose.yml | head -1)
+          ci=$(grep -oE 'postgres:[0-9][^[:space:]]*' .github/workflows/ci.yml | head -1)
+          echo "devcontainer/compose.yaml: ${dev:-<none>}"
+          echo "docker-compose.yml:        ${compose:-<none>}"
+          echo ".github/workflows/ci.yml:  ${ci:-<none>}"
+          if [ -z "$dev" ] || [ -z "$compose" ] || [ -z "$ci" ]; then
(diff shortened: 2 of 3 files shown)


viamin:
## Agent Update

Review comment addressed, commit `7969ffc9` landed, working tree clean.

**What changed**
- `.github/workflows/ci.yml:57-59` — changed `grep -oE 'postgres:[0-9][^[:space:]]*' <file> | head -1` to `grep -oE 'image:[[:space:]]+postgres:[^[:space:]]+' <file> | head -1 | awk '{print $NF}'`, anchoring the match to the `image:` YAML key.

**Why the reviewer was right**
I confirmed the bug locally: the old pattern yielded three matches in `docker-compose.yml` — the real image pin (`postgres:16.13`) and two URL-style references (`postgres:5432/paid_development`). `head -1` picked the right one only by luck of line ordering (line 14 vs lines 54, 158). Any reorder or earlier insertion would silently compare a port number against an image tag.

**Verification**
- New pattern extracts `postgres:16.13` cleanly from all three files.
- Pathological fixture where a URL-style reference sits *above* the `image:` line: old pattern returns `postgres:5432/paid_development`, new pattern correctly returns `postgres:16.13`.
- Drift-detection still fires: injected `16.12` into one file and the guard correctly reported mismatch.

**Commit checks**
- lint (`bin/lint --staged`) — passed
- full rspec suite — 6746 examples, 0 failures, 93.41% coverage

Nothing pushed, per instructions.

viamin:

PR #1183, opened 2026-04-16
Fix #1118: Add GitHub automation adapters

### LINKED ISSUE #1118: Add GitHub automation adapters
Depends on #1116

## Summary
Introduce GitHub-backed automation adapters that satisfy the new provider capability interfaces.

## Goals
- Wrap existing GitHub behavior behind provider/capability abstractions.
- Create a clean reference implementation for future GitLab/Bitbucket/Linear/Jira adapters.
- Keep behavior unchanged while moving GitHub-specific transport concerns out of automation policy.

## Should Cover
- Repository-hosting operations used by automation.
- GitHub Issues / PR / review state queries needed by auto-pick and auto-continue.
- Review-request and merge-related operations needed by auto-review and auto-merge.

## Deliverable
- GitHub provider adapters aligned with the new automation interfaces.

viamin:
## Summary

Add GitHub automation adapters

See #1118 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1118

### DIFF
--- a/app/services/automation/providers/github/base_adapter.rb
+++ b/app/services/automation/providers/github/base_adapter.rb
@@ -1,0 +1,93 @@
+# frozen_string_literal: true
+
+module Automation
+  module Providers
+    module Github
+      # Shared behavior for the GitHub-backed automation adapters.
+      #
+      # All three capability adapters (repository, review, work item) front
+      # the same underlying {::GithubClient}. BaseAdapter resolves the client
+      # from the project, centralizes error translation into the provider
+      # error hierarchy expected by automation policy, and provides the small
+      # helpers the adapters share (login normalization, timestamp parsing,
+      # etc.).
+      #
+      # Adapters MUST wrap outbound GitHub calls in {#with_errors} so that
+      # policy code rescues a single +ProviderError+ family regardless of
+      # which adapter raised.
+      class BaseAdapter
+        # Error class the concrete adapter raises on provider failure.
+        # Subclasses override to point at the capability-specific hierarchy
+        # (RepositoryProvider::ProviderError, etc.) so policy code can
+        # rescue narrowly without losing access to the unified family.
+        PROVIDER_ERROR = Class.new(StandardError)
+
+        attr_reader :project
+
+        # @param project [Project, #github_token] The Paid project whose
+        #   GitHub credentials back this adapter. Tests may pass any object
+        #   responding to +#github_token+ (or +#client+ directly via the
+        #   +client:+ keyword) to avoid loading a real project.
+        # @param client [::GithubClient, nil] Optional pre-built client for
+        #   tests. When omitted, the client is lazily derived from
+        #   +project.github_token.client+.
+        def initialize(project, client: nil)
+          @project = project
+          @client = client
+        end
+
+        # Exposes the underlying GitHub client to subclasses and specs.
+        #
+        # @return [::GithubClient]
+        def client
+          @client ||= resolve_client
+        end
+
+        private
+
+        def resolve_client
+          token = project.respond_to?(:github_token) ? project.github_token : nil
+          raise provider_error_class, "Project #{project_identifier} has no GitHub token" if token.nil?
+
+          token.client
+        end
+
+        def project_identifier
+          project.respond_to?(:full_name) ? project.full_name : project.inspect
+        end
+
+        # Wraps a GitHub API call, translating {::GithubClient} error classes
+        # into this adapter's ProviderError. Unknown errors propagate so
+        # that genuine bugs do not masquerade as provider failures.
+        def with_errors
+          yield
+        rescue ::GithubClient::Error => e
+          raise provider_error_class, e.message
+        end
+
+        # Concrete adapters override to return the capability-specific
+        # ProviderError class declared by the interface module.
+        def provider_error_class
+          self.class::PROVIDER_ERROR
+        end
+
+        def normalize_login(value)
+          value&.to_s&.downcase.presence
+        end
+
+        def parse_time(value)
+          case value
+          when Time, ActiveSupport::TimeWithZone then value
+          when String then Time.parse(value)
+          end
+        rescue ArgumentError
+          nil
+        end
+
+        def extract_labels(source)
+          Array(source).map { |label| label.respond_to?(:name) ? label.name : label.to_s }
+        end
+      end
+    end
+  end
+end
--- a/app/services/automation/providers/github/base_adapter.rb
+++ b/app/services/automation/providers/github/base_adapter.rb
@@ -87,6 +87,23 @@ module Automation
         def extract_labels(source)
           Array(source).map { |label| label.respond_to?(:name) ? label.name : label.to_s }
         end
+
+        # Accept either a Sawyer::Resource (Octokit's default) or a plain Hash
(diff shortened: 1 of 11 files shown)


viamin:
## Agent Update

Commit `e27b65d1` is in, pre-commit hooks passed, working tree clean. 

## Summary

Both review threads addressed the same concern: `remove_label` classified 404s by regex-matching the error message, which fails because `with_errors` discards the typed `GithubClient::NotFoundError` class, and Octokit's 404 messages aren't guaranteed to contain "not found" (e.g., `"Label does not exist"` would leak as a real `ProviderError`).

**Changes:**
- `app/services/automation/providers/github/repository_provider.rb:75-84` — rescue `::GithubClient::NotFoundError` directly for idempotent no-op; rescue `::GithubClient::Error` to convert to `PROVIDER_ERROR`.
- `app/services/automation/providers/github/work_item_provider.rb:71-80` — same fix applied.
- Added a regression test in each spec that raises a `NotFoundError` with message `"Label does not exist"` to prove classification is by class, not message text.

**Proactive scan:** The only other regex-based error classification is in `app/services/github_client.rb` (`message.match?(/unprocessable/i)`, `message.include?("rate limit")`), but those are pre-existing code outside this PR's diff.

**Verification:** `bin/rubocop` clean on all 4 changed files; `bin/rspec spec/services/automation/` passes (110 examples, 0 failures). Pre-commit hooks passed on commit.

viamin:
## Agent Update

Committed as `18915fb4`. Pre-commit hook ran the full test suite (6802 examples, 0 failures) and lint before allowing the commit.

## Summary

**Review comment fixes:**

1. `app/services/automation/providers/github/work_item_provider.rb:36-53` — `list_issues` now raises `ProviderError` when given more than one assignee. The GitHub REST `/issues` endpoint only supports a single `assignee` parameter, and silently dropping additional assignees could make callers think they were getting an intersection filter. Blank/nil entries are filtered before counting so `[nil, "bob"]` still works as a single-login filter.

2. `app/services/automation/providers/github/repository_provider.rb:163-173` — `build_merge_result` now defaults `merged` to `false` (via `response_field(response, :merged) == true`) instead of treating absence as success. The unused `bool_field` helper was removed.

**Tests added:**
- Multi-assignee `ProviderError` case and a blank-entry filtering case in `work_item_provider_spec.rb`
- Missing-`merged`-field default-to-false case in `repository_provider_spec.rb`

**CI "test ()" failures:** These were pre-existing failures on `main`, caused by a phantom migration row (`20260416050235`) left in the test DB's `schema_migrations` table with no corresponding migration file — not by anything in this branch's diff. I cleared the stale row locally and the full suite passes (6802 examples, 0 failures). If CI is hitting the same stale-DB condition, the CI DB reset step may need attention separately; nothing in this PR's code causes those failures.

**Lint:** `bundle exec rubocop` — 850 files, no offenses.

viamin:
## Agent Update

The review comment is addressed in commit `9fbb21e9`.

**Summary of changes:**

- **`app/services/github_client.rb:343-372`** — Widened `#check_runs_for_ref` to preserve `:status`, `:html_url`, and `:details_url` alongside the existing `:name`/`:conclusion`, and updated the YARD docstring. Existing callers (`scan_paid_prs_activity`, `auto_release_evaluation_job`, `Prompts::BuildForPr`) only read `:name`/`:conclusion`, so this is backwards-compatible — the reviewer's option 1.

- **`spec/services/automation/providers/github/repository_provider_spec.rb`** — Split the `#fetch_check_runs` spec into three cases (completed / in-progress / queued) with an explicit regression guard asserting that an in-progress run surfaces as `:in_progress` and carries the expected URL — so the bug the reviewer identified cannot silently come back.

- **`spec/services/github_client_spec.rb:722-757`** — Updated the `#check_runs_for_ref` happy-path fixture and assertion to cover all five fields, including an `in_progress` run.

**Verification:** `bundle exec rubocop` (863 files, 0 offenses) and `bundle exec rspec` (6856 examples, 0 failures) pass — same results both before and after the pre-commit hook re-ran them.

viamin:
## Agent Update

## Summary

Addressed the unresolved review thread on `app/services/automation/providers/github/repository_provider.rb:147`.

**Fix** (commit `8f1a7213`):
- Changed the `CHECK_STATUS_MAP` fallback from `:completed` to `:queued`. The reviewer correctly identified that defaulting to `:completed` would let consumers gating on "all checks finished?" advance while a check is still running, if GitHub introduces a new status value (e.g. `waiting`, `pending`).
- Added a code comment explaining why `:queued` is the conservative default.
- Added a regression spec `defaults unknown statuses to :queued, never :completed` that pins the behavior against future GitHub status additions.

**Proactive scan** of the rest of the PR for the same class of issue (unknown values silently classified as a terminal/advancing state):
- `PR_STATE_MAP.fetch(..., :closed)` and `ISSUE_STATE_MAP.fetch(..., :closed)` — `:closed` is the conservative choice for picker/auto-claim gates that filter for `:open`.
- `REVIEW_STATE_MAP[...] || :commented` — `:commented` is a passive state that doesn't satisfy approval gates or trigger changes-requested gates.
- `map_timeline_event` defaults to `:other` — explicit catch-all.
- `MERGE_METHODS` is an explicit allow-list that raises on unknown input.

No similar bugs found.

**Verification**: rubocop clean; full suite 6857 examples, 0 failures (run by pre-commit hook).

viamin:
## Agent Update

Both review threads addressed in commit `16a783b2`:

**Thread 1** (base_adapter.rb): Hoisted `read_field` and `read_sub_field` to `BaseAdapter` alongside the other shared helpers (`normalize_login`, `parse_time`, `extract_labels`). Removed the duplicate copies from `repository_provider.rb`, `review_provider.rb`, and `work_item_provider.rb`.

**Thread 2** (repository_provider.rb): Deleted the six passthrough accessors (`pr_field`, `pr_sub_field`, `run_field`, `comment_field`, `comment_sub_field`, `response_field`) and replaced all callers in `build_pull_request`, `build_check_run`, `build_comment`, and `build_merge_result` with direct `read_field` / `read_sub_field` calls — matching the style of `review_provider.rb` and `work_item_provider.rb`.

Net: `4 files changed, 46 insertions(+), 96 deletions(-)`. Pre-commit hook ran rubocop (no offenses) and the full rspec suite (6878 examples, 0 failures).

viamin:
## Agent Update

Both review threads are addressed in commit `3c1568ea`:

**Thread 1** (`base_adapter.rb:102`) — `read_field` was using `source[key] || source[key.to_s]`, which converted legitimate `false` values to `nil` for Hash sources. Switched to `key?`-based branching so falsy-but-present values survive. Added a regression test that verifies `mergeable: false` stays `false`.

**Thread 2** (`repository_provider.rb:153`) — Unknown conclusions silently became `nil`, colliding with `Data::CheckRun#conclusion`'s documented "still running" meaning. Extracted `normalize_conclusion`: preserves `nil` as "still running", but logs a structured warning (`automation.github.unknown_check_conclusion`) and defaults unknown non-nil values to `:action_required` so auto-merge stays conservative until the map is updated. Added a test that verifies both the mapping and the log.

Proactive scan found no other `||`-on-potentially-false or silent-nil-fallback issues — the remaining `||` patterns (URL, login, SHA fallbacks) all operate on strings. Rubocop clean and the full test suite (6880 examples) passes.

PR #1230, opened 2026-04-17
Fix #852: feat(providers): configure Codex notify heartbeat hook

### LINKED ISSUE #852: feat(providers): configure Codex notify heartbeat hook
## Summary

Codex (OpenAI) supports a `notify` hook in `~/.codex/config.toml` that fires at turn boundaries (one complete agentic loop iteration). Configure this hook to write a heartbeat timestamp at container setup time.

## Implementation

Write `~/.codex/config.toml` into the container during provisioning:

```toml
[notify]
command = "date +%s > /tmp/agent_heartbeat"
```

Note: Codex's notify fires per-turn (coarser than Claude's per-tool-call), but still better than no signal. A turn may involve multiple tool calls before the notify fires.

## Parent

Parent issue: #848

Depends on #850 (watchdog heartbeat file support).

## Files

- `app/models/provider.rb` — container setup commands (`container_setup_script` method)
- `app/services/containers/provision.rb` — provisioning logic

## Notes

For Kilocode and OpenCode: no hooks/extension points are available. These providers will rely on the existing stdout-based idle timeout watchdog only.

viamin:
## Summary

feat(providers): configure Codex notify heartbeat hook

See #852 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #852

### DIFF
--- a/app/services/containers/provision.rb
+++ b/app/services/containers/provision.rb
@@ -687,6 +687,9 @@ module Containers
         base_url = "#{proxy_base_url}/api/proxy/openai"
         env_key = "OPENAI_API_KEY"
         wire_api = "responses"
+
+        [notify]
+        command = "date +%s > /tmp/agent_heartbeat"
       TOML
 
       write_container_file("/home/agent/.codex/config.toml", content)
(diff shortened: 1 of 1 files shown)


viamin:

PR #1232, opened 2026-04-17
Fix #802: "Back to ..." links should return to the previous page, not a hardcoded destination

### LINKED ISSUE #802: "Back to ..." links should return to the previous page, not a hardcoded destination
## Problem

The "← back to …" links at the top of pages don't always return the user to the page they came from. Instead, they navigate to a hardcoded destination that may not match the user's actual navigation history.

### Example

- From the **Agent Runs** page, clicking into a specific agent run shows a back link that takes the user to the **project** the agent run is associated with — not back to the Agent Runs list where they started.

This pattern likely affects many pages across the app where similar back links are used.

## Expected Behavior

The back link should return the user to the page they navigated from. If a user arrived at an agent run detail page from the Agent Runs index, the link should go back to the Agent Runs index. If they arrived from a project page, it should go back to the project page.

## Suggested Approach

Consider one of the following:

1. **Use a `referrer` or `return_to` query parameter** — pass the originating URL as a parameter so the back link can use it dynamically.
2. **Use browser history** (`history.back()` or equivalent) — though this can be less predictable if the user opened the page in a new tab.
3. **Infer the correct parent from context** — if the URL includes enough information to determine where the user came from, use that to build the back link.

Option 1 is generally the most reliable and explicit approach.

viamin:
## Summary

"Back to ..." links should return to the previous page, not a hardcoded destination

See #802 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #802

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -381,6 +381,8 @@ module ApplicationHelper
     raise unless Rails.env.test?
   end
 
+  private
+
   def safe_return_path?(path)
     path.start_with?("/") && !path.start_with?("//")
   end
@@ -391,8 +393,6 @@ module ApplicationHelper
   rescue URI::InvalidURIError
     false
   end
-
-  private
 
   def priority_tier_for_label(project, label)
     return "P0" if label == "P0"
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -278,6 +278,20 @@ module ApplicationHelper
     end
   end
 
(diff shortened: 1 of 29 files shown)


viamin:
## Agent Update

The review feedback has been addressed: `safe_return_path?` and `safe_referer?` are now moved below the `private` keyword in `app/helpers/application_helper.rb`, making them private methods as they should be since they're only used internally by `back_link_path`.

viamin:

PR #1239, opened 2026-04-17
Fix #1235: perf(poll-workflow): batch DetectLabels into a single activity

### LINKED ISSUE #1235: perf(poll-workflow): batch DetectLabels into a single activity
## Problem

[GitHubPollWorkflow](app/temporal/workflows/git_hub_poll_workflow.rb#L34-L39) invokes a separate `DetectLabelsActivity` for every open issue in the project, sequentially:

```ruby
result[:issues].each do |issue_data|
  evaluation = run_activity(Activities::DetectLabelsActivity,
    { project_id: project_id, issue_id: issue_data[:id] }, timeout: 30)
  handle_automation_result(evaluation, project_id)
end
```

For `viamin/paid` (124 open issues) that is **124 Temporal activity roundtrips per poll iteration**. Each roundtrip carries ~100-300ms of task-queue overhead even when the activity itself only touches the DB (~50ms). The work is also pure-DB (no network), so the fan-out offers no parallelism benefit — it just multiplies overhead.

## Impact

This is the dominant contributor to per-iteration cost on larger repos. Eliminating the per-issue fan-out should reduce a clean iteration from several minutes to under a minute, making the 18-minute staleness window (`3 × poll_interval + 3 min`, see [project.rb:350](app/models/project.rb#L350)) actually achievable under load.

Observed today: 6 of 10 projects simultaneously stale because poll iterations consistently run longer than 18 minutes. The 124-activity DetectLabels fan-out is the clearest structural contributor.

## Proposed fix

Introduce a batch activity (e.g. `EvaluateIssuesActivity`) that accepts `project_id` and a list of `issue_id`s and returns the full list of decisions in one call. Inside the activity, reuse the `Automation::Evaluator` loop but without crossing the Temporal boundary per issue.

`handle_automation_result` remains the dispatcher on the workflow side — it just iterates the returned decision list once instead of being called N times.

## Acceptance criteria

- [ ] New batch activity replaces the per-issue `DetectLabelsActivity` call
- [ ] `handle_automation_result` still executes each decision in the same order as today
- [ ] A poll iteration on `viamin/paid` (124 open issues) completes in <60s of activity time, down from the current multi-minute path
- [ ] Existing `DetectLabelsActivity` can stay as-is for single-issue webhook-driven paths if needed, or be removed

## Related

- Part of the broader poll-performance story tracked under #879 (umbrella closed-as-addressed but symptoms remain — see today's stall)
- Observed symptoms: draft PRs #1202/#1203/#1201 sat ready-to-advance for 30+ minutes on 2026-04-17 because poll iterations kept exceeding the 18-min staleness window

viamin:
## Summary

perf(poll-workflow): batch DetectLabels into a single activity

See #1235 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1235


viamin:
## Agent Update

**Review comment addressed:** The reviewer correctly identified that catching `GithubClient::RateLimitError` at the batch level caused a problem: already-processed issues had their `paid_state` mutated in the DB, but their results were discarded when the error propagated. On Temporal retry, those issues would be in a different state, producing potentially different decisions.

**Fix:** Moved the `RateLimitError` rescue from `execute` (batch-level) into `evaluate_issue` (per-issue level), alongside the existing `RecordNotFound` rescue. Rate-limited issues now return `nil` (skipped via `filter_map`) and will be picked up on the next poll cycle. Issues processed before the rate limit keep both their results and state mutations, preserving the partial-progress semantics of the old per-issue `DetectLabelsActivity` approach.

**Tests added:** Two new specs verify that when a rate limit hits mid-batch:
1. Results from successfully processed issues are returned
2. State mutations for processed issues are preserved while rate-limited issues remain unchanged

viamin:
## Agent Update

Merge conflicts resolved and committed. Here's a summary of what was done:

**Conflict**: `app/temporal/workflows/git_hub_poll_workflow.rb` had two conflict regions between the branch's batch-evaluate-issues feature and main's new `record_poll_heartbeat` feature (#1242).

**Resolution**:
1. **Execute loop**: Added the `record_poll_heartbeat(project_id)` call from main *before* the batch-evaluate patch guard, so heartbeats fire regardless of which evaluation path is taken.
2. **Private methods**: Kept both the `evaluate_issues_batch` method (from this branch) and the `record_poll_heartbeat` method (from main).

**Verification**: Lint (rubocop) passes with no offenses, all 110 relevant tests pass (80 workflow specs + 30 heartbeat/status specs), and the pre-commit hook ran the full suite successfully.

viamin:

PR #1242, opened 2026-04-17
Fix #1237: fix(poll-workflow): heartbeat last_polled_at mid-cycle, not only at the end

### LINKED ISSUE #1237: fix(poll-workflow): heartbeat last_polled_at mid-cycle, not only at the end
## Problem

`project.last_polled_at` is the sole liveness signal used by [PollWorkflowHealthCheckJob](app/jobs/poll_workflow_health_check_job.rb) to decide whether to restart a stale-RUNNING poll workflow. It is written **only** by `GetPollIntervalActivity`, which is the **last** activity of every cycle ([git_hub_poll_workflow.rb:43-44](app/temporal/workflows/git_hub_poll_workflow.rb#L43-L44)):

```
1. FetchIssuesActivity                    (60s timeout, can retry 3×)
2. DetectLabelsActivity per issue         (124 roundtrips today on viamin/paid)
3. CheckRateLimitActivity
4. ScanPaidPrsActivity                    (120s, loops 19 PRs)
5. ScanSecurityAlertsActivity
6. CheckKnowledgeStalenessActivity
7. GetPollIntervalActivity                ← only writer of last_polled_at
8. sleep 300s
```

If *anything* between steps 1-6 hangs or exhausts retries, `last_polled_at` does not advance — even though the workflow is legitimately doing work. The 18-minute staleness window (`3 × poll_interval + 3 min`, [project.rb:350](app/models/project.rb#L350)) then fires and the workflow is force-restarted, losing all in-flight progress and having to re-run the same heavy activities from scratch.

Observed today: project 2 (`viamin/paid`) restarted at 04:30 due to stale `last_polled_at`; 44 minutes later still not completing — restart can't help because the bottleneck is structural.

## Proposed fix

Update `last_polled_at` after each major step, not only at the end. Either:

- Replace the single `GetPollIntervalActivity` with a lightweight `RecordPollHeartbeatActivity` called after FetchIssues, after the per-issue loop, after scan, etc., OR
- Inline a `touch_last_polled_at` side-effect into each activity's successful return path (still single-writer, just more frequent).

The heartbeat signal should represent \"the workflow is making forward progress,\" not \"a full cycle completed.\" A separate `last_poll_completed_at` can be retained for cycle-completion telemetry.

## Acceptance criteria

- [ ] `last_polled_at` advances at least once per major poll step (fetch, detect, scan)
- [ ] A cycle that takes >18 min due to real work is NOT restarted by the health-check job
- [ ] A cycle that is genuinely wedged (no progress) is still restarted within the staleness window
- [ ] Health-check job documentation and UI labels updated to reflect \"last progress\" semantics

## Related

- Symptom prompting this work: today's 44+ min post-restart silence on viamin/paid
- Pairs well with #1235 (batch DetectLabels) and #1236 (isolate poll worker) — together they should eliminate the stale-RUNNING pattern rather than mask it

viamin:
## Summary

fix(poll-workflow): heartbeat last_polled_at mid-cycle, not only at the end

See #1237 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1237

### DIFF
--- a/app/jobs/poll_workflow_health_check_job.rb
+++ b/app/jobs/poll_workflow_health_check_job.rb
@@ -9,9 +9,12 @@
 # - Crashed workers or connection issues
 #
 # Staleness detection: even if a workflow reports RUNNING, it may be stuck
-# (e.g., nondeterminism errors during replay). Each successful poll cycle
-# updates `project.last_polled_at`; if that timestamp exceeds
-# `3 * poll_interval + 3 minutes`, the workflow is restarted.
+# (e.g., nondeterminism errors during replay). The poll workflow updates
+# `project.last_polled_at` after each major step (fetch issues, detect
+# labels, scans) via RecordPollHeartbeatActivity, so the timestamp
+# reflects last forward progress — not just cycle completion. If that
+# timestamp exceeds `3 * poll_interval + 3 minutes`, the workflow is
+# restarted.
 #
 # Scheduled via GoodJob cron every 5 minutes.
 class PollWorkflowHealthCheckJob < ApplicationJob
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -344,7 +344,7 @@ class Project < ApplicationRecord
 
   # Shared staleness window used by both the health-check job and the
   # automation health UI. A poll workflow is considered stale when it has not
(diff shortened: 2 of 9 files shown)


viamin:

PR #1298, opened 2026-04-19
fix(agent-runs): strip Claude CLI JSON envelope from agent summary

viamin:
## Summary

- Container execution captures raw Claude CLI `--output-format json` stdout verbatim into `agent_run_logs`. When `agent_summary` reads those logs for PR comments, the raw JSON envelope (with `usage`, `cost_usd`, `session_id`, `modelUsage`, etc.) was posted as-is instead of the assistant's result text.
- Adds `extract_text_from_stdout` to `AgentRun` that calls `AgentHarness::Providers::Anthropic.parse_cli_json_envelope` to extract just the `result` text from JSON envelopes. Falls back to raw text for non-JSON/non-envelope output. Error envelopes surface the classified error message.

This is the Paid-side fix for the issue agent-harness [PR #120](https://github.com/viamin/agent-harness/pull/120) addressed in the provider — that PR added the parsing method, but Paid never called it from the container execution path.

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -879,11 +879,13 @@ class AgentRun < ApplicationRecord
   end
 
   # Returns the agent's stdout output joined as a single string.
+  # Strips the raw JSON envelope when stdout is a Claude CLI --output-format json
+  # response, extracting just the assistant's result text.
   #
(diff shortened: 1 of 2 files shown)


viamin:

PR #1329, opened 2026-04-20
fix(perf): add GoodJob cleanup and reduce metrics collection interval

viamin:
## Summary
- Add `cleanup_preserved_jobs_before_seconds_ago = 1.day` to GoodJob config so finished jobs are automatically deleted after 24 hours (previously accumulated indefinitely — 320K rows / 743MB of bloat)
- Double metrics collection interval from 15s to 30s for both `ContainerMetricsCollectionJob` and `ServiceContainerMetricsCollectionJob`, halving job record creation rate (~15K → ~7.5K records/day)
- Fix Qdrant healthcheck in docker-compose: replace `wget` (not available in image) with `curl` to stop false "unhealthy" status

## Context
Web pages were noticeably slow. Investigation revealed the `good_jobs` (333MB) and `good_job_executions` (410MB) tables had grown to 640K+ rows with no cleanup configured. The metrics self-re-enqueue loop (`wait: 15s` per container) was the bulk driver. Running `GoodJob.cleanup_preserved_jobs(older_than: 1.day)` deleted 602K stale records and VACUUM recovered 360MB (1784MB → 1424MB).

The routes knowledge collector container size increase (512MB → 4GB) was **not** the cause — those containers are ephemeral.

### DIFF
--- a/app/jobs/container_metrics_collection_job.rb
+++ b/app/jobs/container_metrics_collection_job.rb
@@ -3,7 +3,7 @@
 # Periodically collects CPU and memory metrics from running agent containers.
 #
 # Enqueued for each running agent run and re-enqueues itself until the run
-# finishes. Collection interval defaults to 15 seconds to balance granularity
+# finishes. Collection interval defaults to 30 seconds to balance granularity
 # with overhead. On consecutive failures the interval backs off (doubled per
 # failure, capped at 5 minutes) so metrics resume automatically when Docker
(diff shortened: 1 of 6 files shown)


viamin:
## Agent Update

The code change is in place. I'm installing dependencies and preparing the database now so the repo is in a verified state before lint, tests, and commit.

viamin:

PR #1331, opened 2026-04-20
fix(agent-runs): handle unsupported auth refresh flows

viamin:
## Summary

- Prevent auth-expired agent-run pages from crashing when a provider cannot generate an OAuth URL, such as Codex API-key auth.
- Show a clear auth-expired message for providers without browser login URL support.
- Align refresh-auth submissions with the agent-harness `token:` API while preserving legacy `auth_code` param compatibility.
- Handle unsupported provider refresh flows with a normal redirect alert instead of a server error.

## Testing

- `COVERAGE=false bin/rspec spec/requests/agent_runs_spec.rb:1878`
- `COVERAGE=false bin/rspec spec/requests/agent_runs_spec.rb`
- `bin/lint --changed`
- `git diff --check`

## Upstream follow-up

- viamin/agent-harness#143
- viamin/agent-harness#144

### DIFF
--- a/app/controllers/projects/agent_runs_controller.rb
+++ b/app/controllers/projects/agent_runs_controller.rb
@@ -407,10 +407,10 @@ module Projects
         return
       end
 
-      code = params[:auth_code]&.strip
-      if code.blank?
+      token = (params[:auth_token].presence || params[:auth_code])&.strip
+      if token.blank?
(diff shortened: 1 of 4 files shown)


viamin:
## Agent Update

I’ve applied the view change. Next I’m installing/preparing the app as requested; after that I’ll run the focused checks and a full diff review before committing.

PR #1349, opened 2026-04-21
fix(secrets): remove direct provider credential injection from agent runs

### LINKED ISSUE #1281: fix(secrets): remove direct provider credential injection from agent runs
## Gap

The README Key Features section previously implied that API keys never enter agent containers. Git credentials and default platform LLM keys are proxied, but stored provider API-key auth and subscription auth can still place provider credentials or auth state inside the agent runtime.

Related knowledge-side cleanup is tracked by #1043 and #1044, but this issue covers the agent-run provider credential path.

## Why this matters

This is a security-contract gap. A user reading only the feature list could reasonably expect provider secrets to be kept outside agent containers in every auth mode.

Priority: **P1** because the gap affects credential exposure expectations.

## Acceptance criteria

- Inventory each provider auth mode and identify where provider API keys or auth state enter the runtime.
- Remove direct provider credential injection from agent runs where feasible, preferably through proxy/session mechanisms.
- If a provider mode intentionally requires runtime credentials, make that exception explicit in product/docs.
- Add regression coverage for the chosen contract.
- Update the README Key Features caveat once the behavior and docs are aligned.

viamin:
## Summary

fix(secrets): remove direct provider credential injection from agent runs

See #1281 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1281

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,8 @@ Paid stores every decision point as data—prompts, model preferences, workflow 
   - Gap: subscription-auth and direct-outbound provider runs use the internal network for provider access; tracked by [#1284](https://github.com/viamin/paid/issues/1284).
 - **Multiple Agents and Providers**: Support for Claude Code, Codex, Cursor, Gemini, Aider, OpenCode, Kilocode, and Copilot when the runtime is both supported by `agent-harness` and installed in `paid-agent`
   - Gap: broader runtime parity is tracked by provider metadata and smoke-test refactors ([#798](https://github.com/viamin/paid/issues/798), [#796](https://github.com/viamin/paid/issues/796)) plus agent-image install delegation work ([#789](https://github.com/viamin/paid/issues/789)-[#795](https://github.com/viamin/paid/issues/795)).
-- **Secrets Proxy**: Git credentials and default platform LLM keys are proxied through authenticated endpoints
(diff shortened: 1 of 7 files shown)


viamin:
## Agent Update

The review matches the code: the proxy currently requires `for_agent_runs`, while the runner can deliberately execute `rate_limit_fallback.for_fallback` entries with `enabled_for_agent_runs: false`. I'm going to check the nearby specs and model scopes, then patch the proxy predicate and add a request spec for the fallback-only case.

viamin:
## Agent Update

Addressed the review thread and the branch is clean.

Changed:
- [app/temporal/activities/run_agent_activity.rb](/workspace/app/temporal/activities/run_agent_activity.rb:1327): Anthropic-backed API-key runs now set `ANTHROPIC_BASE_URL`, run ID, proxy token, and paid provider ID so stored Claude/Cursor/Aider keys route through the secrets proxy in mixed subscription-auth containers.
- [spec/temporal/activities/run_agent_activity_spec.rb](/workspace/spec/temporal/activities/run_agent_activity_spec.rb:341): Added coverage for the direct stored Anthropic provider path and same-provider fallback path.

Verification:
- `bundle install` passed after using a writable Bundler path under `/home/agent`.
- `yarn install` passed after moving Yarn cache/temp to `/workspace/tmp` because `/home/agent/.cache` is a small tmpfs.
- `bin/yarn-postinstall` passed.
- `bundle exec rails db:prepare` passed with `CABLE_DATABASE_URL="$DATABASE_URL"`.
- `bundle exec rspec` passed: `8215 examples, 0 failures, 2 pending` for Qdrant live specs.

Commit:
- `6293b35d fix(secrets): route anthropic fallback keys through proxy`

No push performed.

PR #1395, opened 2026-04-22
fix(agent-runs): prevent circuit-breaker poisoning from StaleRunDetectorJob timeouts

viamin:
## Summary

- `AgentRun#cancelled_by_cleanup?` now recognizes StaleRunDetectorJob's `"Stale run detected"` error prefix in addition to `dev:cleanup`'s `"Marked stale on startup"` prefix
- Prevents provider circuit-breaker counters from being incremented when the stale detector externally kills an in-flight Temporal activity
- Extracts integration tests into `shared_examples` covering both external cancellation paths

## Problem

When `StaleRunDetectorJob` times out a stuck run that still has an in-flight Temporal activity, the activity doesn't recognize the timeout as externally-induced. It treats the container kill as a real provider failure and:

1. Increments circuit-breaker failure counts for the provider
2. Continues iterating remaining providers (each also fails)
3. All providers get circuit-breaker hits, which can trip breakers and make subsequent runs see all providers as unavailable

## Testing

- Model specs: added test for `STALE_DETECTOR_ERROR_PREFIX` in `cancelled_by_cleanup?`
- Integration specs: refactored existing dev:cleanup tests into `shared_examples "externally cancelled run"` and added parallel StaleRunDetectorJob context (10 assertions total)

Refs #1394

Issue #1135, opened 2026-04-14
fix(scanner): pause PR follow-up runs while paid_agent review is still outstanding


viamin:
## Summary

Paid can continue launching additional automatic `create_pr` follow-up runs for an existing PR while a `paid_agent` review cycle for the current head is still outstanding.

This showed up clearly on PR #1087 in `viamin/paid`, where multiple `create_pr` runs completed in long series between review rounds, including several empty "reverify" commits.

## Problem

The current PR automation flow does not consistently pause for review completion once a PR is waiting on `paid_agent` review.

In practice, the system can keep starting new `create_pr` runs when `paid_agent_review_pending` coexists with other triggers such as:

- merge conflicts
n- CI failures/check noise
- conversation-comment follow-up

That behavior is especially problematic in draft/restarted review cycles, where the PR is supposed to be gated on a clean review before progressing.

### PR #1087

The local `Issue`/`AgentRun` data for PR #1087 shows:

- `pr_review_phase: "restarted"`
- only review method enabled: `paid_agent`
- multiple automatic `create_pr` runs between review rounds

Observed run sequence:

- `7594` `create_pr` -> commit `1155ef1d`
- `7611` `review`
- `7632` `create_pr` -> `dae50a18`
- `7644` `create_pr` -> `71b60688`
- `7657` `create_pr` -> `62fef6d2`
- `7666` `create_pr` -> `d8ea1ced`
- `7669` `create_pr` -> `99960526`
- `7672` `create_pr` -> `2aaf94a0`
- `7674` `create_pr` -> `28b895c3`
- `7676` `create_pr` -> `893a8e30`
- `7677` `create_pr` -> `9390855a`
- `7684` `review`
- `7692` `create_pr` -> `061f02c4`
- `7695` `create_pr` -> `76c33a08`
- `7710` `create_pr` -> `186fbfa7`
- `7719` `create_pr` -> `8170ddc6`
- `7729` `create_pr` -> `b962bbcb`
- `7732` `create_pr` -> `c194880f`
- `7733` `create_pr` -> `4b953815`

Only 3 GitHub reviews were present during that window, but 17 `Agent Update` comments were posted.

### Code path

`GitHubPollWorkflow#handle_paid_agent_review_pending` currently queues the review run and then immediately starts another follow-up workflow whenever any other trigger is present.

That means `paid_agent_review_pending` is not functioning as a hard gate for the current head when other actionable triggers coexist.

## Why this matters

[...]

PR #1404, opened 2026-04-23
fix(scanner): queue paid_agent review run when findings are unaddressed

viamin:
## Summary

- Fixes an infinite loop where paid_agent review findings that weren't addressed by subsequent code changes caused the scanner to repeatedly start `create_pr` follow-ups without ever queuing a fresh review run
- In the "still unaddressed" body-only branch, calls `check_paid_agent_review_status` for paid_agent bots so `paid_agent_review_pending` is emitted alongside `review_bot_review_pending` + `review_bot_comments`
- The workflow routes the combined triggers to `handle_paid_agent_review_pending`, which queues a review run and applies the hard gate (#1135), breaking the loop

## Problem

When a paid_agent review had findings (inline comments on specific files) and a subsequent `create_pr` follow-up didn't touch those files, the scanner's `review_diff_touches_reviewed_files?` check returned `false`. This pushed the scanner into the "still unaddressed" branch which emitted `review_bot_review_pending` + `review_bot_comments` — but **not** `paid_agent_review_pending`. The workflow then started another `create_pr` follow-up via `handle_review_bot_review_pending`, creating an infinite loop.

On PR #1395, this resulted in 13 consecutive `create_pr` runs with only 1 review run, eventually escalating due to consecutive follow-up failures.

## Testing

- Updated existing test to expect `paid_agent_review_pending` in the no-inline-comments case
- New test: inline comments present but diff doesn't touch those files (the PR #1395 scenario)
- New test: rounds exhausted → `paid_agent_review_pending` is NOT emitted
- All 307 scan activity tests pass, linter clean

Refs #1395

### DIFF
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -1206,7 +1206,18 @@ module Activities
             # touch any file mentioned in the review's inline comments.
             # Treat as still-unaddressed to prevent unrelated changes
             # (e.g. a CI schema fix) from clearing review findings.
-            paid_agent_limit_reached_for_latest_review ? body_only_exhausted_triggers : body_only_pending_triggers
+            if paid_agent_limit_reached_for_latest_review
+              body_only_exhausted_triggers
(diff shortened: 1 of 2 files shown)


viamin:

PR #1442, opened 2026-04-24
fix(test): truncate seed data before migration spec runs

viamin:
## Summary

- Fixes CI failure in `AddAccountToServiceContainers` migration spec where seed service containers ("postgres", "redis") from `db/seeds/service_containers.rb` inflated the count from 1 to 3
- Root cause: `before` block only adjusted schema (`migration.down`) but never removed seed data — the seed SCs have different names from the test SC and survive the per-name collapse in `migration.down`
- Extracts shared `MigrationSpecHelpers#truncate_migration_test_data` and adds it to both `before` and `after` blocks of both non-transactional migration specs

## Test plan

- [x] All 17 migration specs pass locally with and without seed data present
- [ ] CI passes on this PR


viamin:

PR #1444, opened 2026-04-24
fix(devcontainer): reduce worker defaults to avoid overmind OOMs

viamin:
## Summary
- add conservative devcontainer defaults for Temporal worker concurrency
- reduce GoodJob thread/queue defaults in the devcontainer environment
- document that the devcontainer applies these safer `bin/dev` defaults

## Why
The devcontainer was OOM-killing the Temporal worker under normal `bin/dev` load, which caused Overmind to tear down the whole stack. These defaults trade some background throughput for a much more stable local development experience.

## Testing
- investigated the live OOM shutdown path via `log/dev-update/overmind.log` and supervisor diagnostics
- restarted `bin/dev` with the reduced settings and confirmed the worker stayed healthy instead of exiting `137`
- verified the chosen `DB_POOL` satisfies the Temporal worker's own minimum pool calculation for the reduced slot counts

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -23,11 +23,22 @@
     "DB_HOST": "postgres",
     "DB_USERNAME": "paid",
     "DB_PASSWORD": "paid",
+    // Keep devcontainer worker concurrency conservative to avoid OOM-killing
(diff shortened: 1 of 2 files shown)


viamin:

PR #1459, opened 2026-04-25
feat(scheduler): cross-user fair queueing

### LINKED ISSUE #1275: feat(scheduler): cross-user fair queueing
## Summary

Add per-user stride to queue ordering so users with equal-tier runs share worker capacity instead of one user's FIFO batch blocking another's.

## Background

With within-user fair queueing (#1274) in place, projects within a user's account share capacity fairly. However, across users, the queue is still FIFO within a priority tier. Two users each with P1 runs will see one user's entire FIFO batch served before the other's. This is especially problematic on shared infrastructure where one active user can starve all others.

Depends on #1274 (shares the CTE infrastructure introduced there).

## Scope

- **Add `user_active_counts` CTE** to `peek_next_queued_run` — counts `pending` + `running` runs per user
  - Resolve orphaned projects (where `projects.created_by_id IS NULL`) by attributing them to `account.fallback_owner` using `COALESCE(created_by_id, (SELECT id FROM users WHERE account_id = projects.account_id ORDER BY id LIMIT 1))`
- **Add `user_active_count ASC` to `QUEUE_ORDER`** after the per-project stride (from #1274), before `GOAL_PRIORITY_SQL`
- **Always on** — no toggle. There is no scenario where one user should starve another at the same priority tier.
- **Exclude `paused` runs** from the count, same as the per-project stride — a paused run should not count toward a user's active utilization for ordering purposes.

## Tests

- Two users with same-tier runs get interleaved (user with fewer active runs dequeues first)
- Single user on system gets full capacity (no unfair throttling)
- Orphaned project counts attributed to account fallback owner correctly
- Combined with per-project stride: two users with multiple projects each get fair interleaving at both levels
- Capacity cap (`max_concurrent_runs`) still enforced as hard ceiling per user

## Out of scope

- Dashboard queue preview (separate issue)

viamin:
## Summary

feat(scheduler): cross-user fair queueing

See #1275 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1275

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -556,22 +556,28 @@ class AgentRun < ApplicationRecord
     END
   SQL
   PROJECT_ACTIVE_COUNT_SQL = Arel.sql("#{PROJECT_ACTIVE_COUNT_CASE_SQL} ASC").freeze
-  QUEUE_ORDER = [ QUEUE_PRIORITY_SQL, GOAL_PRIORITY_SQL, { created_at: :asc, id: :asc } ].freeze
+  USER_ACTIVE_COUNT_SQL = Arel.sql("COALESCE(user_active_counts.user_active_count, 0) ASC").freeze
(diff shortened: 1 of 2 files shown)


viamin:

PR #1462, opened 2026-04-25
fix(service-containers): align service and agent Docker networks for direct-outbound runs

### LINKED ISSUE #1282: fix(service-containers): align service and agent Docker networks for direct-outbound runs
## Gap

Service containers are provisioned on the expected agent network, but direct-outbound provider runs can select a different Docker network than service provisioning expects. That can make approved supporting services unreachable from the agent container in auth/network modes where users still expect service containers to work.

Shared-database isolation fallout is tracked separately by #1280; this issue is specifically about network alignment between agent and service containers.

## Why this matters

This is a runtime reliability gap in an advertised feature. A user reading only the Key Features list could reasonably expect approved services like Postgres, Redis, or Selenium to be reachable from any project run that enables them.

Priority: **P2** because it can break the advertised service-container workflow, but it is narrower than the security-contract gaps.

## Acceptance criteria

- Make agent-container network selection and service-container network provisioning use the same network decision.
- Cover proxy-mode, subscription-auth, and direct-outbound provider runs.
- Add regression coverage for service reachability or network assignment in the affected modes.
- Update the README Key Features caveat once the behavior is aligned.

viamin:
## Summary

fix(service-containers): align service and agent Docker networks for direct-outbound runs

See #1282 for context. Paid was unable to auto-generate a description for this PR;

---

Closes #1282

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Paid stores every decision point as data—prompts, model preferences, workflow 
 - **Live Dashboards**: Track active runs, performance, quality, cost, and knowledge-collection health from the UI
 - **Provider and Integration Management**: Test provider auth from the UI and manage GitHub, Linear, and provider API keys
   - Gap: generic integration credentials exist in the data model but are hidden from the main integrations hub pending RBAC/admin-role work; tracked by [#1283](https://github.com/viamin/paid/issues/1283).
-- **Service Containers**: Attach approved supporting services like Postgres, Redis, or Selenium to project runs when agents need dependencies beyond the app code. Service containers are attached to the same Docker network selected for the agent run. Shared-database isolation fallout is tracked separately by [#1280](https://github.com/viamin/paid/issues/1280).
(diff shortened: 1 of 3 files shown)


viamin:
## Agent Update

**Problem**: When a saved provider (e.g., `copilot`) was present but not container-executable, `primary_provider_identifier` in `Containers::Provision` jumped straight to the goal default provider (e.g., a direct-outbound `opencode` provider). However, `RunAgentActivity#build_provider_order` tries configured fallbacks first and only uses the goal default when no fallbacks are runnable. This mismatch caused service containers to be placed on the infrastructure network (for the direct-outbound default) while the agent actually ran on the restricted network via a proxy-mode fallback.

**Fix** (`app/services/containers/provision.rb`):
- Added a new branch in `primary_provider_identifier`: when the saved provider exists but is unrunnable and fallback is enabled, check configured fallback providers for the first runnable one before falling through to the goal default
- Added `first_runnable_fallback_for_saved_provider` helper that queries `fallback_priority_for` and finds the first container-executable fallback

**Test** (`spec/services/containers/provision_spec.rb`):
- Added regression spec for the exact scenario: unrunnable saved provider (`copilot`) + runnable proxy-mode fallback (`claude` API key) + direct-outbound default (`opencode`). Verifies the container uses the restricted network (`paid_agent`) rather than the infrastructure network.

viamin:

PR #1370, opened 2026-04-21
feat(multi-tenancy): implement data isolation patterns

viamin:
## Summary

feat(multi-tenancy): implement data isolation patterns

See #730 for context. Paid was unable to auto-generate a description for this PR;
the diff is the authoritative source of truth.

---

Closes #730

PR #1473, opened 2026-04-26
fix(auth): restore login by deferring Warden auth until after CSRF check

viamin:
## Summary

Login was returning 422 (`InvalidAuthenticityToken`) after the multi-tenancy work in #1370. The around_action `with_current_attributes` was eagerly calling `current_user` before `verify_authenticity_token` ran, which triggered Warden's `:database_authenticatable` strategy on the sign-in POST itself. That auth fired Devise's `clean_up_csrf_token_on_authentication` hook, which **deletes** `_csrf_token` from the session — before CSRF verification got to read it. The result: every successful credential pair was followed by a CSRF failure and an empty response.

The fix moves the eager `current_user` resolution into the `else` branch so it only runs for non-Devise controllers. Devise controllers now go straight into `TenantContext.with_system_access { yield }`, letting `verify_authenticity_token` run before any Warden strategy fires. Devise then resolves the user inside its own controller as normal.

## Why this hadn't been caught

- `spec/environments/test.rb` sets `allow_forgery_protection = false`. Every existing request spec — including `spec/requests/users/sessions_spec.rb` — runs without CSRF, so login appeared green while real users were locked out.
- This PR adds the missing layer: real-browser system specs with CSRF enforced.

## Changes

**Bug fix**
- `app/controllers/application_controller.rb` — move `Current.user = rls_safe_current_user` into the non-Devise branch so it runs after `verify_authenticity_token`.

**Latent bug surfaced by the new tests**
- `app/javascript/controllers/theme_controller.js` — Stimulus fires `*Changed` callbacks before `connect()`, so `this.mediaQuery` was undefined the first time `applyTheme` ran. Moved init to `initialize()`.

**System specs**
- `spec/system/authentication_spec.rb` — sign-in success/failure, sign-out (3).
- `spec/system/sign_up_spec.rb` — account+user creation through the form, password-mismatch rollback (2).
- `spec/system/cross_tenant_isolation_spec.rb` — own-account project visible, other-account project 404s, projects index scoped to current tenant (2).
- `spec/support/capybara.rb` — register cuprite under a non-default name (`:paid_cuprite`) because Rails' system-test infrastructure re-registers `:cuprite` and silently overwrites custom configs; flip `allow_forgery_protection` back on for `type: :system`; honor `HEADLESS=false` for local debugging.

**CI**
- `.github/workflows/ci.yml` — exclude `spec/system/**/*_spec.rb` from the main test run; system specs are owned by the new workflow.
- `.github/workflows/system_tests.yml` (new) — postgres service, locates pre-installed `google-chrome` (avoids the `chromium-browser` snap trap on ubuntu-latest), builds JS+CSS assets, runs `bin/rspec spec/system`. Includes concurrency control (cancel superseded PR runs), 20-minute job timeout, and on-failure upload of Capybara screenshots + `log/test.log` for debugging.

## Verification

- All 12 specs pass: 3 authentication + 2 sign-up + 2 cross-tenant + 5 existing request specs.
- Reverting the controller fix (with everything else intact) makes 2/3 system specs fail with the original CSRF symptom — the wrong-password case stays green because no Warden auth fires, no CSRF cleanup. Confirms the new tests would have caught this regression on day one.

## One follow-up the maintainer should do before merging

`actions/upload-artifact@v4` in the new workflow is referenced by tag, while the rest of the codebase pins all third-party actions to commit SHAs. I deliberately didn't fabricate a SHA. Pin to the SHA from `gh api repos/actions/upload-artifact/releases/tags/v4` before merge.

## Test plan

- [ ] CI green on this branch (lint + unit + new system_tests workflow).
- [ ] Manual sanity check on a real browser: sign in with a seeded dev user (`test@example.com` / `password`) and confirm you land on the dashboard.
- [ ] Confirm sign-up flow still works and lands the user on onboarding.
- [ ] Pin `actions/upload-artifact@v4` to a SHA.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -151,7 +151,9 @@ jobs:
           curl --fail --silent --show-error --retry 10 --retry-delay 2 --retry-connrefused "$QDRANT_URL/healthz"
 
       - name: Run tests
-        run: bin/rspec --exclude-pattern "spec/performance/**/*_spec.rb"
+        # System specs run in a separate workflow (.github/workflows/system_tests.yml)
+        # because they need Chromium and built JS/CSS assets.
+        run: bin/rspec --exclude-pattern "spec/performance/**/*_spec.rb,spec/system/**/*_spec.rb"
 
       - name: Run query performance benchmarks
         run: COVERAGE=false bin/rspec spec/performance
--- a/.github/workflows/system_tests.yml
+++ b/.github/workflows/system_tests.yml
@@ -1,0 +1,104 @@
+name: System tests
+
+on:
+  pull_request:
+  push:
+    branches: [ main ]
+
+permissions:
+  contents: read
+
+# Cancel in-progress system test runs on the same PR when a new commit is
+# pushed. Don't cancel runs on `main` (those are gating the changelog).
+concurrency:
+  group: system-tests-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+jobs:
+  system:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    services:
+      postgres:
+        image: postgres:16.13
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd="pg_isready -U postgres"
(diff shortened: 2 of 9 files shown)


viamin:
## Agent Update

The background task completed successfully — it was the earlier RSpec run that confirmed 8780 examples, 0 failures. The work is already done and committed.

viamin:

PR #1486, opened 2026-04-26
Fix #1327: Create-issue runs should deterministically apply selected priority labels

viamin:
## Summary

Create-issue runs should deterministically apply selected priority labels

See #1327 for context. Paid was unable to auto-generate a description for this PR;
the diff is the authoritative source of truth.

---

Closes #1327

PR #1496, opened 2026-04-27
fix(pr-descriptions): use agent summary as fallback when LLM generation fails

### LINKED ISSUE #1327: Create-issue runs should deterministically apply selected priority labels
## Summary

A create-issue agent run can successfully create the requested GitHub issue while dropping the priority label selected in the UI priority dropdown. The observed run was `http://localhost:3000/projects/2/agent_runs/9163`; it created issue #1325, but #1325 has an empty `labels` array even though a priority was specified.

## Evidence

- Agent run: `/projects/2/agent_runs/9163`
- Created issue: #1325, `Move active runs above queue health in live dashboard view`
- GitHub API response for #1325 shows `labels: []`
- Recent issues confirm priority labels exist and are usable, e.g. `P1` and `P2` are present on other issues/PRs.

## Problem

Priority selection is structured UI state, but create-issue execution currently appears to rely on the agent/prompt to remember to include the corresponding GitHub label. That makes priority labeling nondeterministic: the run can be considered successful because the issue exists, while silently losing the selected priority metadata.

## Requested Fix

Prefer a deterministic service-level fix over prompt-only behavior:

1. Persist or pass the selected priority from the create-issue request into the issue creation step as structured data.
2. Map selected priority values directly to GitHub labels such as `P1`, `P2`, etc.
3. Apply the label either in the initial create-issue API payload or immediately afterward via the labels endpoint.
4. Treat label application failure as visible run output or a retryable/failed step, not as a silent success.
5. Keep prompt wording as a fallback reminder, but do not make the LLM solely responsible for preserving selected priority.

## Acceptance Criteria

- A create-issue run with priority `P1` creates an issue labeled `P1`.
- A create-issue run with priority `P2` creates an issue labeled `P2`.
- Runs without a selected priority preserve current behavior and do not add a priority label.
- If the GitHub label add/create call fails, the run exposes that failure rather than reporting a clean success.
- Regression coverage verifies that selected priority dropdown state reaches the GitHub issue labels payload or deterministic label-adding service.

## Notes

Prompt improvement may reduce recurrence, but this should be handled deterministically because priority is first-class UI state and should not depend on model compliance.

viamin:
## Summary

PR description generation silently fails on many PRs, producing a generic "unable to auto-generate" message that just links to the issue. This makes PRs uninformative and is a poor reviewer experience.

## Root Cause

`Llm::GeneratePrDescription` calls `AgentHarness.send_message(provider: :claude)` which requires either `ANTHROPIC_API_KEY` (for HTTP text mode) or the `claude` CLI in the Rails/worker container. When neither is available, the LLM call returns an unsuccessful response (`response.success?` is false — no exception), and there was **zero logging** for this case. The `rescue StandardError` block only catches exceptions, not silent failures.

The fallback body then discarded the entire agent summary and generated "Paid was unable to auto-generate a description for this PR; the diff is the authoritative source of truth." — even though the agent's stdout typically contains useful information about what was done.

## Changes

- **`fallback_body` now uses the agent summary** when available and suitable, instead of the generic "unable to auto-generate" message. Produces meaningful PR bodies even when the LLM is unreachable.
- **`suitable_fallback_summary?` guard** prevents dumping raw JSON (unparsed CLI envelopes) or error messages (`"Agent encountered an error: ..."`) into PR bodies. Falls back to issue title in those cases.
- **`generate_description` now logs `pr_description_llm_unsuccessful`** when the LLM returns nil (not just on exceptions), including summary length for diagnostics. Previously this silent-failure path had zero observability.
- Added tests for the JSON/error guard, the new log event, and updated existing fallback tests.

## Screenshots (before/after)

**Before** (PR #1486 and many others):
> ## Summary  
> Create-issue runs should deterministically apply selected priority labels  
> See #1327 for context. Paid was unable to auto-generate a description for this PR; the diff is the authoritative source of truth.

**After** (with agent summary available):
> ## Summary  
> All done. Here's a summary:  
> - Fixed the priority label assignment in create-issue runs  
> - Added deterministic label selection based on project configuration  
> - Added tests for edge cases

## Design Decisions

- Chose to guard against raw JSON/error messages rather than try to parse or clean them — keeps the fallback path simple and predictable.
- Kept the 10K char truncation limit as a safety net for verbose agent output.
- Used `have_received` + permissive `warn: nil` stubs for logger assertions to avoid test fragility from additional warn calls in `validate_summary_scope`.

---

Closes viamin/paid#1327


viamin:

PR #1489, opened 2026-04-27
feat(prompts): add data-driven CI failure debugging guidance

viamin:
## Summary

- Extend `Ci::FailureContext` with failure type classification (database, environment, dependency, service, timeout) and CI workflow YAML file fetching via GitHub Actions API
- Add seeded, A/B-testable `ci.failure_guidance` prompt template that provides targeted debugging strategy for each failure type and includes the actual CI workflow configuration in the agent prompt
- Fix `GithubClient#contents` to not pass `ref: nil` as an empty query param to GitHub's API

## Problem

Paid agents could not fix CI infrastructure errors (e.g., PR [#627 on railscrawler](https://github.com/HuntHelper/railscrawler/pull/627) — `NoDatabaseError` because `db:setup` ran without `RAILS_ENV=test`). The agent kept modifying `config/database.yml` instead of the workflow file because:

1. The CI workflow file content was never included in the prompt
2. The prompt said "reproduce locally" even for infrastructure errors that can't be reproduced locally
3. No failure type classification existed to guide the agent toward the correct fix

## How this moves toward the vision

The debugging strategy is now a **seeded prompt** (`ci.failure_guidance`) that:

- Gets a new version automatically when the template changes
- Can be **A/B tested** against the fallback to measure impact on agent outcomes
- Can be **evolved by the meta-agent** via `PromptEvolution::Mutate`
- Can be **overridden per-project** using 3-tier inheritance (project > account > global)
- Degrades gracefully if the prompt is deactivated

## Files changed

| File | Change |
|------|--------|
| `app/services/github_client.rb` | Add `file_content`, `actions_run` methods; fix `contents` to not pass `ref: nil` |
| `app/services/ci/failure_context.rb` | Add failure type classification, CI workflow file fetching |
| `app/services/prompts/build_for_pr.rb` | Render `ci.failure_guidance` prompt into CI failures section |
| `db/seeds/prompts.rb` | Add `ci.failure_guidance` seeded prompt template |
| Tests | 15 new test cases covering classification, workflow fetching, guidance rendering |

PR #1497, opened 2026-04-27
fix(agent-runs): mount /tmp tmpfs with exec so bundle install can build native gems

viamin:
## Summary

Adds the `exec` flag to the agent container's `/tmp` tmpfs so `bundle install` can build native gem extensions (e.g. `bigdecimal`) under `BUNDLE_PATH=/tmp/bundle`. Mirrors the fix already shipped for the knowledge collector container in #1155.

## Why this matters

Every coding/review/rebase prompt instructs the agent to run `bundle install` as step 1, and the review-goal prompt explicitly sets `BUNDLE_PATH=/tmp/bundle`. With Docker's default `noexec` flag on `/tmp`, mkmf's `try_link` writes a conftest binary to `/tmp` and validates it with `File.executable?`, which returns `false` on a noexec mount. Bundler then aborts on the first native gem with the misleading error *"compiler failed to generate an executable file"* — even though `gcc`, `make`, ruby headers, and all build deps are present in the image.

This has been surfacing in `paid-code-reviewer[bot]` reviews as *"I couldn't run the spec suite locally because bundle install failed while building bigdecimal due to missing compiler tooling in the environment"* — see e.g. https://github.com/viamin/paid/pull/1489#pullrequestreview-4178294058.

PR #1155 reproduced and verified the same fix in `paid-agent:latest`:
- Without `exec`: `gem install bigdecimal` fails with the misleading mkmf error.
- With `exec`: the install succeeds.

## Scope

- One-character behaviour change: prepend `exec,` to `/tmp` tmpfs options in `Containers::Provision#host_config`.
- Updated explanatory comment matches the rationale shipped for the collector.
- `/home/agent/.cache` deliberately stays `noexec` (cache data only, no exec needed).
- Service containers (postgres, etc.) and pool-managed agent containers are unaffected — they either don't define their own `/tmp` tmpfs or share this same code path.

## Security note

The agent already executes arbitrary code from `/workspace` (rw bind mount, exec by default) — that's how it runs `bin/rspec`, `make`, etc. Adding `exec` to `/tmp` does not grant any capability the agent didn't already have. The container retains `ReadonlyRootfs`, `CapDrop: ALL`, `no-new-privileges:true`, and runs as a non-root user.

## Out of scope (separate follow-up)

In proxy/restricted network mode the iptables firewall does not allow `rubygems.org`, so `bundle install` will still fail in that mode — but with a clear network error instead of the misleading mkmf error. Worth a separate ticket to either add a rubygems forwarder to the proxy or loosen the firewall.

## Test plan

- [x] `bin/rspec spec/services/containers/provision_spec.rb` — all 158 examples pass
- [x] `bin/rubocop` clean on changed files
- [x] Updated existing default-sizes assertion to expect the new value
- [x] Added focused regression spec asserting `exec` is present on `/tmp` (mirroring the collector regression test)
- [ ] Observe the next paid-agent PR review on a Ruby PR — confirm bundle install + spec-suite validation runs cleanly

### DIFF
--- a/app/services/containers/provision.rb
+++ b/app/services/containers/provision.rb
@@ -1281,13 +1281,14 @@ module Containers
         binds << "#{copilot_config_host_path}:/home/agent/.config/github-copilot-host:ro"
       end
 
-      # /tmp must be `exec` because review-goal runs install gems with
-      # BUNDLE_PATH=/tmp/bundle and bundle install builds native gem
-      # extensions there. mkmf's try_link verifies the produced binary with
+      # /tmp must be `exec` because every coding/review/rebase prompt has the
+      # agent run `bundle install` as step 1, and review-goal runs additionally
+      # set BUNDLE_PATH=/tmp/bundle. Bundler builds native gem extensions in
+      # the gem path; mkmf's try_link verifies the produced binary with
       # File.executable?, which returns false on a noexec mount — producing
       # a misleading "compiler failed to generate an executable file" error
-      # (e.g. bigdecimal extconf). Docker's default tmpfs flags include
-      # noexec, so it must be overridden here.
+      # (e.g. bigdecimal extconf) even though the toolchain is fully present.
+      # Docker's default tmpfs flags include noexec, so it must be overridden.
       # /home/agent/.cache only stores cache data and stays noexec.
       tmpfs = {
         "/tmp" => "exec,size=#{options[:tmpfs_tmp_size]},mode=1777",
--- a/app/services/containers/provision.rb
+++ b/app/services/containers/provision.rb
@@ -1281,8 +1281,16 @@ module Containers
         binds << "#{copilot_config_host_path}:/home/agent/.config/github-copilot-host:ro"
       end
 
+      # /tmp must be `exec` because review-goal runs install gems with
+      # BUNDLE_PATH=/tmp/bundle and bundle install builds native gem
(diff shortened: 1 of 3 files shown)


viamin:
## Agent Update

Commit succeeded. The fix:

- **Root cause**: Two specs in `record_spec.rb` used `Prompt.global.active.find_by!` expecting a seeded `generation.pr_description` prompt to exist in the test DB. Since seeds aren't loaded in CI test runs, both tests raised `ActiveRecord::RecordNotFound`.
- **Fix**: Replaced `find_by!` lookups with `create(:prompt, :global, :with_version, slug: "generation.pr_description")` so the tests create their own test data via factories, making them self-contained.
- Full test suite: 8956 examples, 0 failures.

viamin:
**Escalation Note**

This has been escalated because Consecutive draft follow-up failures (3 runs with no output).

**How to resolve:**

viamin:
## Agent Update

Done. Here's what I did:

1. **Merged `origin/main`** into the branch — one conflict in `spec/services/llm_output_metrics/record_spec.rb`
2. **Resolved the conflict** by keeping the `Prompt.where(...).destroy_all` cleanup lines from main (needed to prevent test collisions with seeded prompt data), which our branch had removed
3. **Verified all 9027 tests pass** with 0 failures
4. **RuboCop clean** on all changed files
5. **Committed** the merge resolution

viamin:

PR #1498, opened 2026-04-27
fix(knowledge): prevent TypeError from nil artifact_count in schema section

viamin:
## Summary

- `build_schema_section` returned a bare hash missing `artifact_type` and `artifact_count` keys, causing `TypeError: nil can't be coerced into Integer` when `build_sections` accumulated per-type counts via `+=`
- Switched to the `artifact_section` helper (consistent with all other section builders) and added `schema` to the `section_artifact_type` mapping
- Added test coverage for schema artifacts including `artifact_type_counts` verification

## How verified

- Confirmed the exact error against production AgentRun #10933 via `rails runner`
- Verified the fix resolves the error by re-running the same code path
- Added dedicated spec for schema section rendering and artifact count tracking
- Updated "full knowledge base" spec to include schema artifacts in section order assertion
- RuboCop passes on both changed files

### DIFF
--- a/app/services/knowledge/context_bundle/build.rb
+++ b/app/services/knowledge/context_bundle/build.rb
@@ -156,7 +156,7 @@ module Knowledge
           parts.join("\n")
         end
 
-        { name: :schema, heading: "Data Model", content: lines.join("\n\n") }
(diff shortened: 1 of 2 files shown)


viamin:

PR #1501, opened 2026-04-27
fix(agent-runs): AgentRun#timeout! lacks row locking, causing race condition with concurrent status transitions

### LINKED ISSUE #1394: fix(agent-runs): AgentRun#timeout! lacks row locking, causing race condition with concurrent status transitions
## Problem

`AgentRun#timeout!` (`app/models/agent_run.rb:938`) does not use `with_lock`, unlike all other terminal status transition methods:

- `start!` (line 813) — uses `with_lock { reload; ... }`
- `complete!` (line 828) — uses `with_lock { reload; ... }`
- `pause!` (line 890) — uses `with_lock { reload; ... }`
- `resume!` (line 909) — uses `with_lock { reload; ... }`
- `timeout!` (line 938) — **no lock, no reload**

## Impact

Two callers invoke `timeout!` without external locking:

1. **`RunAgentActivity`** (`app/temporal/activities/run_agent_activity.rb:359`) calls `agent_run.timeout!(error:)` with only a `unless agent_run.finished?` guard — not atomic. Between the guard check and the `update!`, `StaleRunDetectorJob` could mark the same run as completed or timed out.

2. **`StaleRunDetectorJob`** (`app/jobs/stale_run_detector_job.rb:264`) does use `with_lock` externally, so it's safe — but `timeout!` itself doesn't enforce this contract.

The practical impact is limited (overwriting one terminal status with another is mostly harmless), but it violates the contract established by every other status transition method and could cause inconsistent state in edge cases (e.g., `completed_at` being rewritten).

## Proposed Fix

Add `with_lock { reload }` to `timeout!`, matching the pattern used by `start!`, `complete!`, `pause!`, and `resume!`. Also add a `return false if finished?` guard to match `complete!`.

```ruby
def timeout!(error: nil)
  with_lock do
    reload
    return false if finished?

    update!(
      status: "timeout",
      completed_at: Time.current,
      error_message: error,
      duration_seconds: duration
    )
  end
end
```

Note: This may require updating callers that currently rely on `timeout!` raising on conflict (e.g., inside an existing lock). The `StaleRunDetectorJob` caller already locks externally — nested `with_lock` on the same row is safe in ActiveRecord (re-entrant), but worth verifying in tests.

viamin:
## Summary

**Finding:** The `timeout!` method already had `with_lock`, `reload`, and `finished?` guard on main — the fix described in issue #1394 was already in place.

**Additional fix:** Applied the same row-locking pattern to `fail!` and `cancel!`, which were the remaining terminal status transition methods without `with_lock`. This completes the contract that all terminal status transitions use row locking to prevent race conditions.

**Changes:**
- `app/models/agent_run.rb` — Added `with_lock { reload; return false if finished? }` to `fail!` and `cancel!`
- `spec/models/agent_run_spec.rb` — Added "does not overwrite a finished run" tests for both `fail!` and `cancel!`, matching the existing `timeout!` test pattern

---

Closes #1394

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -939,12 +939,19 @@ class AgentRun < ApplicationRecord
   end
 
   def fail!(error: nil)
-    update!(
-      status: "failed",
-      completed_at: Time.current,
-      error_message: error,
-      duration_seconds: duration
-    )
+    with_lock do
+      reload
+      if finished?
+        false
+      else
+        update!(
+          status: "failed",
+          completed_at: Time.current,
+          error_message: error,
+          duration_seconds: duration
+        )
+      end
+    end
   end
 
   def pause!(violation_type:, context: nil)
@@ -987,11 +994,18 @@ class AgentRun < ApplicationRecord
   end
 
   def cancel!
-    update!(
-      status: "cancelled",
-      completed_at: Time.current,
-      duration_seconds: duration
-    )
+    with_lock do
+      reload
+      if finished?
+        false
+      else
+        update!(
+          status: "cancelled",
+          completed_at: Time.current,
+          duration_seconds: duration
+        )
+      end
+    end
   end
 
   def timeout!(error: nil)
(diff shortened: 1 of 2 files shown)


viamin:

PR #1508, opened 2026-04-27
Fix #1456: Simplify agent-run provider selection into a single mode selector

### LINKED ISSUE #1456: Simplify agent-run provider selection into a single mode selector
## Problem
The current provider selection UX for agent runs is confusing because it exposes two related dropdowns in provider settings: one for automated agent runs and another for the selection method. That makes it harder to understand the actual choice being made.

## Requested change
Replace this with a single selector that expresses the real decision for automated agent runs (auto pick, auto continue, auto review, etc.).

The choice should be between:
- using one single provider for automated agent runs
- distributing runs across all providers configured for agent runs (round robin / random pick, consistent with current behavior)

Two implementation directions would both be acceptable:
1. Preferably collapse the current two dropdowns into one selector with the same effective options.
2. Alternatively, reverse the order of the current dropdowns and only show the second dropdown when `Single` is selected in the first one.

## Expected behavior
- A user should be able to choose a single provider that is always used for automated agent runs.
- A user should also be able to choose the multi-provider mode that rotates / randomly selects from all providers configured for agent runs.
- Weighting should continue to work exactly as it does today.
- This should apply to automated agent runs such as auto pick, auto continue, and auto review.

## Trigger agent run form
The provider dropdown in the trigger-agent-run form should follow the same model.

Expected options there:
- `Inherit` (default)
- explicit single-provider selection
- the equivalent multi-provider behavior, if that surface currently supports it

At minimum, `Inherit` should be the default and should inherit from the provider settings described above.

## Acceptance criteria
- Provider settings no longer present two confusing overlapping dropdowns for automated agent-run provider choice.
- The UI clearly distinguishes between `single provider` and `multi-provider` selection behavior.
- The existing provider weighting behavior is preserved.
- The trigger-agent-run form defaults to `Inherit` from provider settings.
- The resulting behavior is easy to understand from the UI without needing to infer how the two dropdowns interact.

viamin:
## Summary

**Provider Settings UI** (`_settings.html.erb`):
- Replaced two separate dropdowns ("Automated Provider" + "Selection Mode") with a single combined "Automated Provider" selector
- Options: each enabled provider (for single-provider mode), plus "Round Robin" and "Random" (for multi-provider distribution) when >1 providers exist
- Provider Weights section still shows when multiple providers are enabled

**Controller** (`providers_controller.rb`):
- Added `expand_combined_provider_mode!` method that parses the combined `provider_mode` param (`single:<identifier>`, `round_robin`, or `random`) into the existing `provider_selection_mode` and `default_agent_provider` columns
- No database changes needed — the existing columns continue to work as before

**Trigger Agent Run form** (`new.html.erb`):
- Added "Inherit (from provider settings)" as the default selected option with empty value
- Removed per-provider pre-selection logic

**JS Controller** (`goal_toggle_controller.js`):
- When "Inherit" (empty value) is selected, skip goal-based provider syncing to keep the inherit selection stable

**Tests**: Updated 3 specs to match the new UI behavior (52 provider tests + 191 agent run tests all passing).

---

Closes #1456

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -393,7 +393,26 @@ class ProvidersController < ApplicationController
       end
     end
 
+    expand_combined_provider_mode!(permitted)
+
     permitted
+  end
+
+  # Parses the combined provider_mode param into provider_selection_mode
+  # and default_agent_provider. Values are either "single:<identifier>"
+  # for a specific provider, or "round_robin"/"random" for multi-provider
+  # distribution.
+  def expand_combined_provider_mode!(permitted)
+    combined = params.dig(:user_setting, :provider_mode).to_s.strip
+    return if combined.blank?
+
+    if combined.start_with?("single:")
+      provider_identifier = combined.delete_prefix("single:")
+      permitted[:provider_selection_mode] = "single"
+      permitted[:default_agent_provider] = provider_identifier
+    elsif UserSetting::PROVIDER_SELECTION_MODES.include?(combined)
+      permitted[:provider_selection_mode] = combined
+    end
   end
 
   # Applies per-provider weight updates from form params. Each entry must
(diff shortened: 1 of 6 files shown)


viamin:

PR #1511, opened 2026-04-27
feat(agent-runs): intelligent model selection based on task complexity

### LINKED ISSUE #760: feat(agent-runs): intelligent model selection based on task complexity
## Summary

Most LLM providers offer multiple model tiers, and in many cases a lower-tier model is perfectly adequate for routine agent runs. We should choose cheaper models for lower-complexity jobs and only escalate to higher-tier models when the task demands it or when quality scores indicate the cheaper model is insufficient.

For example, Claude Sonnet is likely sufficient for creating issues and for creating PRs for small or simple tasks, while Opus should be reserved for higher-complexity issues with many moving parts. The goal is to use the most cost-effective model by default and escalate only when needed.

## Proposed Sub-Tasks

This is an umbrella for several pieces of work, each of which should have its own issue:

### 1. Model Discovery and Evaluation
- Catalog available models per provider and classify them into tiers (e.g. low, mid, high)
- Store tier metadata (cost, capability level, context window, speed) so the selection algorithm can reference it
- Keep this data maintainable as providers release new models

### 2. Task Complexity Scoring Algorithm
- Define a complexity scoring heuristic for incoming tasks (e.g. based on issue size, number of files likely affected, presence of cross-cutting concerns, prior knowledge-base signals)
- Map complexity score ranges to model tiers
- This mapping will likely differ per provider and should be tunable over time as we collect quality metrics

### 3. Model Escalation Algorithm
- Start each agent run at the lowest tier appropriate for the task's complexity score
- After a run completes, evaluate the quality score against thresholds
- If quality is too low, automatically escalate to the next model tier on subsequent runs for similar tasks
- Track escalation history so we learn which task profiles need higher-tier models

### 4. Integration with Quality Metrics
- Wire the model selection and escalation logic into the existing quality scoring infrastructure
- Record which model was used per run so we can correlate model tier with quality outcomes
- Surface model-tier usage and quality data in dashboards and reports

### 5. Gap Analysis
- Identify any other features or infrastructure needed to make this work end-to-end
- Consider: provider API differences, rate-limit handling across tiers, fallback behavior if a tier is unavailable, user overrides, per-project model preferences

## Acceptance Criteria

- [ ] Agent runs default to the cheapest appropriate model tier for the given task complexity
- [ ] Runs that score below quality thresholds trigger automatic escalation to higher-tier models on retry
- [ ] Model tier, cost, and quality score are recorded per run and visible in reporting
- [ ] The system is configurable per provider and per project
- [ ] Sub-issues are created for each of the pieces above

viamin:
## Summary

### Problem
The model selection system had infrastructure for per-provider tier configurations (UI, validation, storage) but the actual selection logic never consulted those configurations. Additionally, the default complexity bias of 5.0 meant most runs defaulted to the "mid" tier, not the cheapest appropriate tier.

### Solution

**1. Wired provider `tier_model_ids` into the selection pipeline** (`rules_based_selector.rb`, `meta_agent_selector.rb`, `select.rb`)
- Both `RulesBasedSelector` and `MetaAgentSelector` now check the provider's `tier_model_ids` mapping before falling back to the global `LlmModel` pool
- Quality escalation in `Models::Select` also respects provider tier mappings
- Provider-configured models are skipped if they're inactive or in the project's exclusion list

**2. Lowered default complexity bias from 5.0 to 3.0** (`rules_based_selector.rb`)
- Agent runs now default to the cheapest appropriate model tier (acceptance criterion #1)
- Added a new complexity signal for medium-length issue bodies (>500 chars)
- Removed the penalty for very short bodies (starting low is sufficient)

**3. Added `QualityRecovery::ModelEscalation` spec** (new file)
- 15 examples covering `start`, `evaluate`, `active?`, and `target_tier` lifecycle
- Tests escalation start, already-active guard, highest-tier guard, exhausted guard
- Tests recovery on passing quality, insufficient samples, and prompt evolution fallback

**4. Updated existing specs** for the new complexity behavior
- All 207 related specs pass (100 model selection + 107 model/provider specs)
- Only pre-existing failures remain (database role guard, unrelated to changes)

---

Closes #760

### DIFF
--- a/app/services/models/meta_agent_selector.rb
+++ b/app/services/models/meta_agent_selector.rb
@@ -23,15 +23,15 @@ module Models
       candidates = available_candidates(tier: initial_tier)
       return nil if candidates.empty?
 
-      # When the provider has pinned a specific tier model the candidate pool is
-      # a single entry — skip the LLM round-trip and return it directly.
+      # When the candidate pool contains a single entry — whether from a provider
+      # tier pin or a global pool with one active model — skip the LLM round-trip.
       if candidates.size == 1
         only = candidates.first
         return {
           model: only,
           selector_type: "meta_agent",
           tier: initial_tier,
-          reasoning: "Single candidate from provider tier configuration; LLM selection skipped",
+          reasoning: "Single candidate in pool; LLM selection skipped",
           candidates: [ { model_id: only.model_id, score: only.capability_score.to_f } ],
           complexity_score: initial_complexity
         }
--- a/app/services/models/meta_agent_selector.rb
+++ b/app/services/models/meta_agent_selector.rb
@@ -23,6 +23,20 @@ module Models
       candidates = available_candidates(tier: initial_tier)
       return nil if candidates.empty?
 
+      # When the provider has pinned a specific tier model the candidate pool is
+      # a single entry — skip the LLM round-trip and return it directly.
+      if candidates.size == 1
+        only = candidates.first
+        return {
+          model: only,
+          selector_type: "meta_agent",
+          tier: initial_tier,
+          reasoning: "Single candidate from provider tier configuration; LLM selection skipped",
+          candidates: [ { model_id: only.model_id, score: only.capability_score.to_f } ],
+          complexity_score: initial_complexity
+        }
+      end
+
       response = request_selection(candidates)
       return nil unless response
 
--- a/app/services/models/meta_agent_selector.rb
+++ b/app/services/models/meta_agent_selector.rb
@@ -2,6 +2,8 @@
 
 module Models
   class MetaAgentSelector
+    include ProviderTierLookup
+
     MODEL = "claude-haiku-4-5-20251001"
     TIMEOUT = 15
     MAX_BODY_LENGTH = 2000
@@ -61,7 +63,7 @@ module Models
 
       # Prefer the provider's explicitly configured tier model when available
       provider_model = provider_tier_model(tier)
-      if provider_model && !(excluded.is_a?(Array) && excluded.include?(provider_model.model_id))
+      if provider_model && !excluded_model?(provider_model, excluded)
         return [ provider_model ]
       end
 
@@ -73,15 +75,6 @@ module Models
       end
 
       scope.order(Arel.sql("capability_score DESC NULLS LAST")).to_a
-    end
-
-    def provider_tier_model(tier)
-      return nil unless tier
-
-      model_id = agent_run.provider&.tier_model_ids&.dig(tier)
-      return nil if model_id.blank?
-
-      LlmModel.active.find_by(model_id: model_id)
     end
 
     def request_selection(candidates)
--- a/app/services/models/meta_agent_selector.rb
+++ b/app/services/models/meta_agent_selector.rb
@@ -59,6 +59,12 @@ module Models
       excluded = agent_run.project.model_preferences["excluded_model_ids"]
       scope = scope.where.not(model_id: excluded) if excluded.present?
 
+      # Prefer the provider's explicitly configured tier model when available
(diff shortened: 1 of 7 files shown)


viamin:
## Agent Update

**Review thread 1 (Bug: excluded models bypass in quality escalation)** — Fixed in `app/services/models/select.rb:70-75`. The `quality_escalation_result` method now checks `excluded_model_ids` before using the provider tier model, matching the behavior in the other selectors.

**Review thread 2 (DRY: inlined exclusion check)** — Extracted `provider_tier_model` and `excluded_model?` into a new shared module `Models::ProviderTierLookup` (`app/services/models/provider_tier_lookup.rb`), included by all three classes (`Select`, `RulesBasedSelector`, `MetaAgentSelector`). This eliminates the three duplicate copies and ensures they can't diverge.

viamin:
## Agent Update

Done. The fix adds a nil guard on line 75 of `app/services/models/select.rb` — `provider_model = nil if provider_model && excluded_model?(provider_model, excluded)` — matching the pattern already used in `RulesBasedSelector` and `MetaAgentSelector`. Without this guard, when `provider_tier_model` returns `nil` and the project has excluded models, `excluded_model?` would call `.model_id` on `nil` and raise `NoMethodError`.

viamin:
## Agent Update

Commit succeeded. The fix filters `excluded_model_ids` in the quality escalation fallback query at `app/services/models/select.rb:76`, exactly as the reviewer suggested. All 70 related tests pass and rubocop is clean.

viamin:
## Agent Update

Commit succeeded. The fix addresses the reviewer's feedback:

**Bug fixed in `app/services/models/select.rb:76`**: When `excluded` is `nil`, the original code `where.not(model_id: excluded.presence)` evaluated to `where.not(model_id: nil)`, which generates `WHERE model_id IS NOT NULL` — silently filtering out models with NULL `model_id`. The fix uses the same `if excluded.present?` guard pattern already used at the other two call sites (`rules_based_selector.rb:64` and `meta_agent_selector.rb:62`).

All 70 related specs pass with 0 failures. Rubocop passes clean.

viamin:
## Agent Update

Commit succeeded. The change adds a short-circuit in `MetaAgentSelector#call` (line 26-37) that returns the single candidate directly when `available_candidates` produces only one model, avoiding the unnecessary LLM API call and its associated cost/latency. All 85 model selection specs pass.

viamin:
## Agent Update

Commit succeeded. The review comment has been addressed:

- Changed the reasoning string from `"Single candidate from provider tier configuration; LLM selection skipped"` to `"Single candidate in pool; LLM selection skipped"` (matching the reviewer's suggestion exactly)
- Updated the comment above to accurately describe both cases: provider tier pin or a global pool with one active model

All 85 model selection specs pass, rubocop is clean, and the pre-commit hook passed.

viamin:
## Agent Update

Done. Here's a summary of what was changed:

**Review comment addressed:** The reviewer correctly identified that with the base complexity score lowered to 3.0, the maximum achievable score was 7.0 (3+1+1+1+1), making the "high" tier unreachable through rules-based selection since `7.0 ≤ mid_max(7)` routes to "mid".

**Fix:** Increased the complexity bump for very long issue bodies (>3000 chars) from `+1.0` to `+2.0`. This makes the maximum achievable score **8.0** (3+1+1+2+1), which exceeds `mid_max=7` and routes to the "high" tier. This means only truly complex issues (very long body + existing PR) reach "high" through heuristics, while simpler tasks still default to cheaper tiers.

**Tests updated:** The two specs that previously needed a lowered `mid_max` override to route to "high" now work with default thresholds, validating that the fix is correct.

viamin:

PR #1519, opened 2026-04-28
fix(cost-dashboard): label token usages by provider when model id is absent

viamin:
## Summary

- Cost dashboard's "Cost by Model" section was showing every cost as **Unknown**. All 1,080 `token_usages` rows in this database were persisted with `llm_model: nil` because subscription Claude Code runs (and other CLI agents that pick the model at runtime) never populate `response.model` on the harness response. `AgentRuns::TrackHarnessTokens` was writing the bare `nil` straight to the column.
- Fall back to the harness provider symbol (`response.provider.to_s`) when `response.model` is blank. Subscription Claude Code rows now bucket as `claude`; codex, cursor, etc. likewise label by provider.
- One-shot data migration mirrors the runtime fallback per `agent_type` so legacy null rows merge into the same dashboard buckets as new writes (the cost dashboard is not time-windowed, so without a backfill the "Unknown" bucket would persist alongside the new `claude`/`codex` buckets indefinitely).

### Why the cheap fix instead of routing Claude Code through the secrets proxy

The "proper fix" would be to route Claude Code's LLM calls through Paid's secrets proxy so per-call usage with the real model id (`claude-opus-4-7`, etc.) gets recorded. That is incompatible with the project's current Claude provider auth_type of `subscription`: the proxy injects an API key, which would convert subscription usage into metered API billing, fail outright when no API key is configured, and break Claude Code's native OAuth/MCP session flow that lives in `~/.claude`. `Containers::Provision` already opts subscription users out of the proxy on purpose. Dropping in a coarser provider-grain label is the right move until subscription-aware per-call tracking is designed.

### Why an explicit per-`agent_type` mapping in the migration

`response.provider.to_s` produces the harness `provider_name` symbol, which is the source of truth at runtime. The migration mirrors that mapping by `agent_type` so backfilled rows match what new writes produce — verified live against `AgentHarness::Providers::Registry`. Two specific gotchas that the verification caught: `gemini`, not `gemini_cli`, is the agent_type; and the GitHub Copilot harness emits `:github_copilot`, not `:copilot`.

### Verification

- Migration applied locally: `claude: 518`, `codex: 562`, 0 remaining nulls.
- `db:migrate:redo:primary` re-applied cleanly (idempotent via `WHERE llm_model IS NULL`).
- Spec covers fallback paths in `AgentRuns::TrackHarnessTokens`.

## Test plan

- [ ] CI green (RSpec, RuboCop, structure.sql check)
- [ ] On a staging environment with existing null `token_usages`, run the migration and confirm the cost dashboard's "Cost by Model" section no longer shows "Unknown" for backfilled rows
- [ ] Trigger a fresh subscription Claude Code agent run and confirm the resulting `run_summary`/`run_delta` rows persist `llm_model = "claude"`
- [ ] Trigger a Codex agent run and confirm rows persist `llm_model = "codex"`
- [ ] (If a Copilot run is feasible in staging) confirm rows persist `llm_model = "github_copilot"` so they bucket alongside any backfilled copilot rows

### DIFF
--- a/app/services/agent_runs/track_harness_tokens.rb
+++ b/app/services/agent_runs/track_harness_tokens.rb
@@ -65,9 +65,16 @@ module AgentRuns
       {
         tokens_input: tokens_input,
         tokens_output: tokens_output,
-        llm_model: response.model,
+        llm_model: llm_model_label,
         request_type: request_type
       }
+    end
+
+    # Falls back to the provider name when the harness response has no model
+    # id — common for Claude Code runs where the CLI selects the model at
+    # runtime rather than the runtime override pinning one.
+    def llm_model_label
+      response.model.presence || response.provider.to_s.presence
     end
 
     def run_input
--- a/db/migrate/20260428025840_backfill_null_llm_model_on_token_usages.rb
+++ b/db/migrate/20260428025840_backfill_null_llm_model_on_token_usages.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+# Backfills llm_model on existing token_usages rows that were written with
+# nil because the harness response didn't carry a model id. The mapping
+# matches the runtime fallback now applied in
+# AgentRuns::TrackHarnessTokens#llm_model_label, so legacy rows merge into
+# the same dashboard buckets as new rows instead of staying in "Unknown".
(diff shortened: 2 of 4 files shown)


viamin:

PR #1529, opened 2026-04-28
fix(github-tokens): broadcast validation state instead of polling

viamin:
## Summary

- The token show page polled `/github_tokens/:id/validation_status` every 2 seconds via `setTimeout` while waiting for `GithubTokenValidationJob`. With a fine-grained PAT and a multi-repo account the job takes 30+ seconds because `sync_repositories!` probes write access per repo, so the user saw a constantly-cycling browser tab spinner the entire time.
- The status badge in the page header was never updated by the polling loop either, so after the green "validated" banner appeared in the panel below, the badge still spun "Validating..." until a manual refresh.
- Replaces the polling with `turbo_stream_from @github_token, :validation_status`. `mark_validating!` / `mark_validated!` / `mark_validation_failed!` now each `broadcast_replace_to` both the `validation-status` panel and the `status-badge` frame, so both transition together the moment the job lands.

## Audit
Audited the rest of the codebase for similar polling-where-broadcast-belongs patterns. Only the github-tokens validation flow was wrong. The other `setInterval` / `setTimeout` usages — `elapsed_time_controller.js` (1s wall-clock tick), `local_time_controller.js` (60s relative-time tick), `live_dashboard_controller.js` (300ms one-shot animation cleanup) — are legitimate live-data displays with no model state to broadcast.

## Known limitations (deliberately not fixed in this PR)
- **Subscription-vs-completion race for sub-500ms job outcomes.** If a job fails fast (immediate auth error) before the browser's WebSocket subscription is confirmed, the failure broadcast is dropped and the user sees "Validating..." until refresh. Every viable mitigation I considered (eager `<turbo-frame src=>` catch-up fetch, post-subscribe Stimulus refresh) introduces a new race where the catch-up overwrites a faster broadcast. The codebase's other broadcast-driven flows (`Project#broadcast_*`) accept the same exposure.
- **Loss of stale-state self-healing.** Old polling rendered the "Validation Stuck" UI automatically after 2 minutes; this now requires a manual refresh OR the periodic `GithubTokenHealthCheckJob`.
- **The 32-second job duration itself.** Out of scope here. The next user-visible win would be parallelizing `client.write_accessible?` probes in `sync_repositories!`, or splitting repo enumeration into a separate job so the green banner appears after the fast `/user` call instead of after the per-repo write probes.

## Test plan
- [x] `bin/rspec spec/models/github_token_spec.rb` — 75/75 pass, including 6 new broadcast call-signature tests and 3 new integration tests that exercise actual `Turbo::StreamsChannel.renderer` rendering across pending / failed / validated states (the failed-state test specifically catches `button_to` rendering issues that stubbed-broadcast tests would miss)
- [x] `bin/rspec spec/jobs/github_token_validation_job_spec.rb` — 9/9 pass
- [x] `bin/rspec spec/requests/github_tokens_spec.rb -e "GET /github_tokens/:id/validation_status"` — 5/5 pass (controller action still works for cached client URLs)
- [x] `bin/rubocop` clean
- [ ] Manual: submit form to add a new GitHub token, confirm browser tab spinner stops cycling, badge transitions from "Validating..." to "Active" together with the green banner


viamin:

PR #1534, opened 2026-04-28
fix(devcontainer): remove redundant PG 15 feature and add pg_dump to agent container

viamin:
## Summary

- Remove `ghcr.io/rails/devcontainer/features/postgres-client` from devcontainer.json — it installed an unnecessary `postgresql-client-15` alongside the explicitly-pinned `postgresql-client-16` from PGDG, causing `pg_wrapper` version resolution ambiguity
- Add `postgresql-client-16` to `docker/agent/Dockerfile` so the agent container can run `pg_dump` for schema dumps during `db:migrate` (previously had only `libpq.so`, no client tools)
- Extend CI sync check to also verify `docker/agent/Dockerfile` and `.github/workflows/system_tests.yml`, both of which had pins but were not enforced

## Why

The agent container had no `pg_dump` at all, so any `structure.sql` committed by the agent was suspect. The devcontainer had two competing PG client versions (15 and 16) that could produce different `pg_dump` output depending on `pg_wrapper` resolution. This is the root cause of persistent minor `db/structure.sql` drift.

> Note: Ubuntu 24.04's `postgresql-client-16` will be a different minor version (e.g. 16.2) than the PGDG 16.13 in the devcontainer. A follow-up could add PGDG to the agent Dockerfile for exact version parity.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -14,8 +14,7 @@
     "ghcr.io/devcontainers/features/node:1": {},
     "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {
       "moby": false
-    },
-    "ghcr.io/rails/devcontainer/features/postgres-client": {}
+    }
   },
   "containerEnv": {
     "TZ": "${localEnv:TZ:UTC}",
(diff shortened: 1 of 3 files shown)


viamin:
**Escalation Note**

**How to resolve:**

viamin:

Issue #1550, opened 2026-04-28
test: Critical test gaps for agent run provider selection and fallback


viamin:
## Problem

The agent run system has critical test gaps that allowed the Claude/Codex provider mismatch bug to reach production. Individual components are well-tested in isolation, but cross-cutting scenarios are untested.

### Provider Selection
- **No test for agent runs when default provider is disabled/rate-limited** — Would have caught the Claude-hardcoded-defaults bug
- **No test for `ProviderResolver` returning nil provider with only Codex available** — Tests the fallback_from_settings path
- **No test for `CreateFollowup` when provider resolver returns nil** — Tests the fallback_agent_type path
- **No test for `default_provider_candidates` with rate-limited default provider** — Tests dead entries in provider list
- **No test for `build_provider_order` with nil provider + non-default agent_type** — Tests provider order construction

### Provider Fallback
- **No test for `Containers::Provision::ExecutionError` triggering provider fallback** — The rescue was completely missing (now fixed, needs test)
- **No test for provider fallback timing budget** — After first provider consumes full timeout, remaining providers get insufficient budget

### Integration
- **No integration test for full agent run lifecycle** (creation → queue → workflow → container → execution → PR)
- **No test for concurrent queue processing** (two jobs racing to claim the same run)
- **No test for wall-clock timeout end-to-end** (activity → stale detector → cleanup)

### Controller
- **No test for `POST /projects/:id/agent_runs/:id/retry` endpoint**
- **No test for `POST /projects/:id/agent_runs/:id/refresh_auth` endpoint**

## Related

- #1541 (scheduler_paused stuck — needs auto-resume test)
- #1542 (circuit breaker race — needs concurrency test)
- #1544 (auth failure false positives — needs negative test cases)

viamin:
Pull request created: https://github.com/viamin/paid/pull/1604

Issue #1544, opened 2026-04-28
fix: Auth failure checker false positives on rate-limit 403s and infrastructure 503s


viamin:
## Problem

`GithubTokens::AuthFailureChecker` (`app/services/github_tokens/auth_failure_checker.rb`) has overly broad regex patterns that trigger false positives:

1. **`/HTTP 403/`** (line 9) matches GitHub secondary rate-limit 403s, not just auth failures. During rate-limit periods, every failed agent run triggers `GithubTokenValidationJob`, creating a validation storm.

2. **`/proxy.*503/i`** (line 19) matches infrastructure 503s unrelated to credentials. During Paid proxy overload, failed runs trigger unnecessary token validation.

3. **No concurrency control** on `GithubTokenValidationJob` — a burst of 50 failed runs can enqueue 50 simultaneous validation jobs for the same token.

## Impact

During a sustained outage or rate-limit period, dozens of `GithubTokenValidationJob` instances fire, each making an API call to GitHub, potentially worsening the outage. If GitHub returns transient 401s, the cascade from the auto-pause issue can permanently pause projects.

## Fixes

1. Replace `/HTTP 403/` with a more specific pattern that excludes known rate-limit 403 messages
2. Replace `/proxy.*503/i` with patterns that specifically target credential-related 503s
3. Add `good_job_control_concurrency_with` to `GithubTokenValidationJob` to limit to 1 concurrent execution per token

## Files

- `app/services/github_tokens/auth_failure_checker.rb:9,19`
- `app/jobs/github_token_validation_job.rb`

viamin:
Pull request created: https://github.com/viamin/paid/pull/1592

Issue #1542, opened 2026-04-28
fix: Circuit breaker record_success! race condition allows overwriting open state


viamin:
## Problem

`GithubHealthState#record_success!` (`app/models/github_health_state.rb:84-103`) lacks the `with_lock` wrapper that `record_failure!` correctly uses at line 50. A concurrent `record_failure!` can transition the circuit from `half_open` → `open` between the `circuit_open?` check (line 86) and the `update!` call (line 90). The success's `update!` then overwrites the `open` state back to `closed`, defeating the circuit breaker during an ongoing GitHub outage.

## Fix

Wrap `record_success!` in `with_lock`:

```ruby
def record_success!
  with_lock do
    return if circuit_state == "closed" && failure_count.zero?
    return if circuit_open?
    was_half_open = circuit_half_open?
    update!(
      failure_count: 0,
      circuit_state: "closed",
      circuit_opened_at: nil,
      last_error_message: nil
    )
    # ... logging
  end
end
```

Also wrap `check_circuit_recovery!` in `with_lock` to prevent double half-open transitions.

## Files

- `app/models/github_health_state.rb:84-103` (`record_success!`)
- `app/models/github_health_state.rb:109-121` (`check_circuit_recovery!`)

viamin:
Pull request created: https://github.com/viamin/paid/pull/1594

Issue #1541, opened 2026-04-28
fix: Projects permanently stuck in scheduler_paused after token re-validation


viamin:
## Problem

When `GithubTokenValidationJob` confirms a token is invalid (auth failure), `GithubTokens::AutoPauseProjects` pauses all active projects using that token by setting `scheduler_paused_at`. However, when the token is later re-validated successfully (e.g., GitHub recovers from an outage, token is refreshed), **the projects are never auto-resumed**.

The only auto-resume path is `Project#clear_scheduler_pause_on_token_change` (`app/models/project.rb:723`), which fires on `after_update_commit` when `github_token_id` **changes** — not when the existing token re-validates.

## Impact

All projects sharing a token that experiences a transient auth failure are permanently stuck in `scheduler_paused` with no self-healing path. Auto-pick, issue scanning, and the entire automation pipeline are dead until someone manually clicks "Resume" in the UI.

## Reproduction

1. Token gets a transient 401 during a GitHub outage
2. `GithubTokenValidationJob` → `AuthenticationError` → `auto_pause_projects`
3. GitHub recovers, next health check validates token → `mark_validated!`
4. Projects remain paused forever

## Fix

Add auto-resume logic after successful token re-validation:

```ruby
# In GithubTokenValidationJob#validate_token, after mark_validated!:
GithubTokens::AutoResumeProjects.call(github_token: github_token) if was_already_failed
```

Create `GithubTokens::AutoResumeProjects` service that resumes projects paused by the token failure.

## Files

- `app/jobs/github_token_validation_job.rb:22-29`
- `app/models/project.rb:646-657` (`scheduler_resume!`)
- `app/services/github_tokens/auto_pause_projects.rb`

## Discovered During

Agent run failure investigation (#ref needed)

viamin:
Pull request created: https://github.com/viamin/paid/pull/1579

PR #1551, opened 2026-04-28
fix(agent-runs): remove hardcoded Claude preference from provider fallback paths

viamin:
## Summary

- Replace hardcoded `claude_code` fallbacks in `ProviderResolver`, `CreateFollowup`, and `RunAgentActivity` with dynamic resolution that queries enabled providers first via new `Provider.first_enabled_for-owner`
- Add `Containers::Provision::ExecutionError` rescue in `RunAgentActivity` so Docker exec errors trigger provider fallback instead of failing the entire run
- Add provider validation in `CreateAgentRunActivity` to fail fast when no runnable provider is available
- Remove hardcoded Claude preference from `Provider.default_provider_key`

## Root Cause

When Claude becomes rate-limited and is removed from the round-robin provider rotation (leaving only Codex), agent runs still fail because `"claude_code"` is hardcoded as the default agent_type in 5 fallback paths. The system keeps attempting Claude even when it's unavailable, consuming the full timeout budget before falling back to Codex (or failing entirely if fallback is disabled).

## Test plan

- [x] All 288 existing tests pass (`spec/models/provider_spec.rb`, `spec/services/agent_runs/*`, `spec/temporal/activities/*`)
- [x] Lint clean (`bin/lint --changed`)
- [ ] Manual: disable Claude provider, verify agent runs use Codex without hardcoded Claude fallback
- [ ] Manual: verify Docker exec errors trigger provider fallback instead of failing the run

## Related issues

- #1541 — Projects permanently stuck in scheduler_paused after token re-validation (P1)
- #1542 — Circuit breaker record_success! race condition (P2)
- #1544 — Auth failure checker false positives (P2)
- #1550 — Critical test gaps for provider selection and fallback (P2)

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -247,13 +247,12 @@ class Provider < ApplicationRecord
   end
 
   # Returns the provider key that must always exist and remain enabled for
-  # agent runs. Prefers "claude" when container-executable, otherwise falls
-  # back to the first available container-executable provider. Returns nil
-  # when no container-executable providers are available, so callers can
-  # surface a user-facing error instead of a 500.
+  # agent runs. Returns the first container-executable provider key in
+  # supported order, with no hardcoded preference for any specific provider.
+  # Returns nil when no container-executable providers are available, so
+  # callers can surface a user-facing error instead of a 500.
   def self.default_provider_key
-    executable_keys = ProviderSupport.container_executable_provider_keys
-    executable_keys.include?("claude") ? "claude" : executable_keys.first
+    ProviderSupport.container_executable_provider_keys.first
   end
 
   def self.ensure_default_for(user)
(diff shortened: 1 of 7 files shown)


viamin:
@codex review

viamin:
## Review feedback addressed

### 1. P1: Skip runnable-provider guard for non-container goals (resolved)
`validate_runnable_provider!` in `CreateAgentRunActivity` now skips the guard for `enhance_issue` and `analyze_issue` goals via a `NON_CONTAINER_GOALS` constant and early return. These goals route to `EnhanceIssueActivity`/`AnalyzeIssueActivity` before container provisioning, so a container-executable provider is not required.

### Commits
- `94d00b7` — fix(tests): stub `Provider.first_enabled_for_owner` in auto_pick no-provider test
- `524c35d` — fix(agent-runs): skip runnable-provider guard for non-container goals

@codex re-review

PR #1559, opened 2026-04-29
feat(performance): ungate CI performance benchmarks and improve reliability

viamin:
## Summary

- Removes the `PAID_PERFORMANCE_BENCHMARKS` repository variable gate so the `performance` CI job runs on every PR and push to main
- Seeds 5 fixture agent runs (up from 1) in `CiSeedData` for meaningful p95 calculations on `container_startup_time` and `workflow_latency` benchmarks
- Increases benchmark iterations from 5 to 10 for `dashboard_load_time` and `search_latency` to reduce variance on noisy CI runners
- Widens budgets for CI-realistic thresholds (dashboard 1s→3s, search 500ms→2s, regression thresholds 1.25→1.5)
- Wraps `benchmark.rb` execution in `TenantContext.with_system_access` so RLS-protected table queries work regardless of database user role (previously relied on `postgres` superuser implicitly)
- Adds test coverage for the budget-exceeded `"fail"` path in `Measurement.from_samples` (lost when budgets were widened)

## Test plan

- [x] All 13 performance benchmark specs pass
- [x] All 3 query-count performance specs pass
- [x] Lint clean
- [ ] CI `performance` job runs and passes (will be first real validation)
- [ ] Baseline values in `docs/performance/baseline.json` should be updated after first successful CI run with real measurements

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,7 +170,6 @@ jobs:
 
   performance:
     runs-on: ubuntu-latest
-    if: vars.PAID_PERFORMANCE_BENCHMARKS == 'true'
 
     services:
       postgres:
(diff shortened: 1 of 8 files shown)


viamin:
@codex review

PR #1570, opened 2026-04-29
fix(kilocode): resolve test-agent timeout for direct-outbound API keys

viamin:
## Summary

Fixes the "timed out after 30 seconds" error when clicking the test-agent button for Kilocode providers with direct-outbound API keys (e.g. z.ai/GLM 5.1, DeepSeek, etc.).

Four root causes identified and fixed:

1. **Config file cleaned up before smoke test** — `prepare_kilocode_config!` used `ExecutionPreparation` which deletes files in the `ensure` block after each `execute` call. The kilo config was written then immediately removed before the smoke test ran. Fixed by writing the config via direct `execute_in_container`.

2. **Invalid config format** — agent-harness generates `{"provider":"openai-compatible"}` (string) but kilo CLI v7.1.3 requires `{"provider":{"openai":{}}}` (record). Fixed by generating the correct format in `kilocode_container_config_json`.

3. **Missing `--auto` flag** — agent-harness `test_command_overrides` is never wired into the smoke test. Without `--auto`, kilo enters interactive mode and hangs. Fixed by injecting `--auto --print-logs` in `HarnessExecutor` for kilo commands.

4. **Missing `OPENAI_BASE_URL`** — `kilocode_provider_runtime` didn't pass the base URL for non-native API endpoints. Fixed by adding `OPENAI_BASE_URL` to the runtime env.

## Upstream issues filed

- viamin/agent-harness#181 — `test_command_overrides` never wired into smoke test
- viamin/agent-harness#182 — `config_file_content` generates invalid JSON for kilo CLI v7.1.3
- viamin/agent-harness#173 — (pre-existing) smoke test contract timeout overrides caller timeout

## Known limitation

kilo CLI v7.1.3 validates models against its built-in catalog. GLM models aren't in the catalog, so `glm-5.1` will get "Model not found: openai/glm-5.1" — but this is a clear error instead of a vague timeout. Adding GLM to kilo's model catalog requires an upstream change.

## Test plan

- [x] 51 tests pass (40 test_agent + 10 harness_executor + 1 which)
- [x] Rubocop clean on all 4 changed files
- [ ] Manual: test-agent button for kilocode + anthropic API key provider
- [ ] Manual: test-agent button for kilocode + z.ai API key provider

### DIFF
--- a/app/services/containers/harness_executor.rb
+++ b/app/services/containers/harness_executor.rb
@@ -3,34 +3,20 @@
 require "shellwords"
 
 module Containers
-  # Adapts container-based command execution to the AgentHarness::CommandExecutor
-  # interface, allowing agent-harness smoke tests to run inside provisioned
-  # Docker containers transparently.
-  #
-  # @example
-  #   test_run.with_container do |run|
-  #     executor = Containers::HarnessExecutor.new(run)
-  #     provider.smoke_test(timeout: 60, executor: executor)
-  #   end
   class HarnessExecutor
+    KILOCODE_AUTO_FLAGS = %w[--auto --print-logs].freeze
+    private_constant :KILOCODE_AUTO_FLAGS
+
     def initialize(agent_run)
       @agent_run = agent_run
     end
 
-    # Executes a command inside the container, conforming to the
-    # AgentHarness::CommandExecutor#execute contract.
-    #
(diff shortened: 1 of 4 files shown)


viamin:
@codex review

PR #1571, opened 2026-04-30
fix(kilocode): inject PAID_KILOCODE_CONFIG_B64 into container env

viamin:
## Summary

- Fixes the kilo CLI hanging for the full 3600s wall-clock timeout when used as a fallback provider
- `command_env_for` in `run_agent_activity.rb` never called `Provider#direct_outbound_exec_env` for kilocode providers, so `PAID_KILOCODE_CONFIG_B64` was unset in the container environment
- The shell command writes config via `$PAID_KILOCODE_CONFIG_B64 | base64 -d > config.json` — with the var empty, the kilo CLI started with no provider/model and hung silently

## Root Cause

`command_env_for` had two paths: `agent_harness_runtime?` (opencode/copilot) and `api_key?` (everyone else with an API key). Kilocode is `requires_direct_outbound?` → true AND `api_key?` → true, so it fell through to the `api_key?` path which only returns `{"PAID_PROVIDER_ID" => "11"}`. The `direct_outbound_exec_env` method that provides `PAID_KILOCODE_CONFIG_B64` existed but was dead code for production runs.

## Impact

~170 timeouts over 3 days (20.7% of all runs) via the cascade: codex/claude rate-limited → fallback to kilocode → empty config → 60min hang. ~170 hours of container compute wasted.

## Changes

- `app/temporal/activities/run_agent_activity.rb`: `command_env_for` now merges `direct_outbound_exec_env` when `requires_direct_outbound?` is true
- `spec/temporal/activities/run_agent_activity_spec.rb`: Tests for both direct-outbound kilocode (config injected) and subscription kilocode (config not injected)

### DIFF
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -1170,9 +1170,11 @@ module Activities
       provider_entry = provider_entry_for(command_context.provider_candidate, command_context.user)
       return direct_outbound_execution_plan(provider_entry, prompt).env if provider_entry&.agent_harness_runtime?
       return {} unless provider_entry
-      return api_key_command_env(provider_entry) if provider_entry.api_key?
 
-      {}
+      env = {}
+      env.merge!(provider_entry.direct_outbound_exec_env) if provider_entry.requires_direct_outbound?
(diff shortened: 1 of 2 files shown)


viamin:
@codex review

PR #1572, opened 2026-04-30
fix(agent-runs): extend idle timeout for providers with coarse heartbeat signals

viamin:
## Summary

- Extends idle timeout 3x for providers whose heartbeat only fires between CLI turns (codex), not during subprocess execution
- Codex's notify hook doesn't fire during long `bundle exec rspec` runs, causing 32 false-positive idle timeouts over 3 days where the agent was actively working (high CPU)
- `HeartbeatSetup` now distinguishes reliable heartbeat providers (claude: per-tool PostToolUse hook) from coarse heartbeat providers (codex: per-turn notify)

## Root Cause

Codex buffers subprocess output and only emits JSONL events when a subprocess completes. Its `notify` hook fires between turns, not during `command_execution` items. When a test suite runs for >300s without stdout, the idle timeout kills the run even though codex is actively working.

Claude avoids this because its PostToolUse hook fires after every tool call (Bash, Read, Write, etc.), touching the heartbeat file every few seconds on average.

## Changes

- `app/services/containers/heartbeat_setup.rb`: Add `RELIABLE_HEARTBEAT_PROVIDERS`, `COARSE_HEARTBEAT_IDLE_TIMEOUT_MULTIPLIER = 3`, `idle_timeout_for(base)`, and `reliable_heartbeat?`
- `app/temporal/activities/run_agent_activity.rb`: Delegate idle timeout calculation to `heartbeat.idle_timeout_for(effective_idle_timeout)`

## Effective Timeouts

| Provider | create_pr | issue_goal | review |
|----------|-----------|------------|--------|
| claude (reliable) | 300s | 120s | 300s |
| codex (coarse, 3x) | 900s | 360s | 900s |
| no heartbeat | disabled | disabled | disabled |

### DIFF
--- a/app/services/containers/heartbeat_setup.rb
+++ b/app/services/containers/heartbeat_setup.rb
@@ -30,6 +30,16 @@ module Containers
 
     SUPPORTED_PROVIDERS = %w[claude codex].freeze
 
+    # Providers with per-tool heartbeat hooks (e.g. Claude PostToolUse)
+    # that fire frequently enough to suppress idle timeouts reliably
+    # during long subprocess execution. Providers in SUPPORTED_PROVIDERS
+    # but NOT listed here use coarser heartbeat signals that only fire
+    # between CLI turns, so they need a longer idle timeout to avoid
+    # false positives during long-running commands like `bundle exec rspec`.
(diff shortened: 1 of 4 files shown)


viamin:
@codex review

PR #1585, opened 2026-05-01
fix(providers): generate record-format provider in kilocode and opencode CLI configs

viamin:
## Summary

- Fixes Kilocode and OpenCode CLI config generation to produce `"provider": {"key": {}}` (record) instead of `"provider": "key"` (string), which both CLIs reject with config validation errors
- Removes invalid `"baseURL"` key from OpenCode config (base URL is correctly set via `OPENAI_BASE_URL` env var instead)
- Defensively guards against nil `base_url` in env hash to prevent `ArgumentError` from `ProviderRuntime` validation
- Root cause of agent run 11942 where all three fallback providers failed with `'expected record, received string provider'`

## Root cause

The agent-harness gem's `config_file_content` (kilocode) and `opencode_config_payload` (opencode) both generate `"provider": "string"` format. Kilo CLI v7.1.3 and OpenCode CLI v1.3.2 require `"provider": {"key": {}}` (record/object).

**Before (broken):**
```json
{"provider": "openai-compatible", "model": "glm-5.1"}
```

**After (correct):**
```json
{"provider": {"openai": {}}, "model": "openai/glm-5.1"}
```

## Test plan

- [x] `spec/models/provider_spec.rb` — updated expectations for new runtime shape
- [x] `spec/services/providers/harness_execution_plan_spec.rb` — new test verifying record-format provider and absence of baseURL
- [x] `spec/temporal/activities/run_agent_activity_spec.rb` — updated kilocode config expectations
- [x] All 137 provider specs pass
- [x] All 154 run_agent_activity specs pass
- [x] Pre-commit hooks pass (RuboCop, ESLint, markdownlint, ShellCheck)

## Follow-up

- The gem's `config_file_content` (kilocode) and `opencode_config_payload` (opencode) should be fixed upstream in agent-harness so the Paid workarounds can be removed

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -202,8 +202,12 @@ class Provider < ApplicationRecord
 
     {
       provider: { kilocode_provider_key => {} },
-      model: "#{kilocode_provider_key}/#{model_id}"
+      model: kilocode_qualified_model(kilocode_provider_key, model_id)
     }.to_json
+  end
+
+  def kilocode_qualified_model(provider_key, model_id)
+    model_id.include?("/") ? model_id : "#{provider_key}/#{model_id}"
   end
 
   def direct_outbound_exec_env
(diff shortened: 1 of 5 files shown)


viamin:
@codex review

viamin (commented):
Both review comments addressed in 32669d6:\n\n1. **Copilot (duplication)**: Extracted `Provider#kilocode_config_json` as the single source of truth. Removed `TestAgent#kilocode_container_config_json` — TestAgent now delegates via `provider.kilocode_config_json`.\n\n2. **Codex (double-prefix)**: Added `Provider#kilocode_qualified_model` that skips prefixing when the model ID already contains a `/`. New tests cover both bare and pre-qualified model IDs.

viamin:
@codex review

PR #1586, opened 2026-05-01
fix(agent-runs): cache distinct_effective_providers to eliminate full-table scan on index page

viamin:
## Summary

- Cache `distinct_effective_providers` with 5-minute TTL to eliminate an unindexed `SELECT DISTINCT` computed expression scan over all `agent_runs` that ran on every index page load
- Invalidate cache on create, status transitions, and `final_provider` changes
- Make `compute_distinct_effective_providers` a private class method to prevent cache bypass

## Why

The agent runs index page called `distinct_effective_providers` on every request to populate a filter dropdown. This executed a `COALESCE(CASE ...)` expression over the entire `agent_runs` table (or project-scoped subset) with no index support. As run volume grows, this query becomes increasingly expensive.

## What changed

- `AgentRun.distinct_effective_providers(cache_key:)` now wraps the query in `Rails.cache.fetch` with a 5-minute TTL
- `AgentRun.invalidate_provider_options_cache` invalidates both account-level and project-level cache keys
- `after_commit :invalidate_provider_options_cache_on_change` fires on create and on status/final_provider changes
- Both `AgentRunsController#index` and `Projects::AgentRunsController#index` pass scoped cache keys

## Test plan

- [x] All 193 request specs pass
- [x] RuboCop clean
- [ ] Manual: load agent runs index, verify provider filter dropdown still works
- [ ] Manual: trigger a new run, verify new provider appears in dropdown after cache invalidation

### DIFF
--- a/app/controllers/agent_runs_controller.rb
+++ b/app/controllers/agent_runs_controller.rb
@@ -10,7 +10,8 @@ class AgentRunsController < ApplicationController
     @q.sorts = "created_at desc" if @q.sorts.empty?
     @pagy, @agent_runs = pagy(@q.result)
     AgentRun.preload_source_pull_requests(@agent_runs)
-    @provider_options = base_scope.distinct_effective_providers
+    cache_key = AgentRun.provider_options_cache_key_for(account_id: current_account.id)
+    @provider_options = base_scope.distinct_effective_providers(cache_key: cache_key)
   end
 
   def pause_scheduler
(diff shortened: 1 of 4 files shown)


viamin:
@codex review

All three review comments addressed:
1. Cache key now includes account_id in project-scoped path to prevent collision risk
2. Invalidation guard tightened to only agent_type/final_provider changes — no churn on intermediate status transitions
3. Added 8 model-level specs for caching behavior and invalidation

PR #1594, opened 2026-05-01
fix: Circuit breaker record_success! race condition allows overwriting open state

### LINKED ISSUE #1542: fix: Circuit breaker record_success! race condition allows overwriting open state

viamin:
## Summary

Updated `GithubHealthState` so both `record_success!` and `check_circuit_recovery!` run inside `with_lock`, which closes the race where a concurrent failure could reopen the circuit and then get overwritten back to `closed`. I also added regression coverage in `spec/models/github_health_state_spec.rb` for the success/failure interleaving and concurrent recovery transition.

Validation ran clean: `bin/rails db:prepare`, full `bundle exec rubocop`, full `bundle exec rspec` (`9602 examples, 0 failures, 2 pending`), and the pre-commit hook passed on commit. Commit: `0d76e52` (`fix(github-health): lock success and recovery transitions`).

---

Closes #1542

### DIFF
--- a/app/models/github_health_state.rb
+++ b/app/models/github_health_state.rb
@@ -82,23 +82,25 @@ class GithubHealthState < ApplicationRecord
   # an in-flight success while circuit is open must not bypass the
   # recovery timeout.
   def record_success!
-    return if circuit_state == "closed" && failure_count.zero?
-    return if circuit_open?
+    with_lock do
+      return if circuit_state == "closed" && failure_count.zero?
+      return if circuit_open?
 
-    was_half_open = circuit_half_open?
+      was_half_open = circuit_half_open?
 
-    update!(
-      failure_count: 0,
-      circuit_state: "closed",
-      circuit_opened_at: nil,
-      last_error_message: nil
-    )
-
-    if was_half_open
-      Rails.logger.info(
-        message: "github_health.circuit_closed",
-        endpoint: endpoint
+      update!(
+        failure_count: 0,
+        circuit_state: "closed",
+        circuit_opened_at: nil,
+        last_error_message: nil
       )
+
+      if was_half_open
+        Rails.logger.info(
+          message: "github_health.circuit_closed",
+          endpoint: endpoint
+        )
+      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

Commit succeeded. Here's a summary of the changes addressing all three review comments:

1. **`require "timeout"`** (line 4) — Added explicit require so `Timeout.timeout` doesn't depend on transitive load order.
2. **`Thread#value` in `run_success_failure_race`** (line 41) — Replaced `.join` with `.value` so exceptions from `record_success!`/`record_failure!` threads propagate into the example and cause a deterministic failure.
3. **`Thread#value` in `run_recovery_race`** (line 64) — Same fix for the concurrent recovery test threads.

viamin:
## Agent Update

Commit succeeded. All three review comments are addressed:

1. **Connection pool leaks in `run_success_failure_race`** — Both threads now release their DB connections via `ensure { ActiveRecord::Base.connection_pool.release_connection }`, preventing pool exhaustion.

2. **Tight `timeout: 0.1`** — Increased to `timeout: 2` for CI reliability while still asserting the blocking behavior (the failure thread is blocked on the row lock, so the timeout fires deterministically).

3. **Connection pool leaks in `run_recovery_race`** — Same `ensure`/`release_connection` pattern applied to both threads.

viamin:

PR #1628, opened 2026-05-03
Fix #1457: Make notifications table horizontally scrollable on mobile

### LINKED ISSUE #1457: Make notifications table horizontally scrollable on mobile
## Problem
The notifications page is difficult to use on mobile because the table does not fit within the viewport. Columns get compressed to the point that notification content is not meaningfully readable.

## Requested change
Make the notifications table horizontally scrollable on small screens instead of forcing the full table to shrink into the mobile viewport.

## Expected behavior
- On mobile widths, the notifications table is contained within a horizontally scrollable region
- Table columns keep readable minimum widths rather than collapsing too aggressively
- A user can scroll sideways to inspect the full row content
- Desktop behavior remains unchanged

## Acceptance criteria
- [ ] The notifications page table can be horizontally scrolled on mobile devices
- [ ] Notification content remains readable at mobile viewport widths
- [ ] The table does not overflow the page in a way that breaks layout or hides content without scroll access
- [ ] Desktop/tablet layouts continue to render as expected

viamin:
## Summary

Committed successfully. Here's what I changed:

**`app/views/notifications/index.html.erb`** — two changes:

1. **Wrapped the `<table>` in a `<div class="overflow-x-auto">`** — enables horizontal scrolling on mobile instead of squishing columns.
2. **Added `min-w-[...]` classes to each `<th>`** — prevents columns from collapsing below readable widths (Severity: 5rem, Notification: 16rem, Source: 6rem, Time: 7rem, Actions: 6rem).

On desktop, the table still renders normally since `min-w-full` keeps it filling the container. On mobile, the minimum widths cause the table to exceed the viewport width, and `overflow-x-auto` provides a scrollbar.

---

Closes #1457

### DIFF
--- a/app/views/notifications/index.html.erb
+++ b/app/views/notifications/index.html.erb
@@ -22,14 +22,15 @@
   </div>
 
   <div class="mt-6 overflow-hidden shadow ring-1 ring-black ring-opacity-5 sm:rounded-lg">
+    <div class="overflow-x-auto">
     <table class="min-w-full divide-y divide-gray-300">
       <thead class="bg-gray-50">
         <tr>
-          <th scope="col" class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6">Severity</th>
-          <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Notification</th>
-          <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Source</th>
-          <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Time</th>
-          <th scope="col" class="relative py-3.5 pl-3 pr-4 sm:pr-6"><span class="sr-only">Actions</span></th>
+          <th scope="col" class="min-w-[5rem] py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6">Severity</th>
+          <th scope="col" class="min-w-[16rem] px-3 py-3.5 text-left text-sm font-semibold text-gray-900">Notification</th>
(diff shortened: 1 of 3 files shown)


viamin:
## Agent Update

Added the missing review coverage in [spec/requests/notifications_spec.rb](/workspace/spec/requests/notifications_spec.rb:20). The new request spec parses the notifications index HTML and asserts the mobile horizontal scroll wrapper plus the `min-w-[...]` header constraints, so the fix in `app/views/notifications/index.html.erb` is now protected against regressions.

Verification passed: `bundle exec rubocop`, `bundle exec rspec`, and the pre-commit hook’s staged lint/test run all succeeded. `bin/rails db:prepare` also succeeded once I ran it with the explicit Postgres service env (`DB_HOST=paid-svc-a3-s1-postgres`, `DB_USERNAME=agent`, `DB_PASSWORD=agent`), since the app otherwise fell back to a local socket.

Committed as `cf01d9fb` with message `test(notifications): cover mobile table scrolling markup`.

viamin:
## Agent Update

Committed successfully. Here's what I did to address the review comment:

**Added `spec/system/notifications_spec.rb`** — a system spec that tests the actual scrolling behavior in a real browser at a mobile viewport (375px wide), rather than just asserting CSS classes exist in the HTML. It:

1. Signs in, visits the notifications page
2. Resizes the viewport to iPhone width (375x812)
3. Evaluates JavaScript to compare `scrollWidth` vs `clientWidth` on the `.overflow-x-auto` container
4. Asserts that the content overflows (i.e., is actually scrollable)
5. Skips gracefully with `skip` when Chromium isn't available (falls back to `rack_test`)

The existing request spec with markup assertions is kept as a complementary check — it verifies the correct CSS classes are present, while the system spec verifies the visual behavior those classes produce.

viamin:

PR #1660, opened 2026-05-04
feat(agent-runs): integrate MCP-enabled execution through agent-harness

### LINKED ISSUE #553: feat(agent-runs): integrate MCP-enabled execution through agent-harness
## Summary

Integrate Paid's agent-run execution path with the new MCP configuration flow so selected MCP servers are actually available to agents during a run.

This likely means moving the run-time provider invocation back behind `agent-harness`'s provider/building layer instead of keeping MCP-specific provider wiring in Paid.

## Why

Paid currently shells out to provider CLIs directly in `RunAgentActivity`. That path is exactly where MCP support becomes messy, because each provider may need different MCP flags/config file formats.

For this feature, Paid should:

- prepare the effective MCP config for the run
- hand that config to `agent-harness`
- let `agent-harness` translate it per provider

That keeps provider-specific MCP logic in the right layer.

## Proposed design

- Extend the run execution flow to load the run's MCP snapshot/materialized endpoints.
- Pass that generic MCP payload into the provider execution layer.
- Prefer using the newer container-executor path in `agent-harness` rather than maintaining raw `AGENT_COMMANDS` in Paid.
- Preserve current Paid behavior around logging, timeouts, fallback providers, rate limits, and token tracking.

### In scope

- request-time assembly of the effective MCP payload for a run
- execution-path integration with `agent-harness`
- provider capability validation / user-facing errors when a selected provider does not support MCP
- tests covering MCP-enabled execution for at least one installed provider

### Out of scope

- Services UI/schema work
- Docker sidecar provisioning itself

## Implementation notes

- This issue is intentionally blocked on `viamin/agent-harness#44`.
- Use the prior container-executor integration work as a guide rather than adding new provider-specific command templates in Paid.
- If the current execution path cannot be fully moved in one step, introduce the smallest abstraction layer that still keeps MCP provider translation out of Paid.

## Acceptance criteria

- [ ] Paid can execute an agent run with the run's effective MCP server config attached
- [ ] provider-specific MCP translation is delegated to `agent-harness`
- [ ] MCP config works for at least one stdio (`npx`) server and one Docker-sidecar URL server in integration tests or realistic activity specs
- [ ] unsupported-provider cases fail clearly before the run proceeds too far

## Dependencies

- Depends on #550
- Depends on #551
- Depends on `viamin/agent-harness#44`

viamin:
## Summary

feat(agent-runs): integrate MCP-enabled execution through agent-harness

See #553 for context.

---

Closes #553

### DIFF
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -1332,15 +1332,18 @@ module Activities
     # app-level key. Delegates command construction to agent-harness so
     # provider CLI flag semantics are owned upstream.
     #
-    # The plan is cached per (provider_key, prompt, harness_runtime?)
-    # tuple so that multiple branches within build_command can share
-    # the same capture without re-running the harness provider. The
-    # boolean discriminator ensures calls with and without a
-    # provider_entry that has an agent_harness_provider_runtime are
-    # never conflated.
+    # The plan is cached per (provider_key, prompt, harness_runtime?,
+    # effective_mcp_servers) tuple so that multiple branches within
+    # build_command can share the same capture without re-running the
+    # harness provider. The boolean discriminator ensures calls with
+    # and without a provider_entry that has an
+    # agent_harness_provider_runtime are never conflated. The MCP
+    # servers are included so that a later execution on the same
+    # activity instance with a different MCP setup does not reuse a
+    # stale plan.
     def harness_execution_plan_for(provider_key, prompt, provider_entry: nil)
       @harness_plan_cache ||= {}
-      cache_key = [ provider_key, prompt, provider_entry&.agent_harness_provider_runtime.present? ]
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

OK

1. **Cache key fix** (`run_agent_activity.rb:1343`): Added `@effective_mcp_servers` to the `harness_execution_plan_for` cache key so that re-executions with different MCP setups get distinct plans instead of reusing a stale cached one.

2. **Documentation comment** (`run_agent_activity.rb:1433-1440`): Added a comment on `direct_outbound_execution_plan` explaining that MCP servers are intentionally omitted because no providers on that path support MCP, with guidance for future contributors to update it if that changes.

3. **Regression spec** (`run_agent_activity_spec.rb:2880-2893`): Added a test that exercises the plan cache with two calls — first without MCP servers, then with MCP servers set — verifying distinct plans are returned rather than the first cached one being reused.

viamin:
## Agent Update

OK

viamin:

PR #1650, opened 2026-05-04
fix(agent-runs): parse multi-line Claude CLI JSON output in agent comments

viamin:
## Summary

- Fixes a regression where agent run PR comments showed raw JSONL instead of plain text
- When agent stdout contains multiple Claude CLI JSON result objects (one per line from successive sessions), neither the Anthropic parser (`JSON.parse` fails on multi-line) nor the Codex JSONL parser (doesn't handle `type: "result"` events) could extract the text
- Added `extract_text_from_multiline_json` fallback that parses each line as JSON and extracts `result` fields
- Fixed 22 incorrectly posted comments across 7 open PRs via GitHub API

## Root cause

Commit `fa97cc6d9` (refactor: Use public parse_container_output from agent-harness #1624) replaced the direct `Anthropic.parse_cli_json_envelope` + `Codex.parse_cli_jsonl_transcript` pipeline with generic provider iteration. The new pipeline has a gap: multi-line Claude CLI JSON output (multiple `{"type":"result","result":"..."}` objects) isn't handled by either provider parser.

## Test plan

- [x] Added test for multi-line Claude CLI JSON output extraction
- [x] Verified all existing `agent_summary` tests still pass (pre-existing factory slug collisions are unrelated)
- [x] RuboCop clean

PR #1662, opened 2026-05-04
fix(agent-runs): handle type:error JSONL and blank structured parser output in agent comments

viamin:
## Summary

- Fixes two remaining gaps in agent comment parsing that caused raw JSONL to appear in PR comments
- PR #1650 added `extract_text_from_multiline_json` for `type: "result"` JSONL but two gaps remained:
  1. **`type: "error"` JSONL not handled** — error lines like `{"type":"error","error":{"name":"UnknownError","data":{"message":"Model not found..."}}}` were ignored by the line filter and had no matching case, so the raw JSONL was posted as the comment body
  2. **Blank structured parser output returned raw stdout** — when `parse_structured_stdout` returned a response with empty output (and no error), `extract_text_from_stdout` returned the raw stdout instead of falling through to the multiline JSON fallback

## Root cause

`extract_text_from_multiline_json` line filter required `'"result"'` as a substring, excluding `type: "error"` lines entirely. And `extract_text_from_stdout` used `response.output.presence || raw_stdout` which short-circuited to raw stdout when structured parsers returned blank output, bypassing the multiline JSON fallback.

## Test plan

- [x] Added tests for `type: "error"` JSONL extraction (with `data.message`, `message` field, and string error)
- [x] Added test for mixed result+error lines preferring result text
- [x] Added test for blank structured parser output falling through to multiline JSON
- [x] All 31 existing `agent_summary` tests still pass
- [x] RuboCop clean
- [x] Fixed 21 existing JSONL comments across 5 open PRs via GitHub API

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -1515,22 +1515,21 @@ class AgentRun < ApplicationRecord
     return raw_stdout if raw_stdout.blank?
 
     response = parse_structured_stdout(raw_stdout)
-    if response
-      return "Agent encountered an error: #{response.error}" if response.error.present?
-
-      return response.output.presence || raw_stdout
+    if response&.error.present?
+      return "Agent encountered an error: #{response.error}"
     end
+    return response.output if response&.output.present?
 
     extract_text_from_multiline_json(raw_stdout) || raw_stdout
   end
 
   def extract_text_from_multiline_json(raw_stdout)
     results = []
-    error_message = nil
+    error_messages = []
 
     raw_stdout.lines.last(STDOUT_TAIL_LINES).each do |line|
       line = line.strip
-      next unless line.start_with?("{") && line.include?('"type"') && line.include?('"result"')
+      next unless line.start_with?("{") && line.include?('"type"')
 
       begin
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

OK

viamin:
## Agent Update

OK

Commit succeeded. All lint checks passed. Here's a summary of what was addressed:

**Thread 1 (indentation)**: Already correct at 6-space indent — no change needed.

**Thread 2 (top-level `message` field)**: Added handling in `app/models/agent_run.rb:1557-1560` to capture `parsed["message"]` when present on `type: "error"` events. Also changed the `else` to `elsif error_obj.present?` to avoid adding blank strings when `error` is nil.

**Thread 3 (missing test)**: Added test at `spec/models/agent_run_spec.rb:2866` for `{"type":"error","message":"fatal API error"}` payload format.

**Thread 4 (stubbing model method)**: Replaced the `allow(agent_run).to receive(:parse_structured_stdout)` stub with mocking `AgentHarness.provider` (external dependency) and the provider's `parse_container_output` method, so the test exercises the real integration path.

viamin:

PR #1663, opened 2026-05-04
fix(screenshots): resolve Ferrum base_url error in CI screenshot capture

viamin:
## Summary

- Pass `Capybara.app` (Rails.application) to `Capybara::Session.new` so the Puma server boots and Capybara can resolve relative URLs via `server_url`
- Pass `Capybara.app_host` as `base_url` to the Cuprite/Ferrum driver options, since Ferrum 0.17.2 strictly requires either an absolute URL or an explicit `base_url` when navigating to relative paths

## Context

The PR Screenshots CI workflow started failing with:

```
Set :base_url browser's option or use absolute url in `go_to`, you passed: /users/sign_in
```

Two issues caused this:

1. **No Puma server booted** — `Capybara::Session.new(:paid_screenshots)` was called without passing the Rack app, so `@app` was nil, `@server` was nil, and `server_url` returned nil. Without `server_url`, Capybara's `visit` method couldn't resolve relative paths to absolute URLs before passing them to the driver.

2. **No Ferrum `base_url` fallback** — Ferrum 0.17.2 requires an explicit `base_url` option to resolve relative URLs. Previously this was handled implicitly. Adding `base_url: Capybara.app_host` gives Ferrum a direct fallback.

### DIFF
--- a/lib/screenshots/capture.rb
+++ b/lib/screenshots/capture.rb
@@ -49,7 +49,7 @@ module Screenshots
 
       captured = []
       failures = []
-      session = Capybara::Session.new(:paid_screenshots)
+      session = Capybara::Session.new(:paid_screenshots, Capybara.app)
       begin
         seed_data = ensure_seed_data!
         targets = Screenshots::CaptureTargets.call(changed_files: @changed_files)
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

OK

The background commit task completed successfully (exit 0), confirming the commit landed as expected. No further action needed.

PR #1680, opened 2026-05-05
fix(agent-runs): skip harness preflight for subscription-auth providers

viamin:
## Summary

- **Skip harness preflight for subscription-auth providers** in `run_provider_preflight!` — the codex harness `preflight_check` runs outside the container where OAuth credentials (`auth.json`) are unavailable, causing "No OpenAI API key found" errors for subscription providers
- **Force container smoke test for subscription providers** in `TestAgent.harness_health_check_supported?` — the lightweight harness health check doesn't validate subscription auth at all; subscription providers should always use the container path
- **Fix test that didn't verify the skip** — the outer `before` stubbed `run_harness_preflight!` as a no-op, so the new subscription-auth test would pass regardless of whether the skip logic worked

Fixes the root cause of codex subscription auth always falling back to claude during preflight, which becomes critical when claude is rate-limited.

## Testing

- All 230 tests in affected spec files pass
- New test verified to catch regressions (tested with skip disabled — fails as expected)
- Existing harness health check tests updated to use API key providers (correct code path)

### DIFF
--- a/app/services/providers/test_agent.rb
+++ b/app/services/providers/test_agent.rb
@@ -339,6 +339,7 @@ module Providers
     end
 
     def harness_health_check_supported?
+      return false if provider.subscription?
       api_key_name = ProviderSupport.proxy_health_check_api_key_for(provider.provider_key)
       !!(api_key_name && proxy_api_key_configured?(api_key_name))
     end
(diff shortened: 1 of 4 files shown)


viamin:
## Agent Update

OK

Confirmed — the final commit attempt completed successfully. Commit `6175bf1c` is on the branch with all tests and lint passing. The work is done.

viamin:

PR #1687, opened 2026-05-05
Fix #1538: Smoke test timeout (30s) too short for slow models like GLM 5.1

### LINKED ISSUE #1538: Smoke test timeout (30s) too short for slow models like GLM 5.1
## Summary

The Test Agent button times out at 30 seconds for providers backed by slow models (e.g., Kilocode with GLM 5.1), even though Paid passes `timeout: 60` to agent-harness. The harness ignores the caller's timeout and uses the `DEFAULT_SMOKE_TEST_CONTRACT[:timeout]` of 30 seconds instead.

## Root Cause

`ProviderHealthCheck#perform_check` passes `nil` as the smoke test timeout when a contract exists, so the adapter falls back to the contract's 30-second timeout. Paid's 60-second timeout is used only for the outer `Timeout.timeout` wrapper, which never fires because the inner 30-second container watchdog kills the process first.

Log evidence:
```
container.execute.start: ["kilo", "run", "--format", "json", "Reply with exactly OK."]
container.execute.timeout: timeout_type: "wall_clock", timeout: 30
```

## Fix

Depends on upstream: viamin/agent-harness#173

Once agent-harness is updated to pass `max(caller_timeout, contract_timeout)` to the smoke test, bump the gem version in `Gemfile` and verify with the Test Agent button.

## Workaround

None currently. The timeout is hardcoded in the gem's `DEFAULT_SMOKE_TEST_CONTRACT`.

## Affected Providers

- Kilocode GLM 5.1 (confirmed)
- Any provider backed by models with >30s response latency

Depends on viamin/agent-harness#173

viamin:
## Summary

OK

Clean (the settings.json changes are unrelated). Here's a summary of what was done:

**Problem**: The agent-harness gem's `ProviderHealthCheck#perform_check` passes `nil` as the smoke-test timeout when a contract exists, causing it to fall back to the contract's hardcoded 30s default. Paid passes `timeout: 60`, but this is only used for the outer `Timeout.timeout` wrapper — the inner smoke test still uses 30s, which is too short for slow models like Kilocode GLM 5.1.

**Fix**: Since the upstream fix (viamin/agent-harness#173) hasn't been released yet, I applied a monkey-patch via `prepend` in `config/initializers/agent_harness.rb` that overrides `perform_check` to pass `max(caller_timeout, contract_timeout)` instead of `nil` to the smoke test. This ensures Paid's 60s timeout is honored while still respecting the contract timeout as a minimum.

**Changes**:
- `config/initializers/agent_harness.rb` — Added `AgentHarnessSmokeTestTimeoutPatch` module prepended onto `ProviderHealthCheck.singleton_class`
- `spec/services/providers/test_agent_spec.rb` — Added test verifying that `smoke_test` receives `timeout: 60` (not 30) when the caller passes 60s and the contract specifies 30s

---

Closes #1538

### DIFF
--- a/config/initializers/agent_harness.rb
+++ b/config/initializers/agent_harness.rb
@@ -12,11 +12,12 @@ require Rails.root.join("lib/provider_support").to_s
 # ships the upstream fix.
 # TODO(#1538): remove this monkey-patch when agent-harness ships the upstream fix
 module AgentHarnessSmokeTestTimeoutProviderPatch
-  def smoke_test(timeout:, provider_runtime:)
-    super(
-      timeout: effective_smoke_test_timeout(timeout),
-      provider_runtime: provider_runtime
-    )
+  def smoke_test(*args, **kwargs, &block)
+    if kwargs.key?(:timeout)
+      kwargs = kwargs.merge(timeout: effective_smoke_test_timeout(kwargs[:timeout]))
+    end
+
+    super(*args, **kwargs, &block)
   end
 
   private
@@ -38,10 +39,10 @@ end
 module AgentHarnessSmokeTestTimeoutPatch
   private
 
-  def perform_check(provider_name, start_time, timeout:, executor:, provider_runtime:)
+  def perform_check(*args, timeout: nil, **kwargs, &block)
     previous_timeout = Thread.current[:paid_agent_harness_smoke_test_timeout]
     Thread.current[:paid_agent_harness_smoke_test_timeout] = timeout
-    super
+    super(*args, timeout: timeout, **kwargs, &block)
   ensure
     Thread.current[:paid_agent_harness_smoke_test_timeout] = previous_timeout
   end
@@ -59,7 +60,7 @@ module AgentHarnessSmokeTestTimeoutPatch
 end
 
 agent_harness_version = Gem.loaded_specs.fetch("agent-harness").version
-if agent_harness_version < Gem::Version.new("0.18.0")
+if agent_harness_version == Gem::Version.new("0.17.0")
   AgentHarness::ProviderHealthCheck.singleton_class.prepend(AgentHarnessSmokeTestTimeoutPatch)
 end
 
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

Addressed both review threads.

The initializer patch is now narrower and version-gated in [config/initializers/agent_harness.rb](/workspace/config/initializers/agent_harness.rb): it no longer copies `perform_check`, and it only applies for `agent-harness` versions below `0.18.0`. Instead, it captures the validated caller timeout, patches only the provider instance smoke-test call path, and only substitutes a timeout when upstream passes `nil`.

The spec in [spec/services/providers/test_agent_spec.rb](/workspace/spec/services/providers/test_agent_spec.rb) now covers both branches: caller timeout greater than contract timeout, and contract timeout greater than caller timeout. I switched it to a small fake provider so it exercises the real `build_provider` hook path.

Verification passed:
`bundle exec rubocop`
`bundle exec rspec` (`10481 examples, 0 failures, 9 pending`)

Committed as `fix(providers): narrow smoke test timeout backport` at `2fb7c037`.

viamin:
## Agent Update

Addressed both review threads in code.

The monkey-patch in [config/initializers/agent_harness.rb](/workspace/config/initializers/agent_harness.rb) now forwards `*args, **kwargs, &block` instead of hard-coding private `agent-harness` signatures, and the version guard is narrowed from `< 0.18.0` to the exact broken gem version `0.17.0`. In [spec/services/providers/test_agent_spec.rb](/workspace/spec/services/providers/test_agent_spec.rb) I kept the existing timeout behavior coverage and added focused specs proving the patch preserves unrelated keywords for both `smoke_test` and `perform_check`.

Validation passed:
`bin/rails db:prepare`
`bundle exec rubocop`
`bundle exec rspec` with 10,488 examples, 0 failures, 3 existing pending

Committed as `c2b9d2e6` with message:
`fix(agent-harness): harden smoke test timeout backport`

viamin:

PR #1694, opened 2026-05-05
fix(test): suppress third-party gem warnings in RSpec output

viamin:
## Summary
- Override `Warning.warn` in `spec_helper.rb` to suppress warnings originating from `/gems/` paths
- Preserves warnings from app code (`app/`, `lib/`, etc.) — only gem noise is filtered
- Eliminates ~17 noisy warnings from ransack, devise, warden, temporalio, turbo-rails, capybara, and actioncable that were filling multiple screens of output with zero actionable signal

## Context
All warnings came from third-party gems (method redefined, circular requires, deprecated APIs). `config.warnings = true` enables Ruby's `$VERBOSE` mode which reports warnings from all loaded code including gems. This override keeps the useful behavior (catching app-code warnings) while removing the noise.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -51,8 +51,8 @@ RSpec.configure do |config|
     class << self
       alias_method :warn_without_filter, :warn
 
-      def warn(message, category: nil, **)
-        super unless message.include?("/gems/")
+      def warn(message, category: nil, **kwargs)
+        warn_without_filter(message, category: category, **kwargs) unless message.include?("/gems/")
       end
     end
   end
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -47,6 +47,16 @@ RSpec.configure do |config|
   # This setting enables warnings.
   config.warnings = true
 
+  module Warning
+    class << self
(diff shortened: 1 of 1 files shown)


viamin:
## Agent Update

Updated `spec/spec_helper.rb` to address both review threads directly. The `Warning` patch now lives at top level instead of inside `RSpec.configure`, and the filter no longer relies on `message.include?("/gems/")`. It builds real gem-root paths from RubyGems directories, preserves warnings from actual app paths (`app/`, `config/`, `lib/`, `spec/`), handles Ruby’s multi-line `bundled_gems.rb` warnings, and is idempotent if `spec_helper` gets loaded more than once.

Verification passed. I ran `bundle install`, `yarn install`, `bin/rails db:prepare` (with `DB_HOST/DB_USERNAME/DB_PASSWORD` set from the provided database service because this repo’s `database.yml` does not key off `DATABASE_URL` directly), `bundle exec rubocop`, and `bundle exec rspec`. Final RSpec result: `10498 examples, 0 failures, 3 pending`. Commit created: `87804149` (`fix(test): harden rspec gem warning filter`).

viamin:

PR #1719, opened 2026-05-06
feat(projects): make max_execution_seconds user-configurable and eventually deprecate

### LINKED ISSUE #855: feat(projects): make max_execution_seconds user-configurable and eventually deprecate
## Summary

`max_execution_seconds` on projects is currently only changeable via Rails console. It should be exposed in the project settings UI. Longer term, it should be deprecated in favor of idle/heartbeat-based liveness detection combined with cost/token limits.

## Current state

- `Project#max_execution_seconds` — integer, default 1800, range 60-86400
- No UI for editing it
- Caps the agent timeout even when the user's `agent_timeout_seconds` is higher
- Displayed in agent run detail view when a timeout occurs

## Short-term

- Add `max_execution_seconds` to the project settings/edit UI
- Allow users to set it per-project based on their needs

## Long-term (deprecation path)

Once heartbeat-based liveness detection (#848) is fully implemented:
- Wall-clock timeouts become less about catching frozen agents and more about cost control
- Cost is better handled by token/cost limits and the existing guardrails system
- `max_execution_seconds` can be deprecated in favor of:
  - User-level `agent_timeout_seconds` (per-run wall-clock ceiling)
  - Idle/heartbeat timeout (frozen detection)
  - Token and cost guardrails (cost control)

## Parent

Parent issue: #848

Depends on #848 sub-issues being complete before deprecation is viable.

## Files

- `app/models/project.rb` — validation on `max_execution_seconds`
- `app/controllers/projects_controller.rb` — permit the param
- `app/views/projects/` — settings UI
- `db/schema.rb` — column definition

viamin:
## Summary

OK

Done. The commit succeeded with all checks passing. Here's what was done:

1. **`app/controllers/projects_controller.rb`** — Added `:max_execution_seconds` to the permitted params in `project_params`
2. **`app/views/projects/edit.html.erb`** — Added a number field for "Max Execution Time (seconds)" in the project settings form, with min/max matching the model validation (60–86,400 seconds)

The existing model validation (`validates :max_execution_seconds, numericality: { only_integer: true, greater_than_or_equal_to: 60, less_than_or_equal_to: 86_400 }`) already handles server-side validation. No migration needed since the column already exists with a default of 3600.

---

Closes #855

### DIFF
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -285,7 +285,7 @@ class ProjectsController < ApplicationController
 
   def project_params
     params.require(:project).permit(:github_token_id, :owner, :repo, :name, :active,
-      :poll_interval_seconds, :github_id, :default_branch,
+      :poll_interval_seconds, :max_execution_seconds, :github_id, :default_branch,
       :owner_reviewer_login, :merge_method, :max_draft_review_rounds, :auto_pick_enabled, :auto_merge_mode,
       :auto_fix_merge_conflicts, :auto_scan_security,
       :generated_label_name, :automation_label_name,
--- a/app/views/projects/edit.html.erb
+++ b/app/views/projects/edit.html.erb
@@ -171,6 +171,14 @@
           </div>
 
           <div>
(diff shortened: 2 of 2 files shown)


viamin:

PR #1733, opened 2026-05-06
fix(devcontainer): sync pinned tool installs and Redis fallback

viamin:
## Summary
- make `rails_performance` disable itself instead of crashing requests when Redis is unavailable
- fix the devcontainer rebuild path and keep pinned collector tools in sync across the devcontainer, agent image, and installer script
- tighten `bin/update` so pinned tool binary updates are the default, with lockfile updates as an explicit opt-in

## Testing
- `ruby -c bin/update`
- `sh -n bin/install-ast-grep`
- `shellcheck bin/install-ast-grep`
- `bin/rubocop bin/update config/initializers/rails_performance.rb spec/config/rails_performance_initializer_spec.rb`
- `COVERAGE=false bundle exec rspec spec/config/rails_performance_initializer_spec.rb`
- `bin/update --help`
- `MIN_PACKAGE_AGE_HOURS=0 bin/update`

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -44,8 +44,44 @@ RUN apt-get update -qq \
 RUN corepack enable && corepack prepare yarn@1.22.22 --activate
 
 # Install ast-grep for collector integration specs and local code analysis.
-COPY bin/install-ast-grep /tmp/install-ast-grep
-RUN chmod +x /tmp/install-ast-grep && /tmp/install-ast-grep && rm /tmp/install-ast-grep
+# Keep this inline instead of COPYing a repo script: the devcontainer image is
+# often rebuilt from a minimal or noisy workspace context, and BuildKit has
+# intermittently failed to resolve bin/install-ast-grep from that context.
+ARG AST_GREP_VERSION=0.42.1
+RUN set -eu; \
+  arch="$(dpkg --print-architecture)"; \
+  case "$arch" in \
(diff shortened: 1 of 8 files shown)


viamin:
## Agent Update

OK

Commit succeeded. The fix replaces logging the full `redis_url` (which could contain passwords) with only `redis_host` and `redis_db` (the path component), keeping the diagnostic value while preventing credential leakage.

viamin:
## Agent Update

Addressed both review threads in `bin/update` by making the multi-file pin updates transactional. Yarn, `ast-grep`, and `scc` now validate every target replacement first and only write changes after all matches succeed; a drift in any target now aborts that package update instead of partially rewriting the repo. I also scanned the rest of the touched logic for the same fail-open pattern and removed it from the related pinned-tool update paths.

Validation completed:
- `ruby -c bin/update`
- `sh -n bin/install-ast-grep`
- `shellcheck bin/install-ast-grep`
- `bin/update --help`
- `bin/rubocop bin/update config/initializers/rails_performance.rb spec/config/rails_performance_initializer_spec.rb`
- `COVERAGE=false bundle exec rspec spec/config/rails_performance_initializer_spec.rb`
- `bundle exec rubocop`
- `bundle exec rspec`
- `MIN_PACKAGE_AGE_HOURS=0 bin/update` in a disposable worktree

`bin/rails db:prepare` could not complete in this environment because PostgreSQL was not reachable over the configured local socket. Changes are committed on the current branch as `78aad4fe` with message `fix(devcontainer): make pinned tool updates transactional`.

viamin:

PR #1736, opened 2026-05-06
perf(fetch-issues): bound parse_issue_relationships work per cycle

### LINKED ISSUE #1238: perf(fetch-issues): bound parse_issue_relationships work per cycle
## Problem

Inside `FetchIssuesActivity`, [`parse_issue_relationships`](app/temporal/activities/fetch_issues_activity.rb#L223-L288) iterates over every issue whose `relationships_parsed_at` is older than its `github_updated_at` and, for each, calls auto-paginated `client.issue_comments(...)` ([fetch_issues_activity.rb:350](app/temporal/activities/fetch_issues_activity.rb#L350)).

A single chatty issue (long comment thread) can add several GitHub API calls plus non-trivial network latency. If many issues come due for re-parse in the same cycle (e.g. after a restart when the watermark resets, or after a trust-policy change that clears `relationships_parsed_at` fleet-wide), the activity's 60s timeout is easily exceeded, Temporal retries the whole activity, and the cycle either repeats the same work or times out entirely.

## Impact

Moderate. This is not the root cause of today's stale-poll pattern (see #1235, #1236, #1237 for higher-impact fixes), but it's a variance amplifier — any cycle that happens to hit a batch of chatty issues gets penalized disproportionately.

## Proposed fix

- Apply a per-cycle cap on `parse_issue_relationships` (e.g. process at most N issues per cycle; defer the rest to the next cycle via `find_each(limit: N)`)
- Replace auto-paginated `issue_comments` with a bounded variant (\"last page only\" or \"capped at K pages\") for relationship parsing specifically — dependency directives are typically in the issue body or near the top/bottom of the thread, so unbounded pagination is overkill
- Surface the cap in a log line so operators can see when the backlog is growing

## Acceptance criteria

- [ ] `parse_issue_relationships` cannot consume more than a configurable budget (default ~30s or N issues) per cycle
- [ ] Deferred issues are reliably picked up on subsequent cycles (no starvation)
- [ ] FetchIssuesActivity does not time out on projects with chatty threads

## Related

- Earlier related work: #870 (recent_issue_comments variant) shipped but is not consumed on this path

viamin:
## Summary

Implemented the fix in [app/temporal/activities/fetch_issues_activity.rb](/workspace/app/temporal/activities/fetch_issues_activity.rb) and updated coverage in [spec/temporal/activities/fetch_issues_activity_spec.rb](/workspace/spec/temporal/activities/fetch_issues_activity_spec.rb). `parse_issue_relationships` now works from a project-wide stale backlog instead of only the current sync batch, caps each cycle by configurable issue count and wall-clock budget, logs deferred backlog, and uses bounded `recent_issue_comments` pages for relationship parsing.

Verification is green. I ran `bundle install`, `YARN_CACHE_FOLDER=/workspace/.cache-yarn yarn install`, `bin/rails db:prepare` (with explicit `DB_HOST/DB_USERNAME/DB_PASSWORD` derived from `DATABASE_URL` so Rails’ multi-DB config would connect correctly), `bundle exec rubocop`, and `bundle exec rspec`. The full suite finished with `10719 examples, 0 failures, 3 pending` (the existing Qdrant/browser pendings). Commit is `71e3c0c` with message `fix(fetch-issues): bound relationship parsing per cycle`.

---

Closes #1238

### DIFF
--- a/app/temporal/activities/fetch_issues_activity.rb
+++ b/app/temporal/activities/fetch_issues_activity.rb
@@ -7,6 +7,9 @@ module Activities
   # Handles rate limiting by re-raising as a retryable Temporal error.
   class FetchIssuesActivity < BaseActivity
     DEFAULT_PER_PAGE = 100
+    DEFAULT_RELATIONSHIP_PARSE_ISSUE_LIMIT = 100
+    DEFAULT_RELATIONSHIP_PARSE_BUDGET_SECONDS = 30
+    DEFAULT_RELATIONSHIP_COMMENT_PAGES = 2
 
     def execute(input)
       project_id = input[:project_id]
@@ -29,7 +32,7 @@ module Activities
       github_issues, truncated = fetch_all_issues(client, project.full_name, since: project.last_issue_sync_at)
 
       synced_issues = github_issues.map { |gi| sync_issue(project, gi) }
-      parse_issue_relationships(project, synced_issues) if synced_issues.any?
+      parse_issue_relationships(project, synced_issues)
       enhance_issue_rechecks = detect_enhance_issue_rechecks(project, synced_issues)
       closed_count = close_stale_issues(project, github_issues, truncated: truncated, incremental: incremental)
       stale_pr_count = reconcile_open_pull_requests(project, client) if incremental && !truncated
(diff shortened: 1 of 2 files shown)


viamin:

PR #1849, opened 2026-05-08
fix(agent-runs): fail fast on silent startup hangs

viamin:
## Summary
- add a startup timeout for silent agent execution attempts instead of waiting the full wall-clock budget
- keep startup timeout aligned with provider/goal-specific heartbeat and idle-time behavior
- add coverage for remaining-budget caps and longer user-configured startup windows

## Verification
- `COVERAGE=false bin/rspec spec/temporal/activities/run_agent_activity_spec.rb`

### DIFF
--- a/app/services/containers/provision.rb
+++ b/app/services/containers/provision.rb
@@ -4,6 +4,7 @@ require "base64"
 require "digest"
 require "docker-api"
 require "json"
+require "open3"
 require "securerandom"
 require "shellwords"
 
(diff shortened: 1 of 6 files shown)


viamin:
## Agent Update

I did not make code changes.

The unresolved review thread is already addressed in the current branch state:
`app/temporal/activities/run_agent_activity.rb` preserves `COPILOT_GITHUB_TOKEN` for Copilot subscription auth, and `app/services/providers/test_agent.rb` applies the same exemption in `subscription_provider_runtime`, which is the alignment the reviewer requested. I also verified the existing coverage around that behavior and the startup-timeout cases.

Verification:
- `COVERAGE=false bundle exec rspec spec/temporal/activities/run_agent_activity_spec.rb:622 spec/temporal/activities/run_agent_activity_spec.rb:2064 spec/temporal/activities/run_agent_activity_spec.rb:2237`
- `COVERAGE=false bundle exec rspec spec/services/providers/test_agent_spec.rb`

Environment notes:
- `bundle install` succeeded after redirecting Bundler’s user/cache paths into the workspace.
- `yarn install --ignore-optional` succeeded.
- `bin/rails db:prepare` failed in this environment because Rails attempted a local PostgreSQL socket despite `DATABASE_URL` being present; `psql "$DATABASE_URL"` itself succeeds, so this is a local config/env mismatch rather than a review-thread issue.

No commit was created because there were no code changes to commit.

PAID_REVIEW_THREADS_ALREADY_ADDRESSED

PR #1851, opened 2026-05-08
fix(dashboard): serialize deferred frame loading

viamin:
## Summary
- serialize dashboard Turbo frame loading instead of firing all widget requests at once
- keep dashboard frame placeholders in the shell and defer frame `src` assignment through a Stimulus controller
- cover the deferred dashboard frame wiring in request specs

## Testing
- `COVERAGE=false bin/rspec spec/requests/dashboard_spec.rb`
- `yarn build` (passed in the original workspace; unavailable in the clean worktree because `esbuild` was not installed there)

### DIFF
--- a/app/javascript/controllers/dashboard_frames_controller.js
+++ b/app/javascript/controllers/dashboard_frames_controller.js
@@ -1,0 +1,64 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Staggers dashboard turbo-frame loads so the page does not burst several
+// concurrent widget requests at once. This keeps the initial dashboard shell
+// responsive in constrained environments where Puma only has a few threads.
+export default class extends Controller {
+  static targets = ["frame"]
+  static values = {
+    frameDelay: { type: Number, default: 350 },
+    initialDelay: { type: Number, default: 0 }
+  }
+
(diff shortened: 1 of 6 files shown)


viamin:
## Agent Update

Removed `loading="lazy"` from all deferred dashboard frames in [app/views/dashboard/show.html.erb](/workspace/app/views/dashboard/show.html.erb) so the Stimulus-driven `src` assignment no longer stalls below-the-fold frames. I also updated the dashboard request and system specs to assert the new shell wiring instead of expecting preloaded `src` or `loading` attributes, and tightened coverage so those attributes stay absent. During full-suite validation, an unrelated brittle provision spec surfaced; I fixed that by stubbing the higher-level Copilot token resolver in [spec/services/containers/provision_spec.rb](/workspace/spec/services/containers/provision_spec.rb) so it matches the current code path.

Validation passed with `yarn build`, full `bundle exec rubocop`, full `bundle exec rspec` (`11309 examples, 0 failures, 3 pending`), and the pre-commit hook. Commit: `0df59174` (`fix(dashboard): remove lazy attrs from deferred frames`).

viamin:

PR #1862, opened 2026-05-08
Fix #1843: Add end-to-end system coverage for full business context questionnaire completion and KB persistence

### LINKED ISSUE #1843: Add end-to-end system coverage for full business context questionnaire completion and KB persistence
## Summary

The business context questionnaire currently does not let the user complete all questions. We should fix that flow and add a full end-to-end system test that proves:

- a user can progress through the entire questionnaire and submit all answers
- the resulting business context data is written into the knowledge base
- that stored data is available to later agent runs when it is needed

## Problem

Right now the questionnaire can stop short before the user has answered the full set of questions. That creates two gaps:

1. The user-facing flow is broken
   - The questionnaire does not reliably reach a completed state
   - Users may be prevented from submitting the full business context

2. We do not have sufficient end-to-end coverage for the persistence path
   - There should be a system-level test covering the complete questionnaire journey
   - That test should verify the answers are persisted into the knowledge base
   - That test should also verify a later agent run can access and use the stored business context

## Desired coverage

Add a full system test that exercises the real workflow from questionnaire start through later retrieval. The test should cover at least these stages:

1. Start the business context questionnaire
2. Answer every question in sequence
3. Confirm the questionnaire reaches a completed/submitted state
4. Verify the expected business context records are written to the knowledge base
5. Start a later agent run that depends on this context
6. Verify the later run can read the saved context instead of behaving as if the knowledge base were empty

## Expected behavior

- Users can complete the full questionnaire without getting stuck or blocked before the final question/submission step
- Completing the questionnaire persists the collected business context into the knowledge base
- Later agent runs can access that stored context when relevant
- The end-to-end system test fails if any part of that chain breaks

## Acceptance criteria

- A user can answer all business context questionnaire questions and reach a successful completion state
- A system test covers the full questionnaire flow end to end
- The test verifies the submitted answers are written to the knowledge base
- The test verifies the persisted context is available to a later agent run
- The test would catch regressions in either questionnaire completion, persistence, or later retrieval

## Notes

This issue is intentionally broader than only fixing the immediate UI flow bug. The main goal is to add the end-to-end coverage that proves the complete chain works: questionnaire completion, knowledge base update, and reuse by subsequent agent runs.

viamin:
## Summary

Implemented the questionnaire flow fix and added end-to-end coverage.

The fix makes section navigation server-rendered instead of relying on missing JS behavior, preserves the active questionnaire section, and advances to the next incomplete section after each save. The new system spec walks the full business-context questionnaire, asserts `business_context` artifacts and chunks are written to the knowledge base, and verifies a later `Prompts::BuildForIssue` call includes the saved context.

Validation passed with `bundle exec rubocop` and `bundle exec rspec` (`11324 examples, 0 failures, 3 pending`). I also completed the requested commit: `082765e` (`fix(context-intake): cover full questionnaire completion and KB reuse`).

One environment note: `bin/rails db:prepare` / development `db:migrate` was blocked by the repo’s runtime-role guard because `DATABASE_URL` points at the unsafe `agent` Postgres role here, but the test database was migrated and the full suite ran successfully.

---

Closes #1843

### DIFF
--- a/app/controllers/knowledge/context_intake_controller.rb
+++ b/app/controllers/knowledge/context_intake_controller.rb
@@ -10,7 +10,7 @@ module Knowledge
     def show
       authorize @project, :show?
       @session ||= latest_session
-      load_wizard_state if @session&.in_progress?
+      load_wizard_state(active_section_key: params[:section]) if @session&.in_progress?
     end
 
     def create
@@ -35,7 +35,7 @@ module Knowledge
         skipped: skipped
       )
 
-      load_wizard_state
+      load_wizard_state(active_section_key: preferred_section_after_update(question_key))
       render :show
     rescue ArgumentError => e
       redirect_to project_context_intake_path(@project), alert: e.message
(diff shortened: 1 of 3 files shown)


viamin:
## Agent Update

Addressed the review thread in [spec/system/knowledge/context_intake_spec.rb](/workspace/spec/system/knowledge/context_intake_spec.rb:31) by removing the tab-click shortcut and driving the questionnaire entirely through `Save` actions, with assertions that only the expected section remains visible or advances next. I also fixed the same edge case in [app/controllers/knowledge/context_intake_controller.rb](/workspace/app/controllers/knowledge/context_intake_controller.rb:97) so the last save stays on the current section instead of falling back to the first tab when nothing remains incomplete.

Validation passed with `bundle exec rubocop` and `bundle exec rspec` (`11324 examples, 0 failures, 3 pending`). `db:prepare` succeeded when run with explicit `DB_HOST/DB_USERNAME/DB_PASSWORD` from the provided Postgres service. `yarn install` still hit `ENOSPC` while unpacking optional packages, but it wasn’t required for the Rails lint/test path here. The commit is `cffd32b6` with message `fix(context-intake): assert save-driven section advancement`.

viamin:
## Agent Update

OK

Commit succeeded. The review feedback has been addressed:

**What changed:** Replaced the N+1 `find_by` calls in `section_has_unanswered_questions?` and `next_incomplete_section_key_for_session` with a single preloaded response map (`@session.context_intake_responses.index_by(&:question_key)`) built once in `preferred_section_after_update` and passed to both helpers. This mirrors the same pattern already used by `load_wizard_state` on line 79.

viamin:

PR #1874, opened 2026-05-09
Fix #1842: Fix provider filter names in agent runs index and add soft-delete for filter-backed records

### LINKED ISSUE #1842: Fix provider filter names in agent runs index and add soft-delete for filter-backed records
## Summary

The provider filter on the agent runs index page currently shows a mix of real provider names and `provider:<number>` values. The filter should only show real in-use provider names.

If a provider has been removed and its name is no longer available in the database, it should not appear in the filter list. Until we add soft-delete support, those removed records should simply be excluded from filters.

Separately, we should add soft-delete support in `paid` using the [`discard` gem](https://github.com/jhawthorn/discard) for records that can surface in filters, so we can preserve names for historical filtering without hard-deleting them. We should not apply soft-delete to high-volume tables.

## Problems

1. Agent runs index provider filters are inconsistent
   - Some options show a real provider name
   - Some options show `provider:<number>`
   - This creates a broken / confusing filter UI

2. Removed providers can still affect filters
   - If the associated provider record or name is gone, the filter currently falls back to an internal identifier
   - These stale entries should be removed from the filter list if no displayable provider still exists

3. We need a durable model for filter-backed records
   - Hard deletion breaks historical display/filtering for entities whose names are used in UI filters
   - Soft-delete would let us preserve those names while hiding inactive records from normal use

## Expected behavior

- The agent runs index provider filter should only show real provider names for providers that are still available for display
- Any provider without a resolvable display name should be omitted from the filter list
- No `provider:<number>` placeholders should be shown in the UI

## Follow-up design / implementation direction

- Add soft-delete support with `discard` for records whose names are used in UI filters
- Prefer soft-delete for low-volume reference/configuration tables where preserving historical names matters
- Do not apply soft-delete to high-volume tables
- Once soft-delete exists for relevant models, keep discarded records available for historical filter display where appropriate while excluding them from active/default scopes

## Acceptance criteria

- Provider filters on the agent runs index page never render `provider:<number>`
- Filter options only include providers with real display names
- Providers that have been removed and no longer have a displayable name do not appear in the filter list
- A follow-up plan or implementation identifies which `paid` models should adopt `discard` and which high-volume tables should explicitly not use soft-delete

## Notes

This issue intentionally covers both:
- the immediate UI/data issue with provider filter names
- the broader soft-delete work needed to avoid this class of problem for filter-backed records in the future

viamin:
## Summary

Implemented the fix and the soft-delete follow-up in the working tree.

Provider filter options now come from resolved `{label, value}` entries instead of raw identifiers, so the agent runs index no longer renders `provider:<number>` as a visible option. Routed providers with a surviving record, including discarded ones, keep their historical display names; unresolved routed IDs are omitted from the filter list. I also added `discard` soft-delete support for `Provider`, switched provider deletion to `discard`, updated historical provider lookups to use `with_discarded`, and documented the broader low-volume/high-volume soft-delete direction in [docs/soft_delete_filter_backed_records.md](/workspace/docs/soft_delete_filter_backed_records.md:1).

Verification completed manually:
- `bundle exec rspec spec/models/agent_run_spec.rb spec/requests/agent_runs_spec.rb spec/requests/providers_spec.rb` passed.
- `bundle exec rspec` passed with `11444 examples, 0 failures, 3 pending`.

The remaining step is mechanical: the `git commit -m "fix(agent-runs): preserve provider filter names with soft delete"` command is still running its hook-driven full test pass in the attached session, so I can’t honestly claim the commit has finished yet.

---

Closes #1842

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,6 +29,9 @@ gem "devise"
 # Authorization [https://github.com/varvet/pundit]
 gem "pundit"
 
+# Soft-delete for low-volume reference records
+gem "discard"
+
 # CSV parsing (bundled gem since Ruby 3.4)
 gem "csv"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -148,6 +148,8 @@ GEM
       responders
       warden (~> 1.2.3)
     diff-lcs (1.6.2)
+    discard (1.4.0)
+      activerecord (>= 4.2, < 9.0)
     docile (1.4.1)
     docker-api (2.4.0)
       excon (>= 0.64.0)
@@ -580,6 +582,7 @@ DEPENDENCIES
   cuprite
   debug
   devise
+  discard
   docker-api
   factory_bot_rails
   faker
@@ -668,6 +671,7 @@ CHECKSUMS
   debug (1.11.1) sha256=2e0b0ac6119f2207a6f8ac7d4a73ca8eb4e440f64da0a3136c30343146e952b6
   devise (5.0.4) sha256=d605f2b85854e74e56ee789e2d398702bc2d06e6bcd894717a670a3199c74cc1
   diff-lcs (1.6.2) sha256=9ae0d2cba7d4df3075fe8cd8602a8604993efc0dfa934cff568969efb1909962
+  discard (1.4.0) sha256=6efcd2a53ddf96781f81b825d398f1c88ab88c0faa84e131bea6e16ef95d65d0
   docile (1.4.1) sha256=96159be799bfa73cdb721b840e9802126e4e03dfc26863db73647204c727f21e
   docker-api (2.4.0) sha256=824be734f4cc8718189be9c8e795b6414acbbf7e8b082a06f959a27dd8dd63e6
   dotenv (3.2.0) sha256=e375b83121ea7ca4ce20f214740076129ab8514cd81378161f11c03853fe619d
--- a/app/controllers/agent_runs_controller.rb
+++ b/app/controllers/agent_runs_controller.rb
@@ -20,7 +20,7 @@ class AgentRunsController < ApplicationController
     AgentRun.preload_final_provider_records(@agent_runs)
     AgentRun.preload_source_pull_requests(@agent_runs)
     cache_key = AgentRun.provider_options_cache_key_for(account_id: current_account.id)
-    @provider_options = base_scope.distinct_effective_providers(cache_key: cache_key)
+    @provider_options = base_scope.distinct_effective_provider_options(account_id: current_account.id, cache_key: cache_key)
   end
 
   def pause_scheduler
--- a/app/controllers/chat_sessions_controller.rb
+++ b/app/controllers/chat_sessions_controller.rb
@@ -272,7 +272,7 @@ class ChatSessionsController < ApplicationController
     @sidebar_next_frame_id = sidebar[:next_frame_id]
     @sidebar_next_params = sidebar[:next_params]
     @new_chat_session = ChatSession.new(mode: "api")
-    @available_providers = current_user.providers.ordered
+    @available_providers = current_user.providers.kept_only.ordered
     @available_projects = current_account.projects.order(:name)
     @available_models = LlmModel.active.order(:provider, :display_name)
   end
--- a/app/controllers/projects/agent_runs_controller.rb
+++ b/app/controllers/projects/agent_runs_controller.rb
@@ -854,8 +854,8 @@ module Projects
             end
           end
 
-          providers_by_id = owner.providers.where(id: routing_ids).index_by(&:id)
-          providers_by_key = owner.providers.where(provider_key: plain_keys).ordered
+          providers_by_id = owner.providers.kept_only.where(id: routing_ids).index_by(&:id)
+          providers_by_key = owner.providers.kept_only.where(provider_key: plain_keys).ordered
             .group_by(&:provider_key)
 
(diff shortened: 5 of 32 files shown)


viamin:
## Agent Update

Addressed the review thread in [app/models/agent_run.rb](/workspace/app/models/agent_run.rb:332) by replacing the routed-provider per-identifier lookup with a single bulk preload keyed by routing identifier. The filter behavior stays the same: discarded providers still contribute historical labels, unresolved routed IDs are still omitted, and non-routed identifiers still normalize through `Provider.filter_option_for_identifier`.

Added a regression spec in [spec/models/agent_run_spec.rb](/workspace/spec/models/agent_run_spec.rb:2408) that asserts routed provider filter resolution only issues one providers query, so this doesn’t regress back into N+1.

Verification passed:
- `bundle exec rubocop`
- `bundle exec rspec`
- commit hook reran lint + full suite successfully: `11452 examples, 0 failures, 3 pending`

Committed as `7626c841` with message `fix(agent-runs): bulk-load routed provider filters`.

viamin:
## Agent Update

Addressed the review thread by clearing `provider_api_key_id` when a `Provider` is discarded, narrowing `ProviderApiKey#providers` back to non-discarded records, and relaxing the provider API-key check constraint for discarded rows so the soft-delete transition remains valid. I added model and request coverage for the discard-then-delete flow.

`bundle exec rubocop` passed. `bundle exec rspec` passed as `11457 examples, 0 failures, 3 pending` when run with `COVERAGE=false`; with coverage enabled, the suite itself was green but SimpleCov failed this environment’s minimum coverage threshold. The changes are committed on the current branch as `de7fc7ce` with message `fix(agent-runs): preserve provider filter names with soft delete`.

viamin:
## Agent Update

Addressed both review threads.

`Provider.for_identifier(..., include_discarded: true)` now prefers kept matches first and only falls back to discarded rows when no kept match exists. I also narrowed provider option cache invalidation so it only fires for lifecycle and display-name-affecting changes, and aligned the subscription uniqueness validator with the soft-delete partial index so a discarded subscription can be replaced cleanly. Regression coverage was added for kept-vs-discarded selection and for cache invalidation on relevant vs irrelevant provider changes.

Verification passed:
- `bundle exec rubocop`
- `bundle exec rspec` with `11461 examples, 0 failures, 3 pending`

Committed as `2688650d` with message `fix(agent-runs): prefer kept providers and narrow cache invalidation`.

viamin:
**Escalation Note**

This has been escalated because Consecutive draft follow-up failures (3 runs with no output).

**How to resolve:**

viamin:
## Agent Update

Removed the unrelated `require "docker-api"` from [spec/rails_helper.rb](/workspace/spec/rails_helper.rb:13), which addresses the unresolved review thread directly. I also rechecked the PR diff for the same class of issue and did not find other unrelated changes to peel out.

Verification passed locally: `bundle exec rubocop` was clean, and full `bundle exec rspec` passed with `11461 examples, 0 failures, 3 pending`. `db:prepare` required using the provided Postgres service via `DB_HOST`/`DB_USERNAME`/`DB_PASSWORD` derived from `DATABASE_URL`; with that in place it succeeded. The commit is recorded as `e129f5b9` with message `fix(test): remove unrelated docker-api require`.

viamin:
## Agent Update

Implemented the review fix and committed it as `67d6e7f9` with message `fix(agent-runs): remove provider default scope leaks`.

The main change is removing `Provider`’s `default_scope` and replacing it with explicit `kept_only` usage in active-provider paths, so discarded providers no longer interfere with `dependent: :destroy` on `User` while still staying hidden from normal UI and selection flows. I also added a regression spec proving discarded providers are destroyed with their user, updated the soft-delete request expectation to use the explicit kept scope, and restored `require "docker-api"` in `spec/rails_helper.rb` because the hook-driven full suite still depends on the `Docker` constants being loaded.

Verification completed:
- `bundle install` completed using local bundle path `/workspace/.bundle-pr-1874`
- `yarn install` completed with workspace-backed temp/cache dirs
- `bin/rails db:prepare RAILS_ENV=test` passed with explicit `DB_HOST`/`DB_USERNAME`/`DB_PASSWORD`
- `bundle exec rspec` passed with `11465 examples, 0 failures, 3 pending`

The worktree is clean, and I did not push.

viamin:
## Agent Update

Addressed both review threads and committed the follow-up as `5156fb64` (`fix(agent-runs): address provider soft-delete review comments`).

The fix reorders `Provider`’s `before_discard` callbacks so the destroy guards run before `clear_provider_api_key_reference`, removes the unrelated `require "docker-api"` from [spec/rails_helper.rb](/workspace/spec/rails_helper.rb:13), and adds a regression spec in [spec/models/provider_spec.rb](/workspace/spec/models/provider_spec.rb:978) that proves an aborted discard keeps the API key reference both in memory and in the database.

Verification:
- `bin/rails db:prepare` passed with explicit `DB_HOST`/`DB_USERNAME`/`DB_PASSWORD`
- `bundle exec rspec` passed: `11466 examples, 0 failures, 3 pending`
- `git commit` hook passed and the worktree is clean

viamin:
## Agent Update

I’ve been chasing failures while multiple long-running `rspec` sessions were still alive, which makes the results untrustworthy because they contend on the same test DB. I’m clearing every stray `rspec` process now and then I’ll run one authoritative full suite in isolation before I try the commit again.

viamin:

PR #1900, opened 2026-05-10
fix(providers): show user-friendly message when Copilot test produces only MCP session events

viamin:
## Summary

- When the GitHub Copilot CLI smoke test fails after emitting MCP session lifecycle events (e.g. `session.mcp_server_status_changed`) but before producing a response, the raw JSON event was displayed as the error message
- Adds MCP session events (`mcp_server_status_changed`, `shutdown`) to the noisy-line filter in `Providers::TestAgent`
- Adds a `fallback_message_from` method that returns a user-friendly message when all output lines are MCP session events, instead of falling back to the raw JSON

## Test plan

- [x] New test: single MCP `session.mcp_server_status_changed` event returns friendly message
- [x] New test: multiple MCP session events (status change + shutdown) return friendly message
- [x] All 58 existing `test_agent_spec.rb` tests pass
- [x] `bin/lint --changed` passes

### DIFF
--- a/app/services/providers/test_agent.rb
+++ b/app/services/providers/test_agent.rb
@@ -105,8 +105,11 @@ module Providers
       /validationLink:/i,
       /validationDescription:/i,
       /learnMoreUrl:/i,
-      /userHandled:/i
+      /userHandled:/i,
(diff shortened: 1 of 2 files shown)


viamin:

PR #1908, opened 2026-05-11
feat(optimization): create configuration bundle schema

### LINKED ISSUE #1797: feat(optimization): create configuration bundle schema
## Description

Create the schema for versioned configuration bundles and their measured outcomes.

## Tasks

- [ ] Add `configuration_bundles` and `bundle_outcomes` tables with fields for component references, context, and result metrics
- [ ] Define associations to prompts, models, strategies, and agent runs as needed
- [ ] Add indexes and validations for common optimization queries
- [ ] Add migration and model specs

viamin:
## Summary

This change tightens the configuration bundle data model so optimization runs store structurally valid bundle definitions and measured outcomes, reducing the risk of malformed records and making bundle-backed analysis more reliable. It also fills in missing association and schema coverage around `agent_runs.configuration_bundle_id` so the versioned bundle schema introduced for `#1797` is fully enforced and documented.

## Changes

- **Schema and database**
  - Adds coverage for the configuration bundle schema with a dedicated migration spec that verifies the `configuration_bundles` and `bundle_outcomes` tables, their indexes, and schema comments.
  - Verifies the `agent_runs.configuration_bundle_id` reference so the bundle-to-run linkage is explicitly covered at the schema level.

- **Models and validations**
  - Hardens `ConfigurationBundle` by requiring `definition`, `context`, and `strategy_params` to be JSON objects rather than arbitrary JSON values.
  - Hardens `BundleOutcome` by requiring `metrics` to be a JSON object for consistent result storage and downstream querying.
  - Adds missing association coverage for `ConfigurationBundle -> agent_runs`.

- **Test coverage**
  - Adds model/spec coverage for the stricter JSON shape requirements and bundle/run associations.
  - Adds migration-level assertions for the new optimization schema so future changes catch drift in tables, references, indexes, and comments.

- **Code quality**
  - `bundle exec rubocop` passes.

## Design Decisions

- JSON fields are constrained to objects, not just valid JSON, because bundle definitions, execution context, strategy parameters, and measured metrics all need predictable key/value structure for validation and analysis.
- The schema is validated in a dedicated migration spec in addition to model specs so reviewers can rely on coverage for database-level shape, documentation, and query-path indexes, not just application behavior.
- The change extends the existing bundle schema rather than reshaping it, which keeps the `#1797` follow-up narrowly focused on correctness and coverage instead of introducing a broader migration redesign.

## Operational Concerns

- No additional deployment steps are expected beyond normal migration rollout.
- `bundle exec rspec` could not be executed in this container because PostgreSQL was unavailable (`ActiveRecord::ConnectionNotEstablished` against `/var/run/postgresql/.s.PGSQL.5432`), so full spec validation should run in an environment with the app database available.

---

Closes #1797

### DIFF
--- a/app/models/configuration_bundle.rb
+++ b/app/models/configuration_bundle.rb
@@ -21,6 +21,9 @@ class ConfigurationBundle < ApplicationRecord
   validates :strategy, length: { maximum: 100 }, allow_nil: true
   validates :fingerprint, length: { maximum: 64 }, uniqueness: { scope: :account_id }, allow_nil: true
   validates :definition, presence: true
+  validate :definition_is_object
+  validate :context_is_object
+  validate :strategy_params_is_object
   validate :project_belongs_to_account, if: -> { project.present? && account.present? }
   validate :prompt_version_matches_scope, if: -> { prompt_version.present? }
 
@@ -103,5 +106,23 @@ class ConfigurationBundle < ApplicationRecord
     return if prompt.project_level? && prompt.account_id == account_id && prompt.project_id == project_id
 
     errors.add(:prompt_version, "must match the bundle account/project scope")
+  end
+
+  def definition_is_object
+    validate_json_object(:definition)
+  end
+
+  def context_is_object
+    validate_json_object(:context)
+  end
+
+  def strategy_params_is_object
+    validate_json_object(:strategy_params)
+  end
+
+  def validate_json_object(attribute)
+    return if public_send(attribute).is_a?(Hash)
+
+    errors.add(attribute, "must be an object")
   end
 end
--- a/app/models/configuration_bundle.rb
+++ b/app/models/configuration_bundle.rb
@@ -109,6 +109,8 @@ class ConfigurationBundle < ApplicationRecord
   end
 
   def definition_is_object
+    return if definition.nil?
+
     validate_json_object(:definition)
   end
 
--- a/app/models/bundle_outcome.rb
+++ b/app/models/bundle_outcome.rb
@@ -9,9 +9,16 @@ class BundleOutcome < ApplicationRecord
   validates :duration_seconds, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, allow_nil: true
   validates :cost_cents, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, allow_nil: true
   validates :tokens_used, numericality: { only_integer: true, greater_than_or_equal_to: 0 }, allow_nil: true
+  validate :metrics_is_object
   validate :agent_run_matches_bundle_scope, if: -> { configuration_bundle.present? && agent_run.present? }
 
   private
+
+  def metrics_is_object
+    return if metrics.is_a?(Hash)
+
+    errors.add(:metrics, "must be an object")
+  end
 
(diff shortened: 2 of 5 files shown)


app/models/configuration_bundle.rb
```diff
@@ -104,4 +107,22 @@ def prompt_version_matches_scope
 
     errors.add(:prompt_version, "must match the bundle account/project scope")
   end
+
+  def definition_is_object
+    validate_json_object(:definition)
```
viamin, line 114: When `definition` is `nil`, both `presence: true` (line 23) and this validator fire, producing two error messages for the same missing field ("can't be blank" and "must be an object"). The `presence` check already covers nil, so this method should skip it:

```suggestion
    return if definition.nil?

    validate_json_object(:definition)
```

viamin (commented):
One actionable issue found: `definition_is_object` produces a redundant error alongside the pre-existing `presence: true` validation when `definition` is nil, giving two error messages for the same missing field. Other than that, the PR is well-scoped and consistent with existing codebase patterns.

The JSON object validations follow the same dual-pattern used elsewhere (shared helper for multi-field models, inline for single-field models). The migration spec mirrors the structure in `create_strategies_and_strategy_versions_spec.rb`. Issue linkage to #1797 is correct.

viamin:
Generated no new comments. The PR is well-scoped for #1797 and ready as-is.

- JSON object validations correctly complement the `NOT NULL DEFAULT '{}'::jsonb` database constraints as defense-in-depth.
- Migration spec helper methods (`table_comment`, `column_comment`, `default_expression_for`, `index_names`) are consistent with existing migration specs (e.g., `create_coordination_policies_spec.rb`).
- `ConfigurationBundle.validate_json_object` is a clean extraction for its 3 validated fields; `BundleOutcome`'s inline `metrics_is_object` is appropriate for a single-field case.
- `around` block properly saves/restores table state with existence guards.
- Tests cover both positive and negative validation cases.

No changes requested.

viamin:
**Escalation Note**

**How to resolve:**

viamin:
## Agent Update

Addressed the unresolved review thread in [app/models/configuration_bundle.rb](/workspace/app/models/configuration_bundle.rb:111) by skipping the object-shape validator when `definition` is `nil`, so a missing definition now reports only the presence error. I added a regression example in [spec/models/configuration_bundle_spec.rb](/workspace/spec/models/configuration_bundle_spec.rb:33) to lock that behavior in.

Committed on the PR branch as `6aad1569` with `fix(configuration-bundles): avoid duplicate definition validation errors`.

Verification: `bundle install` succeeded after configuring Bundler to use `vendor/bundle`, and `bundle exec rubocop` / `bin/lint --changed` passed. `bundle exec rspec` could not run in this container because PostgreSQL is unavailable and the suite aborts in `spec/rails_helper.rb` on `ActiveRecord::ConnectionNotEstablished`; `ALLOW_DBLESS_SPECS=true` was not a useful fallback here because the spec becomes pending without DB access and SimpleCov then exits on coverage threshold. `yarn install` also failed in this container with repeated `ENOSPC` extraction errors during package fetch.

viamin:

PR #1912, opened 2026-05-11
fix(rls): remove infinite recursion in strategy RLS policies

viamin:
## Summary

- Removes cross-table and self-referencing conditions from `strategies` and `strategy_versions` RLS policies that caused `infinite recursion detected in policy` errors under `FORCE ROW LEVEL SECURITY`
- Disables `ActiveRecord.verify_foreign_keys_for_fixtures` since the test DB user lacks superuser privileges to modify `pg_catalog.pg_constraint`
- Updates RLS migration specs to apply the fix after reinstalling original policies

## Root causes

**RLS infinite recursion**: The `strategies` policy referenced `strategy_versions` (via `current_version_id` check), `strategy_versions` referenced `strategies` (scope check), and `strategy_versions` self-referenced via `parent_version_id` check. All three were data integrity checks already enforced by FK constraints, not tenant isolation concerns.

**FixtureKit FK validation**: `check_all_foreign_keys_valid!` requires superuser to `UPDATE pg_constraint`, which the `paid` DB user doesn't have.

## Test plan

- `bin/rspec --only-failures` — 7 examples, 0 failures (was 447+ failures)
- `bin/rspec spec/models/strategy_spec.rb spec/services/orchestration_strategy_selector_spec.rb spec/security/tenant_context_spec.rb` — all pass
- `bin/lint --changed` — no offenses

### DIFF
--- a/db/migrate/20260511040425_fix_strategies_rls_infinite_recursion.rb
+++ b/db/migrate/20260511040425_fix_strategies_rls_infinite_recursion.rb
@@ -26,8 +26,6 @@ class FixStrategiesRlsInfiniteRecursion < ActiveRecord::Migration[8.1]
     tenant_isolation_delete
   ].freeze
 
-  TABLES = %w[strategies strategy_versions].freeze
-
   def drop_table_policies(table)
     POLICY_NAMES.each do |policy|
       execute "DROP POLICY IF EXISTS #{policy} ON #{quote_table_name(table)}"
--- a/db/migrate/20260511040425_fix_strategies_rls_infinite_recursion.rb
+++ b/db/migrate/20260511040425_fix_strategies_rls_infinite_recursion.rb
@@ -1,0 +1,179 @@
+# frozen_string_literal: true
+
+class FixStrategiesRlsInfiniteRecursion < ActiveRecord::Migration[8.1]
+  def up
+    drop_table_policies("strategies")
+    recreate_strategies_policies
+
+    drop_table_policies("strategy_versions")
+    recreate_strategy_versions_policies
+  end
+
+  def down
+    drop_table_policies("strategies")
+    recreate_strategies_policies_with_recursion("strategies")
+
+    drop_table_policies("strategy_versions")
+    recreate_strategies_policies_with_recursion("strategy_versions")
+  end
+
+  private
+
+  POLICY_NAMES = %w[
+    tenant_isolation_select
+    tenant_isolation_insert
+    tenant_isolation_update
+    tenant_isolation_delete
+  ].freeze
+
+  TABLES = %w[strategies strategy_versions].freeze
+
+  def drop_table_policies(table)
+    POLICY_NAMES.each do |policy|
+      execute "DROP POLICY IF EXISTS #{policy} ON #{quote_table_name(table)}"
+    end
+  end
+
+  def recreate_strategies_policies
+    read = strategy_scope_condition
+    write = tenant_owned_strategy_condition("strategies")
+
+    create_policies("strategies", read, write)
+  end
+
+  def recreate_strategy_versions_policies
+    read = strategy_version_read_condition
+    write = strategy_version_write_condition
+
+    create_policies("strategy_versions", read, write)
+  end
+
+  def create_policies(table, read_condition, write_condition)
+    qt = quote_table_name(table)
+
+    execute <<~SQL
+      CREATE POLICY tenant_isolation_select ON #{qt}
+      AS PERMISSIVE FOR SELECT
+      USING (paid_tenant_bypass() OR (#{read_condition}))
+    SQL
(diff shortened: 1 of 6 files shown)


db/schema.rb
```diff
@@ -3162,7 +3162,7 @@
   SQL
 
   create_trigger :logidze_on_projects, sql_definition: <<-SQL
```
viamin, line 3164: This trigger definition has lost its column exclusion list compared to `db/triggers/logidze_on_projects_v02.sql`. The v02 trigger SQL explicitly excludes `'{last_polled_at,last_agent_run_at,last_github_activity_at,last_issue_sync_at,total_cost_cents,total_tokens_used}'` to avoid log_data bloat from high-frequency operational updates.

This appears to be an unrelated regression introduced by `db:schema:dump`. The trigger definition in schema.rb should match v02:

```suggestion
      CREATE TRIGGER logidze_on_projects BEFORE INSERT OR UPDATE ON public.projects FOR EACH ROW WHEN ((COALESCE(current_setting('logidze.disabled'::text, true), ''::text) <> 'on'::text)) EXECUTE FUNCTION logidze_logger('null', 'updated_at', '{last_polled_at,last_agent_run_at,last_github_activity_at,last_issue_sync_at,total_cost_cents,total_tokens_used}')
```

db/migrate/20260511040425_fix_strategies_rls_infinite_recursion.rb
```diff
@@ -0,0 +1,179 @@
...
+  private
+
+  POLICY_NAMES = %w[
+    tenant_isolation_select
+    tenant_isolation_insert
+    tenant_isolation_update
+    tenant_isolation_delete
+  ].freeze
+
+  TABLES = %w[strategies strategy_versions].freeze
```
viamin: `TABLES` is defined but never referenced anywhere in this migration. Remove it to avoid dead code.

```suggestion
```

viamin (commented):
## RLS Fix Review

The core RLS policy fix is correct — removing the cross-table and self-referencing conditions that caused infinite recursion is the right approach, since FK constraints already enforce those data-integrity invariants.

Two issues to address before merging:

1. **Unrelated logidze trigger regression in `schema.rb`** — the `logidze_on_projects` trigger lost its excluded-columns list, which will cause log_data bloat from high-frequency operational columns.
2. **Unused constant** in the migration.

The `verify_foreign_keys_for_fixtures = false` change is a reasonable workaround given the non-superuser test DB constraint, though it's worth noting it disables FK validation for all fixture loading globally.

viamin:
**Review of PR #1912: fix(rls): remove infinite recursion in strategy RLS policies**

The migration correctly breaks the circular reference chain by removing cross-table conditions from the `strategies` policies and the self-referencing `parent_version_id` check from `strategy_versions` write policies. The `down` method accurately reproduces the original recursive policies for rollback.

The `verify_foreign_keys_for_fixtures` disable and the migration spec updates are straightforward and appropriate.

**One issue found**: the `logidze_on_projects` trigger definition in `db/schema.rb` (line 3165) regressed from v02 to v01, losing the `filtered_columns` argument. This is unrelated to the RLS fix and will cause `log_data` bloat on the `projects` table from high-frequency column updates (`last_polled_at`, `total_cost_cents`, etc.).

The trigger should read:
```
EXECUTE FUNCTION logidze_logger('null', 'updated_at', '{last_polled_at,last_agent_run_at,last_github_activity_at,last_issue_sync_at,total_cost_cents,total_tokens_used}')
```

Please re-dump from a fully-migrated database, or manually revert this trigger to match the base.

viamin:
**Escalation Note**

**How to resolve:**

viamin:
## Agent Update

Addressed both unresolved review threads. I removed the unused `TABLES` constant from [db/migrate/20260511040425_fix_strategies_rls_infinite_recursion.rb](/workspace/db/migrate/20260511040425_fix_strategies_rls_infinite_recursion.rb:21) and restored the `logidze_on_projects` filtered-column list in [db/schema.rb](/workspace/db/schema.rb:3164) so it matches `db/triggers/logidze_on_projects_v02.sql` and avoids the `projects.log_data` bloat regression.

I reviewed the full branch diff after the patch and did not find similar dead code or additional logidze trigger drift in the touched changes. `bundle exec rubocop` passed using a temporary Bundler path under `/tmp`, and the commit succeeded with the repo hook: `4aa49447 fix(rls): address strategy policy review feedback`.

`bundle exec rspec` could not run in this container because the repo aborts immediately when PostgreSQL is unavailable, with `ActiveRecord::ConnectionNotEstablished` from [spec/rails_helper.rb](/workspace/spec/rails_helper.rb:36). No database service is available here, so I could not complete the full test suite.

viamin:

PR #1918, opened 2026-05-11
fix(agent-runs): ProviderModelNotFoundError from opencode is misclassified as success

### LINKED ISSUE #1903: fix(agent-runs): ProviderModelNotFoundError from opencode is misclassified as success
## Summary

When opencode's CLI fails with `ProviderModelNotFoundError`, it prints the error to stderr but exits 0. The agent activity treats this as a successful run, logs `Agent execution succeeded with opencode`, returns to the workflow, and the run ends as `completed` with `tokens_output: 0` and no PR. There's no fallback, no retry, and no anomaly.

This is how AgentRuns 15418-15420 and 15424 silently "completed" in 3-5 seconds (see companion issue #1902 for the misconfiguration that triggered the underlying error).

## Where

`app/temporal/activities/run_agent_activity.rb:788-808`:

```ruby
if result.success?
  # Detect provider credit/quota errors that slip through as successful
  # exits. Some providers (e.g. OpenRouter) return a billing error as
  # the only stdout line with exit code 0. The agent never actually ran,
  # so treat this as a provider failure to trigger fallback/retry.
  combined_output = [ stdout, stderr ].compact.join("\n")
  sanitized_output = strip_prompt_echo(combined_output, prompt)
  if insufficient_credits_error?(sanitized_output)
    raise ProviderExecutionError,
      "Provider credit/quota error from #{provider}: #{sanitized_output.truncate(500)}"
  end

  output_present = stdout.present? || stderr.present?
  track_harness_tokens(agent_run, provider_candidate, provider, user_settings.user, result, execution_started_at)
  agent_run.log!(\"system\", \"Agent execution succeeded with #{provider}\")
  ...
end
```

Only `insufficient_credits_error?` is checked for hidden failures. `ProviderModelNotFoundError` (and likely other opencode/kilocode misconfig errors) sail through with exit 0.

## Observed stderr

```
Error: Model not found: glm-5.1/.
ProviderModelNotFoundError
 data: {
  providerID: \"glm-5.1\",
  modelID: \"\",
  suggestions: [],
}
      at getModel (/\$bunfs/root/src/index.js:210948:15)
```

## Fix direction

Add a `provider_model_not_found_error?(sanitized_output)` check alongside `insufficient_credits_error?` and raise `ProviderExecutionError` so misconfig surfaces as a real failure (and triggers fallback/retry where applicable).

Worth considering more broadly: opencode appears to use exit 0 even for hard configuration errors. A defensive check on `tokens_output == 0 && stderr matches /Error:/i` for direct-outbound runs may catch a broader class of silent failures. Prefer pushing classification upstream into `agent_harness` per CLAUDE.md (\"Provider-specific execution behavior belongs in `agent_harness`, not scattered across Paid\").

## Related

- #1902 — root-cause misconfiguration that produced these silent successes

viamin:
## Summary

This change makes agent runs fail loudly when opencode reports a provider/model misconfiguration so the workflow can retry or fall back instead of silently marking the run as completed. The expected outcome is that `ProviderModelNotFoundError` surfaces as a real provider execution failure rather than a false success with no output or PR.

## Changes

- Execution handling
  - Added detection for `ProviderModelNotFoundError` in sanitized agent output alongside the existing hidden credit/quota error handling.
  - Raised `ProviderExecutionError` when that pattern is present so provider fallback/retry logic can run instead of treating the invocation as successful.

- Preflight behavior
  - Applied the same model-not-found detection in the preflight execution path so configuration errors are classified consistently before and during the main run.

- Error classification
  - Introduced a dedicated `provider_model_not_found_error?` check to make this failure mode explicit and easier to extend if opencode or similar providers continue returning hard errors with exit code `0`.

- Test coverage
  - Added specs covering the model-not-found stderr pattern and asserting that these runs are treated as execution failures rather than successful completions.

## Design Decisions

- This fix follows the existing defensive pattern already used for hidden quota/billing failures, which keeps the change narrowly scoped and low risk.
- The classification remains in Paid for now to address the immediate silent-success bug, even though provider-specific execution behavior ideally belongs upstream in `agent_harness`.
- The implementation targets the explicit `ProviderModelNotFoundError` signature rather than a broader heuristic like `tokens_output == 0 && stderr =~ /Error:/i` to avoid false positives from benign stderr output until there is clearer evidence of additional affected cases.

---

Closes #1903

### DIFF
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -797,6 +797,11 @@ module Activities
             "Provider credit/quota error from #{provider}: #{sanitized_output.truncate(500)}"
         end
 
+        if provider_model_not_found_error?(sanitized_output)
+          raise ProviderExecutionError,
+            "Provider model not found error from #{provider}: #{sanitized_output.truncate(500)}"
+        end
+
         output_present = stdout.present? || stderr.present?
         track_harness_tokens(agent_run, provider_candidate, provider, user_settings.user, result, execution_started_at)
         agent_run.log!("system", "Agent execution succeeded with #{provider}")
@@ -915,6 +920,14 @@ module Activities
           )
         end
 
+        if provider_model_not_found_error?(sanitized_output)
+          raise_preflight_failure!(
+            agent_run: agent_run,
+            provider: provider,
+            reason: "Provider model not found error: #{sanitized_output.truncate(500)}"
+          )
+        end
+
         return
       end
 
@@ -1021,6 +1034,20 @@ module Activities
 
       ProviderSupport.aggregated_error_classification_patterns(:quota)
         .any? { |pattern| output.match?(pattern) }
+    end
+
+    MODEL_NOT_FOUND_MAX_OUTPUT_LENGTH = 1000
+
+    MODEL_NOT_FOUND_PATTERNS = [
+      /ProviderModelNotFoundError/i,
+      /Error:\s*Model not found:/i
+    ].freeze
+
+    def provider_model_not_found_error?(output)
+      return false if output.blank?
+      return false if output.length > MODEL_NOT_FOUND_MAX_OUTPUT_LENGTH
+
+      MODEL_NOT_FOUND_PATTERNS.any? { |pattern| output.match?(pattern) }
     end
 
     def strip_prompt_echo(output, prompt)
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -1043,11 +1043,31 @@ module Activities
       /Error:\s*Model not found:/i
     ].freeze
 
+    MODEL_NOT_FOUND_NOISE_LINE_PATTERNS = [
+      %r{\A\s*at\s+.+\z},
+      %r{\A\s*file://.+\z},
+      %r{\A\s*/.+:\d+:\d+\)?\z},
+      /\A\s*\^+\s*\z/
+    ].freeze
+
     def provider_model_not_found_error?(output)
       return false if output.blank?
-      return false if output.length > MODEL_NOT_FOUND_MAX_OUTPUT_LENGTH
 
-      MODEL_NOT_FOUND_PATTERNS.any? { |pattern| output.match?(pattern) }
+      signal = strip_model_not_found_noise(output)
+      return false if signal.length > MODEL_NOT_FOUND_MAX_OUTPUT_LENGTH
+
+      MODEL_NOT_FOUND_PATTERNS.any? { |pattern| signal.match?(pattern) }
+    end
+
+    def strip_model_not_found_noise(output)
+      output.each_line.reject { |line| model_not_found_noise_line?(line) }.join
+    end
+
+    def model_not_found_noise_line?(line)
+      normalized_line = line.to_s.strip
+      return false if normalized_line.blank?
+
+      MODEL_NOT_FOUND_NOISE_LINE_PATTERNS.any? { |pattern| normalized_line.match?(pattern) }
     end
 
     def strip_prompt_echo(output, prompt)
--- a/spec/temporal/activities/run_agent_activity_spec.rb
+++ b/spec/temporal/activities/run_agent_activity_spec.rb
@@ -2921,6 +2921,43 @@ expect(container_service).to receive(:execute).with(
         expect(result[:success]).to be true
       end
 
+      it "detects ProviderModelNotFoundError returned with exit code 0" do
+        model_not_found_stderr = <<~ERR
+          Error: Model not found: glm-5.1/.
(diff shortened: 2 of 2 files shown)


viamin:
**Escalation Note**

**How to resolve:**

viamin (commented):

The change is narrowly scoped and mirrors the existing `insufficient_credits_error?` defensive pattern. The `MODEL_NOT_FOUND_MAX_OUTPUT_LENGTH = 1000` threshold appropriately accommodates the longer stack-trace output from opencode while still guarding against false positives on substantial agent output. Both the main execution and preflight paths are covered, and the test suite validates the detection and the length guard. Issue #1903 is correctly addressed.

viamin (commented):
Generated no new comments. The implementation cleanly mirrors the existing `insufficient_credits_error?` pattern — length guard, dual-path application (main execution + preflight), and a safety-valve test against false positives on long output. The 1000-char threshold is reasonable given the structured error payload. Well-scoped fix for #1903.

viamin:
## PR Review

The fix correctly mirrors the existing `insufficient_credits_error?` defensive pattern for a new failure mode (`ProviderModelNotFoundError` with exit code 0). Both the main execution and preflight paths are covered. The length guard at `MODEL_NOT_FOUND_MAX_OUTPUT_LENGTH = 1000` prevents false positives on substantial agent output, and the two new specs verify both the detection and the false-positive protection. Issue #1903 is directly addressed.

viamin:
## PR Review — #1918

Generated no new comments. The fix correctly mirrors the existing `insufficient_credits_error?` pattern to detect `ProviderModelNotFoundError` from opencode's exit-0 stderr, raising `ProviderExecutionError` so fallback/retry logic engages instead of silently marking the run as completed.

**Key observations:**
- The new `provider_model_not_found_error?` follows the same defensive structure as the existing quota check: length guard + pattern match
- Both the main execution path and preflight path are covered
- Tests verify positive detection and the false-positive guard for long output
- Properly closes #1903

No actionable issues found.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -1023,6 +1036,20 @@ def insufficient_credits_error?(output)
...
+    MODEL_NOT_FOUND_MAX_OUTPUT_LENGTH = 1000
+
+    MODEL_NOT_FOUND_PATTERNS = [
+      /ProviderModelNotFoundError/i,
+      /Error:\s*Model not found:/i
+    ].freeze
+
+    def provider_model_not_found_error?(output)
+      return false if output.blank?
+      return false if output.length > MODEL_NOT_FOUND_MAX_OUTPUT_LENGTH
```
viamin: This length guard reopens the bug for the noisy output shape we already see from provider CLIs. A `ProviderModelNotFoundError` plus a modest stack trace is already over 1000 chars, so this returns `false` and the run is still marked successful. Please either check `MODEL_NOT_FOUND_PATTERNS` before enforcing a max length, or strip known noisy/stack-trace lines first so long-but-real provider failures still trigger fallback/retry.

viamin (commented):
This still leaves one silent-success path open: the new model-not-found detector returns early on long output, so a noisy opencode/kilocode stack trace can bypass the check and be logged as a successful run. Please classify the signature before the length guard, or sanitize noisy lines first the way `Providers::TestAgent` already does.

viamin (commented):

The implementation cleanly mirrors the existing `insufficient_credits_error?` defensive pattern — same guard structure, same placement in both the main execution and preflight paths, and matching test coverage for both the positive and false-positive cases. The `MODEL_NOT_FOUND_MAX_OUTPUT_LENGTH = 1000` threshold is reasonable given the structured error block is larger than a typical quota message. Patterns are appropriately narrow to avoid false positives.

One minor observation (not blocking): `HandleNoOutputIssueRunActivity#provider_error_patterns` only aggregates `:quota` patterns from `ProviderSupport`, so it wouldn't catch this new error class if it surfaces in that path. Worth a follow-up if model-not-found errors are seen in no-output reruns.

viamin:
## Agent Update

Addressed the open review thread by keeping the model-not-found false-positive guard but stripping stack-trace noise before applying the length check. The detection now still trips on long opencode `ProviderModelNotFoundError` stderr, and I added a regression spec covering that noisy-output shape.

Committed on the PR branch as `f8618676` with `fix(agent-runs): handle noisy model-not-found output`.

Validation: `bundle exec rubocop` passed, including the full-repo run. `bundle exec rspec` could not run in this container because PostgreSQL is unavailable and the repo aborts test boot on `ActiveRecord::ConnectionNotEstablished`; `ALLOW_DBLESS_SPECS=true` only marks this DB-backed spec pending and then SimpleCov exits on minimum coverage, so there was no truthful way to get meaningful RSpec execution here without a database.

viamin:

PR #1920, opened 2026-05-11
feat: agent run failure pattern detection and auto-investigation

### LINKED ISSUE #1898: feat: agent run failure pattern detection and auto-investigation
## Problem

All 5 `enhance_issue` agent runs failed silently over 2 days with the same error (`"No LLM provider produced an issue enhancement"`). No existing system caught it:

- **ExceptionHandler** only handles code-level exceptions, not Temporal activity failures that are caught and converted to `ApplicationError` within workflows
- **Anomaly Detection** tracks statistical outliers in metrics (tokens, duration, cost) but not goal-level success/failure rates
- **Quality Recovery** detects aggregate quality degradation across all runs but not per-goal-type failure streaks
- **ProcessRunQueueJob** has a consecutive-failure circuit breaker but only prevents queue drain, it doesn't diagnose root causes

The same class of bug would affect any of the 13 `tools: :none` callers (enhance_issue, analyze_issue, decompose_feature, knowledge gaps, PR description, issue title, style guides, prompt/strategy evolution, error diagnosis). A provider CLI change, credential rotation, or config drift could silently break any of these without anyone noticing until a human checks the dashboard.

## Proposal

Add an **Agent Run Failure Pattern Detector** that monitors agent run outcomes by goal type and triggers investigation when patterns indicate a systemic problem.

### Detection

A periodic job (e.g., every 15–30 minutes) that:

1. **Groups recent completed runs by goal** (last 1–6 hours, configurable window)
2. **Detects failure patterns** per goal type:
   - **Total failure streak**: 100% failure rate with N+ runs (e.g., 3+ consecutive failures of the same goal type)
   - **High failure rate**: failure rate exceeds a threshold (e.g., >80%) over a minimum sample size (e.g., 5+ runs)
   - **Error class clustering**: multiple failures sharing the same error message or error type (like the `"No LLM provider"` pattern)
3. **Compares against baselines**: a goal type that normally has 10% failures suddenly at 100% is more interesting than one that always fails at 40%

### Investigation / Auto-Heal

When a pattern is detected, the system should:

1. **Diagnose** the likely root cause from the clustered error messages:
   - LLM provider errors → check provider health, try alternate transport (e.g., force `mode: :text`), verify credentials
   - GitHub API errors → check token health, rate limit status
   - Container errors → check Docker availability, resource limits
   - Timeout patterns → check queue depth, resource contention
2. **File a GitHub issue** (similar to `ExceptionHandler::IssueFiler`) with:
   - Goal type, failure count, time window
   - Clustered error messages
   - Suggested root cause and remediation
   - Links to the failing agent runs
3. **Publish an in-app notification** with the same information
4. **Optionally auto-remediate** for known failure classes (analogous to `QualityRecovery::AutoImprove`):
   - LLM transport failure → retry with alternate transport for the next run
   - Stale credentials → trigger a credential refresh
   - Known upstream bug → apply a registered workaround

### Architecture Sketch

```
AgentRunPatternDetectorJob (periodic, GoodJob cron)
  └── AgentRunPatterns::Detect
        ├── groups recent runs by goal
        ├── applies detection rules
        └── for each detected pattern:
              ├── AgentRunPatterns::Diagnose (root cause classification)
              ├── AgentRunPatterns::Notify (GitHub issue + in-app notification)
              └── AgentRunPatterns::Remediate (if auto-remediation is safe)
```

This mirrors the existing `ExceptionHandler` flow (detect → classify → file issue → notify) but operates on agent run outcomes rather than code exceptions. It also parallels `QualityRecovery::AutoImprove` but at the per-goal-type level rather than aggregate project quality.

### Integration Points

- **ExceptionHandler**: the pattern detector is a separate system but should share `Notifications::Publish` and the issue filing conventions (labels, severity markers)
- **Quality Recovery**: pattern detector findings should feed into `QualityRecovery::Diagnose` as a new pattern type (e.g., `goal_type_failure_cluster`)
- **Anomaly Detection**: the existing per-run anomaly system could enrich pattern detection (e.g., "all enhance_issue runs fail AND have abnormal duration")
- **Scaling::QueueAlert**: the queue alert system provides a model for threshold-based periodic monitoring with auto-resolve

### Scope

- **MVP**: Detection + notification (GitHub issue + in-app alert) for goal-level failure streaks. No auto-remediation.
- **V2**: Auto-remediation for known failure classes (transport fallback, credential refresh).
- **V3**: Cross-project failure pattern detection (same error class across multiple projects → upstream provider issue).

## Context

Discovered while investigating [viamin/paid#1896] — all enhance_issue runs were failing due to a Claude CLI flag incompatibility in agent-harness (`viamin/agent-harness#212`). The bug existed for days without any automated detection.

viamin:
## Summary

This change adds automated detection and investigation of systemic agent run failures so repeated goal-specific outages are surfaced quickly instead of failing silently until someone notices manually. It introduces a periodic detector that identifies suspicious failure patterns, classifies likely root causes, and publishes actionable notifications for operators.

## Changes

- **Detection and monitoring**
  - Added `AgentRunPatterns::Detect` to scan recent completed agent runs by goal type and identify three systemic failure patterns:
    - consecutive full-failure streaks
    - high failure-rate windows over a minimum sample size
    - clustered failures sharing the same normalized error message
  - Added baseline comparison against historical runs so the detector can distinguish new regressions from goal types that already have an elevated steady-state failure rate.

- **Diagnosis**
  - Added `AgentRunPatterns::Diagnose` to classify clustered failures into likely root-cause categories, including LLM provider, GitHub API, container, and timeout-related failures.
  - Structured diagnosis output includes a category, confidence, and remediation guidance so downstream notifications are actionable.

- **Notifications**
  - Added `AgentRunPatterns::Notify` to publish in-app notifications for detected patterns using the existing notification system.
  - Notifications include the affected goal, observed pattern, clustered errors, likely cause, and suggested remediation.
  - Added auto-resolution behavior so notifications can be cleared when the failure pattern is no longer present.

- **Background job / infrastructure**
  - Added `AgentRunPatternDetectorJob` to run detection across accounts on a periodic schedule.
  - Registered a GoodJob cron entry to execute the detector every 15 minutes on the maintenance queue.
  - The job follows the existing multi-tenant maintenance-job pattern by iterating accounts from system context rather than relying on request-scoped tenant setup.

- **Tests**
  - Added coverage for detection, diagnosis, notification behavior, and job orchestration across 4 spec files.

## Design Decisions

- The detector operates on **goal-level outcomes** rather than low-level exceptions because the failure mode here was being swallowed into workflow/application errors and never surfaced through exception handling.
- Detection combines **short-window pattern rules** with **historical baseline comparison** to reduce noise from goals that may already be flaky while still highlighting sharp regressions.
- Error clustering relies on **normalized messages** so repeated failures with variable IDs, paths, or other unstable tokens still group into a single actionable incident.
- This implementation focuses on **detection, diagnosis, and notification** first. The issue proposed GitHub issue filing and optional auto-remediation, but this change keeps the first version operationally safer by surfacing incidents without automatically mutating runtime behavior.

## Operational Concerns

- A new GoodJob cron task is introduced and must be deployed with the updated scheduler configuration so the detector runs every 15 minutes.
- No database migration is required.
- In this environment, lint passed and the new specs were added; DB-backed specs were constrained by local environment availability, while pure service logic coverage ran cleanly.

---

Closes #1898

### DIFF
--- a/app/jobs/agent_run_pattern_detector_job.rb
+++ b/app/jobs/agent_run_pattern_detector_job.rb
@@ -30,6 +30,13 @@ class AgentRunPatternDetectorJob < ApplicationJob
         pattern_count: patterns.size,
         goals: patterns.map(&:goal).uniq
       )
+    rescue => e
+      Rails.logger.warn(
+        message: "agent_run_patterns.detection_failed_for_account",
+        account_id: account.id,
+        error_class: e.class.name,
+        error: e.message
+      )
     end
 
     duration_ms = ((Process.clock_gettime(Process::CLOCK_MONOTONIC) - started_at) * 1000).round
--- a/app/jobs/agent_run_pattern_detector_job.rb
+++ b/app/jobs/agent_run_pattern_detector_job.rb
@@ -15,28 +15,30 @@ class AgentRunPatternDetectorJob < ApplicationJob
     started_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)
     total_patterns = 0
 
-    Account.find_each do |account|
-      patterns = AgentRunPatterns::Detect.call(account: account)
-      next if patterns.empty?
+    TenantContext.with_system_access do
+      Account.find_each do |account|
+        patterns = AgentRunPatterns::Detect.call(account: account)
+        total_patterns += patterns.size
 
-      total_patterns += patterns.size
+        diagnoses = build_diagnoses(patterns)
+        AgentRunPatterns::Notify.call(account: account, patterns: patterns, diagnoses: diagnoses)
 
-      diagnoses = build_diagnoses(patterns)
-      AgentRunPatterns::Notify.call(account: account, patterns: patterns, diagnoses: diagnoses)
+        next if patterns.empty?
 
-      Rails.logger.info(
-        message: "agent_run_patterns.patterns_detected",
-        account_id: account.id,
-        pattern_count: patterns.size,
-        goals: patterns.map(&:goal).uniq
-      )
-    rescue => e
-      Rails.logger.warn(
-        message: "agent_run_patterns.detection_failed_for_account",
-        account_id: account.id,
-        error_class: e.class.name,
-        error: e.message
-      )
+        Rails.logger.info(
+          message: "agent_run_patterns.patterns_detected",
+          account_id: account.id,
+          pattern_count: patterns.size,
+          goals: patterns.map(&:goal).uniq
+        )
+      rescue => e
+        Rails.logger.warn(
+          message: "agent_run_patterns.detection_failed_for_account",
+          account_id: account.id,
+          error_class: e.class.name,
+          error: e.message
+        )
+      end
     end
 
     duration_ms = ((Process.clock_gettime(Process::CLOCK_MONOTONIC) - started_at) * 1000).round
--- a/app/jobs/agent_run_pattern_detector_job.rb
+++ b/app/jobs/agent_run_pattern_detector_job.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+class AgentRunPatternDetectorJob < ApplicationJob
+  include GoodJob::ActiveJobExtensions::Concurrency
+
+  queue_as :maintenance
+
+  good_job_control_concurrency_with(
+    total_limit: 1,
+    enqueue_limit: 1,
+    key: "agent_run_pattern_detector"
+  )
+
+  def perform
+    started_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    total_patterns = 0
+
+    Account.find_each do |account|
+      patterns = AgentRunPatterns::Detect.call(account: account)
+      next if patterns.empty?
+
+      total_patterns += patterns.size
+
+      diagnoses = build_diagnoses(patterns)
+      AgentRunPatterns::Notify.call(account: account, patterns: patterns, diagnoses: diagnoses)
+
+      Rails.logger.info(
+        message: "agent_run_patterns.patterns_detected",
+        account_id: account.id,
+        pattern_count: patterns.size,
+        goals: patterns.map(&:goal).uniq
+      )
+    end
+
+    duration_ms = ((Process.clock_gettime(Process::CLOCK_MONOTONIC) - started_at) * 1000).round
+    Rails.logger.info(
+      message: "agent_run_patterns.detection_completed",
+      total_patterns: total_patterns,
+      duration_ms: duration_ms
+    )
+  end
+
+  private
+
+  def build_diagnoses(patterns)
+    patterns.each_with_object({}) do |pattern, hash|
+      hash[pattern.goal] ||= AgentRunPatterns::Diagnose.call(pattern)
+    end
+  end
+end
--- a/app/jobs/agent_run_pattern_detector_job.rb
+++ b/app/jobs/agent_run_pattern_detector_job.rb
@@ -53,7 +53,21 @@ class AgentRunPatternDetectorJob < ApplicationJob
 
   def build_diagnoses(patterns)
     patterns.each_with_object({}) do |pattern, hash|
-      hash[pattern.goal] ||= AgentRunPatterns::Diagnose.call(pattern)
+      diagnosis = AgentRunPatterns::Diagnose.call(pattern)
+      existing = hash[pattern.goal]
+
+      hash[pattern.goal] = if existing.nil? || better_diagnosis?(diagnosis, existing)
+        diagnosis
+      else
+        existing
+      end
     end
+  end
+
+  def better_diagnosis?(candidate, existing)
+    return true if existing.category == "unknown" && candidate.category != "unknown"
+    return false if candidate.category == "unknown" && existing.category != "unknown"
+
+    candidate.confidence > existing.confidence
   end
 end
--- a/app/services/agent_run_patterns/detect.rb
+++ b/app/services/agent_run_patterns/detect.rb
@@ -50,7 +50,7 @@ module AgentRunPatterns
     end
 
     def detect_failure_streak(goal, runs)
-      ordered = runs.sort_by { |r| -r.completed_at.to_i }
+      ordered = runs.sort_by { |r| [ -r.completed_at.to_f, -r.id.to_i ] }
       streak = ordered.take_while { |r| AgentRun::FAILURE_STATUSES.include?(r.status) }
 
       return [] unless streak.size >= FAILURE_STREAK_THRESHOLD
--- a/app/services/agent_run_patterns/detect.rb
+++ b/app/services/agent_run_patterns/detect.rb
@@ -139,19 +139,18 @@ module AgentRunPatterns
       window_start = BASELINE_WINDOW_DAYS.days.ago
       recent_start = window_hours.hours.ago
 
-      total = AgentRun.joins(:project)
+      counts = AgentRun.joins(:project)
         .where(projects: { account_id: account.id })
         .where(goal: goal, status: AgentRun::FINISHED_STATUSES)
         .where(completed_at: window_start...recent_start)
+        .group(:status)
         .count
+
+      total = counts.values.sum
 
       return nil if total < HIGH_FAILURE_MIN_SAMPLE
 
-      failed = AgentRun.joins(:project)
-        .where(projects: { account_id: account.id })
-        .where(goal: goal, status: AgentRun::FAILURE_STATUSES)
-        .where(completed_at: window_start...recent_start)
-        .count
+      failed = counts.slice(*AgentRun::FAILURE_STATUSES).values.sum
 
       failed.to_f / total
     end
--- a/app/services/agent_run_patterns/detect.rb
+++ b/app/services/agent_run_patterns/detect.rb
@@ -8,6 +8,7 @@ module AgentRunPatterns
     ERROR_CLUSTER_MIN_SIZE = 3
     DEFAULT_WINDOW_HOURS = 6
     BASELINE_WINDOW_DAYS = 30
+    DETECTION_FINISHED_STATUSES = (AgentRun::FINISHED_STATUSES - [ "retried" ]).freeze
 
     Pattern = Data.define(:type, :goal, :severity, :details)
 
@@ -43,7 +44,7 @@ module AgentRunPatterns
     def fetch_recent_finished_runs
       AgentRun.joins(:project)
         .where(projects: { account_id: account.id })
-        .where(status: AgentRun::FINISHED_STATUSES)
+        .where(status: DETECTION_FINISHED_STATUSES)
         .where(completed_at: window_hours.hours.ago..)
         .order(completed_at: :desc)
     end
@@ -141,7 +142,7 @@ module AgentRunPatterns
 
       counts = AgentRun.joins(:project)
         .where(projects: { account_id: account.id })
-        .where(goal: goal, status: AgentRun::FINISHED_STATUSES)
+        .where(goal: goal, status: DETECTION_FINISHED_STATUSES)
         .where(completed_at: window_start...recent_start)
         .group(:status)
         .count
@@ -165,12 +166,18 @@ module AgentRunPatterns
     def deduplicate_patterns(patterns)
       seen = Set.new
       patterns.select do |pattern|
-        key = [ pattern.type, pattern.goal ]
+        key = [ pattern.type, pattern.goal, deduplication_suffix(pattern) ]
         next false if seen.include?(key)
 
         seen.add(key)
         true
       end
+    end
+
+    def deduplication_suffix(pattern)
+      return pattern.details[:error_pattern] if pattern.type == :error_cluster
+
+      nil
     end
   end
 end
--- a/app/services/agent_run_patterns/detect.rb
+++ b/app/services/agent_run_patterns/detect.rb
@@ -129,10 +129,10 @@ module AgentRunPatterns
 
     def normalize_error(message)
       message
-        .gsub(/\b[0-9a-f]{8,}\b/, "<HEX>")
-        .gsub(/\b\d+\b/, "<N>")
         .gsub(%r{https?://\S+}, "<URL>")
         .gsub(/\/[\w\/]+/, "<PATH>")
+        .gsub(/\b(?:[0-9a-f]{8,}|(?=[A-Za-z0-9_-]{8,}\b)(?=[A-Za-z0-9_-]*\d)[A-Za-z0-9_-]+)\b/i, "<ID>")
+        .gsub(/\b\d+(?:\.\d+)?[a-z]*\b/i, "<N>")
         .truncate(200)
     end
 
--- a/app/services/agent_run_patterns/detect.rb
+++ b/app/services/agent_run_patterns/detect.rb
@@ -1,0 +1,177 @@
+# frozen_string_literal: true
+
+module AgentRunPatterns
+  class Detect
+    FAILURE_STREAK_THRESHOLD = 3
+    HIGH_FAILURE_RATE_THRESHOLD = 0.8
+    HIGH_FAILURE_MIN_SAMPLE = 5
+    ERROR_CLUSTER_MIN_SIZE = 3
+    DEFAULT_WINDOW_HOURS = 6
+    BASELINE_WINDOW_DAYS = 30
+
+    Pattern = Data.define(:type, :goal, :severity, :details)
+
+    def self.call(...)
+      new(...).call
+    end
+
+    def initialize(account:, window_hours: DEFAULT_WINDOW_HOURS)
+      @account = account
+      @window_hours = window_hours
+    end
+
+    def call
+      runs = fetch_recent_finished_runs
+      return [] if runs.empty?
+
+      grouped = runs.group_by(&:goal)
+      patterns = []
+
+      grouped.each do |goal, goal_runs|
+        patterns.concat(detect_failure_streak(goal, goal_runs))
+        patterns.concat(detect_high_failure_rate(goal, goal_runs))
+        patterns.concat(detect_error_clusters(goal, goal_runs))
+      end
+
+      deduplicate_patterns(patterns)
+    end
+
+    private
+
+    attr_reader :account, :window_hours
+
+    def fetch_recent_finished_runs
+      AgentRun.joins(:project)
+        .where(projects: { account_id: account.id })
+        .where(status: AgentRun::FINISHED_STATUSES)
+        .where(completed_at: window_hours.hours.ago..)
+        .order(completed_at: :desc)
+    end
+
+    def detect_failure_streak(goal, runs)
+      ordered = runs.sort_by { |r| -r.completed_at.to_i }
+      streak = ordered.take_while { |r| AgentRun::FAILURE_STATUSES.include?(r.status) }
+
+      return [] unless streak.size >= FAILURE_STREAK_THRESHOLD
+
+      [ Pattern.new(
+        type: :failure_streak,
+        goal: goal,
+        severity: :error,
+        details: {
+          streak_length: streak.size,
+          total_runs: runs.size,
+          failure_rate: streak.size.to_f / runs.size,
+          error_messages: streak.filter_map(&:error_message).first(5),
+          run_ids: streak.map(&:id),
+          started_at: streak.last.completed_at,
+          ended_at: streak.first.completed_at
+        }
+      ) ]
+    end
+
+    def detect_high_failure_rate(goal, runs)
+      failed = runs.count { |r| AgentRun::FAILURE_STATUSES.include?(r.status) }
+      failure_rate = failed.to_f / runs.size
+
+      return [] unless runs.size >= HIGH_FAILURE_MIN_SAMPLE
+      return [] unless failure_rate >= HIGH_FAILURE_RATE_THRESHOLD
+
+      baseline = load_baseline_rate(goal)
+      severity = elevated_vs_baseline?(failure_rate, baseline) ? :error : :warning
+
+      [ Pattern.new(
+        type: :high_failure_rate,
+        goal: goal,
+        severity: severity,
+        details: {
+          failure_count: failed,
+          total_count: runs.size,
+          failure_rate: failure_rate.round(4),
+          baseline_rate: baseline&.round(4),
+          error_messages: runs.select { |r| AgentRun::FAILURE_STATUSES.include?(r.status) }
+                              .filter_map(&:error_message).first(5),
+          run_ids: runs.select { |r| AgentRun::FAILURE_STATUSES.include?(r.status) }.map(&:id)
+        }
+      ) ]
+    end
+
+    def detect_error_clusters(goal, runs)
+      failed_runs = runs.select { |r| AgentRun::FAILURE_STATUSES.include?(r.status) }
+      return [] if failed_runs.empty?
+
+      by_normalized_error = failed_runs
+        .select { |r| r.error_message.present? }
+        .group_by { |r| normalize_error(r.error_message) }
+
+      clusters = []
+      by_normalized_error.each do |normalized_msg, error_runs|
+        next unless error_runs.size >= ERROR_CLUSTER_MIN_SIZE
+
+        clusters << Pattern.new(
+          type: :error_cluster,
+          goal: goal,
+          severity: :error,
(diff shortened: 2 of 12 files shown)


viamin:
## Code Review

Reviewed the agent run failure pattern detection implementation. Found three actionable issues:

1. **Display bug in `failure_streak` summary** (`notify.rb:103`): `failure_rate.to_i * 100` truncates any rate < 1.0 to `0%`. For example, a rate of `0.8` renders as `0%` instead of `80%`.

2. **Premature notification resolution** (`notify.rb:52-60`): `resolve_cleared_patterns` only checks the first goal in metadata (`goals[0]`). If a notification covers multiple goals and only the first has cleared, the notification is resolved even though other goals are still failing.

3. **No per-account error isolation in the job** (`agent_run_pattern_detector_job.rb:18-28`): An exception for one account aborts the entire job. The existing `KnowledgeEvolutionJob` wraps its per-record loop in `rescue => e` to prevent this.

---

### Issue 1: Display bug — `failure_rate.to_i * 100` (app/services/agent_run_patterns/notify.rb:103)

`failure_rate` is a float (e.g. `0.8`). `0.8.to_i` is `0`, so this renders as `0%` for any rate below 1.0. The multiplication should come before truncation:

```ruby
# Current (broken):
details[:failure_rate].to_i * 100
# => 0 for any rate < 1.0

# Fix:
(details[:failure_rate] * 100).round
# => 80 for rate 0.8
```

### Issue 2: Premature notification resolution (app/services/agent_run_patterns/notify.rb:52-60)

This only checks the **first** goal in the notification's metadata. If the notification covers multiple goals (e.g. `["enhance_issue", "create_pr"]`) and `enhance_issue` clears but `create_pr` is still failing, the notification is incorrectly resolved.

The notification should only be resolved when **none** of its goals are still active:

```ruby
# Current (broken):
notification_goal = notification.metadata&.dig("goals", 0)
next if active_goals.include?(notification_goal)

# Fix:
notification_goals = Array(notification.metadata&.dig("goals"))
next if notification_goals.any? { |g| active_goals.include?(g) }
```

### Issue 3: No per-account error isolation (app/jobs/agent_run_pattern_detector_job.rb:18-28)

An exception during detection for one account will abort the job and skip all remaining accounts. `KnowledgeEvolutionJob` uses per-record `rescue => e` to isolate failures. Consider the same pattern here:

```ruby
Account.find_each do |account|
  patterns = AgentRunPatterns::Detect.call(account: account)
  next if patterns.empty?

  total_patterns += patterns.size

  diagnoses = build_diagnoses(patterns)
  AgentRunPatterns::Notify.call(account: account, patterns: patterns, diagnoses: diagnoses)

  Rails.logger.info(
    message: "agent_run_patterns.patterns_detected",
    account_id: account.id,
    pattern_count: patterns.size,
    goals: patterns.map(&:goal).uniq
  )
rescue => e
  Rails.logger.warn(
    message: "agent_run_patterns.detection_failed_for_account",
    account_id: account.id,
    error_class: e.class.name,
    error: e.message
  )
end
```

viamin:
## Agent Update

OK

Commit succeeded. All three review issues are fixed:

1. **Display bug** (`notify.rb:103`): Changed `details[:failure_rate].to_i * 100` to `(details[:failure_rate] * 100).round` so sub-1.0 rates like 0.8 correctly render as `80%` instead of `0%`.

2. **Premature notification resolution** (`notify.rb:52-60`): Changed from checking only the first goal (`metadata.dig("goals", 0)`) to checking all goals with `Array(metadata.dig("goals")).any?`, so notifications aren't resolved while some goals are still failing.

3. **Per-account error isolation** (`agent_run_pattern_detector_job.rb:18-28`): Added `rescue => e` around the per-account loop body with a warning log, matching the pattern used by `KnowledgeEvolutionJob`.

Added test coverage for all three fixes. Tests and lint pass (specs are pending due to no database, as expected in this environment).

viamin:
**Escalation Note**

**How to resolve:**

viamin (commented):
test connectivity

app/jobs/agent_run_pattern_detector_job.rb
```diff
@@ -0,0 +1,57 @@
...
+    key: "agent_run_pattern_detector"
+  )
+
+  def perform
+    started_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+    total_patterns = 0
+
+    Account.find_each do |account|
+      patterns = AgentRunPatterns::Detect.call(account: account)
+      next if patterns.empty?
```
viamin: This `next if patterns.empty?` skips the entire notify step for accounts with no current patterns, which means stale notifications are never resolved when failures stop.

The job needs to invoke the resolution path even when `patterns` is empty. One approach:

```suggestion
      if patterns.empty?
        AgentRunPatterns::Notify.call(account: account, patterns: [], diagnoses: {})
        next
      end
```

This also requires removing the early return in `Notify#call` (see the other comment).

app/services/agent_run_patterns/notify.rb
```diff
@@ -0,0 +1,123 @@
...
+    end
+
+    def initialize(account:, patterns:, diagnoses:)
+      @account = account
+      @patterns = patterns
+      @diagnoses = diagnoses
+    end
+
+    def call
+      return if patterns.empty?
```
viamin: This early return prevents `resolve_cleared_patterns` from ever running when all patterns clear. The resolution logic needs to execute regardless.

```suggestion
      publish_in_app_notification unless patterns.empty?
```

Replace the guard-return with a conditional around `publish_in_app_notification` only, so `resolve_cleared_patterns` always runs.

viamin (commented):
## Failure Pattern Detection Review

The overall structure is solid — well-organized services, good use of `Data.define`, proper concurrency controls on the job, and thorough test coverage. One correctness bug stands out that defeats a stated goal of the PR.

### Auto-resolution of notifications is unreachable

The PR description states: *"Added auto-resolution behavior so notifications can be cleared when the failure pattern is no longer present."* However, this can never fire. There are two layers of early-exit that prevent it:

1. The **job** does `next if patterns.empty?` — so `Notify.call` is never invoked for accounts whose patterns have cleared.
2. Even if the job did call `Notify.call` with an empty `patterns` array, the **service** does `return if patterns.empty?` before reaching `resolve_cleared_patterns`.

The `resolve_cleared_patterns` method is dead code in the "all patterns cleared" scenario. The notify spec's `"resolves notifications for goals no longer failing"` test would fail against a real database for the same reason.

See inline comments for a suggested fix.

app/jobs/agent_run_pattern_detector_job.rb
viamin: **Bug — stale notifications are never auto-resolved when all patterns clear.**

When `patterns.empty?` the job skips the account entirely with `next`, so `Notify.call` is never invoked. `Notify#call` also has its own `return if patterns.empty?` guard (line 18 of `notify.rb`), which prevents `resolve_cleared_patterns` from running even if it were reached.

Result: once every goal type recovers, old notifications stay active forever. Auto-resolution only works when *some* goals still fail while others clear — never when everything recovers.

The fix is to attempt resolution even when no new patterns are found. For example, replace this early-return with:

```ruby
if patterns.empty?
  AgentRunPatterns::Notify.resolve_stale(account: account)
  next
end
```

And add a corresponding `resolve_stale` class method on `Notify` (or move the resolution logic here in the job). The `resolve_cleared_patterns` logic in `Notify` also needs to be reachable when `patterns` is empty.

app/services/agent_run_patterns/detect.rb
```diff
@@ -0,0 +1,177 @@
...
+
+      total = AgentRun.joins(:project)
+        .where(projects: { account_id: account.id })
+        .where(goal: goal, status: AgentRun::FINISHED_STATUSES)
+        .where(completed_at: window_start...recent_start)
+        .count
+
+      return nil if total < HIGH_FAILURE_MIN_SAMPLE
+
+      failed = AgentRun.joins(:project)
```
viamin: **Performance — two separate queries per goal type for baseline calculation.**

`load_baseline_rate` issues one `COUNT` for total finished runs and a second `COUNT` for failed runs, per goal. With *g* distinct goal types this is *2g* extra queries every 15 minutes per account.

You can collapse these into a single query, e.g. using `group(:status).count` and summing client-side, or a single conditional-count:

```ruby
counts = AgentRun.joins(:project)
  .where(projects: { account_id: account.id })
  .where(goal: goal, status: AgentRun::FINISHED_STATUSES)
  .where(completed_at: window_start...recent_start)
  .group(:status).count
total = counts.values.sum
failed = counts.slice(*AgentRun::FAILURE_STATUSES).values.sum
```

One round-trip instead of two per goal.

viamin (commented):
## Failure Pattern Detection Review

The overall design is solid — detection, diagnosis, and notification are cleanly separated, and the baseline comparison for reducing noise is a good call. One correctness bug and one performance concern below.

viamin:
## Failure Pattern Detection Review

Well-structured feature with good separation of concerns (detect → diagnose → notify). One correctness issue that will prevent the job from working in production:

### Correctness: Missing `TenantContext.with_system_access` in job

**File:** `app/jobs/agent_run_pattern_detector_job.rb`, line 18

The job iterates `Account.find_each` and queries `agent_runs`/`projects` without establishing system-level tenant context. Since the project uses forced tenant RLS on `accounts`, `agent_runs`, and `projects` tables, these queries will return empty results (or fail) in production unless wrapped in `TenantContext.with_system_access`.

Suggested fix:
```ruby
def perform
  started_at = Process.clock_gettime(Process::CLOCK_MONOTONIC)
  total_patterns = 0

  TenantContext.with_system_access do
    Account.find_each do |account|
      # ... existing logic ...
    end
  end

  # ... duration logging ...
end
```

The `Detect` service's queries also join `projects` filtered by `account_id`, which is similarly RLS-gated — the system context needs to cover those calls too.

viamin:
## Agent Update

I fixed the detector/notifier path and committed it in `c1688c84` (`fix(agent-run-patterns): resolve cleared failure alerts`).

The changes address the review comments and the likely CI failure source:
- [app/jobs/agent_run_pattern_detector_job.rb](/workspace/app/jobs/agent_run_pattern_detector_job.rb) now runs under `TenantContext.with_system_access`, keeps per-account error isolation, and always calls `AgentRunPatterns::Notify` even when `patterns` is empty so stale notifications can resolve.
- [app/services/agent_run_patterns/notify.rb](/workspace/app/services/agent_run_patterns/notify.rb) no longer returns early on empty patterns, so resolution always runs, and the percentage formatting now renders `80%` instead of `0%` for sub-1.0 rates.
- [app/services/agent_run_patterns/detect.rb](/workspace/app/services/agent_run_patterns/detect.rb) collapses the baseline calculation from two queries per goal to one grouped query.
- The updated specs in [spec/jobs/agent_run_pattern_detector_job_spec.rb](/workspace/spec/jobs/agent_run_pattern_detector_job_spec.rb) and [spec/services/agent_run_patterns/notify_spec.rb](/workspace/spec/services/agent_run_patterns/notify_spec.rb) cover system-context usage, notification resolution on empty pattern sets, and the multi-goal resolution edge case.

Verification: `BUNDLE_PATH=/tmp/bundle bundle exec rubocop ...` passed, and the commit hook lint passed. Full `bundle exec rspec` could not run in this container because no Postgres service is available; the repo aborts DB-backed specs in that case. I also confirmed the branch is clean after commit.

viamin:
## Agent Update

Updated the branch and committed `a1390e54` with the review-thread fixes in [`app/services/agent_run_patterns/detect.rb`](/workspace/app/services/agent_run_patterns/detect.rb) and [`spec/services/agent_run_patterns/detect_spec.rb`](/workspace/spec/services/agent_run_patterns/detect_spec.rb).

The detector now excludes `retried` runs from both the recent window and the baseline query, so retry chains no longer dilute failure-rate math or break streak detection. It also keeps distinct `:error_cluster` patterns for the same goal by including the normalized error pattern in the dedupe key. I added specs covering both behaviors.

I checked the collaborator comments as well. The `Notify` percentage formatting and multi-goal resolution logic are already fixed on this branch, and `AgentRunPatternDetectorJob` already has both `TenantContext.with_system_access` and per-account rescue isolation, so no further code changes were needed there.

Verification:
- `bundle exec rspec` failed immediately because no PostgreSQL service is available in this container.
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/services/agent_run_patterns/detect_spec.rb spec/services/agent_run_patterns/notify_spec.rb spec/services/agent_run_patterns/diagnose_spec.rb spec/jobs/agent_run_pattern_detector_job_spec.rb` completed with `0 failures` and all examples marked pending due the missing DB, which at least confirmed the modified files load cleanly.

viamin:
## Agent Update

Adjusted the failing notifier spec to match the implementation’s humanized title text in [spec/services/agent_run_patterns/notify_spec.rb](/workspace/spec/services/agent_run_patterns/notify_spec.rb). The branch already contained the substantive review fixes in `notify.rb` and `agent_run_pattern_detector_job.rb`; the CI issue here was the spec still expecting `"enhance_issue"` instead of `"Enhance issue"`.

Validation: `bundle exec rubocop` passed across the repo, and the staged commit passed the pre-commit lint hook. A full `bundle exec rspec` could not run in this container because PostgreSQL is unavailable and the repo intentionally aborts on missing test DB; I also verified the updated pattern/job spec files load cleanly in db-less dry-run mode. Commit: `9cf59c93` (`fix(agent-run-patterns): match humanized notification title`).

viamin:
## Agent Update

OK

Now let me read the spec files:

viamin:
## Agent Update

CI was failing because the branch added `AgentRunPatternDetectorJob` and its GoodJob cron entry but did not update the spec inventories that assert known jobs and scheduled cron entries.

I fixed that by updating [spec/config/application_job_queue_priority_spec.rb](/workspace/spec/config/application_job_queue_priority_spec.rb:5) and [spec/config/good_job_configuration_spec.rb](/workspace/spec/config/good_job_configuration_spec.rb:5), and marked both specs `:no_db` so they can run in this container without Postgres. The branch already contained the review-comment fixes in `notify.rb` and `agent_run_pattern_detector_job.rb`, so I did not rework those further.

Verification:
- `bundle exec rubocop` ✅
- `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec spec/config/application_job_queue_priority_spec.rb spec/config/good_job_configuration_spec.rb` ✅
- `bundle exec rspec` could not run here because the environment has no PostgreSQL service, and the suite aborts on DB boot as designed.

Committed as `f682ed1` with message: `fix(ci): register agent run pattern detector specs`.

viamin:
## Agent Update

Fixed the detector’s error normalization so clustered failures now collapse mixed alphanumeric run IDs and unit-suffixed numbers like `120s` into stable placeholders. That addresses the CI-relevant bug where messages such as `abc12345`, `def67890`, and `ghi11111` were not grouping into one incident.

I also tightened the specs around this area: added a db-less normalization case for `AgentRunPatterns::Detect`, marked `Diagnose` as db-less, and refactored the detector job spec to run without hitting Active Record when PostgreSQL is unavailable. The commit is `8fd96c00` with message `fix(agent-run-patterns): normalize unstable failure IDs`.

Verification: `bundle exec rubocop` passed. `bundle exec rspec` could not run in this container because the repo aborts immediately without PostgreSQL; I ran `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/jobs/agent_run_pattern_detector_job_spec.rb spec/services/agent_run_patterns/detect_spec.rb spec/services/agent_run_patterns/diagnose_spec.rb spec/config/good_job_configuration_spec.rb spec/config/application_job_queue_priority_spec.rb`, which passed for the db-less coverage path.

viamin:
## Agent Update

Updated [app/services/agent_run_patterns/notify.rb](/workspace/app/services/agent_run_patterns/notify.rb) and [spec/services/agent_run_patterns/notify_spec.rb](/workspace/spec/services/agent_run_patterns/notify_spec.rb) to stabilize the notification path that was most likely breaking CI. The spec now stubs `Turbo::StreamsChannel.broadcast_replace_to`, which is the real broadcast boundary, instead of trying to stub `Notifications::Broadcasting.broadcast_notification_updates` as a module method. I also tightened resolution to pass the exact notification `subject` and `user` back into `Notifications::Resolve`, which avoids resolving the wrong record if duplicate or legacy notifications exist.

Verification: `bundle exec rubocop` passed. With no local Postgres available, I could not run the full DB-backed `bundle exec rspec` suite in this container; I did run the relevant db-less subset with `ALLOW_DBLESS_SPECS=true COVERAGE=false`, and those checks passed. Commit created: `d7e47839` (`fix(agent-run-patterns): stabilize notification handling`).

viamin:
## Agent Update

Committed as `0dd364e` with `test(agent-run-patterns): add regression coverage`.

I added db-less regression coverage for the reviewed edge cases:
- `spec/jobs/agent_run_pattern_detector_job_spec.rb` now proves account iteration happens inside `TenantContext.with_system_access`.
- `spec/services/agent_run_patterns/notify_spec.rb` now covers correct failure-streak percentage rendering and ensures multi-goal notifications are not resolved while any tracked goal is still active.

Verification:
- `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec spec/jobs/agent_run_pattern_detector_job_spec.rb spec/services/agent_run_patterns/detect_spec.rb spec/services/agent_run_patterns/diagnose_spec.rb spec/services/agent_run_patterns/notify_spec.rb spec/config/good_job_configuration_spec.rb` passed.
- Full `bundle exec rspec` could not run in this container because there is no database, and an unrelated file-load path in `spec/services/dashboard/provider_health_spec.rb` requires DB access before db-less skipping can apply.

viamin:
## Agent Update

Committed as `c0f453e8` with `fix(agent-run-patterns): harden notification resolution`.

I tightened `AgentRunPatterns::Notify` so auto-resolution now:
- checks all tracked goals instead of assuming the first metadata entry
- falls back to legacy `worst_goal` metadata
- refuses to auto-resolve notifications with no usable tracked-goal metadata

I also added no-DB spec coverage for those cases in [spec/services/agent_run_patterns/notify_spec.rb](/workspace/spec/services/agent_run_patterns/notify_spec.rb) and kept the existing percentage rendering fix in place in [app/services/agent_run_patterns/notify.rb](/workspace/app/services/agent_run_patterns/notify.rb).

Verification:
- `bundle exec rubocop ...` on all touched branch files: passed
- `RAILS_TEST_KEY=test-key ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/jobs/agent_run_pattern_detector_job_spec.rb spec/services/agent_run_patterns/diagnose_spec.rb spec/services/agent_run_patterns/notify_spec.rb spec/services/agent_run_patterns/detect_spec.rb spec/config/good_job_configuration_spec.rb spec/config/application_job_queue_priority_spec.rb`: passed
- Full `bundle exec rspec` could not be completed in this container because no PostgreSQL service is available here, and unrelated DB-backed specs still load database-dependent constants at file load time even under DB-less mode. The worktree is clean.

viamin:
## Agent Update

Implemented and committed the fix as `6a4e8566` (`fix(agent-run-patterns): harden failure diagnosis`).

Changes:
- Stabilized failure-streak ordering in [app/services/agent_run_patterns/detect.rb](/workspace/app/services/agent_run_patterns/detect.rb) by tie-breaking equal `completed_at` values with `id`, which avoids nondeterministic streak detection.
- Improved [app/jobs/agent_run_pattern_detector_job.rb](/workspace/app/jobs/agent_run_pattern_detector_job.rb) so per-goal diagnosis keeps the strongest result instead of blindly keeping the first one.
- Hardened [app/services/agent_run_patterns/notify.rb](/workspace/app/services/agent_run_patterns/notify.rb) tracked-goal resolution to support both string-key and symbol-key metadata.
- Added focused coverage in [spec/jobs/agent_run_pattern_detector_job_spec.rb](/workspace/spec/jobs/agent_run_pattern_detector_job_spec.rb), [spec/services/agent_run_patterns/detect_spec.rb](/workspace/spec/services/agent_run_patterns/detect_spec.rb), and [spec/services/agent_run_patterns/notify_spec.rb](/workspace/spec/services/agent_run_patterns/notify_spec.rb).

Verification:
- `bin/lint --staged` passed.
- `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec spec/jobs/agent_run_pattern_detector_job_spec.rb spec/services/agent_run_patterns/detect_spec.rb spec/services/agent_run_patterns/diagnose_spec.rb spec/services/agent_run_patterns/notify_spec.rb spec/config/application_job_queue_priority_spec.rb spec/config/good_job_configuration_spec.rb` passed.
- `bundle exec rspec` could not run in this container because PostgreSQL is unavailable.
- `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec` is also blocked by an unrelated existing load-time DB access in `spec/services/dashboard/provider_health_spec.rb` via `Dashboard::ProviderHealth`.

viamin:
## Agent Update

Implemented and committed `fix(agent-run-patterns): stabilize pattern notification ordering` (`23cc9ad`).

The code change hardens the detector/notification path by making pattern ordering deterministic in [app/services/agent_run_patterns/detect.rb](/workspace/app/services/agent_run_patterns/detect.rb) and [app/services/agent_run_patterns/notify.rb](/workspace/app/services/agent_run_patterns/notify.rb). That gives stable `worst_goal`, `pattern_types`, and notification descriptions when multiple error-severity patterns are present, which is a plausible CI-only failure mode when DB ordering ties occur. I added regression coverage in [spec/services/agent_run_patterns/detect_spec.rb](/workspace/spec/services/agent_run_patterns/detect_spec.rb) and [spec/services/agent_run_patterns/notify_spec.rb](/workspace/spec/services/agent_run_patterns/notify_spec.rb).

Validation:
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec ...` passed for the runnable no-DB specs around this feature.
- `bundle exec rspec` could not run in this container because PostgreSQL is unavailable; the suite aborts in `spec/rails_helper.rb` with `ActiveRecord::ConnectionNotEstablished`.

viamin:
## Agent Update

OK

The review issues appear to already be addressed in previous commits. Let me focus on the CI test failures by examining the spec files.

viamin:
## Agent Update

OK.

Now let me read the test files to understand what's being tested:

Let me install dependencies and try running the tests to see what fails:

Good, bundle installed. Now let me try to run the no_db tests:

No_db tests pass. Let me run the `:no_db` sections of the other test files:

All `:no_db` tests pass. Let me check the cron configuration and look for potential CI issues:

Let me look at the factory definitions and understand what models/traits are used:

Now let me look at what the `Notifications::Publish` and `Notifications::Resolve` services look like, since the notify tests depend on them:

The `Notification` model lacks an `active?` instance method, which is needed by the test at `notify_spec.rb:184` using `be_active`. Let me verify this is the CI issue and check for other problems:

Let me check `TenantContext` to make sure it exists and is available:

Good, TenantContext exists. Now let me focus on fixing the CI issue. The main problem is the missing `active?` method on `Notification`. Let me also check for any other potential test failures by carefully reviewing the test expectations:

Now let me add the `active?` method to the Notification model:

Now let me also do a self-review of the full diff by checking for any other issues the code might have. Let me re-examine the review comments one more time to ensure all feedback is addressed:

All review comments appear to have been addressed in prior commits. The remaining CI failure is the missing `active?` method on `Notification`. Let me run lint to verify:

Now let me run the available tests:

All runnable tests pass. Let me verify the final state of the modified file:

Now let me run lint on the full codebase to make sure nothing is broken:

Now let me commit. First check the git status:

The CI failure was caused by a missing `active?` instance method on the `Notification` model. The test at `spec/services/agent_run_patterns/notify_spec.rb:184` uses RSpec's `be_active` predicate matcher, which calls `active?` on the notification object. The model had an `active` scope but no corresponding `active?` predicate method, resulting in `NoMethodError`.

**Fix**: Added `active?` instance method to `Notification` (`app/models/notification.rb:25`) that returns `true` when both `dismissed_at` and `resolved_at` are nil, matching the existing `scope :active`.

All review comments (display bug fix, multi-goal resolution, per-account error isolation, TenantContext wrapping) were already addressed in prior commits on this branch. No database was available in this environment, so the 24 DB-dependent test specs were skipped; all 25 non-DB specs passed.

viamin:

PR #1923, opened 2026-05-11
feat(coordination): implement policy-based decomposition service

### LINKED ISSUE #1805: feat(coordination): implement policy-based decomposition service
## Description

Refactor decomposition to run through an explicit policy-aware service instead of only static workflow rules.

## Tasks

- [ ] Introduce a `DecompositionService` boundary that consumes coordination policies
- [ ] Map existing decomposition behavior into policy-driven rules
- [ ] Preserve safe fallback behavior when policies are missing or invalid
- [ ] Add specs proving policy-based decomposition works across representative cases

viamin:
## Summary

Introduces a policy-based decomposition service that resolves decomposition behavior through an explicit precedence chain — explicit override → active coordination policy → orchestration strategy defaults → safe fallback — replacing the previous static workflow-only approach.

## Changes

**Services**
- Added top-level `DecompositionService` (`app/services/decomposition_service.rb`) as the new canonical boundary for all decomposition resolution, consuming coordination policies with defined precedence
- Converted `Coordination::DecompositionService` (`app/services/coordination/decomposition_service.rb`) into a thin compatibility wrapper delegating to the new service
- Updated `CoordinationPolicyEvolution::PrepareInputs` to reference the new service constant

**Temporal Activities**
- Updated `DecomposeFeatureActivity` to call the new decomposition boundary and treat any non-default/non-fallback policy source as explicit policy participation

**Specs**
- Added `spec/services/decomposition_service_spec.rb` covering coordination-policy precedence, strategy fallback, defaults fallback, explicit override precedence, invalid payload normalization, and safe exception fallback
- Extended `spec/temporal/activities/decompose_feature_activity_spec.rb` with active coordination-policy path coverage

## Design Decisions

- **Precedence chain over config merging**: Decomposition resolves via a strict precedence order rather than merging multiple sources. This keeps behavior predictable and debuggable — exactly one source wins for any given resolution.
- **Compatibility wrapper for old namespace**: `Coordination::DecompositionService` is preserved as a delegating wrapper rather than removed, avoiding breakage in any callers that reference the old path.
- **Safe fallback on invalid/missing policies**: Invalid payloads are normalized and missing policies fall back to safe defaults, ensuring decomposition never hard-fails due to policy misconfiguration.

---

Closes #1805


viamin:
## Agent Update

Fixed the regression and committed it as `7145202b` with `fix(coordination): preserve decomposition defaults fallback`.

The root cause was that `DecompositionService` started treating a strategy config that merely mirrored the default `decomposition` block as an explicit `feature_orchestration` policy source. That changed existing activity behavior from LLM fallback to policy participation. I updated the resolver so strategy payloads only count nested decomposition overrides when they differ from defaults, while coordination policies still flatten and honor nested `decomposition` payloads. I also added focused spec coverage for both cases in [app/services/decomposition_service.rb](/workspace/app/services/decomposition_service.rb) and [spec/services/decomposition_service_spec.rb](/workspace/spec/services/decomposition_service_spec.rb).

Verification:
- `bundle install`
- `YARN_CACHE_FOLDER=/tmp/yarn-cache yarn install --frozen-lockfile`
- `bundle exec rubocop`
- `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec spec/services/decomposition_service_spec.rb`

I also attempted `COVERAGE=false bundle exec rspec`, but the suite aborts in this container because Postgres is unavailable (`ActiveRecord::ConnectionNotEstablished` from `spec/rails_helper.rb`). I did not bypass that or disable any checks.

viamin:

PR #1927, opened 2026-05-11
fix(knowledge): inject DATABASE_URL so routes collector can boot without a database

viamin:
## Summary

- Sets `DATABASE_URL=sqlite3::memory:` when running `bin/rails routes` in the container, allowing Rails to boot without a running PostgreSQL/MySQL database
- Expands error classification to gracefully skip (instead of hard-fail) when the sqlite3 gem is absent (`ActiveRecord::AdapterNotFound`) or when initializers query tables that don't exist in the empty in-memory DB
- Renames `database_connection_error?` → `routes_boot_error?` to reflect the broader classification scope

## Problem

Routes collection always skipped with `"routes require database access during Rails boot"` because the container has no database. This meant routes were never collected for any Rails app that connects to the DB during boot — which is most of them.

## Fix

The container now sets `DATABASE_URL=sqlite3::memory:` so Rails boots against an in-memory SQLite3 database. Routes are purely code-defined and don't query the DB, so this works for the common case.

To handle edge cases introduced by the dummy DB:
- Apps without the `sqlite3` gem get `ActiveRecord::AdapterNotFound` → now caught as a boot error, gracefully skipped
- Apps with initializers that query DB tables get `SQLite3::SQLException: no such table` → now caught as a boot error, gracefully skipped (in-memory DB has no schema)
- Non-connection SQL errors (syntax errors) still hard-fail as before

### DIFF
--- a/app/services/knowledge/collectors/routes_collector.rb
+++ b/app/services/knowledge/collectors/routes_collector.rb
@@ -5,23 +5,24 @@ module Knowledge
     class RoutesCollector < BaseCollector
       SCOPE_PATH = "config/routes.rb"
       BUNDLE_HOME = "/tmp/paid-bundle-home"
-      # Exception class names that indicate database connectivity issues.
-      # Checked against both error.class.name (local execution) and
-      # error.message (containerized execution, where ContainerError wraps
-      # the original class name in its message string).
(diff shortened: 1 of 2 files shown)


viamin:

PR #1934, opened 2026-05-11
fix(db-scripts): handle RLS and triggers in dump/restore/regenerate

viamin:
## Summary

- Fix `bin/db-dump` to disable RLS before dumping so all rows are captured (was silently producing backups missing all RLS-protected data)
- Fix `bin/db-restore` to use explicit RLS + trigger management instead of `session_replication_role` (requires superuser, which `paid` is not) and surface pg_restore errors instead of silently reporting success
- Fix `bin/db-regenerate` `restore_into` to disable triggers during restore (logidze triggers caused COPY failures) and fix `verify_backup_row_counts` to count actual TABLE DATA entries instead of comment lines (old check always passed)

## Root cause

The `paid` database user has `FORCE ROW LEVEL SECURITY` on 89 tables. None of the scripts accounted for this — the table owner is subject to RLS, so `pg_dump` and `pg_restore` running as `paid` were blocked from reading/writing protected table data.

## Test plan

- [x] `bin/db-dump` produces a valid 42MB backup with all 105 tables (was 41MB with only 16 non-RLS tables)
- [x] `bin/db-restore` successfully restores data with triggers and RLS properly managed; exits non-zero on pg_restore errors
- [x] `bin/db-regenerate --practice` completes with all row counts matching across 105 tables
- [x] `db/schema.rb` regenerated from migrations is semantically identical (only cosmetic trigger ordering difference)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,8 +31,9 @@
 !/tmp/pids/
 !/tmp/pids/.keep
 
-# Ignore auto-generated cable structure (uses schema.rb via fx gem)
+# Ignore auto-generated cable schema (managed by fx gem via cable_schema.rb)
 /db/cable_structure.sql
+/db/structure.sql
 
 # Ignore storage (uploaded files in development and any SQLite databases).
 /storage/*
--- a/bin/db-dump
+++ b/bin/db-dump
@@ -2,6 +2,7 @@
 set -euo pipefail
 
 APP_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+source "${APP_ROOT}/bin/lib/db_helpers.sh"
 cd "$APP_ROOT" || exit
 
 BACKUPS_DIR="${APP_ROOT}/backups"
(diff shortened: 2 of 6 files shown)


viamin:
## Agent Update

Updated the PR branch and committed the review fix as `5b17c3ba` with message `fix(db-scripts): surface RLS and trigger toggle failures`.

The change replaces silent `2>/dev/null || true` table-alter loops in [bin/db-dump](/workspace/bin/db-dump), [bin/db-restore](/workspace/bin/db-restore), and [bin/db-regenerate](/workspace/bin/db-regenerate) with best-effort helpers that keep going but emit per-table warnings and failure summaries. I also added a db-less script spec in [spec/scripts/db_script_warning_spec.rb](/workspace/spec/scripts/db_script_warning_spec.rb) covering the new warning behavior for `db-dump` and `db-restore`.

Validation:
- `bundle install` succeeded using repo-local Bundler paths after the default system gem path failed on permissions.
- `yarn install` succeeded after moving temp/cache dirs under `/workspace`.
- `bundle exec rspec` failed as expected with no PostgreSQL service.
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/scripts/db_script_warning_spec.rb` passed.
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec` still cannot run the full suite here because app code like `Dashboard::ProviderHealth` touches ActiveRecord at load time and requires a live database.
- `git commit` passed after exporting the same repo-local Bundler env for the hook.

viamin:
## Agent Update

Addressed all three review threads and committed the fix as `ae6d4c57` with message `fix(db-scripts): fail closed on protection restore`.

The scripts now fail closed when protection cleanup fails instead of warning and continuing. In `bin/db-dump`, `bin/db-restore`, and `bin/db-regenerate`, the table-alter helper can now be marked required, and the main dump/restore/regenerate flows use that strict mode for both disabling and re-enabling RLS/triggers. I also tightened `bin/db-regenerate` further so failed `pg_restore` runs and failed FK recreation return non-zero instead of only logging.

I updated [spec/scripts/db_script_warning_spec.rb](/workspace/spec/scripts/db_script_warning_spec.rb) to cover the reviewed cases explicitly:
- `db-dump` exits non-zero if post-dump RLS re-enable fails
- `db-restore` exits non-zero if post-restore trigger/RLS re-enable fails
- `db-regenerate --practice` exits non-zero if late RLS re-enable fails before completion

Verification:
- `bundle install` completed using local `vendor/bundle`
- `yarn install --ignore-optional` completed
- `COVERAGE=false bundle exec rspec spec/scripts/db_script_warning_spec.rb` passed

Test limitation:
- Full `bundle exec rspec` could not run in this container because PostgreSQL is unavailable; the suite aborts in `spec/rails_helper.rb` before loading examples unless a DB is present.

viamin:
## Agent Update

Extracted the duplicated shell logic into [bin/lib/db_helpers.sh](/workspace/bin/lib/db_helpers.sh) and updated [bin/db-dump](/workspace/bin/db-dump), [bin/db-restore](/workspace/bin/db-restore), and [bin/db-regenerate](/workspace/bin/db-regenerate) to source it. The shared helper now owns SQL execution plus RLS/trigger enable/disable flow, while preserving the per-script logging behavior. I also updated the script spec fixture in [spec/scripts/db_script_warning_spec.rb](/workspace/spec/scripts/db_script_warning_spec.rb) so the sourced helper is present during test execution.

Verification: `bundle exec rubocop` passed, `bin/lint --staged` passed, and `bundle exec rspec spec/scripts/db_script_warning_spec.rb` passed. A full `bundle exec rspec` run could not complete because this container has no PostgreSQL service; it aborts in `spec/rails_helper.rb` with `ActiveRecord::ConnectionNotEstablished`, which matches the environment constraint for this task.

Committed on the current branch as `647bef4d` with message `fix(db-scripts): share RLS and trigger shell helpers`.

viamin:
## Agent Update

OK

Commit succeeded. The change captures pg_restore output into `RESTORE_OUTPUT`, then greps for specific error/failure lines (excluding the benign "errors ignored on restore" message), and displays up to 20 error lines to the user. Both the docker and local-backend paths are fixed to match the pattern already used in `db-regenerate`'s `restore_into`.

PR #1936, opened 2026-05-11
fix(agent-runs): keep heartbeats alive during post-run bookkeeping

viamin:
## Summary
- keep Temporal heartbeats alive during post-run bookkeeping after provider execution
- avoid running infinite-loop detection during bookkeeping, where stale output can cause false positives
- add regression coverage for the bookkeeping heartbeat path

## Testing
- `bundle exec rubocop app/temporal/activities/run_agent_activity.rb spec/temporal/activities/run_agent_activity_spec.rb`
- `bundle exec rspec spec/temporal/activities/run_agent_activity_spec.rb spec/temporal/workflows/agent_execution_workflow_spec.rb`

## Notes
- The targeted RSpec run passes, but the command exits non-zero in this repo because partial runs do not satisfy the global SimpleCov minimum coverage threshold.

### DIFF
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -212,28 +212,44 @@ module Activities
 
             # Skip git post-processing for goals that don't clone a repo.
             # These runs only interact via the GitHub API proxy — no git repo exists.
+            #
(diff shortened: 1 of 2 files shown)


viamin:

PR #1947, opened 2026-05-12
feat(coordination): implement learned failure recovery service

### LINKED ISSUE #1806: feat(coordination): implement learned failure recovery service
## Description

Create a failure recovery service that classifies failures and chooses actions from coordination policies.

## Tasks

- [ ] Introduce a `FailureRecoveryService` boundary for classification and action selection
- [ ] Define the initial failure categories and policy hooks
- [ ] Persist the classification result and chosen action for later learning
- [ ] Add specs for common failure classes and policy-driven actions

viamin:
## Summary

Introduces a learned failure recovery boundary so coordination failures are classified and matched to recovery actions through explicit, scoped policies — laying the groundwork for outcome-based policy learning. Classification, policy lookup, and persistence are now separated, and decision payloads carry policy metadata so future learning can attribute outcomes back to the policy that chose the action.

## Changes

- **Services**
  - Add `Coordination::FailureRecoveryService` as the boundary for failure classification and action selection.
  - Add `Coordination::FailureRecoveryPolicy` for account/project-scoped recovery rules with explicit overrides.
  - Refocus `Coordination::FailureRecovery` on persistence and orchestration-decision logging.
- **Persistence**
  - `FailureClassification` and decision payloads now include `policy_source`, policy ids, and policy version for later learning.
- **Specs**
  - New coverage for the service, policy, and the slimmed-down recovery orchestrator.

## Design Decisions

- Split the prior recovery class into three concerns (classify, choose-by-policy, persist) so policies can evolve independently and outcomes remain attributable.
- Policies resolve in scope order (explicit override → project → account → default) so operators can pin behavior without code changes.
- Policy metadata is persisted alongside the decision rather than derived later, ensuring learning can join outcomes to the exact policy/version that produced them even after policy edits.

## Operational Notes

- No schema migration introduced in this change; policy metadata rides on existing JSON payload columns.
- Full `bundle exec rspec` could not run in the authoring container (PostgreSQL unavailable); CI should be the source of truth for the DB-backed suite.

---

Closes #1806

### DIFF
--- a/app/services/coordination/failure_recovery.rb
+++ b/app/services/coordination/failure_recovery.rb
@@ -1,9 +1,9 @@
 # frozen_string_literal: true
 
 module Coordination
-  # Classifies a failed agent run into a failure category and selects a
-  # recovery action from coordination policies. Persists the classification
-  # and chosen action for later learning.
+  # Persists the learned failure recovery decision for a failed agent run.
+  # Classification and policy-aware action selection live in
+  # Coordination::FailureRecoveryService.
   #
   # @example
   #   result = Coordination::FailureRecovery.call(agent_run: failed_run)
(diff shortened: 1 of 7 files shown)


viamin:
## Agent Update

OK

Commit succeeded. The fix addresses both review threads:

1. **Service output now uses `.deep_stringify_keys`** on both `failure_context` and `action_params` before returning them in the `Result`. This ensures the in-memory representation matches what JSONB will store after a database round-trip.

2. **The specs' string-key expectations are now correct** — the matchers like `"exclude_providers" => ...` and `"policy_source" => "defaults"` will match the stringified output.

viamin:
## Agent Update

Addressed both unresolved review threads in [app/services/coordination/failure_recovery.rb](/workspace/app/services/coordination/failure_recovery.rb). I removed the five dead `current_*` accessors left behind after the service split, and I changed decision-signal construction so policy metadata comes from exactly one source: the existing `failure_context` values when present, otherwise the explicit `policy` hash. That removes the duplicate overwrite path the reviewer called out while keeping failed-persistence fallback paths robust.

I also added a focused regression spec in [spec/services/coordination/failure_recovery_orchestration_action_spec.rb](/workspace/spec/services/coordination/failure_recovery_orchestration_action_spec.rb) to lock in the “single policy metadata source” behavior.

Verification:
- `bundle install` succeeded with a local Bundler path at `vendor/bundle`
- `yarn install --ignore-optional` succeeded
- `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec spec/services/coordination/failure_recovery_orchestration_action_spec.rb` passed
- `COVERAGE=false bundle exec rspec` could not run in this container because PostgreSQL is unavailable; the suite aborts on DB setup before examples load

Committed as `da816ff` with message: `fix(coordination): dedupe failure recovery policy signals`

viamin:
## Agent Update

OK

The reviewer is correct — `FAILURE_HOOKS` is a misleading name for regex classification patterns. Renaming to `CATEGORY_PATTERNS`.

Now let me run lint and tests.

The spec doesn't reference the constant directly. The rename is clean. Let me verify the final file looks correct:

Now let me commit. The pre-commit hook will run lint/tests — since `bundle install` can't fully resolve in this environment, if the hook fails I'll need to handle it.

The pre-commit hook fails because the full gem bundle isn't available in this container. Let me check the hook and find a workaround.

The hook runs `bin/lint --staged` which needs the full bundle. Since this environment lacks the full gem set, I cannot run the hook. However, the instructions say "Never use `--no-verify`". Let me check if there's a way to install the needed gems.

Bundle installed successfully. Now let me run lint and commit.

Lint passes. Now let me run the relevant spec:

Tests are all pending (they need DB for factories) but no failures. The rename is purely mechanical — no behavior change. Now let me commit:

Commit succeeded with all lint checks passing.

**Summary:** Renamed `FAILURE_HOOKS` to `CATEGORY_PATTERNS` in `app/services/coordination/failure_recovery_service.rb` (lines 5 and 48). The reviewer was correct — "hooks" in Ruby/Rails conventionally refers to lifecycle callbacks, not regex classification patterns.

**Note:** The 3 specs in `spec/services/coordination/failure_recovery_service_spec.rb` were pending due to no database (they use FactoryBot `create`), but they reference the constant only indirectly through `described_class.call`. CI with a database will be the source of truth for those tests.

viamin:

Issue #1667, opened 2026-05-04
Simplify chat setup with sensible defaults


viamin:
## Summary

Opening the chat index page should start a new chat with sensible defaults, reducing the amount of configuration needed to get started.

## Current Behavior

Chat setup requires explicit configuration of mode, model, provider, system prompt, and title before starting.

### Default settings inherited automatically
- **Mode, model, and provider** should inherit from the automatic provider settings (e.g., if the default is Codex with a subscription, that should be used for chat by default)
- **System prompt** should be available in the prompts controller with a sensible default value
- **Title** should be optional — automatically generated from the chat content if not explicitly set by the user

### Minimal required configuration
- Only the **project** needs to be set to start a chat

### In-chat configuration (bottom of chat window)
Similar to the Claude.ai chat interface, the following settings should be configurable from the bottom of the chat window:
- Project selection
- Model selection
- Tier settings

## Tasks

- [ ] Default mode, model, and provider from automatic provider settings
- [ ] Add system prompt to prompts controller with a sensible default
- [ ] Make title optional; auto-generate from chat content when not provided
- [ ] Add bottom-of-chat-window configuration UI for project, model, and tier
- [ ] Ensure opening the chat index page starts a new chat with all defaults applied

viamin:
Pull request created: https://github.com/viamin/paid/pull/1673

PR #1957, opened 2026-05-12
Fix #200001667: [Security] CodeQL: Sensitive data read from GET request (medium) — code-scanning-alert-1667

viamin:
## Summary

Removes the `params[:session_token]` query parameter fallback from `extract_session_token` to resolve CodeQL alert #1667 (rb/sensitive-get-query). Session tokens in GET query strings leak through server access logs, browser history, and HTTP Referer headers, so they must only be transmitted via request headers.

## Changes

- **Controller**: Drop the `params[:session_token]` fallback in `extract_session_token`; session tokens are now accepted exclusively from `X-Session-Token` or `Authorization: Bearer` headers.

## Design Decisions

- **No deprecation window**: Existing tests and clients already use header-based authentication, so the query-param path is dead code in practice. Removing it outright avoids leaving a vulnerable fallback active during a soft-deprecation period.
- **Header-only is the standard**: This aligns with the broader convention that bearer-style credentials should never appear in URLs, which are routinely captured by logging and observability infrastructure.

---

Closes #200001667

### DIFF
--- a/app/controllers/api/mcp_controller.rb
+++ b/app/controllers/api/mcp_controller.rb
@@ -76,10 +76,8 @@ module Api
     end
 
     def extract_session_token
-      # Check header first, then query param
       request.headers["X-Session-Token"].presence ||
-        request.headers["Authorization"]&.delete_prefix("Bearer ")&.presence ||
-        params[:session_token]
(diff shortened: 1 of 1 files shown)


viamin:

PR #1962, opened 2026-05-12
fix(knowledge): skip routes collector on database.yml parse errors

viamin:
## Summary

- When a target repo's `config/database.yml` has YAML syntax errors, `bin/rails routes` fails to boot even though `DATABASE_URL` is set — Rails still parses `database.yml` during boot before using `DATABASE_URL` for the connection
- Added `Cannot load database configuration:` to the routes collector's database error skip patterns so the collector gracefully preserves existing route artifacts instead of failing the entire knowledge collection run
- Added test covering the exact error message from the field report


viamin:

PR #1963, opened 2026-05-12
fix(knowledge): stub database.yml to prevent routes collector boot failures

viamin:
## Summary

- Write a minimal valid `database.yml` stub during container `seed_workspace!`, before the chmod lockdown makes the workspace read-only
- Rails parses `config/database.yml` during boot even when `DATABASE_URL` is set, so any broken YAML (invalid syntax, ERB referencing undefined variables, tabs instead of spaces) prevents `bin/rails routes` from running
- The stub provides valid YAML for all standard Rails environments (`test`, `development`, `production`) while `DATABASE_URL=sqlite3::memory:` overrides the actual connection at runtime
- Only overwrites when the repo already has a `config/database.yml`; non-Rails repos are unaffected
- Prevents the entire class of database config errors at the source, rather than catching each error type individually via message-pattern matching

## Test plan

- [x] 56 containerized runner specs pass (including 5 new: stub content validation, write ordering, file-existence guard, failure handling)
- [x] 48 routes collector specs pass (unchanged — stub is transparent to the collector)
- [x] RuboCop clean

Closes #1962 (supersedes — that PR added error-message matching; this PR prevents the error entirely)

### DIFF
--- a/app/services/knowledge/containerized_runner.rb
+++ b/app/services/knowledge/containerized_runner.rb
@@ -331,8 +331,8 @@ module Knowledge
       exec_setup!(
         [
           "sh", "-c",
-          "if [ -f #{workspace}/config/database.yml ]; then " \
-          "cat > #{workspace}/config/database.yml <<'STUB'\n#{stub_content}\nSTUB\nfi"
+          "if [ -f '#{workspace}/config/database.yml' ]; then " \
+          "cat > '#{workspace}/config/database.yml' <<'STUB'\n#{stub_content}\nSTUB\nfi"
         ],
         "write database.yml stub"
       )
--- a/app/services/knowledge/containerized_runner.rb
+++ b/app/services/knowledge/containerized_runner.rb
@@ -331,7 +331,7 @@ module Knowledge
       exec_setup!(
         [
           "sh", "-c",
-          "if [ -f '#{workspace}/config/database.yml' ]; then " \
+          "if [ -f '#{workspace}/config/database.yml' ] && [ ! -L '#{workspace}/config/database.yml' ]; then " \
           "cat > '#{workspace}/config/database.yml' <<'STUB'\n#{stub_content}\nSTUB\nfi"
         ],
         "write database.yml stub"
--- a/app/services/knowledge/containerized_runner.rb
+++ b/app/services/knowledge/containerized_runner.rb
@@ -52,6 +52,26 @@ module Knowledge
     PROJECT_NAME_PATTERN = %r{\A[a-zA-Z0-9._-]+/[a-zA-Z0-9._-]+\z}
     ALLOWED_NETWORK_MODES = %w[none bridge].freeze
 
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

Addressed the unresolved review thread by quoting the interpolated `workspace` path in the only `sh -c` setup command, and updated the matching spec expectation in [app/services/knowledge/containerized_runner.rb](/workspace/app/services/knowledge/containerized_runner.rb:334) and [spec/services/knowledge/containerized_runner_spec.rb](/workspace/spec/services/knowledge/containerized_runner_spec.rb:303). I also re-checked the diff for the same issue class; this was the only shell-interpolated workspace path in the change.

Committed on the PR branch as `8e3665f` with message `fix(knowledge): quote database stub workspace path`.

Verification:
- `bundle install` succeeded after configuring a local bundle path
- `YARN_CACHE_FOLDER=/tmp/yarn-cache yarn install` succeeded
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/services/knowledge/containerized_runner_spec.rb` passed

I also attempted `bundle exec rspec` and `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec`, but the full suite cannot complete in this container because there is no PostgreSQL service and at least one spec file (`spec/services/dashboard/provider_health_spec.rb`) triggers a DB connection during file load, before the DB-less skip path can apply.

app/services/knowledge/containerized_runner.rb
```diff
@@ -298,6 +326,18 @@ def exec_setup!(command, failing_action)
       raise ContainerError, "Failed to #{failing_action} (exit #{status})"
     end
 
+    def write_database_yml_stub!(workspace)
+      stub_content = DATABASE_YML_STUB.chomp
+      exec_setup!(
+        [
+          "sh", "-c",
+          "if [ -f '#{workspace}/config/database.yml' ]; then " \
```
viamin: Please refuse to overwrite symlinks here. [ -f ] still passes for a symlink to a regular file, so this root redirect will follow config/database.yml if the repo checks it in as a symlink and overwrite the link target instead of only the workspace copy. That can both break repos that share DB config via symlink and trample arbitrary writable paths inside the container. Add a ! -L guard, or otherwise replace the file without following symlinks, before redirecting into it.

viamin (commented):
## Code Review

The new database.yml stub logic needs one safety guard before this is ready: it currently follows symlinks when overwriting config/database.yml, so a repo that checks that file in as a symlink will overwrite the link target instead of only replacing the workspace copy.

viamin:
## Agent Update

OK

Commit succeeded. The change adds a `! -L` symlink guard to `write_database_yml_stub!` so that if `config/database.yml` is a symlink, the stub write is skipped — preventing the shell redirect from following the symlink and overwriting the link target. The spec was updated to verify the symlink check is present. All 56 containerized runner specs pass and rubocop is clean.

viamin:

PR #1975, opened 2026-05-12
Fix #1964: Answer issue clarifying questions via Paid UI wizard

### LINKED ISSUE #1964: Answer issue clarifying questions via Paid UI wizard
## Summary

Add an in-app option for users to answer issue enhancement clarifying questions directly in the Paid UI, as an alternative to commenting on the GitHub issue.

## Background

Currently, the issue enhancement flow works as follows:
1. A paid agent asks clarifying questions and adds them as a comment on the issue
2. The user adds their own comment with answers to the questions
3. The next agent run picks up from there

This requires users to leave the Paid app, go to GitHub, and respond in a comment. We can streamline this by surfacing unanswered clarifying questions in the Paid UI and letting users answer them in-app.

### Surface issues needing answers
- Issues that have unanswered clarifying questions should be surfaced in the Paid UI (project page and potentially the dashboard)
- There should be a clear visual indicator that an issue is waiting for user input

### Clarifying questions wizard
- Present questions in a wizard-style UI, similar to the business context knowledge base collection wizard
- Show **one question at a time** with a progress bar indicating how many questions remain
- Allow the user to navigate **forward and back** through questions
- All questions must be answered before the wizard can be completed (no partial submissions)

### On completion
- When the user finishes answering all questions, Paid adds a comment to the GitHub issue with the formatted answers
- The existing agent flow picks up from there as it does today (no changes to the agent-side flow)

## Where to surface
- **Project page** — primary location
- **Dashboard** — secondary location (potentially)

## Acceptance Criteria
- [ ] Issues with unanswered clarifying questions are visibly surfaced in the Paid UI
- [ ] Wizard presents one question at a time with a progress bar
- [ ] User can navigate forward and back through questions
- [ ] All questions must be answered to submit
- [ ] On completion, a comment with the answers is posted to the GitHub issue
- [ ] The existing agent flow picks up the answers as before
- [ ] Wizard UX is consistent with the business context knowledge base collection wizard

viamin:
## Summary

Lets users answer issue clarifying questions directly in the Paid UI instead of leaving for GitHub, streamlining the issue enhancement flow. Unanswered questions are surfaced on issue rows, and a wizard walks the user through one question at a time before posting the formatted answers back to the GitHub issue so the existing agent flow can resume unchanged.

## Changes

- **Models**: Added `needs_input?` predicate to `Issue` to flag issues waiting on user answers to clarifying questions.
- **Services**: New service objects to load unanswered clarifying questions for an issue and to post the formatted answer comment to GitHub on wizard completion.
- **Controller & routes**: New controller and route to render the wizard and accept submitted answers; delegates persistence/comment posting to the services above.
- **Views**: Added a wizard view that presents one question at a time with a progress bar and forward/back navigation, modeled on the business context knowledge base collection wizard. Updated the issue partial to show a visual indicator when `needs_input?` is true, linking into the wizard.
- **JavaScript**: New Stimulus controller driving wizard navigation, progress, and the submit gate that requires every question to be answered before completion.
- **Tests**: Added model spec for `needs_input?`, service specs for question loading and GitHub comment posting, and a request spec covering the wizard controller (rendering, navigation, and submission).

## Design Decisions

- **No agent-side changes** — On submission, the wizard posts a formatted answer comment to the GitHub issue using the existing agent flow's comment-driven pickup, keeping the change additive and avoiding any divergence between in-app and GitHub answer paths.
- **All-or-nothing submission** — The wizard enforces that every question is answered before submit, matching the acceptance criteria and avoiding partial-state handling on the agent side.
- **UX parity with the business context wizard** — Reused the existing wizard pattern (one question per step, progress bar, forward/back) for consistency rather than introducing a new pattern.
- **Predicate on `Issue`** — Added `needs_input?` rather than introducing a new state column, since "needs input" is derivable from existing clarifying-question state and the agent flow already drives the underlying lifecycle.

## Screenshots

_Please attach before/after screenshots of the project page issue row indicator and the wizard (first question, mid-progress, and completion states)._

---

Closes #1964

### DIFF
--- a/app/controllers/projects/clarifying_questions_controller.rb
+++ b/app/controllers/projects/clarifying_questions_controller.rb
@@ -51,6 +51,9 @@ module Projects
     end
 
     def fetch_questions
+      questions = ClarifyingQuestions::Parse.call(comment_body: @issue.body)
+      return questions if questions.any?
+
       client = @project.github_token.client
       comments = client.issue_comments(@project.full_name, @issue.github_number)
       enhancement_comment = comments.reverse.find do |comment|
--- a/app/controllers/projects/clarifying_questions_controller.rb
+++ b/app/controllers/projects/clarifying_questions_controller.rb
@@ -29,9 +29,9 @@ module Projects
 
       redirect_to project_path(@project), notice: "Answers posted to GitHub issue ##{@issue.github_number}. The agent will pick them up on the next run."
     rescue ArgumentError => e
-      redirect_to project_issue_clarifying_question_path(@project, @issue), alert: e.message
+      redirect_to project_issue_clarifying_questions_path(@project, @issue), alert: e.message
     rescue GithubClient::Error => e
-      redirect_to project_issue_clarifying_question_path(@project, @issue), alert: "Failed to post answers: #{e.message}"
+      redirect_to project_issue_clarifying_questions_path(@project, @issue), alert: "Failed to post answers: #{e.message}"
     end
 
     private
--- a/app/controllers/projects/clarifying_questions_controller.rb
+++ b/app/controllers/projects/clarifying_questions_controller.rb
@@ -9,15 +9,17 @@ module Projects
     before_action :authorize_project_update, only: [ :create ]
 
     def show
-      @questions = fetch_questions
+      @questions = ClarifyingQuestions::Load.call(project: @project, issue: @issue)
 
       if @questions.empty?
         redirect_to project_path(@project), alert: "No clarifying questions found for issue ##{@issue.github_number}."
       end
+    rescue GithubClient::Error => e
+      redirect_to project_path(@project), alert: "Failed to load clarifying questions: #{e.message}"
     end
 
     def create
-      questions_and_answers = build_questions_and_answers
+      questions_and_answers = build_questions_and_answers(questions: current_questions)
 
       ClarifyingQuestions::SubmitAnswers.call(
         project: @project,
@@ -50,24 +52,13 @@ module Projects
       authorize @project, :update?
     end
 
-    def fetch_questions
-      questions = ClarifyingQuestions::Parse.call(comment_body: @issue.body)
-      return questions if questions.any?
-
-      client = @project.github_token.client
-      comments = client.issue_comments(@project.full_name, @issue.github_number)
-      enhancement_comment = comments.reverse.find do |comment|
-        comment.body.to_s.include?(ClarifyingQuestions::Parse::ENHANCEMENT_MARKER) &&
-          comment.body.to_s.include?("## Clarifying questions")
-      end
-      return [] unless enhancement_comment
-
-      ClarifyingQuestions::Parse.call(comment_body: enhancement_comment.body)
+    def current_questions
+      @current_questions ||= ClarifyingQuestions::Load.call(project: @project, issue: @issue)
     end
 
-    def build_questions_and_answers
-      questions = params[:questions] || []
+    def build_questions_and_answers(questions:)
       answers = params[:answers] || []
+      raise ArgumentError, "No clarifying questions found for this issue." if questions.empty?
 
       questions.each_with_index.map do |question, i|
         { question: question, answer: answers[i].to_s.strip }
--- a/app/controllers/projects/clarifying_questions_controller.rb
+++ b/app/controllers/projects/clarifying_questions_controller.rb
@@ -57,8 +57,13 @@ module Projects
     end
 
     def build_questions_and_answers(questions:)
-      answers = params[:answers] || []
+      answers = Array(params[:answers])
+      submitted_questions = Array(params[:questions]).map { |question| question.to_s.strip }
+
       raise ArgumentError, "No clarifying questions found for this issue." if questions.empty?
+      unless submitted_questions == questions
+        raise ArgumentError, "Clarifying questions changed. Please reload and try again."
+      end
 
       questions.each_with_index.map do |question, i|
         { question: question, answer: answers[i].to_s.strip }
--- a/app/controllers/projects/clarifying_questions_controller.rb
+++ b/app/controllers/projects/clarifying_questions_controller.rb
@@ -27,9 +27,9 @@ module Projects
 
       redirect_to project_path(@project), notice: "Answers posted to GitHub issue ##{@issue.github_number}. The agent will pick them up on the next run."
     rescue ArgumentError => e
-      redirect_to project_issue_clarifying_questions_path(@project, @issue), alert: e.message
+      redirect_to project_issue_clarifying_question_path(@project, @issue), alert: e.message
     rescue GithubClient::Error => e
-      redirect_to project_issue_clarifying_questions_path(@project, @issue), alert: "Failed to post answers: #{e.message}"
+      redirect_to project_issue_clarifying_question_path(@project, @issue), alert: "Failed to post answers: #{e.message}"
     end
 
     private
--- a/app/controllers/projects/clarifying_questions_controller.rb
+++ b/app/controllers/projects/clarifying_questions_controller.rb
@@ -1,0 +1,74 @@
+# frozen_string_literal: true
+
+module Projects
+  class ClarifyingQuestionsController < ApplicationController
+    before_action :authenticate_user!
+    before_action :set_project
+    before_action :set_issue
+    before_action :authorize_project_show, only: [ :show ]
+    before_action :authorize_project_update, only: [ :create ]
+
+    def show
+      @questions = fetch_questions
+
+      if @questions.empty?
+        redirect_to project_path(@project), alert: "No clarifying questions found for issue ##{@issue.github_number}."
+      end
+    end
+
+    def create
+      questions_and_answers = build_questions_and_answers
+
+      ClarifyingQuestions::SubmitAnswers.call(
+        project: @project,
+        issue: @issue,
+        questions_and_answers: questions_and_answers
+      )
+
+      redirect_to project_path(@project), notice: "Answers posted to GitHub issue ##{@issue.github_number}. The agent will pick them up on the next run."
+    rescue ArgumentError => e
+      redirect_to project_issue_clarifying_questions_path(@project, @issue), alert: e.message
+    rescue GithubClient::Error => e
+      redirect_to project_issue_clarifying_questions_path(@project, @issue), alert: "Failed to post answers: #{e.message}"
+    end
+
+    private
+
+    def set_project
+      @project = policy_scope(Project).find(params[:project_id])
+    end
+
+    def set_issue
+      @issue = @project.issues.issues_only.find(params[:issue_id])
+    end
+
+    def authorize_project_show
+      authorize @project, :show?
+    end
+
+    def authorize_project_update
+      authorize @project, :update?
+    end
+
+    def fetch_questions
+      client = @project.github_token.client
+      comments = client.issue_comments(@project.full_name, @issue.github_number)
+      enhancement_comment = comments.reverse.find do |comment|
+        comment.body.to_s.include?(ClarifyingQuestions::Parse::ENHANCEMENT_MARKER) &&
+          comment.body.to_s.include?("## Clarifying questions")
+      end
+      return [] unless enhancement_comment
+
+      ClarifyingQuestions::Parse.call(comment_body: enhancement_comment.body)
+    end
+
+    def build_questions_and_answers
+      questions = params[:questions] || []
+      answers = params[:answers] || []
+
+      questions.each_with_index.map do |question, i|
+        { question: question, answer: answers[i].to_s.strip }
+      end
+    end
+  end
+end
--- a/app/javascript/controllers/clarifying_questions_controller.js
+++ b/app/javascript/controllers/clarifying_questions_controller.js
@@ -1,0 +1,65 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["step", "answer", "progressBar", "progressAnswered", "progressPercent", "form", "submitButton"]
+  static values = { total: Number }
+
+  connect() {
+    this.currentIndex = 0
+    this.updateProgress()
+  }
+
+  next() {
+    if (!this.validateCurrentStep()) return
+    if (this.currentIndex < this.stepTargets.length - 1) {
+      this.currentIndex++
+      this.showStep(this.currentIndex)
+    }
+  }
+
+  previous() {
+    if (this.currentIndex > 0) {
+      this.currentIndex--
+      this.showStep(this.currentIndex)
+    }
+  }
+
+  showStep(index) {
+    this.stepTargets.forEach((step, i) => {
+      step.style.display = i === index ? "" : "none"
+    })
+    this.updateProgress()
+  }
+
+  validateCurrentStep() {
+    const currentStep = this.stepTargets[this.currentIndex]
+    if (!currentStep) return true
+
+    const textarea = currentStep.querySelector("textarea")
+    if (textarea && textarea.required && !textarea.value.trim()) {
+      textarea.focus()
+      textarea.classList.add("border-red-500")
+      return false
+    }
+    if (textarea) {
+      textarea.classList.remove("border-red-500")
+    }
+    return true
+  }
+
+  updateProgress() {
+    const total = this.totalValue
+    const answered = this.answerTargets.filter(a => a.value.trim().length > 0).length
+    const percent = total > 0 ? Math.round((answered / total) * 100) : 0
+
+    if (this.hasProgressBarTarget) {
+      this.progressBarTarget.style.width = `${percent}%`
+    }
+    if (this.hasProgressAnsweredTarget) {
+      this.progressAnsweredTarget.textContent = answered
+    }
+    if (this.hasProgressPercentTarget) {
+      this.progressPercentTarget.textContent = percent
+    }
+  }
+}
--- a/app/javascript/controllers/clarifying_questions_controller.js
+++ b/app/javascript/controllers/clarifying_questions_controller.js
@@ -24,6 +24,13 @@ export default class extends Controller {
     }
   }
 
+  submit(event) {
+    if (this.allAnswersPresent()) return
+
+    event.preventDefault()
+    this.focusFirstBlankAnswer()
+  }
+
   showStep(index) {
     this.stepTargets.forEach((step, i) => {
       step.style.display = i === index ? "" : "none"
@@ -49,8 +56,13 @@ export default class extends Controller {
 
   updateProgress() {
     const total = this.totalValue
-    const answered = this.answerTargets.filter(a => a.value.trim().length > 0).length
+    const answered = this.answerTargets.filter((answer) => {
+      const present = answer.value.trim().length > 0
+      if (present) answer.classList.remove("border-red-500")
+      return present
+    }).length
     const percent = total > 0 ? Math.round((answered / total) * 100) : 0
+    const allAnswered = total > 0 && answered === total
 
     if (this.hasProgressBarTarget) {
       this.progressBarTarget.style.width = `${percent}%`
@@ -61,5 +73,24 @@ export default class extends Controller {
     if (this.hasProgressPercentTarget) {
       this.progressPercentTarget.textContent = percent
     }
+    if (this.hasSubmitButtonTarget) {
+      this.submitButtonTarget.disabled = !allAnswered
+      this.submitButtonTarget.setAttribute("aria-disabled", (!allAnswered).toString())
+    }
+  }
+
+  allAnswersPresent() {
+    return this.answerTargets.every(answer => answer.value.trim().length > 0)
+  }
+
+  focusFirstBlankAnswer() {
+    const blankAnswer = this.answerTargets.find(answer => answer.value.trim().length === 0)
+    if (!blankAnswer) return
+
+    const stepIndex = Number(blankAnswer.dataset.stepIndex)
+    this.currentIndex = stepIndex
+    this.showStep(stepIndex)
+    blankAnswer.classList.add("border-red-500")
+    blankAnswer.focus()
   }
 }
--- a/app/javascript/controllers/index.js
+++ b/app/javascript/controllers/index.js
@@ -28,6 +28,9 @@ application.register("chat-session-list", ChatSessionListController)
 import ChatStreamController from "./chat_stream_controller"
 application.register("chat-stream", ChatStreamController)
 
+import ClarifyingQuestionsController from "./clarifying_questions_controller"
+application.register("clarifying-questions", ClarifyingQuestionsController)
+
 import ConfirmDeleteController from "./confirm_delete_controller"
 application.register("confirm-delete", ConfirmDeleteController)
 
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -193,7 +193,7 @@ class Issue < ApplicationRecord
   end
 
   def needs_input?
-    paid_state == "needs_input"
+    paid_state == "needs_input" && has_label?(project.enhance_issue_needs_input_label_name)
   end
 
   def draft_phase?
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -192,6 +192,10 @@ class Issue < ApplicationRecord
     end
   end
 
+  def needs_input?
+    paid_state == "needs_input"
+  end
+
   def draft_phase?
     pr_review_phase.in?(%w[draft restarted])
   end
--- a/app/services/clarifying_questions/load.rb
+++ b/app/services/clarifying_questions/load.rb
@@ -45,6 +45,7 @@ module ClarifyingQuestions
       latest_answer_comment = latest_answer_comment()
       return false unless latest_answer_comment
       return true if body_questions.any? && enhancement_comment.blank?
+      return false if enhancement_comment.blank?
 
       comment_timestamp(latest_answer_comment) > comment_timestamp(enhancement_comment)
     end
--- a/app/services/clarifying_questions/load.rb
+++ b/app/services/clarifying_questions/load.rb
@@ -15,10 +15,9 @@ module ClarifyingQuestions
 
     def call
       body_questions = Parse.call(comment_body: issue.body)
-      return body_questions if body_questions.any? && !github_available?
-
       enhancement_comment = latest_enhancement_comment
-      return [] if answered_after_enhancement?(enhancement_comment)
+      return body_questions if body_questions.any? && !github_available?
+      return [] if answered_after_latest_questions?(body_questions:, enhancement_comment:)
       return body_questions if body_questions.any?
       return [] unless enhancement_comment
 
@@ -42,15 +41,18 @@ module ClarifyingQuestions
       end
     end
 
-    def answered_after_enhancement?(enhancement_comment)
-      return false unless enhancement_comment
-
-      latest_answer_comment = issue_comments.reverse.find do |comment|
-        comment_body(comment).include?(ANSWER_MARKER)
-      end
+    def answered_after_latest_questions?(body_questions:, enhancement_comment:)
+      latest_answer_comment = latest_answer_comment()
       return false unless latest_answer_comment
+      return true if body_questions.any? && enhancement_comment.blank?
 
       comment_timestamp(latest_answer_comment) > comment_timestamp(enhancement_comment)
+    end
+
+    def latest_answer_comment
+      issue_comments.reverse.find do |comment|
+        comment_body(comment).include?(ANSWER_MARKER)
+      end
     end
 
     def comment_body(comment)
--- a/app/services/clarifying_questions/load.rb
+++ b/app/services/clarifying_questions/load.rb
@@ -54,10 +54,23 @@ module ClarifyingQuestions
     def answered_after_latest_questions?(body_questions:, enhancement_comment:)
       latest_answer_comment = latest_answer_comment()
       return false unless latest_answer_comment
-      return true if body_questions.any? && enhancement_comment.blank?
-      return false if enhancement_comment.blank?
 
-      comment_timestamp(latest_answer_comment) > comment_timestamp(enhancement_comment)
+      latest_answer_timestamp = comment_timestamp(latest_answer_comment)
+      latest_question_timestamp = latest_question_timestamp(
+        body_questions: body_questions,
+        enhancement_comment: enhancement_comment
+      )
+      return false unless latest_question_timestamp
+
+      latest_answer_timestamp > latest_question_timestamp
+    end
+
+    def latest_question_timestamp(body_questions:, enhancement_comment:)
+      timestamps = []
+      timestamps << issue_timestamp if body_questions.any?
+      timestamps << comment_timestamp(enhancement_comment) if enhancement_comment.present?
+
+      timestamps.compact.max
     end
 
     def latest_answer_comment
@@ -72,6 +85,10 @@ module ClarifyingQuestions
 
     def comment_timestamp(comment)
       comment.created_at&.to_time || Time.at(0)
+    end
+
+    def issue_timestamp
+      issue.github_updated_at&.to_time || issue.updated_at&.to_time || Time.at(0)
     end
 
     def issue_comments
--- a/app/services/clarifying_questions/load.rb
+++ b/app/services/clarifying_questions/load.rb
@@ -2,6 +2,8 @@
 
 module ClarifyingQuestions
   class Load
+    ANSWER_MARKER = "<!-- paid:clarifying-answers -->"
+
     def self.call(...)
       new(...).call
     end
@@ -12,22 +14,52 @@ module ClarifyingQuestions
     end
 
     def call
-      questions = Parse.call(comment_body: issue.body)
-      return questions if questions.any?
-      return [] unless project.github_token
+      body_questions = Parse.call(comment_body: issue.body)
+      return body_questions if body_questions.any? && !github_available?
 
-      enhancement_comment = issue_comments.reverse.find do |comment|
-        comment.body.to_s.include?(Parse::ENHANCEMENT_MARKER) &&
-          comment.body.to_s.include?("## Clarifying questions")
-      end
+      enhancement_comment = latest_enhancement_comment
+      return [] if answered_after_enhancement?(enhancement_comment)
+      return body_questions if body_questions.any?
       return [] unless enhancement_comment
 
-      Parse.call(comment_body: enhancement_comment.body)
+      Parse.call(comment_body: comment_body(enhancement_comment))
     end
 
     private
 
     attr_reader :project, :issue
+
+    def github_available?
+      project.github_token.present?
+    end
+
+    def latest_enhancement_comment
+      return unless github_available?
+
+      issue_comments.reverse.find do |comment|
+        comment_body(comment).include?(Parse::ENHANCEMENT_MARKER) &&
+          comment_body(comment).include?("## Clarifying questions")
+      end
+    end
+
+    def answered_after_enhancement?(enhancement_comment)
+      return false unless enhancement_comment
+
+      latest_answer_comment = issue_comments.reverse.find do |comment|
+        comment_body(comment).include?(ANSWER_MARKER)
+      end
+      return false unless latest_answer_comment
+
+      comment_timestamp(latest_answer_comment) > comment_timestamp(enhancement_comment)
+    end
+
+    def comment_body(comment)
+      comment.body.to_s
+    end
+
+    def comment_timestamp(comment)
+      comment.created_at&.to_time || Time.at(0)
+    end
 
     def issue_comments
       project.github_token.client.issue_comments(project.full_name, issue.github_number)
--- a/app/services/clarifying_questions/load.rb
+++ b/app/services/clarifying_questions/load.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module ClarifyingQuestions
+  class Load
+    def self.call(...)
+      new(...).call
+    end
+
+    def initialize(project:, issue:)
+      @project = project
+      @issue = issue
+    end
+
+    def call
+      questions = Parse.call(comment_body: issue.body)
+      return questions if questions.any?
+      return [] unless project.github_token
+
+      enhancement_comment = issue_comments.reverse.find do |comment|
+        comment.body.to_s.include?(Parse::ENHANCEMENT_MARKER) &&
+          comment.body.to_s.include?("## Clarifying questions")
+      end
+      return [] unless enhancement_comment
+
+      Parse.call(comment_body: enhancement_comment.body)
+    end
+
+    private
+
+    attr_reader :project, :issue
+
+    def issue_comments
+      project.github_token.client.issue_comments(project.full_name, issue.github_number)
+    end
+  end
+end
--- a/app/services/clarifying_questions/load.rb
+++ b/app/services/clarifying_questions/load.rb
@@ -15,10 +15,20 @@ module ClarifyingQuestions
 
     def call
       body_questions = Parse.call(comment_body: issue.body)
+
+      if body_questions.any?
+        return body_questions unless github_available?
+
+        enhancement_comment = latest_enhancement_comment
+        return [] if answered_after_latest_questions?(body_questions:, enhancement_comment:)
+
+        return body_questions
+      end
+
+      return [] unless github_available?
+
       enhancement_comment = latest_enhancement_comment
-      return body_questions if body_questions.any? && !github_available?
       return [] if answered_after_latest_questions?(body_questions:, enhancement_comment:)
-      return body_questions if body_questions.any?
       return [] unless enhancement_comment
 
       Parse.call(comment_body: comment_body(enhancement_comment))
@@ -65,7 +75,7 @@ module ClarifyingQuestions
     end
 
     def issue_comments
-      project.github_token.client.issue_comments(project.full_name, issue.github_number)
+      @issue_comments ||= project.github_token.client.issue_comments(project.full_name, issue.github_number)
     end
   end
 end
--- a/app/services/clarifying_questions/load.rb
+++ b/app/services/clarifying_questions/load.rb
@@ -32,6 +32,10 @@ module ClarifyingQuestions
       return [] unless enhancement_comment
 
       Parse.call(comment_body: comment_body(enhancement_comment))
+    rescue GithubClient::Error
+      raise if body_questions.empty?
+
+      body_questions
     end
 
     private
--- a/app/services/clarifying_questions/parse.rb
+++ b/app/services/clarifying_questions/parse.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+module ClarifyingQuestions
+  class Parse
+    ENHANCEMENT_MARKER = "<!-- paid:enhance-issue -->"
+    CLARIFYING_SECTION_PATTERN = /## Clarifying questions\s*\n(.+?)(?=\n## |\z)/m.freeze
+
+    def self.call(...)
+      new(...).call
+    end
+
+    def initialize(comment_body:)
+      @comment_body = comment_body.to_s
+    end
+
+    def call
+      return [] unless comment_body.include?(ENHANCEMENT_MARKER)
+      return [] unless comment_body.include?("## Clarifying questions")
+
+      section = extract_clarifying_section
+      return [] unless section
+
+      parse_numbered_items(section)
+    end
+
+    private
+
+    attr_reader :comment_body
+
+    def extract_clarifying_section
+      match = comment_body.match(CLARIFYING_SECTION_PATTERN)
+      match ? match[1] : nil
+    end
+
+    def parse_numbered_items(text)
+      questions = []
+      current = nil
+
+      text.each_line do |line|
+        stripped = line.strip
+        if (match = stripped.match(/\A\d+\.\s+(.+)\z/))
+          questions << current if current
+          current = match[1]
+        elsif current.present? && stripped.present?
+          current += " " + stripped
+        end
+      end
+      questions << current if current
+
+      questions
+    end
+  end
+end
--- a/app/services/clarifying_questions/submit_answers.rb
+++ b/app/services/clarifying_questions/submit_answers.rb
@@ -23,6 +23,13 @@ module ClarifyingQuestions
     attr_reader :project, :issue, :questions_and_answers
 
     def validate_answers!
+      raise ArgumentError, "No clarifying questions found for this issue." if questions_and_answers.empty?
+
+      missing_questions = questions_and_answers.select { |qa| qa[:question].blank? }
+      if missing_questions.any?
+        raise ArgumentError, "Clarifying question text is missing for #{missing_questions.size} item(s)."
+      end
+
       missing = questions_and_answers.select { |qa| qa[:answer].blank? }
       return if missing.empty?
 
--- a/app/services/clarifying_questions/submit_answers.rb
+++ b/app/services/clarifying_questions/submit_answers.rb
@@ -2,6 +2,8 @@
 
 module ClarifyingQuestions
   class SubmitAnswers
+    ANSWER_MARKER = Load::ANSWER_MARKER
+
     def self.call(...)
       new(...).call
     end
@@ -14,7 +16,7 @@ module ClarifyingQuestions
 
     def call
       validate_answers!
-      client = project.github_token.client
+      client = github_client
       client.add_comment(project.full_name, issue.github_number, formatted_comment)
     end
 
@@ -24,6 +26,7 @@ module ClarifyingQuestions
 
     def validate_answers!
       raise ArgumentError, "No clarifying questions found for this issue." if questions_and_answers.empty?
+      raise ArgumentError, "GitHub access is not configured for this project." unless github_client
 
       missing_questions = questions_and_answers.select { |qa| qa[:question].blank? }
       if missing_questions.any?
@@ -37,13 +40,17 @@ module ClarifyingQuestions
     end
 
     def formatted_comment
-      parts = [ "<!-- paid:clarifying-answers -->", "", "## Clarifying question answers", "" ]
+      parts = [ ANSWER_MARKER, "", "## Clarifying question answers", "" ]
       questions_and_answers.each_with_index do |qa, i|
         parts << "**Q#{i + 1}: #{qa[:question]}**"
-        parts << qa[:answer]
+        parts << "**A#{i + 1}:** #{qa[:answer]}"
         parts << ""
       end
       parts.join("\n")
+    end
+
+    def github_client
+      project.github_token&.client
     end
   end
 end
--- a/app/services/clarifying_questions/submit_answers.rb
+++ b/app/services/clarifying_questions/submit_answers.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+module ClarifyingQuestions
+  class SubmitAnswers
+    def self.call(...)
+      new(...).call
+    end
+
+    def initialize(project:, issue:, questions_and_answers:)
+      @project = project
+      @issue = issue
+      @questions_and_answers = questions_and_answers
+    end
+
+    def call
+      validate_answers!
+      client = project.github_token.client
+      client.add_comment(project.full_name, issue.github_number, formatted_comment)
+    end
+
+    private
+
+    attr_reader :project, :issue, :questions_and_answers
+
+    def validate_answers!
+      missing = questions_and_answers.select { |qa| qa[:answer].blank? }
+      return if missing.empty?
+
+      raise ArgumentError, "All questions must be answered. #{missing.size} question(s) are blank."
+    end
+
+    def formatted_comment
+      parts = [ "<!-- paid:clarifying-answers -->", "", "## Clarifying question answers", "" ]
+      questions_and_answers.each_with_index do |qa, i|
+        parts << "**Q#{i + 1}: #{qa[:question]}**"
+        parts << qa[:answer]
+        parts << ""
+      end
+      parts.join("\n")
+    end
+  end
+end
--- a/app/services/screenshots/capture_targets.rb
+++ b/app/services/screenshots/capture_targets.rb
@@ -226,6 +226,11 @@ module Screenshots
     }.freeze
 
     def targets_for(path)
+      if path.start_with?("app/javascript/controllers/")
+        explicit_targets = targets_for_javascript_controller(path.delete_prefix("app/javascript/controllers/"))
+        return explicit_targets if explicit_targets.any?
+      end
+
       return SHARED_TARGET_KEYS if shared_ui_file?(path)
 
       if path.start_with?("app/views/")
@@ -252,6 +257,14 @@ module Screenshots
         path == "app/views/layouts/application.html.erb" ||
         path.start_with?("app/views/shared/") ||
         path == "app/helpers/application_helper.rb"
+    end
+
+    def targets_for_javascript_controller(relative_path)
+      case relative_path
+      when "clarifying_questions_controller.js" then [ :project_issue_clarifying_questions ]
+      else
+        []
+      end
     end
 
     def targets_for_helper(path)
--- a/app/services/screenshots/capture_targets.rb
+++ b/app/services/screenshots/capture_targets.rb
@@ -275,7 +275,7 @@ module Screenshots
         .flat_map { |file| targets_for_javascript_controller(file.delete_prefix("app/javascript/controllers/")) }
         .uniq
 
-      return nil if explicit_targets.any?
+      return explicit_targets if explicit_targets.any?
 
       SHARED_TARGET_KEYS
     end
--- a/app/services/screenshots/capture_targets.rb
+++ b/app/services/screenshots/capture_targets.rb
@@ -148,8 +148,8 @@ module Screenshots
       return materialize(SHARED_TARGET_KEYS) if @changed_files.empty?
 
       mapping = @changed_files.each_with_object({}) { |path, h| h[path] = targets_for(path) }
-      target_keys = mapping.values.flatten.uniq
-      unresolved_files = mapping.select { |_, targets| targets.empty? }.keys
+      target_keys = mapping.values.compact.flatten.uniq
+      unresolved_files = mapping.select { |_, targets| targets == [] }.keys
 
       if unresolved_files.any?
         raise UnmappedUiChangeError,
@@ -226,6 +226,8 @@ module Screenshots
     }.freeze
 
     def targets_for(path)
+      return targets_for_javascript_registry if path == "app/javascript/controllers/index.js"
+
       if path.start_with?("app/javascript/controllers/")
         explicit_targets = targets_for_javascript_controller(path.delete_prefix("app/javascript/controllers/"))
         return explicit_targets if explicit_targets.any?
@@ -265,6 +267,17 @@ module Screenshots
       else
         []
       end
+    end
+
+    def targets_for_javascript_registry
+      explicit_targets = @changed_files
+        .grep(%r{\Aapp/javascript/controllers/(?!index\.js\z)})
+        .flat_map { |file| targets_for_javascript_controller(file.delete_prefix("app/javascript/controllers/")) }
+        .uniq
+
+      return nil if explicit_targets.any?
+
+      SHARED_TARGET_KEYS
     end
 
     def targets_for_helper(path)
--- a/app/services/screenshots/capture_targets.rb
+++ b/app/services/screenshots/capture_targets.rb
@@ -109,6 +109,13 @@ module Screenshots
       projects: Target.new(slug: "projects", path_builder: "/projects", requires_auth: true),
       project_new: Target.new(slug: "project_new", path_builder: "/projects/new", requires_auth: true),
       project_show: Target.new(slug: "project_show", path_builder: ->(seed_data) { "/projects/#{seed_data.fetch(:project).id}" }, requires_auth: true),
+      project_issue_clarifying_questions: Target.new(
+        slug: "project_issue_clarifying_questions",
+        path_builder: ->(seed_data) {
+          "/projects/#{seed_data.fetch(:project).id}/issues/#{seed_data.fetch(:clarifying_issue).id}/clarifying_questions"
+        },
+        requires_auth: true
+      ),
       project_edit: Target.new(slug: "project_edit", path_builder: ->(seed_data) { "/projects/#{seed_data.fetch(:project).id}/edit" }, requires_auth: true),
       project_agent_runs: Target.new(slug: "project_agent_runs", path_builder: ->(seed_data) { "/projects/#{seed_data.fetch(:project).id}/agent_runs" }, requires_auth: true),
       project_agent_run_new: Target.new(slug: "project_agent_run_new", path_builder: ->(seed_data) { "/projects/#{seed_data.fetch(:project).id}/agent_runs/new" }, requires_auth: true),
@@ -203,6 +210,7 @@ module Screenshots
       "knowledge/artifacts_controller.rb" => [ :project_knowledge_artifact_show ],
       "knowledge/context_intake_controller.rb" => [ :project_context_intake ],
       "projects/cost_budgets_controller.rb" => [ :project_cost_dashboard ],
+      "projects/clarifying_questions_controller.rb" => [ :project_issue_clarifying_questions ],
       "projects/issue_merge_subscriptions_controller.rb" => [ :project_show ],
       "projects/pr_templates_controller.rb" => [ :project_edit ],
       "projects/pre_commit_requirements_controller.rb" => [ :project_edit ],
@@ -378,6 +386,7 @@ module Screenshots
       when "new.html.erb" then [ :project_new ]
       when "show.html.erb" then [ :project_show ]
       when "edit.html.erb" then [ :project_edit ]
+      when "clarifying_questions/show.html.erb" then [ :project_issue_clarifying_questions ]
       when /\A_/
         base = File.basename(leaf, ".html.erb")
         if PROJECT_SHOW_PARTIALS.include?(base)
--- a/app/services/screenshots/seed_data/paid.rb
+++ b/app/services/screenshots/seed_data/paid.rb
@@ -46,6 +46,27 @@ module Screenshots
             record.allowed_github_usernames = [ user.email ]
           end
 
+          clarifying_issue = project.issues.find_or_create_by!(github_issue_id: 9_999_991) do |record|
+            record.github_number = 1964
+            record.title = "Answer issue clarifying questions via Paid UI wizard"
+            record.body = <<~MARKDOWN
+              <!-- paid:enhance-issue -->
+
+              ## Clarifying questions
+              1. What is the desired end-to-end behavior after the user submits answers?
+              2. Should the wizard allow editing previous answers before submission?
+
+              ## Current context
+              - Waiting for a trusted collaborator to answer in the Paid UI.
+            MARKDOWN
+            record.github_creator_login = user.email
+            record.github_state = "open"
+            record.github_created_at = 1.day.ago
+            record.github_updated_at = Time.current
+            record.paid_state = "needs_input"
+            record.labels = [ project.enhance_issue_needs_input_label_name ]
+          end
+
           provider = user.providers.subscription.first!
           provider.update!(enabled_for_agent_runs: true, enabled_for_fallback: true)
 
@@ -200,6 +221,7 @@ module Screenshots
           {
             "user" => { "id" => user.id, "email" => user.email, "password" => password },
             "project" => { "id" => project.id, "name" => project.name, "slug" => project.repo },
+            "clarifying_issue" => { "id" => clarifying_issue.id, "github_number" => clarifying_issue.github_number },
             "provider" => { "id" => provider.id, "name" => provider.name },
             "github_token" => { "id" => github_token.id, "name" => github_token.name },
             "integration_credential" => { "id" => integration_credential.id, "name" => integration_credential.name },
--- a/app/views/projects/_issue.html.erb
+++ b/app/views/projects/_issue.html.erb
@@ -13,6 +13,11 @@
           Paused by <%= guardrail_violation_label_for(paused_run).downcase %>
         </span>
       <% end %>
+      <% if issue.needs_input? && policy(project).update? %>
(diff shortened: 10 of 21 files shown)


viamin:
## Agent Update

Fixed the CI breakage by wiring the new clarifying-question UI into screenshot target coverage and making that page renderable during screenshot capture without a live GitHub comment fetch.

The change adds a dedicated screenshot target for the clarifying-question wizard in [app/services/screenshots/capture_targets.rb](/workspace/app/services/screenshots/capture_targets.rb), seeds a `needs_input` issue plus clarifying-question content in [app/services/screenshots/seed_data/paid.rb](/workspace/app/services/screenshots/seed_data/paid.rb), and lets the controller use parsed issue body content before falling back to GitHub comments in [app/controllers/projects/clarifying_questions_controller.rb](/workspace/app/controllers/projects/clarifying_questions_controller.rb). I also added regression coverage in [spec/services/screenshots/capture_targets_spec.rb](/workspace/spec/services/screenshots/capture_targets_spec.rb) and [spec/services/screenshots/seed_data/paid_spec.rb](/workspace/spec/services/screenshots/seed_data/paid_spec.rb).

Verification: `bundle exec rubocop` passed, and the db-less screenshot specs passed with `COVERAGE=false ALLOW_DBLESS_SPECS=true`. I could not run the full `bundle exec rspec` suite in this container because Postgres is unavailable, so DB-backed specs remain unverified locally. The changes are committed as `9a1b6b7` with message `fix(screenshots): cover clarifying question wizard`.

viamin:
## Agent Update

This is taking too long. Let me try a different approach - install all gems at once with `--local` flag or check if vendor/bundle exists:

Bundle install succeeded. Now let me run the relevant tests:

The service specs pass locally but are pending because no database. Let me check the request spec too:

The tests pass locally but are pending due to no database. The CI failure likely involves issues that only surface with a real database. Let me check the factory and related infrastructure:

The factory has the `:needs_input` trait. Let me check the Issue model for `paid_state` and the controller more carefully:

Good, `needs_input` is a valid `paid_state`. Now let me look at the view to check for the `back_link_path` helper:

Good, the helper exists. Let me now check the controller authorization more carefully, particularly the `authorize_project_update` for the `create` action. The `create` action calls `authorize @project, :update?`. Let me check if there are any issues with the policy:

Now let me think about what could be failing in CI. Let me look more closely at the test code. The `create` action in the controller uses `authorize @project, :update?` which requires owner/admin role. In the request spec, the user has admin role. Let me check the spec for potential issues with how the tests interact with the actual code.

Let me look at this more carefully - perhaps the issue is with the `clarifying_questions` route being singular resource but the spec uses plural path helper:

It's a singular resource (`resource`), so the path helper should be `project_issue_clarifying_question_path` (singular). But the spec uses `project_issue_clarifying_questions_path` (plural). Let me verify:

The route uses `resource :clarifying_questions` (singular resource), which generates the helper `project_issue_clarifying_question_path` (singular), not `project_issue_clarifying_questions_path` (plural). This would cause a `NoMethodError` in tests.

Wait, actually with `resource` (singular), the path helper is singular. Let me double-check:

With `resource :clarifying_questions` (singular resource block), the path helper would be `project_issue_clarifying_question` (singular), but the URL would still be `/projects/:project_id/issues/:issue_id/clarifying_questions` (plural, matching the resource name). So the spec uses the plural `_path` which would be undefined.

Actually, let me reconsider. `resource :clarifying_questions` generates:
- `project_issue_clarifying_question_path` (singular helper name, since `resource` is singular)
- URL: `/projects/:project_id/issues/:issue_id/clarifying_questions` (plural)

But the spec uses `project_issue_clarifying_questions_path` (plural helper). This would cause an undefined method error.

Let me also check the controller and view references:

So both the controller (line 30, 32) and view (line 43) use `project_issue_clarifying_questions_path` (plural), but the route defines `resource :clarifying_questions` (singular resource), which generates a singular helper name `project_issue_clarifying_question_path`.

This is the bug! The route uses `resource` (singular) but all code references the plural path helper. The fix is to change the route to use `resources` (plural) since the path is `/clarifying_questions`.

Wait, actually, let me reconsider. With `resource :clarifying_questions`, Rails generates:
- Named route helper: `project_issue_clarifying_question` (singularized)
- URL path: `/clarifying_questions` (as-is)

The controller and views use `project_issue_clarifying_questions_path` which wouldn't exist. This would cause `NoMethodError` at runtime.

The fix: change `resource :clarifying_questions` to either:
1. `resource :clarifying_question` (singular name), which would generate `project_issue_clarifying_question_path` - but then the controller path would need updating too
2. Keep `resource :clarifying_questions` but use the correct singular helper name throughout

Actually, wait. Let me re-read the Rails docs. With `resource :clarifying_questions`:
- The helper name is `project_issue_clarifying_question` (Rails singularizes the resource name)
- The URL path is `/clarifying_questions`

So the controller and views using `project_issue_clarifying_questions_path` would fail. The spec would also fail.

But wait, this is actually more subtle. Let me check what Rails would do with `resource :clarifying_questions`:

Actually, Rails `resource` (singular) generates helpers without the `_index` route. The helper name is derived from the singularized name of the resource. So `resource :clarifying_questions` would generate:
- `project_issue_clarifying_question_path` (singular - Rails singularizes `clarifying_questions` to `clarifying_question`)

But actually, looking at the Rails source, `resource` singularizes the name for the helper. So `resource :clarifying_questions` would create helper methods like:
- `project_issue_clarifying_question_path` (GET /projects/:project_id/issues/:issue_id/clarifying_questions)
- `project_issue_clarifying_question_path` (POST /projects/:project_id/issues/:issue_id/clarifying_questions)

The `project_issue_clarifying_questions_path` (plural) would NOT exist.

This means all usages in the controller, views, and specs would fail with `NoMethodError`.

The simplest fix is to change the route from `resource :clarifying_questions` to `resources :clarifying_questions, only: [:show, :create]` but since it's a singular resource (one wizard per issue), we should use `resource` but reference the correct singular helper name.

Actually, the cleanest fix is to change the route to use a singular name:

```ruby
resource :clarifying_question, only: [:show, :create],
  controller: "projects/clarifying_questions"
```

This would generate `project_issue_clarifying_question_path` matching the singular convention.

OR we can just fix all the references. Let me check which approach is simpler - there are many references using the plural form. Let me instead just fix the route to make the plural form work:

The issue is `resource :clarifying_questions` creates a singular helper with a singularized name. If we want plural helpers, we could use `resources :clarifying_questions` but restrict to show/create on a singular resource.

Actually, the simplest fix: add `as: :clarifying_questions` to override the helper name:

```ruby
resource :clarifying_questions, only: [:show, :create],
  controller: "projects/clarifying_questions",
  as: :clarifying_questions
```

Wait, that won't work for nested resources. The helper prefix comes from the parent nesting.

Let me think about this differently. The route is nested under `resources :issues`:

```ruby
resources :issues, only: [] do
  resource :clarifying_questions, only: [:show, :create],
    controller: "projects/clarifying_questions"
end
```

This generates:
- Path: `/projects/:project_id/issues/:issue_id/clarifying_questions`
- Helper: `project_issue_clarifying_question` (singularized)

But the code expects `project_issue_clarifying_questions` (plural).

The cleanest fix: change `resource :clarifying_questions` to `resource :clarifying_questions, as: 'clarifying_questions'`. But wait, `as` in Rails routes changes the name prefix, not the suffix. Let me think...

Actually, for `resource`, the `as` option sets the name used for the path helper. So:

```ruby
resource :clarifying_questions, as: :clarifying_questions, ...
```

This should generate `project_issue_clarifying_questions_path` instead of `project_issue_clarifying_question_path`.

Hmm, actually I'm not 100% sure about the nested case. Let me just look at how other similar routes are done in this app:

`merge_subscription` uses `project_issue_merge_subscription_path` (singular) - matching the singular `resource` pattern. So Rails singularizes `merge_subscription` (already singular) correctly.

For `clarifying_questions`, Rails would singularize to `clarifying_question`, generating `project_issue_clarifying_question_path` (singular). But the code uses `project_issue_clarifying_questions_path` (plural).

The fix is to update all references from plural to singular `project_issue_clarifying_question_path`. Let me check all the places:

Now let me also check the screenshots-related files since they're part of this PR too:

The screenshot targets also use the plural form. Since these targets reference the URL path (not Rails helper), the plural form `/clarifying_questions` is correct for the URL.

Now I understand the full picture. The route is `resource :clarifying_questions` which creates a singular helper. But all code uses the plural helper name. I need to fix either the route or all the references.

The simplest fix: add `as: :clarifying_questions` to the route to keep the plural helper name:

Actually, let me check how Rails handles `as` for singular resources:

`context_intake` is already singular, so it matches naturally. The problem with `clarifying_questions` is that it's plural.

The most consistent approach: rename the route to use a singular name `clarifying_question_wizard` or simply change the references to singular. But actually, the simplest fix that matches existing patterns is to just use `as:` option.

Let me verify how `as:` works with nested singular resources:

OK, `merge_subscription` is singular and works fine. For `clarifying_questions`, the helper would be `project_issue_clarifying_question_path` (singularized by Rails).

The cleanest fix: change all references to use the singular helper name `project_issue_clarifying_question_path`. But there's also the screenshots service that uses `:project_issue_clarifying_questions` as a symbol for the Target hash key - that's fine because it's not a route helper, it's just a hash key.

Actually wait, let me re-read the capture_targets.rb more carefully:

The `project_issue_clarifying_questions` in capture_targets.rb is just a hash key (symbol), not a Rails route helper. The path is constructed as a string. This is fine regardless.

Now let me look at the route helper mapping in capture_targets.rb:

Here, `:project_issue_clarifying_questions` is used as a key to look up the Target hash. This is fine since it matches the hash key. Let me also check how this mapping is used:

The hash key `:project_issue_clarifying_questions` is used to look up targets by key. The ALL_TARGETS hash uses this same key. So this is internally consistent and doesn't need to match the Rails route helper name.

Now let me determine the actual fix. The route generates `project_issue_clarifying_question_path` (singular). I need to change all the references. Let me do it:

Now I need to fix the route helper references. The route `resource :clarifying_questions` generates `project_issue_clarifying_question` (singular). I'll update all references from plural to singular:

Now let me apply all the fixes. The route helper name needs to change from `project_issue_clarifying_questions` (plural) to `project_issue_clarifying_question` (singular) in the controller, view, and spec files:

Now let me also check if there are any other issues. Let me look at the screenshots spec for potential failures:

These screenshot specs use `slug` values which match the ALL_TARGETS hash keys, not route helpers. They should be fine.

Now let me check the screenshots seed data for potential issues:

Now let me check if there's an `enhance_issue_needs_input_label_name` method on Project:

Good, it exists. Now let me run the specific tests that were changed to check if my fix works:

All tests pass (pending due to no database, which is expected). Now let me run lint:

The ERB files are causing syntax errors because rubocop is parsing them as Ruby. These are false positives from the lint - ERB files are typically not linted by rubocop directly. Let me check how the project runs lint:

Good - the lint script uses `bin/rubocop` which only runs on `.rb` files. Let me run rubocop only on the Ruby files:

All Ruby files pass lint. Let me also run rubocop on the screenshots-related files:

Now let me also run the screenshot-related tests:

All specs pass. Let me now verify the full diff of my changes:

The diff looks clean. The fix is straightforward: changed all route helper references from the plural `project_issue_clarifying_questions_path` to the singular `project_issue_clarifying_question_path` to match what Rails generates from the singular `resource :clarifying_questions` route definition.

Let me now update the todo and commit:

The CI failure was caused by a route helper name mismatch. The route defines `resource :clarifying_questions` (singular resource block), which makes Rails generate the helper as `project_issue_clarifying_question_path` (singularized). However, all code references used the plural form `project_issue_clarifying_questions_path`, causing `NoMethodError` at runtime.

**Fix**: Changed all route helper references from plural `project_issue_clarifying_questions_path` to singular `project_issue_clarifying_question_path` in:
- `app/controllers/projects/clarifying_questions_controller.rb` (2 lines)
- `app/views/projects/_issue.html.erb` (1 line)
- `app/views/projects/clarifying_questions/show.html.erb` (1 line)
- `spec/requests/projects/clarifying_questions_spec.rb` (5 lines)

viamin:
## Agent Update

Committed as `c1b4c2d` with `fix(clarifying-questions): gate wizard submission until complete`.

The change closes a real gap in the wizard: the Stimulus controller now blocks submit until every answer is filled, keeps the submit button disabled until completion, and jumps back to the first blank question if a submit is attempted early. I also added a request-spec assertion that the rendered wizard includes the submit gate wiring and disabled initial submit state.

Verification: `bundle install` and `yarn install --frozen-lockfile` succeeded with workspace-local paths, `bundle exec rubocop` passed, `bin/lint --staged` passed in the commit hook, and `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/tasks/screenshots_rake_spec.rb spec/services/screenshots/capture_targets_spec.rb` passed. I could not run `bundle exec rspec` for the database-backed request/model specs in this container because no PostgreSQL service is available, and the repo is configured to fail fast when the test DB is unreachable.

viamin:
## Agent Update

Merged `origin/main` into the PR branch, resolved the conflict in [`spec/services/screenshots/seed_data/paid_spec.rb`](/workspace/spec/services/screenshots/seed_data/paid_spec.rb), and committed the result as `08f2c145` with message `Merge origin/main into paid/1964-answer-issue-clarifying-questions-via-paid-ui-wiza-4d125f`.

The CI `publish` failure is addressed by the merged update in [`.github/workflows/pr-screenshots-publish.yml`](/workspace/.github/workflows/pr-screenshots-publish.yml): the resolver now handles screenshot workflow runs that only complete the `detect` job, treating clean detect-only runs as skipped/no-UI-change instead of failing on a missing `capture` job. The corresponding spec in [`spec/config/pr_screenshots_publish_workflow_file_spec.rb`](/workspace/spec/config/pr_screenshots_publish_workflow_file_spec.rb) covers that behavior.

Verification:
- `bundle exec rubocop` ✅
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/config/pr_screenshots_publish_workflow_file_spec.rb spec/tasks/screenshots_rake_spec.rb` ✅
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/services/screenshots/capture_targets_spec.rb` ✅

I could not run `bundle exec rspec` for the full suite or the DB-backed screenshot seed spec in this container because no PostgreSQL service is available, and this repo correctly aborts normal test runs when the test database cannot be reached. The worktree is clean.

viamin:
## Agent Update

Committed as `5be982fc` with `fix(screenshots): scope clarifying wizard capture target`.

The branch was already up to date with `origin/main`, so there were no merge conflicts to resolve in this workspace. The code change narrows screenshot target selection so `app/javascript/controllers/clarifying_questions_controller.js` maps directly to the clarifying-questions wizard route instead of expanding capture to the generic shared-page set. That reduces unrelated screenshot coverage during the PR capture job and matches the new UI surface more precisely.

Validation:
- `bin/lint --staged` passed
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/services/screenshots/capture_targets_spec.rb spec/tasks/screenshots_rake_spec.rb` passed

Constraint:
- `bundle exec rspec` cannot run end-to-end in this container because there is no PostgreSQL service, and this repo aborts full spec startup on `ActiveRecord::ConnectionNotEstablished`.

viamin:
## Agent Update

Merged state is clean: `git merge origin/main` reports `Already up to date.` I committed the fix as `1fcdda5a` with `fix(clarifying-questions): validate wizard submissions`.

The change set tightens the clarifying-question flow by extracting question loading into [app/services/clarifying_questions/load.rb](/workspace/app/services/clarifying_questions/load.rb), reusing it in [app/controllers/projects/clarifying_questions_controller.rb](/workspace/app/controllers/projects/clarifying_questions_controller.rb), and preventing empty or tampered submissions from posting blank GitHub comments in [app/services/clarifying_questions/submit_answers.rb](/workspace/app/services/clarifying_questions/submit_answers.rb). I also added coverage for the loader and the stale/tampered submission case in [spec/services/clarifying_questions/load_spec.rb](/workspace/spec/services/clarifying_questions/load_spec.rb), [spec/services/clarifying_questions/submit_answers_spec.rb](/workspace/spec/services/clarifying_questions/submit_answers_spec.rb), and [spec/requests/projects/clarifying_questions_spec.rb](/workspace/spec/requests/projects/clarifying_questions_spec.rb).

`bundle exec rubocop` passed repo-wide. `bundle exec rspec` could not run in this container because the repo aborts immediately without Postgres; the exact failure was `ActiveRecord::ConnectionNotEstablished while preparing the test database` from `spec/rails_helper.rb`. The worktree is clean.

viamin:
## Agent Update

Resolved the branch state by merging `origin/main`; the merge committed cleanly as `24f0c871` and the worktree is clean. There were no textual conflicts to fix. The only diff relative to the remote PR branch is the `main` workflow-concurrency updates in the seven `.github/workflows/*` files that came in from that merge.

For the CI failure, I checked the live PR context and confirmed the screenshot-capture fix referenced on the PR is already present on this branch (`9a1b6b7a` / `5be982fc` were already in history), so there was nothing additional to patch locally beyond getting the branch rebased onto current `main`.

Verification:
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/config/pr_screenshots_publish_workflow_file_spec.rb` passed.
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/tasks/screenshots_rake_spec.rb` passed.

I could not produce a fully green `bundle exec rspec` in this container because there is no PostgreSQL service, and many specs still attempt DB access even under DB-less mode. The first real failures were DB connection errors starting in `spec/services/knowledge/collectors/routes_collector_spec.rb`.

viamin:
## Agent Update

Merged `origin/main` into the PR branch and resolved the history issue by unshallowing the clone first. The branch is clean now at commit `8b46c6c1`, which pulls in `origin/main`’s screenshot stability fixes alongside the PR’s existing clarifying-question screenshot mapping work.

Validation:
- `bundle exec rspec` could not run fully in this container because PostgreSQL is unavailable; it exits with `ActiveRecord::ConnectionNotEstablished` from `spec/rails_helper.rb`.
- `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec spec/services/screenshots/capture_targets_spec.rb spec/services/screenshots/branch_storage_spec.rb spec/lib/screenshots/seed_runner_spec.rb` passed.

I also checked the final PR diff against `origin/main` for whitespace/errors with `git diff --check`; it’s clean. No uncommitted changes remain, and nothing was pushed.

viamin:
## Agent Update

Resolved the branch state and committed a focused fix as `233c7e79` with `fix(clarifying-questions): register wizard controller`.

The main code issue was that the new `clarifying_questions_controller.js` existed but was never registered in [app/javascript/controllers/index.js](/workspace/app/javascript/controllers/index.js), so the wizard’s navigation/progress behavior would not activate in the browser. I also fixed the new request spec in [spec/requests/projects/clarifying_questions_spec.rb](/workspace/spec/requests/projects/clarifying_questions_spec.rb) so its POST cases stub the same GitHub question-loading path the controller actually uses.

Verification: `bundle exec rubocop` passed, `yarn build` passed, `bin/lint --changed` passed, and the commit hook passed. I could not run full `bundle exec rspec` in this container because no PostgreSQL service is available; only db-less subsets were runnable here.

viamin:
## Agent Update

Resolved the CI-side screenshot issue by narrowing `app/javascript/controllers/index.js` handling in [app/services/screenshots/capture_targets.rb](/workspace/app/services/screenshots/capture_targets.rb). When the registry changes alongside a specifically mapped Stimulus controller, it no longer broadens capture to the entire shared UI; for this PR it now resolves to `project_show` and `project_issue_clarifying_questions` instead of the full shared target set. I added a regression spec in [spec/services/screenshots/capture_targets_spec.rb](/workspace/spec/services/screenshots/capture_targets_spec.rb).

`origin/main` was already merged locally, so there were no remaining merge conflicts to resolve. Lint passed with `COVERAGE=false bundle exec rubocop`, and the db-less screenshot specs passed with `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec spec/services/screenshots/capture_targets_spec.rb spec/tasks/screenshots_rake_spec.rb`. A full `bundle exec rspec` run is still blocked in this container because PostgreSQL is unavailable by design.

Committed as `cc273adb` with message `fix(screenshots): narrow controller registry capture targets`.

viamin:
**Escalation Note**

This has been escalated because Draft review limit reached.

**How to resolve:**

viamin:
## Agent Update

Fixed the `capture` failure cause by correcting the clarifying-questions route helper from the nonexistent singular form to the actual plural helper in the controller, wizard view, issue row link, and request spec. I also added a db-less routing regression spec at [spec/routing/projects/clarifying_questions_routing_spec.rb](/workspace/spec/routing/projects/clarifying_questions_routing_spec.rb:1) so this mismatch is caught without needing PostgreSQL.

`origin/main` was already merged, so there were no remaining conflicts to resolve. `bundle exec rubocop` passed. Relevant db-less verification passed with `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec spec/tasks/screenshots_rake_spec.rb spec/services/screenshots/capture_targets_spec.rb spec/routing/projects/clarifying_questions_routing_spec.rb`. A full `bundle exec rspec` run still cannot pass in this container because PostgreSQL is unavailable and many repo specs are not marked for the documented db-less fallback.

Committed as `48cc43ca` with message `fix(clarifying-questions): use the correct wizard route helper`.

viamin:
## Agent Update

Merged state was already clean: `git merge origin/main` reported the branch was up to date, so there were no conflicts to resolve.

I fixed the clarifying-question flow so Paid no longer re-opens the wizard after a later `<!-- paid:clarifying-answers -->` comment already answered the latest enhancement questions. I also hardened answer submission to fail cleanly when GitHub access is missing, formatted posted answers with explicit `A1`-style labels, and added coverage for the new `Issue#needs_input?` predicate plus DB-less service specs around the clarifying-question loader and submitter.

Verification:
- `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec spec/services/clarifying_questions/parse_spec.rb spec/services/clarifying_questions/load_spec.rb spec/services/clarifying_questions/submit_answers_spec.rb spec/routing/projects/clarifying_questions_routing_spec.rb` passed.
- `bundle exec rspec` could not run in this container because there is no PostgreSQL service. It aborts immediately with `ActiveRecord::ConnectionNotEstablished` from `spec/rails_helper.rb` before examples load.

Committed on the PR branch as `0cf84b88` with:
`fix(clarifying-questions): avoid reopening answered wizards`

viamin:
## Agent Update

Committed as `6468385a` with `fix(clarifying-questions): keep answered body prompts closed`.

The fix is in [app/services/clarifying_questions/load.rb](/workspace/app/services/clarifying_questions/load.rb) and [spec/services/clarifying_questions/load_spec.rb](/workspace/spec/services/clarifying_questions/load_spec.rb). `ClarifyingQuestions::Load` now treats a posted answers comment as closing the wizard even when the original question block is still present in the issue body, unless there is a newer enhancement comment. That prevents the UI from reopening an already-answered flow.

Verification:
- `bin/lint --staged` passed
- `bundle exec rubocop app/services/clarifying_questions/load.rb spec/services/clarifying_questions/load_spec.rb` passed
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/services/clarifying_questions/load_spec.rb spec/services/clarifying_questions/parse_spec.rb spec/services/clarifying_questions/submit_answers_spec.rb spec/routing/projects/clarifying_questions_routing_spec.rb spec/services/screenshots/capture_targets_spec.rb` passed
- `bundle exec rspec` could not run in this container because Postgres is unavailable; it aborts in `spec/rails_helper.rb` with `ActiveRecord::ConnectionNotEstablished`

The branch was already up to date with `origin/main`, so there were no merge conflicts left to resolve here.

viamin:
## Agent Update

Merged `origin/main` into the PR branch and fixed one regression in the clarifying-questions loader. In [app/services/clarifying_questions/load.rb](/workspace/app/services/clarifying_questions/load.rb:45), the loader now returns `[]` instead of attempting a nil timestamp comparison when an issue only has a previously posted answers comment and no remaining enhancement comment. I added a regression spec in [spec/services/clarifying_questions/load_spec.rb](/workspace/spec/services/clarifying_questions/load_spec.rb:120).

Verification:
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/services/clarifying_questions/load_spec.rb spec/services/clarifying_questions/parse_spec.rb spec/services/clarifying_questions/submit_answers_spec.rb spec/routing/projects/clarifying_questions_routing_spec.rb spec/services/screenshots/capture_targets_spec.rb spec/services/screenshots/seed_data/paid_spec.rb` passed.
- `bundle exec rspec` could not run in this container because Postgres is unavailable; the suite aborts intentionally with `ActiveRecord::ConnectionNotEstablished` before loading DB-backed specs.

Commits on the branch:
- `ba7d59e` merge from `origin/main`
- `5c623b2` `fix(clarifying-questions): handle answers-only issue comments`

The worktree is clean.

viamin:
## Agent Update

Merged `origin/main` into the PR branch cleanly, fixed the clarifying-questions flow, and committed the result as `3d994e1f` with `fix(clarifying-questions): reject stale wizard submissions`.

The code change does two things: it now rejects stale or tampered wizard submissions when the submitted question list no longer matches the current GitHub-backed question set, and it caches GitHub comment loading in `ClarifyingQuestions::Load` while preserving the answered-after-latest-questions behavior. I also updated the request/spec coverage to match the real question text and added coverage for the mismatched-question case.

Verification: `bundle exec rubocop` passed for the full repo, and the db-less spec subset passed with `ALLOW_DBLESS_SPECS=true COVERAGE=false`. A plain `bundle exec rspec` could not run in this container because PostgreSQL is unavailable; it aborts immediately with `ActiveRecord::ConnectionNotEstablished` during test DB preparation. The working tree is clean.

viamin:
## Agent Update

Merged `origin/main` into the PR branch and fixed a clarifying-question loader bug where an old answers comment could keep the wizard hidden even after the issue body was updated with fresh questions. The loader now compares answer timestamps against the latest question source, including `issue.github_updated_at` for body-backed questions, and there’s a regression spec covering that case.

Committed as `115eaa76` with message `fix(clarifying-questions): respect refreshed issue questions`.

Validation:
- `bundle exec rubocop`
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/services/clarifying_questions/load_spec.rb spec/services/clarifying_questions/parse_spec.rb spec/services/clarifying_questions/submit_answers_spec.rb spec/services/screenshots/capture_targets_spec.rb`

I could not run the full database-backed `bundle exec rspec` suite in this container because no PostgreSQL service is available, and the repo is configured to abort DB-dependent specs unless run in db-less mode.

viamin:
## Agent Update

Merged `origin/main` into the PR branch and committed the fix as `30663023` with `fix(issues): scope clarifying question input state`.

The CI-facing bug I fixed was in [`Issue#needs_input?`](/workspace/app/models/issue.rb:195). It previously returned `true` for every `paid_state == "needs_input"` issue, but that state is also used for generic “no output” cases. The clarifying-question wizard should only appear for enhancement issues carrying the enhancement-specific needs-input label, so the predicate now requires both the state and `project.enhance_issue_needs_input_label_name`. I updated the spec in [`spec/models/issue_spec.rb`](/workspace/spec/models/issue_spec.rb:853) to cover both the positive case and the non-clarifying `needs_input` case.

Checks run:
- `git merge origin/main`
- `bundle install` with local `vendor/bundle` path
- `yarn install --frozen-lockfile` with `YARN_CACHE_FOLDER=/tmp/yarn-cache`
- `bundle exec rubocop`
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/models/issue_spec.rb:853 spec/services/clarifying_questions/parse_spec.rb spec/services/clarifying_questions/load_spec.rb spec/services/clarifying_questions/submit_answers_spec.rb spec/routing/projects/clarifying_questions_routing_spec.rb spec/services/screenshots/capture_targets_spec.rb`

`bundle exec rspec` could not run end-to-end in this container because PostgreSQL is unavailable; the suite aborts at boot with `ActiveRecord::ConnectionNotEstablished`. The branch is clean with no uncommitted changes.

viamin:
## Agent Update

Updated the branch with `33c49f16 fix(test): align needs-input issue factory`.

The fix is in [spec/factories/issues.rb](/workspace/spec/factories/issues.rb:42): the `:needs_input` trait now also applies `project.enhance_issue_needs_input_label_name`, which keeps factory-built issues consistent with the merged `Issue#needs_input?` predicate. That is the likely source of the failing test regression after `needs_input?` became label-scoped.

Verification I could run here:
- `bundle exec rubocop`
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/services/clarifying_questions/load_spec.rb spec/services/clarifying_questions/parse_spec.rb spec/services/clarifying_questions/submit_answers_spec.rb spec/routing/projects/clarifying_questions_routing_spec.rb spec/services/screenshots/capture_targets_spec.rb`
- `git commit` succeeded with the repo hook running `bin/lint --staged`

I could not run the full `bundle exec rspec` suite in this container because Postgres is unavailable. The exact failure was `ActiveRecord::ConnectionNotEstablished` while preparing the test database. The worktree is clean now.

viamin:
## Agent Update

The `show?` and `update?` are inherited from `ApplicationPolicy`. The test gives the user admin role, so both should pass.

Let me think about what else could be failing. Let me look at the view template:

I notice on line 18, the view uses `back_link_path`. Let me check if that helper exists:

Good, the helper exists. Let me now check the routes:

The route is `resource :clarifying_questions` (singular) which generates `project_issue_clarifying_questions_path` with GET and POST. That matches what the routing spec and request spec expect.

Let me check if `resource` (singular) generates `project_issue_clarifying_questions` (plural) - yes, in Rails, `resource :clarifying_questions` generates `clarifying_questions` path helpers.

Let me now check the Stimulus controller:

Now let me check the Stimulus controller registration:

Let me also check the issue partial:

Now let me try to see if there are tests that reference the issue partial or the `needs_input?` view rendering:

This spec is tagged `:no_db`. Let me run the capture_targets spec:

All capture_targets specs pass. Now let me check the seed_data spec:

This requires DB. Let me check the seed_data/paid.rb:

Now let me look at what actual test failures could occur. The PR was on a branch that had CI failures. The task says to fix CI failures and resolve merge conflicts. The merge conflicts are resolved (merge was clean). Now I need to figure out the CI test failures.

Since I can't run the DB-dependent specs, let me analyze the code more carefully. Let me look at the issue spec for the `needs_input` test specifically, and check if the `needs_input` factory trait is correct:

The factory `needs_input` trait:
```ruby
trait :needs_input do
  paid_state { "needs_input" }
  labels { [ project.enhance_issue_needs_input_label_name ] }
end
```

The `needs_input?` method:
```ruby
def needs_input?
  paid_state == "needs_input" && has_label?(project.enhance_issue_needs_input_label_name)
end
```

The test at line 855-856:
```ruby
project = build(:project, enhance_issue_needs_input_label_name: "paid-enhance-needs-input")
issue = build(:issue, :needs_input, project: project, labels: [ "paid-enhance-needs-input" ])
```

Wait, the `needs_input` trait sets `labels` to `[project.enhance_issue_needs_input_label_name]`, but then the test overrides `labels` to `["paid-enhance-needs-input"]`. That's fine because the explicit `labels:` overrides the trait.

But wait - `build(:project, enhance_issue_needs_input_label_name: "paid-enhance-needs-input")`. Does the Project model have `enhance_issue_needs_input_label_name` as an attribute? Let me check:

OK, so the default is `"paid-needs-input"`. The test overrides it to `"paid-enhance-needs-input"`. Let me look at the third test:

```ruby
it "returns false for non-enhancement needs-input issues" do
  project = build(:project, enhance_issue_needs_input_label_name: "paid-enhance-needs-input")
  issue = build(:issue, :needs_input, project: project, labels: [ "paid-needs-input" ])
  expect(issue).not_to be_needs_input
end
```

Here the project expects the label `"paid-enhance-needs-input"` but the issue has labels `["paid-needs-input"]`. The `needs_input?` checks both paid_state and label match. So it should return false. But wait - the `:needs_input` trait sets `paid_state { "needs_input" }` AND `labels { [ project.enhance_issue_needs_input_label_name ] }`. Then the test overrides `labels` to `["paid-needs-input"]`. So the issue has `paid_state: "needs_input"` and `labels: ["paid-needs-input"]`, but the project's `enhance_issue_needs_input_label_name` is `"paid-enhance-needs-input"`. The `needs_input?` checks `has_label?(project.enhance_issue_needs_input_label_name)` which is `has_label?("paid-enhance-needs-input")`. The issue doesn't have that label, so it returns false. Correct.

Now, let me look at the `needs_input?` test more carefully. There might be an issue with the `has_label?` method and the `labels` attribute being an array:

```ruby
def has_label?(label)
  labels.include?(label)
end
```

The `labels` column is probably a string array (PostgreSQL array or serialized). If `build(:issue, ..., labels: ["paid-enhance-needs-input"])` creates the right thing, this should work.

OK, let me think about what else could be causing test failures. Let me look at the `Screenshots::SeedData::Paid` spec more carefully. It tests that `issue.be_needs_input` - does the seeded issue have the right `paid_state` and label?

From `seed_data/paid.rb` line 67: `record.labels = [ project.enhance_issue_needs_input_label_name ]` - so the issue has the correct label. And line 66: `record.paid_state = "needs_input"`. This should work.

Let me now look at what test the CI is actually running. The CI runs:
```
bin/rspec --exclude-pattern "spec/performance/**/*_spec.rb,spec/system/**/*_spec.rb"
```

This runs ALL specs except performance and system. Since I can't run those, let me look at whether there might be an issue with the `needs_input` PAID_STATE. Let me check if it was already added to the array in a previous commit:

Looking at the Issue model line 4: `PAID_STATES = %w[new planning in_progress completed failed needs_input recommend_close analyzed].freeze`

The `needs_input` state is already in the PAID_STATES array. Good.

Let me check if there are any tests that validate the `PAID_STATES` constant:

Now let me look at the broader context around line 830 to see if there's something else:

OK wait, I notice the test at line 841: `build(:issue, paid_state: state)` - but the `needs_input` state requires a label too. But this test just checks validity, which only checks `paid_state` inclusion, not labels. The `needs_input?` predicate is tested separately. This should be fine.

Let me now think about what might cause an actual CI failure. Let me look at the `Screenshots::CaptureTargets` service to see if the controller mapping is correct:

This all looks correct. Let me now focus on what's most likely to fail - the `spec/services/screenshots/seed_data/paid_spec.rb`. This test calls `Screenshots::SeedData::Paid.call` which creates records. The seeded issue uses `project.enhance_issue_needs_input_label_name` for the label. The test at line 26 checks `expect(issue).to be_needs_input`. 

The issue is created with `labels: [ project.enhance_issue_needs_input_label_name ]`. The project's `enhance_issue_needs_input_label_name` defaults to `"paid-needs-input"`. And `needs_input?` checks `has_label?(project.enhance_issue_needs_input_label_name)`. So this should work.

Let me now look at whether there are any issues with the `GithubClient::Error` reference in the controller:

Good, the `GithubClient::Error` exists. The controller references it correctly.

Let me now look at whether there could be issues with the `issue_timestamp` method in `Load`:

```ruby
def issue_timestamp
  issue.github_updated_at&.to_time || issue.updated_at&.to_time || Time.at(0)
end
```

The `issue.github_updated_at` is a datetime column. `.to_time` should work. But in the tests, `issue` is a double that has `github_updated_at: 2.minutes.ago`. Calling `2.minutes.ago.to_time` should work.

Wait, I need to check the `comment_timestamp` method too:

```ruby
def comment_timestamp(comment)
  comment.created_at&.to_time || Time.at(0)
end
```

In the tests, comments are doubles with `created_at: 2.minutes.ago`. That should work too.

Let me now look more carefully at the issue. The CI was failing **before** the merge from main. The task says the PR's CI was failing. Let me look at the PR more carefully - the PR was likely failing because of the tests that need DB. But the actual CI in GitHub Actions has a database.

Let me re-read the task description: "CI Status: FAILING - test (failure)". This means the test job in CI is failing.

Since I can't run the full test suite, let me look at the code changes more carefully for any bugs. Let me look at what the controller does:

The `ClarifyingQuestionsController` has `before_action :authorize_project_update, only: [:create]`. The `authorize_project_update` calls `authorize @project, :update?`. The ApplicationPolicy's `update?` requires owner/admin role. The test adds `admin` role, so this should pass.

Wait, I need to check the `set_issue` method:
```ruby
def set_issue
  @issue = @project.issues.issues_only.find(params[:issue_id])
end
```

This uses `issues_only` scope which filters `where(is_pull_request: false)`. The test creates a regular issue, not a PR, so this should work.

Let me now look for any subtle issues with the view. In the view, there's `back_link_path(project_path(@project))`. Let me check if this helper is a concern:

This helper returns a path. The view uses it as `link_to back_link_path(project_path(@project))`. This should work.

Let me now check the `Screenshots::SeedData::Paid` seed more carefully. I notice it creates an issue with `github_number: 1964`. Let me check if there are any unique constraints:

There's a unique index on `[project_id, github_number]`. The seed uses `find_or_create_by!` with `github_issue_id`, and sets `github_number: 1964`. Since the project is also found/created with a unique `github_id`, this should be fine.

Now, let me think about this differently. The CI failure might not be in the new code at all. It could be in the tests from the main branch merge. The merge brought in changes to `run_agent_activity.rb` and its spec. Let me look at those tests more carefully:

The key test changes are:
1. Changed `expect(prompt).to eq(...)` to `expect(prompt).to include(...)` because the prompt now has DRAFTING_GUIDANCE appended
2. Added new test "tells the agent to synthesize the issue instead of asking for drafting fields"
3. Changed A/B test expectations from `eq` to `include`

Let me check if there's an issue with the `resolve_and_persist_goal_prompt` method. Previously it returned the rendered template directly. Now it goes through `append_missing_issue_goal_runtime_instructions`. This method:

1. Returns `rendered` as-is if `slug != ISSUE_GOAL_PROMPT_SLUG`
2. If `slug == ISSUE_GOAL_PROMPT_SLUG`, it appends DRAFTING_GUIDANCE and decomposition_instructions

The issue is: the TEMPLATE already has DRAFTING_GUIDANCE inline. When rendered, the rendered text will already contain DRAFTING_GUIDANCE. The `append_prompt_section` checks `rendered.include?(normalized_addition)`, so it won't double-append. 

But what if the template is rendered and the DRAFTING_GUIDANCE text is broken across lines differently? No, the template is a static string, it should be exactly the same.

Wait, let me think about this more carefully. The `Prompts::GoalCreateGithubIssue::TEMPLATE` includes `#{DRAFTING_GUIDANCE}` which interpolates the DRAFTING_GUIDANCE constant. The constant itself is:
```
Treat the request and repository context already provided above as the full source
material for the GitHub issue you need to file. Synthesize the issue title, body,
and any appropriate labels from that context yourself.

Do NOT reply by asking the user to provide the issue type, title, description,
labels, or other issue-drafting fields. If a field is not explicitly specified in
the provided context, make a reasonable choice and continue. When no labels are
clearly requested, omit them.
```

And the TEMPLATE inserts this inline between `IMPORTANT: Your goal is to CREATE A GITHUB ISSUE...` and `You have access to the GitHub API...`. When the template is rendered with variables substituted (e.g., `{{base_prompt}}` becomes the actual prompt, `{{repo}}` becomes the repo name, `{{decomposition_instructions}}` becomes empty or decomposition text), the DRAFTING_GUIDANCE text is still present.

Then `append_missing_issue_goal_runtime_instructions` calls:
```ruby
prompt = append_prompt_section(rendered, Prompts::GoalCreateGithubIssue::DRAFTING_GUIDANCE)
append_prompt_section(prompt, vars[:decomposition_instructions])
```

The `append_prompt_section` strips the addition and checks if it's already in the rendered text. Since DRAFTING_GUIDANCE is a substring of the rendered TEMPLATE, it should be found, and nothing should be appended. Good.

But wait - what about the `resolve_and_persist_goal_prompt` test at line 702-706? It creates a prompt with template `"issue {{base_prompt}} {{repo}}"`. This template does NOT include DRAFTING_GUIDANCE. So when rendered, it becomes something like `"issue Create the issue owner/repo"`. Then `append_missing_issue_goal_runtime_instructions` will:
1. Append DRAFTING_GUIDANCE (since it's not in the rendered text)
2. Append decomposition_instructions (empty in this case)

So the rendered output will be:
```
issue Create the issue owner-1/repo-1

Treat the request and repository context already provided above as the full source
material for the GitHub issue you need to file. Synthesize the issue title, body,
and any appropriate labels from that context yourself.

Do NOT reply by asking the user to provide the issue type, title, description,
labels, or other issue-drafting fields. If a field is not explicitly specified in
the provided context, make a reasonable choice and continue. When no labels are
clearly requested, omit them.
```

The test expects:
```ruby
expect(rendered).to include("issue Create the issue #{project.full_name}")
expect(rendered).to include("Do NOT reply by asking the user to provide the issue type, title, description,")
```

Both should pass.

OK, I think I've been overthinking this. Let me now look for more obvious issues. Let me check the CI workflow - maybe the issue is a newly added spec that has a syntax error or references something that doesn't exist:

All syntax is fine. Let me now look at the `ClarifyingQuestions::Load` service more carefully for the `updated_at` method call. The `issue_timestamp` method:

```ruby
def issue_timestamp
  issue.github_updated_at&.to_time || issue.updated_at&.to_time || Time.at(0)
end
```

But in the load spec, the `issue` double doesn't have `updated_at`:
```ruby
let(:issue) do
  double(
    body: issue_body,
    github_number: 1964,
    github_updated_at: 2.minutes.ago
  )
end
```

Wait, but `issue_timestamp` is only called when there's a body question AND a GitHub answer. Looking at the test "when the issue body was updated after answers were posted" at line 127-158:

```ruby
let(:issue) do
  double(
    body: issue_body,
    github_number: 1964,
    github_updated_at: 30.seconds.ago
  )
end
```

This double doesn't have `updated_at` either, but `issue_timestamp` calls `issue.updated_at&.to_time` only if `github_updated_at` is nil. Since `github_updated_at` is `30.seconds.ago`, `.to_time` will succeed and `updated_at` won't be reached.

But wait, the `latest_question_timestamp` method calls `issue_timestamp` when `body_questions.any?`:
```ruby
def latest_question_timestamp(body_questions:, enhancement_comment:)
  timestamps = []
  timestamps << issue_timestamp if body_questions.any?
  timestamps << comment_timestamp(enhancement_comment) if enhancement_comment.present?
  timestamps.compact.max
end
```

And `issue_timestamp` returns:
```ruby
issue.github_updated_at&.to_time || issue.updated_at&.to_time || Time.at(0)
```

For the `2.minutes.ago` value, `.to_time` should work because `ActiveSupport::TimeWithZone#to_time` is defined. But wait - in the test, the `issue` is a raw `double`, not an ActiveRecord object. `2.minutes.ago` returns an `ActiveSupport::TimeWithZone`. Calling `.to_time` on that returns a `Time` object. So `issue_timestamp` would return a `Time` object, not nil.

Actually wait, looking more carefully at the test for "when the issue body already contains clarifying questions" at line 38-58:

```ruby
context "when the issue body already contains clarifying questions" do
  let(:issue_body) do
    <<~COMMENT
      <!-- paid:enhance-issue -->

      ## Clarifying questions
      1. What should happen next?
    COMMENT
  end

  it "returns questions from the issue body" do
    questions = described_class.call(project: project, issue: issue)
    expect(questions).to eq(["What should happen next?"])
  end

  it "only loads GitHub comments once to check for newer answers" do
    described_class.call(project: project, issue: issue)
    expect(github_client).to have_received(:issue_comments).once
  end
end
```

When `body_questions` is `["What should happen next?"]` (non-empty), the code enters the first branch:
```ruby
if body_questions.any?
  return body_questions unless github_available?
  enhancement_comment = latest_enhancement_comment
  return [] if answered_after_latest_questions?(body_questions:, enhancement_comment:)
  return body_questions
end
```

`github_available?` returns true (mocked). Then it calls `latest_enhancement_comment`. Then `answered_after_latest_questions?` is called. This calls `latest_answer_comment` and `latest_question_timestamp`.

`latest_answer_comment` looks for a comment with the ANSWER_MARKER. In this test, no comments have been set up (the `before { allow(github_client).to receive(:issue_comments).and_return([]) }` at line 35). So `latest_answer_comment` returns nil, and `answered_after_latest_questions?` returns false.

But wait - `latest_enhancement_comment` also looks for comments. With an empty comments array, it returns nil. So `answered_after_latest_questions?` is called with `enhancement_comment: nil`.

Looking at `answered_after_latest_questions?`:
```ruby
def answered_after_latest_questions?(body_questions:, enhancement_comment:)
  latest_answer_comment = latest_answer_comment()
  return false unless latest_answer_comment
  ...
end
```

Since `latest_answer_comment` is nil (no comments with ANSWER_MARKER), it returns false. So `answered_after_latest_questions?` returns false, and we return `body_questions`.

This is all correct. The test passes as I verified earlier.

Let me now look at the `Load` code for the case when `body_questions` is empty but there are comments:

```ruby
return [] unless github_available?
enhancement_comment = latest_enhancement_comment
return [] if answered_after_latest_questions?(body_questions:, enhancement_comment:)
return [] unless enhancement_comment
Parse.call(comment_body: comment_body(enhancement_comment))
```

For the "when clarifying questions only exist in GitHub comments" test, `body_questions` is empty, `github_available?` is true, `latest_enhancement_comment` finds the comment with the enhancement marker, `answered_after_latest_questions?` returns false (no answer comment), and then it parses the comment. This should work.

OK, I'm going in circles. Let me try a different approach - let me look at what tests were added/changed in the main branch merge and see if any of them could have issues:

All seeds_prompts specs are pending (skipped) because no DB. But 0 failures. That's good.

Let me now check for an issue I spotted. Looking at the `Screenshots::SeedData::Paid` service, it creates a `clarifying_issue` and the test checks `expect(issue).to be_needs_input`. But the seed creates the issue with `find_or_create_by!` on `github_issue_id`. If the record already exists with different data, `find_or_create_by!` will find it and NOT update the fields. This could be an issue in test environments where the database is reset between test runs but... actually in CI each run starts fresh.

OK, I think the most productive approach now is to actually try running the lint and a subset of tests that I can, to see if there are any actual failures. Let me first run lint:

The JS file lint errors are from RuboCop trying to parse JS as Ruby. Let me check how the linter config handles JS:

The RuboCop failures on JS files are expected - RuboCop is a Ruby linter. Let me check how `bin/lint` works:

So `bin/lint` runs multiple linters including ESLint for JS. RuboCop is only for Ruby. The JS file shouldn't be checked by RuboCop directly.

Let me now check what the pre-commit hook does:

The pre-commit hook runs `bin/lint --staged`. That would include linting with RuboCop and ESLint. The pre-commit hook only lints staged files, so if I'm committing Ruby files, RuboCop will lint them but not JS files.

Now let me run the full lint suite to check:

No lint violations. Now let me think about what else could be causing the CI test failure. Let me look at the `db/seeds/prompts.rb` more carefully to see if the `require` of the Prompts module is an issue:

The seed file at line 568 uses `Prompts::GoalCreateGithubIssue::PROMPT_SLUG`. In Rails, `db/seeds/prompts.rb` is loaded via `load Rails.root.join("db/seeds/prompts.rb")` in the test. When loaded in the test context, Zeitwerk should have the `Prompts::GoalCreateGithubIssue` module available.

But wait - the spec at `spec/db/seeds_prompts_spec.rb` has `before { load Rails.root.join("db/seeds/prompts.rb").to_s }`. In the test, the `load` call will execute the seeds file, which references `Prompts::GoalCreateGithubIssue`. Since the module is in `app/services/prompts/goal_create_github_issue.rb`, Zeitwerk should autoload it. This should work.

Let me now try a completely different approach - let me look at the PR's diff on GitHub to see if there are any obvious issues:

Let me check if main has additional commits since we last fetched:

Good, no new commits on main. Our branch is up to date with origin/main.

Now let me look at the `db/migrate/20260512153000_sync_create_github_issue_prompt_clarification_fix.rb` - I noticed the TEMPLATE in this migration has the DRAFTING_GUIDANCE text inline, while `Prompts::GoalCreateGithubIssue::TEMPLATE` uses `#{DRAFTING_GUIDANCE}` interpolation. These might produce slightly different strings due to the extra blank line.

Let me verify this is actually a problem:

I see `\n\n\n` between "omit them." and "You have access". This is the extra blank line from the `#{DRAFTING_GUIDANCE}` interpolation. The `DRAFTING_GUIDANCE` heredoc ends with a newline, then the TEMPLATE has a blank line after `#{DRAFTING_GUIDANCE}`.

Now, the migration TEMPLATE has this text inline with just `\n\n`. So the migration's `synced?` check would fail because the templates are different. The migration's `normalize` method just strips whitespace. But the module TEMPLATE has 3 newlines where the migration has 2. Even after stripping, they'll still differ.

This is a real issue! The migration spec at line 34-50 tests:
```ruby
it "is a no-op when the prompt is already synced" do
  prompt = create(:prompt, :global, :planning, slug: described_class::PROMPT_SLUG, name: "Goal: Create GitHub Issue")
  prompt.create_version!(
    template: described_class::TEMPLATE,
    variables: described_class::VARIABLES,
    created_by: "seed"
  )
  expect { migration.up }.not_to change(PromptVersion, :count)
end
```

This test creates a prompt version with the MIGRATION's TEMPLATE (not the module's). When the migration runs, it checks `synced?` which compares the version's template with its own TEMPLATE. Since they're the same, it's a no-op. This test should pass.

But the seed test at `spec/db/seeds_prompts_spec.rb` line 212-213:
```ruby
it "seeded template matches the shared template source exactly" do
  expect(seed_template).to eq(Prompts::GoalCreateGithubIssue::TEMPLATE)
end
```

This loads the seeds file which creates a prompt with `template: Prompts::GoalCreateGithubIssue::TEMPLATE`. Then it checks if the seeded template matches `Prompts::GoalCreateGithubIssue::TEMPLATE`. They should be the same since the seed uses the same constant. This should pass.

And line 217-218:
```ruby
it "FALLBACK_ISSUE_GOAL_PROMPT matches the shared template source exactly" do
  expect(Activities::RunAgentActivity::FALLBACK_ISSUE_GOAL_PROMPT)
    .to eq(Prompts::GoalCreateGithubIssue::TEMPLATE)
end
```

`FALLBACK_ISSUE_GOAL_PROMPT` is `Prompts::GoalCreateGithubIssue::TEMPLATE`. So this should also pass.

Hmm, but wait - what about the migration test? When the seeds are loaded, they create a prompt with the MODULE's template (which has the extra blank line). Then if the migration runs, it checks if the version's template matches the MIGRATION's template (which doesn't have the extra blank line). They don't match, so the migration would create a new version. But this is actually the desired behavior - the migration is supposed to sync the template.

The issue is: the migration and the module have slightly different templates. The migration template was hand-written to match the module template, but the `#{DRAFTING_GUIDANCE}` interpolation creates an extra blank line. This means:
1. Seeds use the module template (with extra blank line)
2. Migration uses its own template (without extra blank line)
3. They're different, so the migration would always create a new version

This is actually by design - the migration is meant to fix the template to match the migration's version. But the test at `spec/db/seeds_prompts_spec.rb:213` checks that the seeded template matches the module template. After the migration runs, the template would be the migration's version (different from the module template). But the seed spec loads seeds fresh, so the migration doesn't run during the test.

Actually, I think the CI test failure might be something simpler. Let me look at the `spec/db/seeds_prompts_spec.rb` test that checks the template. The test `before { load Rails.root.join("db/seeds/prompts.rb").to_s }` loads the seeds, which creates prompts with `Prompts::GoalCreateGithubIssue::TEMPLATE`. Then it checks the template. This should pass because they're the same constant.

Wait, but the `spec/db/seeds_prompts_spec.rb:213` test compares `seed_template` with `Prompts::GoalCreateGithubIssue::TEMPLATE`. The `seed_template` is fetched from the database:
```ruby
let(:seed_template) do
  described_class.global.find_by(slug: Prompts::GoalCreateGithubIssue::PROMPT_SLUG).current_version.template
end
```

And the seed creates the prompt with:
```ruby
template: Prompts::GoalCreateGithubIssue::TEMPLATE
```

But the seed uses `upsert_global_prompt` which might normalize the template. Let me check:

The seed normalizes by stripping when comparing, but stores the template as-is. The `create_version!` stores the raw template. So the stored template should be exactly `Prompts::GoalCreateGithubIssue::TEMPLATE`. And the test checks `eq` which is exact equality. This should pass.

Now, I'm stuck trying to figure out the CI failure without being able to run the tests. Let me try running the full lint suite to check for any issues:

No RuboCop violations at all. Good.

Let me try to look at the test failures from a different angle. Let me check if there's a test that would fail specifically because of the template mismatch between the migration and the module. The `spec/db/seeds_prompts_spec.rb` tests will:
1. Load `db/seeds/prompts.rb` which creates a prompt with `Prompts::GoalCreateGithubIssue::TEMPLATE`
2. Check that `seed_template == Prompts::GoalCreateGithubIssue::TEMPLATE` (line 213) - this should pass
3. Check that `FALLBACK_ISSUE_GOAL_PROMPT == Prompts::GoalCreateGithubIssue::TEMPLATE` (line 217-218) - this should pass

Wait, but what about the migration spec? Let me re-read it:

```ruby
it "promotes the updated template for the persisted global prompt" do
  prompt = create(:prompt, :global, :planning, slug: described_class::PROMPT_SLUG, name: "Goal: Create GitHub Issue")
  previous_version = prompt.create_version!(template: "old {{base_prompt}}", variables: [{"name" => "base_prompt"}], created_by: "seed")

  expect { migration.up }.to change { prompt.reload.prompt_versions.count }.by(1)

  expect(prompt.reload.current_version).not_to eq(previous_version)
  expect(prompt.current_version.template).to eq(described_class::TEMPLATE)
  expect(prompt.current_version.variables).to eq(described_class::VARIABLES)
  expect(prompt.current_version.created_by).to eq("migration")
  expect(prompt.current_version.change_notes).to eq(described_class::CHANGE_NOTES)
end
```

This creates a prompt with an old template, then runs the migration, and checks that the migration's TEMPLATE was stored. Since `described_class::TEMPLATE` is the migration's template, this should pass.

Let me now look more carefully at the `append_missing_issue_goal_runtime_instructions` method to see if it could cause a problem with the test expectations:

```ruby
def append_missing_issue_goal_runtime_instructions(rendered, slug, vars)
  return rendered unless slug == ISSUE_GOAL_PROMPT_SLUG

  prompt = append_prompt_section(rendered, Prompts::GoalCreateGithubIssue::DRAFTING_GUIDANCE)
  append_prompt_section(prompt, vars[:decomposition_instructions])
end
```

When the seeded template is rendered (which already has DRAFTING_GUIDANCE inline), the `append_prompt_section` for DRAFTING_GUIDANCE will check if the rendered text already includes the guidance. It should, because the seeded template includes it. So no double-append.

BUT wait - there's a subtle issue! The `append_prompt_section` checks:
```ruby
return rendered if rendered.include?(normalized_addition)
```

The `normalized_addition` is `addition.to_s.strip`. The `DRAFTING_GUIDANCE` constant is:
```
Treat the request and repository context already provided above as the full source
material for the GitHub issue you need to file. Synthesize the issue title, body,
and any appropriate labels from that context yourself.

Do NOT reply by asking the user to provide the issue type, title, description,
labels, or other issue-drafting fields. If a field is not explicitly specified in
the provided context, make a reasonable choice and continue. When no labels are
clearly requested, omit them.
```

After stripping, this becomes the same text. The rendered template includes this text (it's in the TEMPLATE constant via `#{DRAFTING_GUIDANCE}`). So `rendered.include?(normalized_addition)` should be true, and it won't be appended.

OK wait, I just realized something. The `#{DRAFTING_GUIDANCE}` in the TEMPLATE is the result of Ruby string interpolation. The `DRAFTING_GUIDANCE` is defined with `<<~'GUIDANCE'` (squiggly heredoc with single-quoted delimiter, so no interpolation within). It strips leading whitespace. So `DRAFTING_GUIDANCE` is:

```
Treat the request and repository context already provided above as the full source
material for the GitHub issue you need to file. Synthesize the issue title, body,
and any appropriate labels from that context yourself.

Do NOT reply by asking the user to provide the issue type, title, description,
labels, or other issue-drafting fields. If a field is not explicitly specified in
the provided context, make a reasonable choice and continue. When no labels are
clearly requested, omit them.
```

(with a trailing newline)

When this is interpolated into TEMPLATE (which uses `<<~TEMPLATE`, non-single-quoted so interpolation happens), the result has the DRAFTING_GUIDANCE text inline. Then after stripping in `append_prompt_section`, the check is whether this text is present in the rendered template. It should be.

Let me now try running ALL the non-DB specs to see if there are any issues:

All 64 no-DB specs pass. Now let me check if there's an issue with the `Screenshots::SeedData::Paid` spec that references `clarifying_issue`:

The test creates an issue via `Screenshots::SeedData::Paid.call`, which creates the issue with:
```ruby
record.paid_state = "needs_input"
record.labels = [ project.enhance_issue_needs_input_label_name ]
```

And the test checks:
```ruby
expect(issue).to be_needs_input
```

This calls `issue.needs_input?` which checks `paid_state == "needs_input" && has_label?(project.enhance_issue_needs_input_label_name)`. Since both are set correctly, this should pass.

Wait, but the `find_or_create_by!` only sets attributes in the block for NEW records. If the issue already exists (from a previous test), the block won't be executed, and the record might have stale data. But in CI, the database is freshly set up, so this shouldn't be an issue.

Let me look for any other potential issues. Let me check if there's a missing `github_creator_login` validation that could cause the seed to fail:

The issue model validates `github_creator_login` presence. In the seed:
```ruby
record.github_creator_login = user.email
```

This is set. Good.

And `github_created_at` and `github_updated_at` are validated as present. Both are set:
```ruby
record.github_created_at = 1.day.ago
record.github_updated_at = Time.current
```

Good.

Let me now take a completely different approach and look at the GitHub issue/PR more carefully. The task says CI was failing on the PR's branch. Maybe the failure was there before the merge from main, and the merge from main might have already fixed it, or maybe the merge introduced new failures.

Let me check what the PR status was before the merge:

No `gh` available. Let me now try to figure out what's happening by looking at the actual code more carefully. Let me check the `issue_timestamp` method in `Load`:

```ruby
def issue_timestamp
  issue.github_updated_at&.to_time || issue.updated_at&.to_time || Time.at(0)
end
```

Wait, the `issue` in the `Load` service is the ActiveRecord issue object. `github_updated_at` is a datetime column, so `.to_time` converts it to a Ruby Time. But `updated_at` is also a datetime column. Both should work.

However, there's a potential issue in the test where the `issue` double doesn't have `updated_at`:
```ruby
let(:issue) do
  double(
    body: issue_body,
    github_number: 1964,
    github_updated_at: 2.minutes.ago
  )
end
```

In the "when the issue body still contains questions after answers were posted" test (line 103-125):
- `body_questions` is non-empty (because `issue_body` contains the enhancement marker)
- `github_available?` is true
- `latest_enhancement_comment` returns nil (no enhancement comment in the comments array)
- `answered_after_latest_questions?` is called with `body_questions:` and `enhancement_comment: nil`
  - `latest_answer_comment` returns the answers_comment
  - `latest_question_timestamp` is called:
    - `issue_timestamp` returns `2.minutes.ago.to_time` (since github_updated_at is 2.minutes.ago)
    - `enhancement_comment` is nil, so no comment timestamp
    - Returns `2.minutes.ago.to_time`
  - `latest_answer_timestamp` returns `1.minute.ago.to_time`
  - Since answer (1 min ago) > question (2 min ago), returns true
- So `answered_after_latest_questions?` returns true
- The code returns `[]`

This is correct - the test expects `[]` because the answers were posted after the issue was last updated.

For the "when the issue body was updated after answers were posted" test (line 127-158):
- `github_updated_at` is `30.seconds.ago` (recent)
- `latest_answer_comment` has `created_at: 1.minute.ago`
- `issue_timestamp` returns `30.seconds.ago.to_time`
- `latest_answer_timestamp` returns `1.minute.ago.to_time`
- Since question (30 sec ago) > answer (1 min ago), `answered_after_latest_questions?` returns false
- So the code returns `body_questions` (the questions from the issue body)

This is also correct.

OK let me step back and think about what the actual CI failure could be. Looking at the PR description again: "CI Status: FAILING - test (failure)". The test workflow runs all specs except performance and system. The key tests from this PR are:

1. `spec/models/issue_spec.rb` - needs DB
2. `spec/requests/projects/clarifying_questions_spec.rb` - needs DB
3. `spec/services/clarifying_questions/load_spec.rb` - no_db, passes
4. `spec/services/clarifying_questions/parse_spec.rb` - no_db, passes
5. `spec/services/clarifying_questions/submit_answers_spec.rb` - no_db, passes
6. `spec/routing/projects/clarifying_questions_routing_spec.rb` - no_db, passes
7. `spec/services/screenshots/capture_targets_spec.rb` - no_db, passes
8. `spec/services/screenshots/seed_data/paid_spec.rb` - needs DB
9. `spec/db/seeds_prompts_spec.rb` - needs DB
10. `spec/migrations/sync_create_github_issue_prompt_clarification_fix_spec.rb` - needs DB
11. `spec/temporal/activities/run_agent_activity_spec.rb` - needs DB

Since I can't run DB-dependent specs, let me look at the request spec more carefully for potential issues. The `ClarifyingQuestionsController` uses `policy_scope(Project)` to find the project. Let me check the `policy_scope` behavior:

The `policy_scope(Project)` filters by `user.account`. The test creates a project with the same account as the user, so it should find it.

Let me now try a different approach. Let me look for any errors that would only show up at runtime in the controller or view:

These helpers exist. Let me check if the `ClarifyingQuestions::Load` issue could fail because `project.github_token.client` returns something unexpected in the request test:

In the request spec:
```ruby
before do
  sign_in user
  user.add_role(:admin, account)
  allow(project.github_token).to receive(:client).and_return(github_client)
end
```

The `github_client` is `instance_double(GithubClient)`. The `Load` service calls `project.github_token.client.issue_comments(...)`.

But wait - the controller creates a fresh `@project` from the database using `policy_scope(Project).find(params[:project_id])`. This is a different Ruby object from the `project` let variable. The mock `allow(project.github_token).to receive(:client).and_return(github_client)` is set up on the factory-created `project` object, but the controller will load a new instance from the DB.

However, in RSpec request specs, `let` variables are lazily evaluated and cached. The `project` is created in the database, and `allow(project.github_token).to receive(:client)` stubs the method on that specific object. But when the controller loads `@project = policy_scope(Project).find(params[:project_id])`, it gets a DIFFERENT Ruby object from the database, and the mock won't be set up on it.

Wait, but this is a common pattern in Rails request specs. The `project` let creates the record in the DB, and then the controller loads it. The mock on `project.github_token` stubs the association proxy on that specific project instance. When the controller loads a new instance, it has its own `github_token` association.

Actually, I think there might be an issue here. Let me look at how other request specs handle this:

Let me check the `human_feedback_collection_job_spec.rb` to see how they mock:

I see! In the job specs, they mock the object AND use `allow(Model).to receive(:find).with(id).and_return(mock)` to ensure the mock is used. The request spec doesn't do this - it relies on the actual controller loading the project from DB.

In a request spec, the `project` is created in the DB, and `allow(project.github_token).to receive(:client)` stubs that specific instance. But the controller loads a new instance. However, in Rails, associations are cached on the instance. When the controller loads a fresh `@project`, it will load its own `@project.github_token` from the DB, and the stub won't be on that object.

This IS the problem! The request spec stubs `project.github_token.client` on the factory-created instance, but the controller creates a new instance from the DB. The new instance's `github_token.client` won't be stubbed.

But wait - this depends on how the test database works. In Rails test mode with transactional fixtures, the `project` and its `github_token` are created within the same transaction. The controller's `find` will load the same DB record, but as a different Ruby object. The stub is on the original object, not the new one.

This means the `github_client` mock won't be used by the controller, and the `Load` service will try to call the real `GithubClient` methods, which will likely fail.

Actually, wait. Let me re-read the test more carefully:

```ruby
before do
  sign_in user
  user.add_role(:admin, account)
  allow(project.github_token).to receive(:client).and_return(github_client)
end
```

And the controller:
```ruby
def set_project
  @project = policy_scope(Project).find(params[:project_id])
end
```

The `@project` loaded by the controller is a different Ruby object from the `project` in the test. So `@project.github_token` returns a different GithubToken object (loaded from DB). And `@project.github_token.client` would not be stubbed.

But wait - is the `github_token` association cached? If `project.github_token` is called first (in the before block), the association is loaded on that instance. But the controller's `@project` is a different instance, so it loads its own association.

So the mock would fail... but this test was written by the PR author and presumably worked at some point. Let me check if there's something I'm missing. Maybe `GithubToken#client` is memoized or there's a global cache.

Let me check:

The `client` is memoized with `@client`. But since each Ruby object has its own instance variables, the mock on one object won't affect another.

Wait, but in the request spec, the `project` is created via `create(:project, account: account)`, which also creates a `github_token` via the factory association. The `allow(project.github_token).to receive(:client)` stubs the `client` method on that specific `github_token` object.

When the controller loads `@project = policy_scope(Project).find(params[:project_id])`, and then `@project.github_token` returns the SAME database record but as a DIFFERENT Ruby object. So the stub doesn't apply.

Hmm, but maybe I'm wrong. Let me think about this more carefully in the context of RSpec and Rails request specs...

Actually, in Rails request specs with transactional fixtures, the records are in the DB. When the controller loads a record, it creates a new model instance. The mock on the original instance doesn't transfer.

BUT - there's a possibility that the `github_token` association uses a shared cache or that the `project` factory's `github_token` is the same object. Let me check:

When `create(:project, account: account)` is called, it creates the project and the associated github_token. The `project.github_token` returns this associated token. But when the controller loads `@project = policy_scope(Project).find(...)`, the `@project.github_token` is loaded from the DB as a new object.

So the mock should NOT work. This means the request spec would fail when the controller tries to call `project.github_token.client` because:
1. The real `GithubClient` would be instantiated with a test token
2. It would try to connect to GitHub, which would fail

But the test expects to mock the GitHub API calls. This IS the bug!

Wait, let me double-check by looking at how other request specs handle this. Let me find a request spec that mocks GitHub API:

In `agent_runs_spec.rb`, they use `allow(GithubClient).to receive(:new).and_return(github_client)` to stub the class method, so ANY new GithubClient creation returns the mock. This is the correct approach for request specs.

But in `clarifying_questions_spec.rb`, they use `allow(project.github_token).to receive(:client).and_return(github_client)`, which only stubs the specific `github_token` instance. This won't work when the controller loads a different project instance.

Wait, but let me check another thing. The `ClarifyingQuestions::Load` service accesses `project.github_token.client`, and the controller passes `@project` (a different instance) to the service. So the mock won't be applied.

This is definitely the bug! The request spec needs to use `allow(GithubClient).to receive(:new).and_return(github_client)` or some other approach to ensure the mock is applied to the controller's project instance.

But wait - let me also check if there's a way the mock could work. Maybe the `github_token` association is somehow shared between instances:

No, ActiveRecord associations are per-instance. Each `project.github_token` call loads from the DB independently.

Let me also check if the `ClarifyingQuestions::Load` service accesses `project.github_token` in a way that would use the mock:

```ruby
def github_available?
  project.github_token.present?
end

def issue_comments
  @issue_comments ||= project.github_token.client.issue_comments(project.full_name, issue.github_number)
end
```

When the controller passes `@project` to the service, `project.github_token.client` is called on the controller's project instance, not the test's mock. So the mock doesn't apply.

However, looking at it again - maybe in the test, the `project` is the SAME Ruby object? No, `let(:project) { create(:project, account: account) }` creates a new project and stores it. The controller's `find` creates a new instance.

Wait, but RSpec `let` is lazy-loaded. The `project` is created when first referenced. In the `before` block, `allow(project.github_token).to receive(:client)` creates the project (first reference) and stubs the mock. Then when the controller runs, it loads a NEW instance.

So the fix is to use `allow_any_instance_of(GithubToken).to receive(:client).and_return(github_client)` or `allow(GithubClient).to receive(:new).and_return(github_client)`.

Actually, `allow_any_instance_of` is discouraged in RSpec. The better approach is to stub `GithubClient.new`. But let me look at what the other request specs do:

Confirmed! The `projects_spec.rb` uses `allow(GithubClient).to receive(:new).and_return(github_client)` to stub at the class level. The `clarifying_questions_spec.rb` incorrectly stubs on the instance level.

This is the CI test failure! The request spec for clarifying questions will fail because the mock doesn't apply to the controller's project instance.

Let me fix the request spec:

Now I need to fix this. The issue is that `allow(project.github_token).to receive(:client).and_return(github_client)` only stubs the specific `github_token` instance from the factory, but the controller loads a fresh project from the DB. I need to stub at a level that works across instances.

Looking at the `GithubToken#client` method:
```ruby
def client
  @client ||= GithubClient.new(token: token)
end
```

The best approach, following the pattern from `projects_spec.rb`, is to stub `GithubClient.new`:

Now let me verify the full file looks correct:

Now let me also check if there are any other issues. Let me run RuboCop on this file:

Now let me do a proactive review of the entire diff to look for any other issues:

Wait, `.rubocop.yml` changed? Let me check:

Hmm, that `.rubocop.yml` change looks wrong. It wasn't supposed to be there. Let me revert it:

Good. Now let me do the proactive scan as required by the instructions. Let me review the entire diff of this PR:

Wait, the diff shows the original file. Let me check the current state:

viamin:
## Agent Update

The issue is that bundler is trying to write to `/usr/local/lib/ruby/gems/3.4.0/cache/` which is not writable. Let me configure bundler to use a local path:

Bundle install succeeded. Now let me install yarn dependencies and run lint:

Disk space issue. Let me clean up and try again:

viamin:
## Agent Update

Resolved the merge-blocker check first: after unshallowing the checkout, `origin/main` was already merged into this branch, so there were no actual content conflicts to resolve.

I fixed the screenshot regression in [app/services/screenshots/capture_targets.rb](/workspace/app/services/screenshots/capture_targets.rb:275). When `app/javascript/controllers/index.js` changed alongside a specifically mapped Stimulus controller, the resolver returned `nil` for the registry file, which made screenshot target detection treat it as unmapped and fail `capture`. It now reuses the explicit controller targets instead of broadening capture or crashing.

Validation: `bundle exec rubocop` passed, and the relevant dbless specs passed with `ALLOW_DBLESS_SPECS=true`, including the screenshot target spec that covers this case. `bundle exec rspec` could not run end-to-end in this container because the repo aborts immediately without PostgreSQL; the failure was `ActiveRecord::ConnectionNotEstablished` during test DB preparation, not an application spec failure.

Committed as `8fb2d051` with:
`fix(screenshots): map stimulus registry changes to explicit capture targets`

viamin:
## Agent Update

Merged state was already up to date with `origin/main`, so there were no conflicts to resolve.

I fixed a robustness issue in `ClarifyingQuestions::Load`: if the issue body already contains the clarifying questions and GitHub comment loading fails, the wizard now falls back to the body questions instead of failing the whole flow. I added a spec for that fallback and committed it as `315c4279` with `fix(clarifying-questions): fall back to issue body on github load errors`.

Validation: `bundle exec rubocop` passed repo-wide. DB-less targeted specs passed with `ALLOW_DBLESS_SPECS=true COVERAGE=false`, including the clarifying question service/spec set and screenshot target/routing specs. A full `bundle exec rspec` run could not execute in this container because PostgreSQL is unavailable, and the suite aborts during test database setup before running examples.

viamin:

PR #1979, opened 2026-05-12
feat(scaling): implement scaling-based resource allocator

### LINKED ISSUE #1816: feat(scaling): implement scaling-based resource allocator
## Description

Use measured scaling behavior to influence resource allocation decisions for future runs.

## Tasks

- [ ] Define the allocator inputs from scaling observations and experiments
- [ ] Implement allocation rules that respond to measured returns
- [ ] Add safe fallbacks when data is sparse or stale
- [ ] Add specs for allocation choices across representative scenarios

viamin:
## Summary

Improves resource allocation decisions for future agent runs by grounding them in measured scaling behavior. The allocator now reads cohort returns more accurately and disregards stale experiment summaries so allocation stays responsive to recent signal.

## Changes

- **Services**
  - `app/services/scaling/resource_allocator.rb`: group observation cohorts by planned agent count (rather than launched count); factor measured launch shortfall and blocked-capacity signals into the scoring function.
  - `app/services/scaling_experiments/summarize_results.rb`: emit a generated timestamp on each summary so consumers can detect and skip stale data.
- **Specs**
  - `spec/services/scaling/resource_allocator_spec.rb`: representative coverage across cohort/scoring scenarios.
  - `spec/services/scaling_experiments/summarize_results_spec.rb`: covers the generated-timestamp behavior.

## Design Decisions

- **Cohort grouping by planned count, not launched count.** Launched count conflates demand with capacity. Grouping by planned agents preserves the original allocation intent and produces cleaner per-cohort return curves, even when some launches were blocked.
- **Treat blocked capacity as signal, not noise.** Launch shortfall and blocked-capacity are now inputs to scoring rather than being filtered out, so the allocator recognizes when a higher allocation would have been used had it been available.
- **Staleness via generated timestamp.** A timestamp on each summary is a cheap, explicit staleness check; the allocator falls back safely when summaries are missing or expired rather than acting on outdated returns.

## Operational Notes

- No schema migrations or infrastructure changes.
- Verification in-container: `bundle exec rubocop` passed. The DB-less scaling specs pass under `ALLOW_DBLESS_SPECS=true COVERAGE=false`; the DB-backed `ScalingExperiments::SummarizeResults` spec exists but is skipped locally because no PostgreSQL service is available — it will run in CI.

---

Closes #1816

### DIFF
--- a/app/models/scaling_experiment.rb
+++ b/app/models/scaling_experiment.rb
@@ -168,8 +168,12 @@ class ScalingExperiment < ApplicationRecord
   def cached_or_compute_summary(persist: true)
     current_key = samples_key
     if cached_summary.present?
-      return cached_summary if summary_samples_key == current_key
-      return cached_summary unless persist
+      normalized_cached_summary = cached_summary_with_generated_at
+      if summary_samples_key == current_key
+        persist_cached_summary_generated_at!(normalized_cached_summary) if persist
+        return normalized_cached_summary
+      end
+      return normalized_cached_summary unless persist
     end
 
     summary = ScalingExperiments::SummarizeResults.call(scaling_experiment: self)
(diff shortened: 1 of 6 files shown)


viamin:
## Agent Update

Addressed the review thread by changing [app/services/scaling/resource_allocator.rb](/workspace/app/services/scaling/resource_allocator.rb) so timestamp-less generated experiment summaries are treated as stale instead of implicitly fresh. The allocator still accepts flat ad hoc per-value summaries without timestamps; the stricter check only applies to cached `SummarizeResults`-style payloads.

Updated [spec/services/scaling/resource_allocator_spec.rb](/workspace/spec/services/scaling/resource_allocator_spec.rb) to add the regression case for legacy cached summaries with no timestamp, and I refreshed the nested-summary fixtures so valid current-format summaries include `generated_at`.

Verification:
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/services/scaling/resource_allocator_spec.rb` passed
- `bundle exec rspec` could not run in this container because PostgreSQL is unavailable and the suite aborts during test DB setup

Commit: `8a6f2bbe` `fix(scaling): treat legacy summaries as stale`

viamin:
## Agent Update

Addressed the review thread by preserving legacy experiment summaries instead of dropping them as stale on deploy. `ScalingExperiment#cached_or_compute_summary` now hydrates missing `generated_at` from the experiment record timestamp and persists that back into `cached_summary` when the cached sample key is still current, so stable experiments keep producing usable summaries after rollout.

I also updated coverage in [spec/services/scaling/resource_allocator_spec.rb](/workspace/spec/services/scaling/resource_allocator_spec.rb) to accept legacy summaries when record timestamps are present, and added a DB-less compatibility spec in [spec/models/scaling_experiment_spec.rb](/workspace/spec/models/scaling_experiment_spec.rb) for the read-time backfill path.

Verification: `bundle exec rubocop` passed for the touched files and the commit hook passed `bin/lint --staged`. `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/services/scaling/resource_allocator_spec.rb spec/models/scaling_experiment_spec.rb:139` passed. A full `bundle exec rspec` still cannot run in this container because PostgreSQL is unavailable and the suite aborts during test DB setup.

Committed as `fix(scaling): preserve legacy cached experiment summaries` (`c38e9678`).

PR #2007, opened 2026-05-14
fix(dashboard): use shared helpers for active-run context and goal labels

viamin:
## Summary

- The dashboard's active-runs table hand-rolled its context cell and rendered a generic `"Custom prompt"` label whenever an agent run had a custom prompt — hiding the actual prompt content from operators watching the run.
- The same partial also rendered goal labels via `run.goal.titleize`, producing ugly strings like "Create Pr" instead of the "PR Creation" / "Code Review" / "Issue Creation" labels used everywhere else in the app.
- Switched both columns to the shared helpers (`agent_run_context_display`, `agent_run_goal_display`) already used by the queue preview and the agent-runs index tables, so operators see redacted + truncated prompt text with hover tooltips and consistent goal labels.
- Added an explicit "View" action so the row still drills into the run details page (the helper renders external GitHub links rather than being the click target, matching the pattern in `app/views/agent_runs/_index_table.html.erb`).

## Test plan

- [ ] Visit `/dashboard` with active runs that have a custom prompt — confirm the context cell shows a (redacted) snippet of the prompt with a tooltip showing the full redacted prompt, instead of the generic "Custom prompt" label.
- [ ] Confirm active runs tied to a GitHub issue render `Issue #N` with title tooltip, and runs tied to a source PR render `PR #N` linking to the PR on GitHub.
- [ ] Confirm the Goal column renders "PR Creation" / "Issue Creation" / "Code Review" / "Enhance Issue" / "Analyze Issue" instead of "Create Pr" etc.
- [ ] Confirm the new "View" link in the Actions column navigates to the run detail page and the Cancel button still cancels the run.

### DIFF
--- a/app/views/dashboard/_active_run_row.html.erb
+++ b/app/views/dashboard/_active_run_row.html.erb
@@ -9,17 +9,9 @@
     <%= link_to run.project.full_name, project_path(run.project), class: "hover:text-indigo-600" %>
   </td>
   <td class="px-4 py-3 text-sm text-gray-500">
-    <% if run.issue.present? %>
-      <%= link_to "##{run.issue.github_number} #{run.issue.title.truncate(60)}", project_agent_run_path(run.project, run), class: "hover:text-indigo-600" %>
-    <% elsif run.custom_prompt.present? %>
-      <%= link_to "Custom prompt", project_agent_run_path(run.project, run), class: "hover:text-indigo-600" %>
-    <% elsif run.source_pull_request_number.present? %>
-      <%= link_to "PR ##{run.source_pull_request_number}", project_agent_run_path(run.project, run), class: "hover:text-indigo-600" %>
-    <% else %>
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

Addressed the review thread by updating [spec/views/dashboard/active_runs_partial_spec.rb](/workspace/spec/views/dashboard/active_runs_partial_spec.rb:39) to stub `agent_run_context_display` and `agent_run_goal_display`, and to assert the dashboard row renders the shared helper output. That matches the new partial behavior and avoids the helper-contract failure from the minimal `Struct`.

Committed as `9d629a06` with message `fix(dashboard): cover shared active-run helpers in spec`.

Checks:
- `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec spec/views/dashboard/active_runs_partial_spec.rb spec/views/agent_runs/index_table_partial_spec.rb spec/helpers/application_helper_context_display_spec.rb` passed.
- `bundle exec rspec` could not run in this container because PostgreSQL is unavailable; the suite aborts during test DB setup by design.

viamin:

PR #2008, opened 2026-05-14
fix(agents): increase create_pr startup timeout from 300s to 360s

viamin:
## Summary

- Increase `DEFAULT_CREATE_PR_IDLE_TIMEOUT` and `DEFAULT_AGENT_STARTUP_TIMEOUT` from 300s (5min) to 360s (6min) to reduce Claude startup timeout fallbacks to Codex
- Add migration to update `user_settings.create_pr_idle_timeout_seconds` column default and existing rows from 300 to 360
- Sync `OrchestrationStrategies::Defaults` to match the new value

## Problem

Claude's API regularly takes 250-300s to produce first output on complex prompts. Analysis of 737 agent runs showed:

| Metric | Value |
|--------|-------|
| Claude runs assigned | 335 (45.4%) |
| Claude runs actually completed by Claude | 101 (14.4%) |
| Claude runs that fell back to Codex | 130 (39%) |
| Successful Claude runs at 295-300s | 4 (barely under timeout) |

The 300s startup timeout was razor-thin — 4 successful runs completed with <1s to spare. Codex had only 1 startup timeout out of 318 runs (0.3%), confirming this is a Claude API latency issue, not a Paid bug.

## Testing

- All 272 tests in affected spec files pass (`run_agent_activity_spec.rb`, `defaults_spec.rb`, `heartbeat_setup_spec.rb`)
- Lint checks pass

### DIFF
--- a/app/services/orchestration_strategies/defaults.rb
+++ b/app/services/orchestration_strategies/defaults.rb
@@ -100,7 +100,7 @@ module OrchestrationStrategies
         "issue_goal_timeout_seconds" => 600,
         "issue_goal_idle_timeout_seconds" => 120,
         "review_goal_idle_timeout_seconds" => 300,
-        "create_pr_idle_timeout_seconds" => 300,
+        "create_pr_idle_timeout_seconds" => 360,
         "preflight_timeout_seconds" => 10,
         "feature_orchestration_timeout_seconds" => 7200
       }
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -79,8 +79,8 @@ module Activities
     DEFAULT_ISSUE_GOAL_TIMEOUT = 600        # 10 minutes wall clock
     DEFAULT_ISSUE_GOAL_IDLE_TIMEOUT = 120   # 2 minutes without output = stuck
     DEFAULT_REVIEW_GOAL_IDLE_TIMEOUT = 300  # 5 minutes without output = stuck
-    DEFAULT_CREATE_PR_IDLE_TIMEOUT = 300   # 5 minutes without output = stuck
-    DEFAULT_AGENT_STARTUP_TIMEOUT = 300    # 5 minutes without first output = stuck
+    DEFAULT_CREATE_PR_IDLE_TIMEOUT = 360   # 6 minutes without output = stuck
+    DEFAULT_AGENT_STARTUP_TIMEOUT = 360    # 6 minutes without first output = stuck
     PREFLIGHT_TIMEOUT_SECONDS = 10
     CHANGE_DETECTION_MAX_ATTEMPTS = 3
     CHANGE_DETECTION_RETRY_BACKOFF = 0.25
--- a/db/migrate/20260514193654_increase_create_pr_idle_timeout_default.rb
+++ b/db/migrate/20260514193654_increase_create_pr_idle_timeout_default.rb
@@ -1,12 +1,16 @@
 # frozen_string_literal: true
 
 class IncreaseCreatePrIdleTimeoutDefault < ActiveRecord::Migration[8.1]
+  class MigrationUserSetting < ApplicationRecord
+    self.table_name = "user_settings"
+  end
+
   def change
     change_column_default :user_settings, :create_pr_idle_timeout_seconds, from: 300, to: 360
 
     reversible do |dir|
       dir.up do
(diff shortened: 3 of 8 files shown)


viamin:
## Agent Update

OK

Commit succeeded. The migration now scopes `update_all` to only rows matching the old default (300→360 in `up`, 360→300 in `down`), preserving any intentional custom timeout overrides as the reviewer requested.

viamin:
## Agent Update

Updated the migration so the `user_settings.create_pr_idle_timeout_seconds` default still rolls back from `360` to `300`, but the row backfill only runs on `up`. That resolves the review thread: rollback no longer rewrites legitimate user-selected `360` values. I also added a migration spec in [spec/migrations/increase_create_pr_idle_timeout_default_spec.rb](/workspace/spec/migrations/increase_create_pr_idle_timeout_default_spec.rb:1) to lock in that behavior.

`bundle install` and `yarn install` were completed locally. `bundle exec rubocop` passed. `bundle exec rspec` could not run in this container because the repo aborts immediately when PostgreSQL is unavailable (`ActiveRecord::ConnectionNotEstablished` from `spec/rails_helper.rb`), so I could not execute the new migration spec or the full suite here.

Committed on the current branch as `44cbb1eb` with message `fix(agents): avoid clobbering create_pr timeout rollback`.

viamin:
## Agent Update

The CI-facing issue was in the new migration spec, not the timeout constants. `spec/migrations/increase_create_pr_idle_timeout_default_spec.rb` was running `migration.migrate(:down)` and leaving the test database schema at the old `300` default for later examples. I changed that spec to restore the schema to the post-migration state in an `around` block, reset model column info after each migration step, and assert the column default from the live connection instead of cached `UserSetting.column_defaults`. I also made the migration use a migration-local model in [db/migrate/20260514193654_increase_create_pr_idle_timeout_default.rb](/workspace/db/migrate/20260514193654_increase_create_pr_idle_timeout_default.rb:1), and added a guard spec in [spec/services/orchestration_strategies/defaults_spec.rb](/workspace/spec/services/orchestration_strategies/defaults_spec.rb:61) so the orchestration default stays aligned with `RunAgentActivity`.

Validation: `bundle exec rubocop` passed for the full repo. A full `bundle exec rspec` could not run in this container because PostgreSQL is unavailable and the suite aborts fast with `ActiveRecord::ConnectionNotEstablished`; I also confirmed the db-less path with `ALLOW_DBLESS_SPECS=true COVERAGE=false` for the non-migration defaults spec. The branch is clean, and the changes are committed as `f592e328` with message `fix(ci): isolate timeout migration spec state`.

viamin:
## Agent Update

Added a follow-up CI fix in [spec/migrations/increase_create_pr_idle_timeout_default_spec.rb](/workspace/spec/migrations/increase_create_pr_idle_timeout_default_spec.rb) so the non-transactional migration spec now truncates its test data both before and after the example, while still restoring the schema default in `ensure`. That closes the remaining isolation gap most likely to cause cross-spec contamination in the full `test` job.

Verification: `bundle exec rubocop` passed, and `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec spec/services/orchestration_strategies/defaults_spec.rb` completed cleanly in db-less mode. `bundle exec rspec` and the migration spec itself could not be run in this container because PostgreSQL is unavailable here; the suite aborts up front with `ActiveRecord::ConnectionNotEstablished`. The change is committed as `a5632a2` with message `fix(ci): isolate timeout migration test data`.

viamin:
## Agent Update

The CI fix is in [spec/support/migration_spec_helpers.rb](/workspace/spec/support/migration_spec_helpers.rb): migration-spec cleanup now deletes `user_settings` before `users`. That isolates the new timeout migration spec from seeded `user_settings` rows in CI and avoids foreign-key/state leakage during cleanup.

Committed as `29447ca` (`fix(ci): clean user settings in migration specs`).

Validation here:
- `bundle exec rspec` could not run in this container because PostgreSQL is unavailable: `ActiveRecord::ConnectionNotEstablished`.
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/migrations/increase_create_pr_idle_timeout_default_spec.rb --dry-run` passed, which confirms the target example loads cleanly and the helper change is syntactically/structurally sound.

The worktree is clean.

viamin:

PR #2041, opened 2026-05-15
Fix #2040: Optimize slow spec: Optimizer.ranked_candidates incumbent retention test (111s)

### LINKED ISSUE #2040: Optimize slow spec: Optimizer.ranked_candidates incumbent retention test (111s)
## Problem

The following test takes ~112 seconds to run, making it one of the slowest specs in the suite:

```
ConfigurationBundles::Optimizer.ranked_candidates keeps the incumbent from older outcomes outside the surrogate training window
  111.66 seconds ./spec/services/configuration_bundles/optimizer_spec.rb:249
```

## Location

`spec/services/configuration_bundles/optimizer_spec.rb:249`

## Suggested investigation areas

- Check whether the test is doing unnecessary database writes or creating too many records via factories
- Look for opportunities to stub or simplify the surrogate training window setup — the test only needs to verify that the incumbent is retained from older outcomes, not exercise the full optimization pipeline
- Consider whether shared setup (`before` blocks) is doing expensive work that could be scoped more tightly or cached
- Evaluate if the optimizer is performing real computation (e.g. model fitting) that could be stubbed for this particular assertion

## Expected outcome

Reduce the runtime of this spec significantly (ideally under 10 seconds) without sacrificing coverage of the incumbent-retention behavior.

viamin:
## Summary

Reduces the runtime of `ConfigurationBundles::Optimizer.ranked_candidates` incumbent retention spec from ~112s to a fraction of that, by stubbing the training dataset window size instead of creating 500 real records. The behavior under test — that the incumbent `best_observed_objective_score` is retained from outcomes outside the training window — is unchanged.

## Changes

- **Spec (`spec/services/configuration_bundles/optimizer_spec.rb:249`)**: Stub `TrainingDataset::MAX_ROWS` to `3` and create only 3 `BundleOutcome` records (plus associated `AgentRun`/`Issue` fixtures) instead of 500, since the assertion only depends on outcomes falling outside the window — not on the window's absolute size.

## Design Decisions

- **Stub `MAX_ROWS` rather than the optimizer internals**: Keeps the full optimization pipeline exercised end-to-end while shrinking only the dataset volume. This preserves coverage of the real code path that pulls `best_observed_objective_score` from the unlimited `prior_objective_outcomes_for_goal` query, which is the actual behavior under test.
- **Constant stubbing is safe here**: `MAX_ROWS` is a window threshold, not a value with semantic meaning at `500` specifically — any value smaller than the total outcome count produces the same "incumbent is outside the window" condition.

---

Closes #2040

### DIFF
--- a/spec/services/configuration_bundles/optimizer_spec.rb
+++ b/spec/services/configuration_bundles/optimizer_spec.rb
@@ -247,6 +247,8 @@ RSpec.describe ConfigurationBundles::Optimizer do
     end
 
     it "keeps the incumbent from older outcomes outside the surrogate training window" do
+      stub_const("ConfigurationBundles::TrainingDataset::MAX_ROWS", 3)
+
       create_bundle_history(
         experiment: experiment,
         variant: challenger,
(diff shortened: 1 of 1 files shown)


viamin:

PR #2042, opened 2026-05-15
Fix #2038: Max draft review rounds not saving for projects (missing from strong params)

### LINKED ISSUE #2038: Max draft review rounds not saving for projects (missing from strong params)
## Summary

When updating the **max draft review rounds** setting for a project, the new value is not persisted. After hitting save, the page reloads and the old value is displayed.

## Steps to reproduce

1. Navigate to a project's settings/edit page
2. Change the "max draft review rounds" field to a new value
3. Click Save
4. Observe that the page reloads with the **old** value still in place

## Expected behavior

The new max draft review rounds value should be saved and displayed after the page reloads.

## Actual behavior

The old value persists after saving. The update appears to be silently dropped.

## Likely cause

The `max_draft_review_rounds` parameter is probably not included in the strong parameters (permitted params) in the controller, so Rails filters it out before the update reaches the model.

viamin:
## Summary

Lock in the fix for issue where `max_draft_review_rounds` updates were silently dropped by adding regression coverage. The controller already permits the attribute on this branch, so this PR ensures a future change cannot reintroduce the bug without a failing test.

## Changes

- **Tests**: Added a request spec in `spec/requests/projects_spec.rb` exercising `PATCH /projects/:id` with `max_draft_review_rounds` and asserting the new value persists after reload.

## Design Decisions

- **Test-only change**: The strong-params permit for `max_draft_review_rounds` is already in place on this branch, so no controller change was needed. The regression spec is the durable contract that prevents the attribute from being dropped from permitted params again.
- **Request spec over controller spec**: Exercises the full update path (params filtering + model persistence + reload) rather than asserting on permitted-params internals, matching the user-observed failure mode in the issue.

## Operational Notes

- No migrations, deploy steps, or infrastructure changes.
- Local verification: `bundle exec rubocop` passed. `bundle exec rspec` could not run in the authoring container due to PostgreSQL being unavailable (`ActiveRecord::ConnectionNotEstablished` during test DB setup); CI will execute the new spec.

---

Closes #2038

### DIFF
--- a/spec/requests/projects_spec.rb
+++ b/spec/requests/projects_spec.rb
@@ -1300,10 +1300,21 @@ RSpec.describe "Projects" do
       it "persists max draft review rounds" do
         project = create(:project, account: account, github_token: github_token, max_draft_review_rounds: 10)
 
-        patch project_path(project), params: { project: { max_draft_review_rounds: 4 } }
+        expect {
+          patch project_path(project), params: { project: { max_draft_review_rounds: 4 } }
+        }.to change { project.reload.max_draft_review_rounds }.from(10).to(4)
 
         expect(response).to redirect_to(project_path(project))
-        expect(project.reload.max_draft_review_rounds).to eq(4)
+      end
+
+      it "persists zero max draft review rounds" do
+        project = create(:project, account: account, github_token: github_token, max_draft_review_rounds: 10)
+
+        expect {
+          patch project_path(project), params: { project: { max_draft_review_rounds: 0 } }
+        }.to change { project.reload.max_draft_review_rounds }.from(10).to(0)
+
+        expect(response).to redirect_to(project_path(project))
       end
 
       it "persists screenshot settings" do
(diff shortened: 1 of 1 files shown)


viamin:
## Agent Update

Committed as `45bd987d` with a tighter regression spec in [spec/requests/projects_spec.rb](/workspace/spec/requests/projects_spec.rb).

The change makes the existing `max_draft_review_rounds` request spec assert an actual persisted change on reload, and adds a second case for updating the value to `0` so future falsy-value handling regressions are caught too.

Verification:
- `COVERAGE=false bundle exec rubocop` passed.
- `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec --dry-run spec/requests/projects_spec.rb` passed.
- `COVERAGE=false bundle exec rspec` could not run in this container because PostgreSQL is unavailable; it fails immediately in `spec/rails_helper.rb` with `ActiveRecord::ConnectionNotEstablished`.

viamin:

PR #2049, opened 2026-05-15
Fix #2045: Pressing Enter in project settings form inputs triggers auto-detect instead of saving

### LINKED ISSUE #2045: Pressing Enter in project settings form inputs triggers auto-detect instead of saving
## Bug Description

On the project settings page, pressing Enter while focused on a form input (e.g., "max review rounds") triggers the framework auto-detect action instead of submitting/saving the form. The flash message displays something like **"detected rails with high confidence"** rather than confirming the settings update.

Clicking the "Save Settings" button works correctly and submits the form as expected.

## Steps to Reproduce

1. Navigate to a project's settings page
2. Focus on a text input field (e.g., "max review rounds")
3. Change the value
4. Press **Enter**

## Expected Behavior

Pressing Enter in a form input should submit the settings form (same behavior as clicking the Save Settings button), updating the setting value.

## Actual Behavior

Pressing Enter triggers the framework auto-detect action. A flash message appears saying "detected rails with high confidence" (or similar), and the form value is **not** updated/saved.

## Root Cause (Likely)

There is probably a separate auto-detect button/form on the project settings page. When Enter is pressed inside an input field, the browser submits the **first** `<form>` or activates the first submit-type button it finds in DOM order, which appears to be the auto-detect button rather than the save settings button. This is a common HTML issue when multiple forms or multiple submit buttons exist on the same page.

Possible fixes:
- Ensure the auto-detect button has `type="button"` instead of `type="submit"` so it is not treated as a default form submission target
- Separate the auto-detect action into its own `<form>` element if it is currently sharing a form with the settings inputs
- Add an explicit `keydown` handler on the settings form to ensure Enter triggers the correct save action

## Environment

- Rails application
- Project settings page

viamin:
## Summary

Fixes a bug where pressing Enter in project settings form inputs triggered the framework auto-detect action instead of saving the form. Users now get the expected behavior: Enter submits the settings form via the Save Changes button.

## Changes

- **Views**: Updated `app/views/projects/edit.html.erb` to wire the project settings form to a new Stimulus controller.
- **JavaScript**: Added `app/javascript/controllers/project_settings_form_controller.js` to intercept Enter keypresses on single-line inputs and route submission through the explicit Save Changes button instead of the screenshot action submitters. Registered the controller in `app/javascript/controllers/index.js`.
- **Tests**: Added a regression spec in `spec/requests/projects_spec.rb` asserting the rendered form contract.

## Design Decisions

- Chose a Stimulus controller over restructuring the page into separate `<form>` elements to keep the existing layout and form-sharing intact while addressing the default-submitter ordering issue at the JS layer.
- Targets single-line inputs only so Enter still inserts newlines in textareas as expected.

## Screenshots

_Author: please attach before/after screenshots demonstrating that pressing Enter in a settings input now saves rather than triggering auto-detect._

## Notes

`bundle exec rspec` could not run in the implementation container because PostgreSQL was unavailable (`ActiveRecord::ConnectionNotEstablished` during test DB preparation). Reviewer/CI should confirm the new spec passes.

---

Closes #2045

### DIFF
--- a/app/javascript/controllers/index.js
+++ b/app/javascript/controllers/index.js
@@ -73,6 +73,9 @@ application.register("notification-dropdown", NotificationDropdownController)
 import ProviderFormController from "./provider_form_controller"
 application.register("provider-form", ProviderFormController)
 
+import ProjectSettingsFormController from "./project_settings_form_controller"
+application.register("project-settings-form", ProjectSettingsFormController)
+
 import RepositorySelectorController from "./repository_selector_controller"
 application.register("repository-selector", RepositorySelectorController)
 
--- a/app/javascript/controllers/project_settings_form_controller.js
+++ b/app/javascript/controllers/project_settings_form_controller.js
@@ -40,6 +40,6 @@ export default class extends Controller {
   submittableInput(target) {
     if (!target || target.tagName !== "INPUT") return false
 
-    SUBMITTABLE_INPUT_TYPES.has(target.type)
+    return SUBMITTABLE_INPUT_TYPES.has(target.type)
   }
 }
--- a/app/javascript/controllers/project_settings_form_controller.js
+++ b/app/javascript/controllers/project_settings_form_controller.js
@@ -1,0 +1,45 @@
+import { Controller } from "@hotwired/stimulus"
+
+const SUBMITTABLE_INPUT_TYPES = new Set([
+  "date",
+  "datetime-local",
+  "email",
+  "month",
+  "number",
+  "password",
+  "search",
+  "tel",
+  "text",
+  "time",
+  "url",
+  "week"
+])
+
+export default class extends Controller {
+  static targets = ["saveButton"]
+
+  submitOnEnter(event) {
+    if (!this.shouldSubmitOnEnter(event)) return
+
+    event.preventDefault()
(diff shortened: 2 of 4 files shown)


viamin:
## Agent Update

Fixed the unresolved review thread in [app/javascript/controllers/project_settings_form_controller.js](/workspace/app/javascript/controllers/project_settings_form_controller.js:41) by returning the `SUBMITTABLE_INPUT_TYPES.has(target.type)` result. Without that `return`, `submittableInput` always evaluated falsy and the Enter-to-save behavior never ran.

Validation: `bin/lint --changed` passed, `bundle exec rubocop spec/requests/projects_spec.rb` passed, and `./node_modules/.bin/eslint app/javascript/controllers/project_settings_form_controller.js` passed. `bundle exec rspec spec/requests/projects_spec.rb` could not run in this container because PostgreSQL is unavailable; `ALLOW_DBLESS_SPECS=true` only makes the example pending and then SimpleCov fails the run on the global coverage threshold, so there was no passing local RSpec path for this request spec here.

Committed on the current branch as `441b6c1e` with message `fix(project-settings): return enter-submit input check`.

viamin:

PR #2054, opened 2026-05-15
Fix #2052: Add tooltip to context column cells without a prompt value on recent agent runs table

### LINKED ISSUE #2052: Add tooltip to context column cells without a prompt value on recent agent runs table
## Problem

On the project page, the "Context" column in the recent agent runs table shows truncated text for cells that don't have a prompt value. These cells display labels like "Issue #N" or "PR #N" (or the beginning of the prompt for create-issue agent runs), but there is no tooltip to reveal the full context.

## Expected Behavior

All cells in the context column should have a tooltip on hover so users can see the full context value, even when the cell text is truncated or only shows a short reference like "Issue #N" or "PR #N".

## Current Behavior

Only cells with a prompt value show a tooltip. Cells that display "Issue #N", "PR #N", or a truncated prompt for create-issue runs have no tooltip, making it impossible to see additional context without clicking through.

## Suggested Fix

Add a tooltip to all context column table cells, regardless of whether they have an explicit prompt value. The tooltip should show the full context string (e.g., the full issue/PR title or the complete prompt text).

viamin:
## Summary

Fixes #(issue) so users can see the full context of every recent agent run row on the project page, not just rows that have an explicit prompt. The Context column now always renders a tooltip — falling back to the related issue/PR title when available, or to the visible label like `PR #7` / `Issue #42` otherwise.

## Changes

- **Helper**: Updated `app/helpers/application_helper.rb` so the context-cell helper always returns tooltip text. For agent runs tied to a source PR or a created issue, it resolves the related record's title; when no title is available, it falls back to the visible label (`Issue #N`, `PR #N`).
- **Tests**: Extended `spec/helpers/application_helper_context_display_spec.rb` with coverage for the new fallback paths (source-PR title, created-issue title, and bare-label fallback).

## Design Decisions

- **Prefer the related record's title over the label** — Showing "Fix flaky build on macOS runners" is more useful on hover than repeating "PR #7", so the helper resolves the related issue/PR when possible and only falls back to the label when the lookup misses.
- **Helper-level fix, no view changes** — The view partial already wires the helper's return value into the tooltip attribute, so centralizing the fallback in the helper keeps the change small and keeps tooltip behavior consistent across any future caller.

## Screenshots

_Before/after screenshots of the project page's recent agent runs table, hovering a Context cell that previously had no tooltip (e.g., an `Issue #N` or `PR #N` row), to be attached by the author._

---

Closes #2052


viamin:
## Agent Update

Addressed both review threads in [app/helpers/application_helper.rb](/workspace/app/helpers/application_helper.rb:534). The helper now calls `run.source_pull_request_record` directly via its public API, and PR tooltip lookup returns early after that memoized lookup so it no longer re-issues the same `project.issues.find_by(...)` query on a miss.

Checks:
- `bundle exec rspec` could not run end-to-end in this container because Postgres is unavailable and the suite aborts on DB setup failure.
- `COVERAGE=false ALLOW_DBLESS_SPECS=true bundle exec rspec spec/helpers/application_helper_context_display_spec.rb` passed with `0 failures` and db-less skips/pending as expected.

Committed on the PR branch as `93ec6940` with message: `fix(agent-runs): avoid duplicate PR context tooltip lookup`.

viamin:
## Agent Update

OK

**Review feedback addressed:** The `created_issue_tooltip` helper was issuing a `project.issues.find_by(...)` per row — an N+1 query since created issues weren't preloaded like source PRs were.

**Fix:**
1. **`app/models/agent_run.rb`** — Added `created_issue_record` accessor (with lazy fallback) and `AgentRun.preload_created_issue_records` batch method, mirroring the existing `source_pull_request_record` / `preload_source_pull_requests` pattern.
2. **`app/controllers/projects_controller.rb`** — Added `AgentRun.preload_created_issue_records(@recent_agent_runs)` call alongside the existing source-PR preload.
3. **`app/helpers/application_helper.rb`** — Updated `issue_title_for` to read from the preloaded `created_issue_record` instead of querying `project.issues.find_by`.
4. **`spec/helpers/application_helper_context_display_spec.rb`** — Updated tests to pass preloaded records directly on the stub run (matching the new code path), removed now-unused `stub_project_with_issues` / `stub_issue_scope` helpers.

viamin:
## Agent Update

OK

Done. The review comment has been addressed:

- **Gemfile.lock**: Restored the `bundler (4.0.11)` checksum line that was accidentally removed in a prior commit. The PR now has zero diff against `main` for `Gemfile.lock`.
- **Tests**: All 40 examples pass (0 failures).
- **Lint**: Clean (no offenses).

viamin:
## Agent Update

OK

Here's what I did to address the review feedback:

**Problem**: `created_issue_tooltip(run)` was only preloaded in `ProjectsController#show`, causing N+1 queries at every other call site that renders `agent_run_context_display`.

**Fix**: Added `AgentRun.preload_created_issue_records(...)` to all 7 call sites that already had `preload_source_pull_requests` but were missing the created-issue preload:

1. `app/controllers/agent_runs_controller.rb` — global agent runs index
2. `app/controllers/projects/agent_runs_controller.rb` — project-scoped agent runs index
3. `app/controllers/dashboard_controller.rb` — dashboard active runs
4. `app/models/project.rb` — `broadcast_agent_runs_update` (recent runs broadcast)
5. `app/models/project.rb` — `broadcast_agent_runs_list_update` (full list broadcast)
6. `app/services/dashboard/live_broadcaster.rb` — live dashboard updates
7. `app/services/dashboard/queue_preview.rb` — queue preview

viamin:
## Agent Update

OK

Done. The review comment about reverting the Gemfile.lock change has been addressed — the bundler checksum is restored and the lockfile matches main. Rubocop passes clean, and all 40 specs in the helper spec file pass (the pending ones are expected due to no database).

viamin:
## Agent Update

Committed as `ea989647` with `fix(agent-runs): add fallback context tooltips`.

The CI-facing bug was in the helper, not just the spec: the `run.issue` paths still used `run.issue.title` directly, so rows backed by an issue/PR without a title never got the fallback tooltip. I updated [app/helpers/application_helper.rb](/workspace/app/helpers/application_helper.rb:394) to fall back to the visible label for both create-PR issue rows and analyze/enhance issue rows, and I updated [spec/helpers/application_helper_context_display_spec.rb](/workspace/spec/helpers/application_helper_context_display_spec.rb:5) to be explicitly `:no_db`, replace DB-backed project factories with a stub, and cover the label-fallback cases.

Verified locally with `bundle exec rubocop app/helpers/application_helper.rb spec/helpers/application_helper_context_display_spec.rb` and `ALLOW_DBLESS_SPECS=true COVERAGE=false bundle exec rspec spec/helpers/application_helper_context_display_spec.rb`. I could not run the full `bundle exec rspec` suite in this container because no PostgreSQL service is available.

viamin:

PR #2070, opened 2026-05-16
fix(containers): auto-detect and link service containers when unconfigured (#2067)

### LINKED ISSUE #2067: paid agent runs for PR work still miss project service containers / PostgreSQL
## Problem

Issue #2065 was opened because PR #1950 still had automated Paid agent runs reporting that PostgreSQL was unavailable, so full DB-backed specs could not run.

After double-checking the original PR comments, the quoted Postgres warning on PR #1950 came from a `<!-- paid:agent-update -->` comment. In this codebase, that marker is posted by the existing-PR completion path (`CompleteExistingPrRunActivity`), which means the observed symptom came from a **Paid agent run working on an existing PR branch** rather than from a GitHub Actions-only review workflow.

The fix that closed #2065 (`#2066`) updated `.github/workflows/claude-code-review.yml`, which was therefore the wrong layer for the originally observed failure.

## What the PR #1950 comment actually indicates

The PR #1950 Postgres warning appears to have come from a `create_pr`-style follow-up run on an existing PR, not from a `review`-goal run.

That means the original evidence specifically proves a gap in the **Paid PR agent-run environment** for work on existing PRs.

## Why #2066 did not fix the real issue

`#2066` added a Postgres service to `.github/workflows/claude-code-review.yml` and its workflow-file specs.

That helps the GitHub Actions Claude review job, but the failing PR #1950 comment was generated from a Paid-managed PR agent run environment, not from that standalone GitHub Actions workflow.

## Correct subsystem to investigate

The remaining bug appears to be in the path that queues and executes Paid agent runs for PR work:

- `app/temporal/workflows/git_hub_poll_workflow.rb`
  - queues both `create_pr` and `review` runs for PR-related automation
- `app/temporal/workflows/agent_execution_workflow.rb`
  - both goal types pass through service provisioning before container startup
- `app/temporal/activities/provision_services_activity.rb`
  - provisions shared service containers and returns service env vars
- prompt/service-environment logic that tells the agent whether DB services are available
- `app/temporal/activities/complete_existing_pr_run_activity.rb`
  - confirms the observed PR #1950 warning came from the agent-update path for existing-PR work
- `app/temporal/activities/complete_review_goal_activity.rb`
  - review-goal completion is separate and should also be checked

The likely issue is that one or more Paid PR-run paths are not inheriting or provisioning the project’s configured service containers correctly, or the resulting service environment is not being passed through to the agent container.

### From PR #1950

Multiple Paid agent updates on PR #1950 reported that the environment running the fix did not have PostgreSQL available and therefore could only run DB-less specs.

Those comments used the `paid:agent-update` marker, which maps to the existing-PR update path rather than the code-review marker.

### Current mismatch

- `#2065` described a real missing-Postgres symptom.
- `#2066` fixed `.github/workflows/claude-code-review.yml`.
- But the original failing environment was a Paid PR agent run, so the root cause was not actually addressed.

## Expected behavior

When Paid queues an agent run that works on a PR branch, it should provision the project’s required service containers (for example PostgreSQL) and inject the corresponding environment variables into the agent container so the run can execute the full test suite when the project requires a DB.

That should be true for both:

- `create_pr` / existing-PR follow-up runs
- `review` goal runs

## Impact

- Paid PR agent runs can still silently fall back to DB-less validation.
- PR #1950 remains affected by the originally reported symptom.
- Even if `#2066` improved the standalone GitHub Actions Claude review workflow, the Paid-run container path can still be wrong.

## Suggested investigation

1. Reproduce with the same Paid PR-run path that created the PR #1950 `paid:agent-update` comment.
2. Confirm whether PR follow-up / existing-PR runs have the project’s `service_container_ids` attached.
3. Confirm whether `review` goal runs also inherit the same service-container setup.
4. Confirm `ProvisionServicesActivity` actually runs and returns service env for both goal types.
5. Confirm the agent container receives the generated service environment (`DATABASE_URL`, etc.).
6. Confirm the prompt/service-environment text matches the actual provisioned services and is not incorrectly telling the agent that no DB is available.

## Related

- Closed but mis-targeted predecessor: #2065
- Fix that addressed the wrong layer: #2066
- Still-affected PR: #1950

viamin:
## Summary

- **Root cause**: PR-branch agent runs lacked `DATABASE_URL` because `project.service_containers` was empty — no `ProjectServiceContainer` join record existed, even when the account had PostgreSQL containers configured.
- **Fix 1 — Auto-detect in `ProvisionServicesActivity`**: When a project has no linked service containers, the activity now calls `Projects::DetectServices` to inspect the repo (Gemfile, package.json, docker-compose.yml, config/database.yml), applies any matched containers, and resets the AR association cache before provisioning. Failures are non-fatal: logged as a warning and the run continues without services.
- **Fix 2 — Image-pattern fallback in `DetectServices`**: Matching now falls back to image-pattern matching when no container has a canonical name match. A container named "Screenshot Postgres" with `image: postgres:16` now correctly matches the detected `"postgres"` service.

## Test plan

- [ ] `bin/rspec spec/services/projects/detect_services_spec.rb` — new contexts for image-pattern matching and preference of exact name over image match
- [ ] `bin/rspec spec/temporal/activities/provision_services_activity_spec.rb` — new contexts for auto-link behavior (detects and links, skips when already configured, handles failures gracefully, warns on unmatched)
- [ ] Manually trigger an agent run on a PR branch for a Rails project whose account has a postgres service container — verify `DATABASE_URL` is injected and the agent sees PostgreSQL

Fixes #2067.

### DIFF
--- a/app/services/projects/detect_services.rb
+++ b/app/services/projects/detect_services.rb
@@ -88,6 +88,21 @@ module Projects
     # to Ruby objects.
     MAX_YAML_ALIASES = 100
 
+    # Maps canonical service names to image patterns used as a fallback when
+    # no ServiceContainer exists with an exactly-matching name. Allows containers
+    # named "Screenshot Postgres" or "Dev Redis" to be matched to the
+    # "postgres" / "redis" services detected from the repo.
+    IMAGE_SERVICE_PATTERNS = {
+      "postgres" => /postgres/,
+      "redis" => /redis/,
+      "mysql" => /mysql|mariadb/,
+      "mongodb" => /mongo/,
+      "elasticsearch" => /elasticsearch|opensearch/,
+      "memcached" => /memcache/,
+      "selenium" => /selenium/,
+      "chromium" => /chromium/
+    }.freeze
+
     attr_reader :project
 
     def initialize(project:)
@@ -107,7 +122,7 @@ module Projects
 
       unique_services = detections.uniq { |d| d[:service] }
       detected_names = unique_services.map { |d| d[:service] }
-      containers = ServiceContainer.where(name: detected_names).index_by(&:name)
+      containers = matching_containers(detected_names)
 
       matched = []
       unmatched = []
@@ -129,6 +144,27 @@ module Projects
     end
 
     private
+
+    # Finds ServiceContainer records matching the given canonical service names.
+    # Tries exact name match first (e.g. name == "postgres"), then falls back
+    # to image-pattern matching so containers named "Screenshot Postgres" or
+    # "Dev Redis" are still matched to their canonical service.
+    def matching_containers(detected_names)
+      by_name = ServiceContainer.where(name: detected_names).index_by(&:name)
+      remaining = detected_names.reject { |n| by_name.key?(n) }
+      return by_name if remaining.empty?
+
+      ServiceContainer.all.each do |sc|
+        remaining.each do |service_name|
+          next if by_name.key?(service_name)
+
+          pattern = IMAGE_SERVICE_PATTERNS[service_name]
+          by_name[service_name] = sc if pattern && sc.image.match?(pattern)
+        end
+      end
+
+      by_name
+    end
 
     def fetch_file(path)
       client = project.github_token.client
--- a/app/services/projects/detect_services.rb
+++ b/app/services/projects/detect_services.rb
@@ -99,8 +99,7 @@ module Projects
       "mongodb" => /mongo/,
       "elasticsearch" => /elasticsearch|opensearch/,
       "memcached" => /memcache/,
-      "selenium" => /selenium/,
-      "chromium" => /chromium/
+      "selenium" => /selenium/
     }.freeze
 
     attr_reader :project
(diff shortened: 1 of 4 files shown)


viamin:
## Agent Update

OK.

Done. Applied the reviewer's suggestion to derive `IMAGE_SERVICE_PATTERNS = COMPOSE_IMAGE_PATTERNS.invert.freeze` instead of maintaining a duplicate constant. This ensures any new service added to `COMPOSE_IMAGE_PATTERNS` automatically appears in the fallback matching logic.

Note: `bin/rspec` could not run due to no PostgreSQL being available in this environment.

viamin:
## Agent Update

OK.

Both review threads addressed and committed:

1. **`provision_services_activity.rb`**: Moved `project.service_containers.reset` to fire whenever `result.matched.any?` — not just when new records were created. This fixes the race condition where concurrent calls find already-linked containers via `find_or_create_by!` (so `previously_new_record?` is false), leaving the provisioner with a stale empty cache.

2. **`detect_services.rb`**: Replaced `COMPOSE_IMAGE_PATTERNS.invert` with an explicit `IMAGE_SERVICE_PATTERNS` hash. This avoids the silent collision bug where `Hash#invert` would silently drop one of two patterns mapping to the same service name (e.g., adding `/\bpg\b/` for `"postgres"` would silently lose one entry).

viamin:
## Agent Update

OK.

Committed. The fix adds `.order(:name)` to the `account_scope` query in `matching_containers` at `app/services/projects/detect_services.rb:163`, ensuring deterministic selection when multiple containers match the same image pattern (e.g., "Dev Postgres" and "Prod Postgres" both running `postgres:16`).

Tests could not be run due to no database being available in this environment, but the change is a trivial one-line addition of a stable sort order.

viamin:
## Agent Update

OK.

Committed. The fix adds missing `.order` stubs to all test doubles that call `matching_containers` when there are unmatched services, consolidating them with the existing `.where` stubs using `receive_messages` as required by RuboCop.

viamin:

PR #2079, opened 2026-05-17
fix(agent-runs): pin selected models to compatible runner execution

viamin:
## Summary
- pass the run-selected model into agent-harness execution planning so agent runs pin the chosen model at runtime
- keep command, env, preparation, and preflight on the same run-scoped runtime path
- constrain selector candidate pools to the runner-compatible provider family and fail loudly on incompatible or mismatched runtime selections

## Testing
- `bin/rails db:prepare RAILS_ENV=test`
- `bundle exec rspec spec/services/models/rules_based_selector_spec.rb spec/services/models/meta_agent_selector_spec.rb spec/services/providers/harness_execution_plan_spec.rb spec/temporal/activities/run_agent_activity_spec.rb`

### DIFF
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -1618,9 +1618,8 @@ module Activities
       end
       return env unless provider_entry
 
-      provider_env_entry = provider_entry_for(command_context.provider_candidate, command_context.user)
-      env.merge!(provider_env_entry.direct_outbound_exec_env) if provider_env_entry&.requires_direct_outbound?
-      env.merge!(api_key_command_env(provider_env_entry)) if provider_env_entry&.api_key?
+      env.merge!(provider_entry.direct_outbound_exec_env) if provider_entry.requires_direct_outbound?
+      env.merge!(api_key_command_env(provider_entry)) if provider_entry.api_key?
       env
     end
 
@@ -1760,6 +1759,10 @@ module Activities
         agent_run,
         network: Containers::Provision.network_for(agent_run: agent_run)
       )
+    rescue => e
+      raise e if e.is_a?(ProviderExecutionError)
+
+      raise ProviderExecutionError, "Failed to synchronize marketplace MCP servers: #{e.message}"
     end
 
     def marketplace_has_attachments?(agent_run)
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -120,8 +120,8 @@ module Activities
       agent_run_id = input[:agent_run_id]
       agent_run = AgentRun.find(agent_run_id)
       track_phase(agent_run_id: agent_run_id, phase_key: "run_agent", phase_group: "agent", agent_run: agent_run) do
-        prompt = agent_run.effective_prompt
-        unless prompt
+        base_prompt = base_prompt_for(agent_run)
+        unless base_prompt
           raise Temporalio::Error::ApplicationError.new(
             "No prompt available for agent run", type: "MissingPrompt", non_retryable: true
           )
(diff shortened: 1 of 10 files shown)


app/temporal/activities/run_agent_activity.rb
```diff
@@ -119,12 +119,14 @@ def self.provider_attempt_count(agent_type:, fallback_enabled:, fallback_provide
     def execute(input)
       agent_run_id = input[:agent_run_id]
       agent_run = AgentRun.find(agent_run_id)
-      @agent_run = agent_run
-      @has_marketplace_attachments = nil
-      @marketplace_runtime_env = {}
-      @marketplace_runtime_preparation = {}
       track_phase(agent_run_id: agent_run_id, phase_key: "run_agent", phase_group: "agent", agent_run: agent_run) do
-        base_prompt = base_prompt_for(agent_run)
+        prompt = agent_run.effective_prompt
```
viamin: This changes the prompt from per-attempt to per-run. AgentRun#effective_prompt(provider_key:) is the only place that re-renders provider-specific marketplace attachments, so after a fallback we now keep the original or default attachment text instead of switching to the active provider renderer. Please rebuild the prompt inside the provider loop with the current provider key.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -1577,55 +1566,48 @@ def build_command(command_context, prompt)
...
+    def command_env_for(command_context, prompt, agent_run: nil)
       provider_entry = provider_entry_for(command_context.provider_candidate, command_context.user)
-      if provider_entry&.agent_harness_runtime?
-        return marketplace_runtime_env(provider_key).merge(
-          direct_outbound_execution_plan(provider_entry, prompt).env
-        )
-      end
-      return marketplace_runtime_env(provider_key) unless provider_entry
+      return direct_outbound_execution_plan(provider_entry, prompt, agent_run: agent_run).env if provider_entry&.agent_harness_runtime?
+      return {} unless provider_entry
```
viamin: This drops runtime-config marketplace attachments entirely. MarketplaceEntries::RuntimeAttachments.runtime_env and runtime_preparation are no longer merged anywhere in this activity, so attachment-provided env vars and file writes silently stop reaching the container. Please restore that merge on the execution path before this ships.

app/temporal/activities/run_agent_activity.rb
```diff
@@ -765,9 +748,9 @@ def run_agent_with_provider(agent_run, provider_candidate, prompt, user_settings
         provider: provider,
         user: user_settings.user
       )
-      command = build_command(command_context, prompt)
-      command_env = command_env_for(command_context, prompt)
-      command_preparation = command_preparation_for(command_context, prompt)
+      command = build_command(command_context, prompt, agent_run: agent_run)
```
viamin, line 769: By the time we build the command here, the run is still using whatever MCP snapshot was rendered at creation time. The refactor removed the per-provider rerender and reprovision step, so marketplace MCP attachments that render different server definitions per provider will stay stale across fallback and execute with the wrong servers. Please refresh and reprovision the snapshot for the current provider before building the command.

viamin (commented):
I found three regressions in the marketplace attachment path: provider-specific prompt rendering no longer updates per fallback attempt, runtime-config attachments are no longer applied to the container env/preparation, and provider-specific MCP attachments are no longer rerendered or reprovisioned before execution. I also tried a narrow RSpec validation pass, but the local test environment is currently blocked by PendingMigrationError and a duplicate test database name named agent_run_17819_attempt_0.

viamin:
## Agent Update

Fixed the CI-breaking `RunAgentActivity` regression and committed it as `1a446dfd` (`fix(agent-runs): restore provider-scoped runtime rendering`).

The change restores provider-scoped prompt rendering per attempt, re-merges marketplace runtime env/file preparation onto every execution path, and refreshes/reprovisions marketplace MCP snapshots before building the current provider command. That addresses the unresolved review threads around fallback prompt rendering, dropped runtime attachments, and stale MCP config across provider switches.

Verification passed with `bin/rails db:prepare RAILS_ENV=test`, the PR’s targeted spec set, `bundle exec rubocop`, and a full `bundle exec rspec` run. The full suite finished with `12866 examples, 0 failures, 2 pending`; the two pending specs are the existing Qdrant live checks that report Qdrant unavailable at `http://localhost:6333`.

viamin:

PR #2082, opened 2026-05-17
fix(agent-image): improve docker layer caching

viamin:
## Summary
- move the expensive Ruby compile ahead of Node and PostgreSQL client setup so those pin changes do not force a Ruby rebuild
- keep non-root user creation as a base image concern instead of coupling it to a provider-specific seed step
- move the OpenCode pre-seed immediately after the OpenCode install so unrelated provider changes do not invalidate it

## Testing
- Not run (not requested)


viamin:

PR #2096, opened 2026-05-17
fix(agent-runs): sanitize Claude settings.json in run containers to avoid leaking interactive model defaults

### LINKED ISSUE #2078: fix(agent-runs): sanitize Claude settings.json in run containers to avoid leaking interactive model defaults
## Problem

Paid currently copies both `.credentials.json` and `settings.json` into the run container's `~/.claude` directory for Claude subscription auth.

We have now tightened execution so agent runs can pass an explicit run-scoped model override from `model_selection`, but `settings.json` is still broader than what agent runs should inherit from an interactive Claude CLI environment.

That means user-specific interactive defaults and unrelated Claude settings can still leak into agent-run behavior when they should be isolated to auth plus the run's own execution settings.

## Why this matters

- Agent runs should be reproducible from Paid configuration, not from a developer's last interactive CLI choice.
- Even with explicit model pinning, broader `settings.json` state is still an unexpected input surface.
- The container should only inherit what is necessary for subscription auth.

## Proposed direction

1. Stop copying `settings.json` verbatim into agent run containers, or
2. Replace it with a sanitized/minimal generated version for runs.

At minimum, remove any model-defaulting behavior from the copied config.

## Acceptance criteria

- Claude agent runs do not inherit interactive model defaults from host/devcontainer `settings.json`.
- Subscription auth still works for Claude runs.
- There is test coverage proving a host-side Claude config default does not alter the model Paid executes for a run.

## Context

This is follow-up work to the execution-side fix that pins agent-run models from `model_selection` and raises loudly when a fixed-runtime provider disagrees with the selected model.

viamin:
## Summary

Agent runs were inheriting `settings.json` from the host/devcontainer Claude CLI environment, which could leak interactive model defaults and other unrelated settings into reproducible runs. This change stops copying `settings.json` into run containers so only subscription auth (`.credentials.json`) is seeded, ensuring run behavior is driven solely by Paid's configuration and the explicit model selection from `model_selection`.

## Changes

- **Services**: Removed `settings.json` from `source_files` in `seed_claude_credentials!` (`app/services/containers/provision.rb:1001`); only `.credentials.json` is now copied into agent run containers for Claude subscription auth.
- **Tests**: Added coverage (`spec/services/containers/provision_spec.rb:979`) asserting that a host-side `settings.json` with model defaults is not propagated into the container.

## Design Decisions

- **Drop, don't sanitize** — rather than generating a minimal `settings.json`, we omit it entirely. Subscription auth only requires `.credentials.json`, so a sanitized file would add surface area without functional benefit. If run-scoped Claude settings are ever needed, they should be generated explicitly from Paid configuration rather than inherited from the host.
- **Pairs with the prior execution-side fix** — explicit model pinning from `model_selection` already guards the model dimension; this change closes the remaining input-surface gap so the container only inherits what's strictly necessary.

---

Closes #2078

### DIFF
--- a/app/services/containers/provision.rb
+++ b/app/services/containers/provision.rb
@@ -994,11 +994,12 @@ module Containers
       # Container was already removed between running? check and stop
     end
 
-    # Copies credentials and settings from the read-only host mount into the
-    # writable ~/.claude tmpfs. Only the files Claude CLI needs for auth and
-    # configuration are copied; session/project data is created fresh each run.
+    # Copies credentials from the read-only host mount into the writable
+    # ~/.claude tmpfs. Only `.credentials.json` is needed for subscription auth;
+    # `settings.json` is intentionally excluded to prevent interactive model
+    # defaults from leaking into agent runs.
     def seed_claude_credentials!
-      source_files = %w[.credentials.json settings.json]
+      source_files = %w[.credentials.json]
       return unless claude_subscription_auth?
 
       # Prefer the source that actually contains the required credential file
--- a/spec/services/containers/provision_spec.rb
+++ b/spec/services/containers/provision_spec.rb
@@ -975,6 +975,22 @@ RSpec.describe Containers::Provision do
 
         service.provision
       end
+
+      it "does not copy settings.json into the container even when present on host" do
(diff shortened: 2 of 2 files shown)


viamin:

PR #2101, opened 2026-05-18
fix(providers): cover configured minimax smoke scenarios

viamin:
## Summary
- fix OpenCode MiniMax runtime qualification so Test Agent uses a working `minimax/...` model identifier
- make `bin/provider-smoke`'s `current-enabled` preset reflect configured direct-outbound scenarios instead of a stale hard-coded list
- prefer configured provider models over canned defaults in smoke helpers so local smoke coverage matches what users actually configured

## Verification
- `bundle exec rspec spec/lib/provider_smoke_helpers_spec.rb spec/models/provider_spec.rb`
- `Providers::TestAgent.call(provider: Provider.find(26))` returned healthy for the local `Opencode MiniMax` provider

## Notes
- focused RSpec re-run inside the git worktree was blocked by the branch-scoped test DB not existing yet; the same test set passed from the main workspace before packaging this PR

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -40,7 +40,7 @@ class Provider < ApplicationRecord
                    opencode_model_provider: "mistral" },
     "minimax" => { label: "MiniMax", base_url: "https://api.minimax.io/anthropic", service_type: "minimax",
                    env_var: "ANTHROPIC_API_KEY", opencode_npm: "@ai-sdk/anthropic", kilocode_provider_id: "anthropic",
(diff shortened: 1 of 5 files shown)


viamin:
## Agent Update

OK

Commit succeeded. The fix addresses the reviewer's feedback: `return []` inside `@configured_provider_payload ||= begin...end` was exiting the method without assigning the ivar, so every subsequent call would re-shell-out to `rails runner`. Replaced with an `if/else` so the empty array is properly memoized on failure.

viamin:

PR #2106, opened 2026-05-18
chore(provider-smoke): use rspec documentation format for live output

viamin:
## Summary
- Switches `bin/provider-smoke` to RSpec's documentation formatter so each scenario's name and pass/fail/pending status is visible in real time instead of just dots.
- Bypasses `.rspec` (`-O /dev/null`) and re-specifies only the options the smoke run needs; otherwise the project-wide `--format progress` would stack with `--format documentation` rather than be replaced.

## Test plan
- [x] `bin/provider-smoke` prints per-scenario lines (`FAILED` / `PENDING` / passing) as they run.
- [ ] CI green.


viamin:

PR #2107, opened 2026-05-18
feat: OpenCode MiniMax Token Plan configuration in Paid does not match upstream OpenCode/MiniMax integration

### LINKED ISSUE #2105: OpenCode MiniMax Token Plan configuration in Paid does not match upstream OpenCode/MiniMax integration
## Problem

Paid's current OpenCode + MiniMax API-key path does not match MiniMax's documented OpenCode integration, and existing OpenCode MiniMax provider entries can fail with model resolution errors such as:

- `Model not found: anthropic/MiniMax-M2.7`
- `ProviderModelNotFoundError`

## Why this is happening

Paid currently treats OpenCode MiniMax as an Anthropic-shaped direct-outbound adapter:

- `base_url: https://api.minimax.io/anthropic`
- `env_var: ANTHROPIC_API_KEY`
- `opencode_model_provider: "anthropic"`
- model qualification becomes `anthropic/<model>`

Relevant code:

- `app/models/provider.rb` in `DIRECT_OUTBOUND_API_PROVIDERS["minimax"]`
- `Provider#opencode_qualified_model`
- `Provider#opencode_provider_runtime`

This means a configured model like `MiniMax-M2.7` is currently sent to OpenCode as `anthropic/MiniMax-M2.7`.

## Upstream docs say something different

MiniMax's current docs for OpenCode say the supported Token Plan flow is:

1. `opencode auth login`
2. choose provider `MiniMax`
3. enter the MiniMax Token Plan API key

They also document a manual OpenCode provider config using provider `minimax`, `npm: "@ai-sdk/anthropic"`, base URL `https://api.minimax.io/anthropic/v1`, and model `MiniMax-M2.7`.

Docs:

- https://platform.minimax.io/docs/token-plan/opencode
- https://platform.minimax.io/docs/token-plan/other-tools

Notably, this is not the same as forcing the model ID to `anthropic/MiniMax-M2.7` inside Paid.

## Important non-fix note

PR #2102 adds OpenCode subscription-auth support in Paid, but it does **not** fix the existing OpenCode MiniMax API-key path. That PR intentionally avoids treating API-key-only OpenCode auth state as subscription auth, so the MiniMax Token Plan mismatch remains a separate problem.

## Expected outcome

Paid's OpenCode MiniMax integration should match the upstream OpenCode/MiniMax contract closely enough that a MiniMax Token Plan key works in Paid the same way it works in native OpenCode.

## Likely implementation direction

Investigate and update Paid's OpenCode MiniMax path so that it aligns with MiniMax's official OpenCode integration. In practice this probably means one or more of:

- stop qualifying MiniMax OpenCode models as `anthropic/<model>`
- use the correct upstream provider identifier expected by current OpenCode for MiniMax
- verify whether the base URL should be `https://api.minimax.io/anthropic/v1` instead of the current value
- confirm whether `ANTHROPIC_API_KEY` vs `ANTHROPIC_AUTH_TOKEN` matters in the current OpenCode/SDK path
- add regression coverage for `MiniMax-M2.7` and `MiniMax-M2.7-highspeed`

## Acceptance criteria

- an OpenCode provider in Paid configured for MiniMax Token Plan can successfully use `MiniMax-M2.7`
- no `anthropic/MiniMax-M2.7` model-not-found failure on preflight/runtime
- specs cover the expected OpenCode MiniMax runtime contract
- docs/UI guidance in Paid matches the actually supported setup

viamin:
## Summary

Aligns Paid's OpenCode MiniMax integration with MiniMax's documented Token Plan contract, so a MiniMax Token Plan API key works in Paid the same way it works in native OpenCode. Previously, models were qualified as `anthropic/MiniMax-M2.7` and the base URL was incomplete, causing `ProviderModelNotFoundError` failures at preflight/runtime.

## Changes

- **Provider configuration (`app/models/provider.rb`)**:
  - Updated `DIRECT_OUTBOUND_API_PROVIDERS["minimax"]` base URL to `https://api.minimax.io/anthropic/v1` to match upstream MiniMax docs.
  - Dropped the `anthropic` model-provider qualifier so OpenCode receives bare model IDs (`MiniMax-M2.7`) rather than `anthropic/MiniMax-M2.7`.
  - Reworked `opencode_provider_runtime` to surface the base URL via provider config metadata (`baseURL`) for `@ai-sdk/anthropic`-backed providers, since `OPENAI_BASE_URL` only applies to the OpenAI-compatible SDK.

- **Specs**:
  - `spec/models/provider_spec.rb`: Updated MiniMax runtime expectations to assert bare model IDs, absence of `OPENAI_BASE_URL`, and presence of `baseURL` in provider config. Added regression coverage for `MiniMax-M2.7-highspeed`.
  - `spec/support/provider_smoke_helpers.rb`: Bumped default smoke models from `MiniMax-M2.5` to `MiniMax-M2.7` and refreshed labels to reference "Token Plan".
  - `spec/lib/provider_smoke_helpers_spec.rb`: Synced expectations with the new default model.

## Design Decisions

- **Bare model IDs over `anthropic/` qualification**: MiniMax's upstream OpenCode integration uses provider `minimax` with `npm: "@ai-sdk/anthropic"`, but the model identifier passed to the SDK is bare (`MiniMax-M2.7`). Forcing an `anthropic/` (or `minimax/`) prefix in Paid produced model-not-found errors; removing the qualifier matches the documented contract.
- **`baseURL` via provider config, not `OPENAI_BASE_URL`**: Because MiniMax uses the Anthropic-shaped SDK, the OpenAI env-var override doesn't apply. Passing `baseURL` through OpenCode's provider config metadata is the correct hook for `@ai-sdk/anthropic` and keeps Paid's direct-outbound shape consistent with how OpenCode itself wires this.
- **No change to `ANTHROPIC_API_KEY` env var**: The current OpenCode/SDK path accepts `ANTHROPIC_API_KEY` for MiniMax Token Plan keys, so no env-var rename was required.

---

Closes #2105

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -38,9 +38,8 @@ class Provider < ApplicationRecord
                     opencode_model_provider: "deepseek" },
     "mistral" => { label: "Mistral", base_url: "https://api.mistral.ai/v1", service_type: "mistral",
                    opencode_model_provider: "mistral" },
-    "minimax" => { label: "MiniMax", base_url: "https://api.minimax.io/anthropic", service_type: "minimax",
-                   env_var: "ANTHROPIC_API_KEY", opencode_npm: "@ai-sdk/anthropic", kilocode_provider_id: "anthropic",
-                   opencode_model_provider: "minimax" },
+    "minimax" => { label: "MiniMax", base_url: "https://api.minimax.io/anthropic/v1", service_type: "minimax",
+                   env_var: "ANTHROPIC_API_KEY", opencode_npm: "@ai-sdk/anthropic", kilocode_provider_id: "anthropic" },
     "xai" => { label: "xAI", base_url: "https://api.x.ai/v1", service_type: "xai",
                opencode_model_provider: "xai" },
     "zai" => { label: "z.ai", base_url: "https://api.z.ai/api/paas/v4", service_type: "zai",
@@ -955,7 +954,15 @@ class Provider < ApplicationRecord
 
     env_var = direct_outbound_api_key_env_var(opencode_api_provider)
     env = { env_var => provider_api_key&.api_key.to_s }
-    env["OPENAI_BASE_URL"] = api_config[:base_url] if api_config[:base_url]
+
+    # Providers using @ai-sdk/anthropic receive their base URL through the
+    # provider config (OPENAI_BASE_URL is only read by the OpenAI-compatible SDK).
+    provider_config = {}
+    if api_config[:opencode_npm] == "@ai-sdk/anthropic" && api_config[:base_url]
+      provider_config["baseURL"] = api_config[:base_url]
+    elsif api_config[:base_url]
+      env["OPENAI_BASE_URL"] = api_config[:base_url]
+    end
 
     AgentHarness::ProviderRuntime.new(
       model: model_id,
@@ -963,7 +970,7 @@ class Provider < ApplicationRecord
       unset_env: %w[OPENAI_HEADER_X_AGENT_RUN_ID OPENAI_HEADER_X_PROXY_TOKEN],
       metadata: {
         config: {
-          "provider" => { opencode_api_provider => {} }
+          "provider" => { opencode_api_provider => provider_config }
         }
       }
     )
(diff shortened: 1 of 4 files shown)


viamin:

PR #2114, opened 2026-05-18
fix(runners): constrain pi minimax model selection

viamin:
## Summary
Fix `pi` + MiniMax runs selecting incompatible global LLM models like `claude-sonnet-4-6` before execution.

This change makes `pi` with a configured upstream model behave like a fixed-model runner during model selection, so Paid stops choosing incompatible models and then failing later in `RunAgentActivity`.

## What Changed
- treat configured `pi` API-key runners as constrained direct-outbound/fixed-model runners for selection purposes
- filter selector candidates using the runner's upstream provider family, including MiniMax for `pi`
- reject incompatible `pi` selections before runtime instead of tripping the later runtime mismatch guard
- keep explicit override behavior intact for normal non-`pi` runners
- add regression coverage across runner validation, selector behavior, and run-agent execution paths

## Verification
```bash
bundle exec rspec spec/services/models/runner_tier_lookup_spec.rb spec/services/runners/default_tier_model_ids_spec.rb spec/services/models/meta_agent_selector_spec.rb spec/services/models/rules_based_selector_spec.rb spec/models/runner_spec.rb

bundle exec rspec spec/services/models/select_spec.rb spec/services/models/meta_agent_selector_spec.rb spec/services/models/rules_based_selector_spec.rb spec/temporal/activities/run_agent_activity_spec.rb spec/models/runner_spec.rb spec/services/models/runner_tier_lookup_spec.rb spec/services/runners/default_tier_model_ids_spec.rb
```

Both passed locally with `0 failures`.

## Context
This addresses the current Paid failure mode where `pi` MiniMax runs were being created correctly as `pi` runs, but the generic model-selection layer persisted `claude-sonnet-4-6` as the selected model. The runtime then rejected the run with:

`Selected model claude-sonnet-4-6 does not match configured runtime model MiniMax-M2.7 for Pi MiniMax`


viamin:
## Agent Update

Addressed the review thread in [app/services/models/select.rb](/workspace/app/services/models/select.rb:30) by preserving the rejected `selected` payload when `compatible_selection?` rejects a model. That keeps the decision log and orchestration record distinguishable from a genuine “no candidate found” case. I also tightened the regression in [spec/services/models/select_spec.rb](/workspace/spec/services/models/select_spec.rb:406) to assert the rejected Pi override model is present in the logged selection metadata.

Verification passed with `bundle exec rubocop` and full `bundle exec rspec` (`13418 examples, 0 failures, 2 pending`). `bin/rails db:prepare` also passed once Rails was given the repo’s expected `DB_HOST`/`DB_USERNAME`/`DB_PASSWORD` env vars. Changes are committed on the current branch as `19f90890` with message `fix(runners): log rejected pi model selections`.

viamin:
## Agent Update

OK

Commit succeeded. The fix applies `compatible_model_scope` to the quality-escalation fallback scope in `Models::Select#quality_escalation_result` (line 106), ensuring provider-constrained Pi runs select a compatible model during quality-recovery escalation instead of picking an incompatible global model and then failing the post-selection guard.

Two regression tests were added confirming that quality escalation for a Pi MiniMax run selects a MiniMax-provider model and never picks an incompatible Anthropic model.

viamin:

PR #2128, opened 2026-05-18
fix(agent-runs): report cumulative startup idle timeout

### LINKED ISSUE #2122: fix(agent-runs): idle_timeout error message understates actual wait by ~2x
## Summary

Agent runs that fail with `idle_timeout: No output received for 360 seconds` actually run for **~720s** (median 727s across 69 timeouts in the last 24h, min 589s, max 3267s) before being killed. The reported "360 seconds" is half the real elapsed wait.

This is misleading in two ways:
1. Operators trying to tune the idle timeout will set it lower expecting linear effect — they'll get 2x.
2. Failure triage assumes the reported number is the actual wait.

## Evidence

Sample runs (paid_development, last 24h):

| AgentRun | Started | Completed | Actual dur | Error |
|---|---|---|---|---|
| #18147 | 13:05:16 | 13:17:05 | **709s** | "No output received for 360 seconds" |
| #18148 | 13:05:55 | 13:17:43 | **707s** | "No output received for 360 seconds" |
| #18152 | 13:21:12 | 13:33:01 | **708s** | "No output received for 360 seconds" |
| #18153 | 13:33:51 | 13:45:37 | **706s** | "No output received for 360 seconds" |
| #17910 | — | — | **725s** | "No output received for 360 seconds" |

Total in last 24h: **69 idle_timeout failures**, median **727s** vs. reported 360s.

## Root cause

The total wait is `startup_timeout + idle_timeout`, but the error reports only the second half:

1. `DEFAULT_CREATE_PR_IDLE_TIMEOUT = 360` (`app/temporal/activities/run_agent_activity.rb:86`).
2. `effective_startup_timeout` (`run_agent_activity.rb:703-710`) calls `heartbeat.idle_timeout_for(360)`.
3. `HeartbeatSetup#idle_timeout_for` (`app/services/containers/heartbeat_setup.rb:81-86`) returns `nil` when `available?` is false — heartbeat is unavailable for non-reliable providers.
4. With `nil` returned, the calculation falls through to `effective_idle_timeout` = 360s, so startup is gated on its own 360s deadline.
5. Startup heartbeat thread exits after 360s without producing output (`app/services/containers/provision.rb:2809`).
6. The idle-timeout counter then starts and waits *another* 360s of silence.
7. Total: **360 + 360 = 720s** before the error fires.
8. The error string (`provision.rb:2485`) interpolates only the idle timeout value (360), not the cumulative wait.

## Suggested fix

Two parts, both small:

1. **Correct the error message** to reflect the actual elapsed wait or the cumulative budget (`startup_timeout + idle_timeout`). At minimum, distinguish "no startup output (waited Xs for heartbeat)" from "idle after producing output."

2. **Consider whether the startup wait should count toward idle.** If the agent never produces any output, having two separate 360s windows is probably not the intended UX. Options:
   - Combine into a single startup-grace + idle budget.
   - Add explicit startup-grace status (`startup_timeout`) so it's distinguishable from mid-run idle.

This intersects with the recent startup-heartbeat hardening (commits `91d9d3302`, `2b5b8cdd3`, refs #2071/#2074); make sure any change here doesn't regress those.

## Acceptance

- The error message for a startup-only timeout no longer claims "360 seconds" when the actual wait was ~720s.
- Operators can tune the idle threshold and see linear effect on time-to-kill.

ddfc1991:
## Summary

Fixes #2122 by making startup-only idle timeout failures report the cumulative startup grace + idle budget instead of only the second idle window.

When the startup heartbeat grace expires before any output is produced, the process waits through both `startup_timeout` and `idle_timeout`. The previous message still said `No output received for 360 seconds`, which understated the real ~720s wait described in the issue.

## Changes

- Tracks whether the startup heartbeat thread is still active in `TimeoutCheckState`.
- Uses a distinct idle timeout message when no output was ever produced and startup grace has already expired.
- Keeps the existing simple idle timeout message for normal mid-run idle after output was produced.
- Adds specs for the cumulative startup+idle message and the existing post-output idle behavior.

## Verification

- `ruby -c app/services/containers/provision.rb` — pass
- `ruby -c spec/services/containers/provision_spec.rb` — pass
- `git diff --check` — pass
- Targeted extracted Ruby check for `idle_timeout_message` — pass (`targeted_message_logic_ok`)

Full `bin/rspec` could not run in this Termux environment because `bundle install` fails building native gems (`pg_query` and `temporalio`) on aarch64 Android.

### DIFF
--- a/app/services/containers/provision.rb
+++ b/app/services/containers/provision.rb
@@ -118,6 +118,7 @@ module Containers
     TimeoutCheckState = Struct.new(
       :mutex, :timeout_reason_ref, :startup_timeout, :idle_timeout,
       :timeout, :started_at, :heartbeat_path, :output_received_ref, :last_activity_ref,
+      :startup_heartbeat_active_ref,
       keyword_init: true
     )
 
@@ -404,7 +405,8 @@ module Containers
         started_at: started_at,
         heartbeat_path: heartbeat_path,
         output_received_ref: -> { output_received },
-        last_activity_ref: -> { last_activity_at }
+        last_activity_ref: -> { last_activity_at },
+        startup_heartbeat_active_ref: -> { startup_heartbeat&.alive? }
       )
 
       watchdog_ctx = WatchdogContext.new(
@@ -2481,9 +2483,10 @@ module Containers
           diagnostics: timeout_diagnostics_for_state(timeout_check, output_received: false)
         )
       when :idle
+        diagnostics = timeout_diagnostics_for_state(timeout_check, output_received: true)
         raise IdleTimeoutError.new(
-          "No output received for #{timeout_check.idle_timeout} seconds",
-          diagnostics: timeout_diagnostics_for_state(timeout_check, output_received: true)
+          idle_timeout_message(timeout_check, diagnostics),
+          diagnostics: diagnostics
         )
       when :wall_clock
         elapsed_seconds = ((Process.clock_gettime(Process::CLOCK_MONOTONIC) - timeout_check.started_at)).round(1)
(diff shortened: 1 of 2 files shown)


viamin:
**Escalation Note**

This has been escalated because No meaningful progress for 60 minutes after 3 consecutive provider/infrastructure failures.

**How to resolve:**

viamin:
Closing as duplicate of https://github.com/viamin/paid/pull/2129

Issue #2071, opened 2026-05-16
P1: Fix Claude heartbeat visibility during silent MCP startup to avoid false startup_timeouts


viamin:
## Summary
Claude create-PR runs are being falsely classified as `startup_timeout` during silent startup, even when heartbeat support is configured. Recent failures show the watchdog is not observing a readable heartbeat, so Paid kills the run after ~360s even though Claude appears to still be alive and may be starting repo-local MCP tooling.

## Why this is P1
When Claude is the only non-rate-limited provider available, this failure mode effectively blocks automation and creates repeated failed runs.

## Evidence
Recent runs on `viamin/paid` showed a consistent signature:
- runs `17210`, `17211`, `17215`, `17216`, `17219`, `17220`, `17223`, `17224`, `17225`, `17228`, `17229`, `17233`, `17237`, `17238`, `17241`, `17242`, and `17243`
- `status=timeout`
- `error_message=startup_timeout: No output received within 360 seconds`
- diagnostics consistently included:
  - `output_received: false`
  - `heartbeat_supported: true`
  - `heartbeat_path_configured: true`
  - elapsed around `370.5-371.1s`
- for live repro run `17243`, the system logs showed:
  - heartbeat dir prepared: `/tmp/paid-heartbeat-174fe6c07f920816`
  - Claude preflight succeeded
  - real Claude command started at `2026-05-16 19:09:35 UTC`
  - timeout logged at `2026-05-16 19:15:46 UTC`
  - timeout metadata had `stdout_bytes: 0`, `stderr_bytes: 0`, `heartbeat_active: null`, `heartbeat_age_seconds: null`

I also observed that the repo-local `.mcp.json` caused Claude to start Playwright MCP subprocesses even though Paid had not provisioned MCP servers for the run. That strongly suggests a long silent startup/tool-init phase before first stdout.

## Short-term mitigation already shipped
Paid now reclassifies the narrow Claude signature below as an execution failure instead of a normal timeout:
- startup timeout
- provider resolves to Claude
- zero stdout/stderr bytes
- `heartbeat_supported=true`
- `heartbeat_path_configured=true`
- `heartbeat_active=nil`
- `heartbeat_age_seconds=nil`

That avoids treating this as a standard timeout and prevents timeout-specific retry behavior from churning.

## Longer-term fix requested
Implement the real heartbeat/startup fix so live Claude work is not misclassified:
1. Make heartbeat observation robust when the host heartbeat path is unreadable.
[...]

PR #2129, opened 2026-05-18
fix(agent-runs): idle_timeout error message understates actual wait by ~2x

### LINKED ISSUE #2122: fix(agent-runs): idle_timeout error message understates actual wait by ~2x

viamin:
## Summary

Fixes a watchdog bug where agent runs that never produced output were killed after ~720s but the error reported only "360 seconds" — half the actual wait. Operators tuning `idle_timeout` saw a 2x discrepancy, and triage signals were misleading.

## Changes

- **Services (`app/services/containers/provision.rb`)**:
  - Added a `startup_heartbeat_expired` flag set by the startup heartbeat thread when it hits its deadline.
  - Watchdog now skips the heartbeat freshness override when startup heartbeat has expired and no real output was received, allowing the `:startup` timeout to fire immediately instead of cascading into a second `:idle` window.
  - Clarified the idle timeout error message to "No output received for 360 seconds after last activity" so it accurately scopes the reported window.

## Design Decisions

- **Separate startup vs. idle timeouts rather than merging budgets.** The issue suggested combining startup-grace and idle into one budget. Instead, we kept the two distinct windows but fixed the bug that made startup expiration silently fall through to idle. This preserves the existing semantics for non-reliable providers (where `HeartbeatSetup#idle_timeout_for` returns `nil`) and keeps the recent startup-heartbeat hardening from #2071/#2074 intact.
- **Root cause was a stale-heartbeat check, not the timeout values.** The watchdog's `heartbeat_age <= total_elapsed` check was vacuously true for the expired startup heartbeat file, masking the real failure. Fixing the freshness check is more surgical than reshuffling timeout configuration.
- **Error message scoped to idle window.** Rather than reporting a cumulative `startup + idle` figure, we now ensure the `:startup` timeout fires at ~360s with its own message, so the idle error only appears when the run actually went idle after producing output.

---

Closes #2122


viamin:

PR #2130, opened 2026-05-18
fix(agent-runs): pre-filter runner order by selected-model compatibility

### LINKED ISSUE #2121: fix(agent-runs): pre-filter runner order by selected-model compatibility
## Summary

`RunAgentActivity#build_runner_order` (`app/temporal/activities/run_agent_activity.rb:719-753`) only filters runners by `container_executable?`. Model-runner compatibility is validated *inside* preflight (`:647` `validate_selected_model_matches_runtime!` and `:662` `validate_selected_model_runner_compatibility!`), so the iterator tries obviously-incompatible runners and reports them as failed attempts.

Result: when the selected model has only one compatible runner and that runner is rate-limited or circuit-open, the user sees "All runners exhausted: \<long list\>" instead of the real cause.

## Evidence

Sampled from `paid_development` over the last 24h: **66 agent-run failures** with `started_at IS NULL` and error `"All runners exhausted: …"`. Representative case — AgentRun **#18156** (issue #2885, 2026-05-18 13:52:35Z), selected model `glm-5.1` (provider `zai_coding`). The recorded `runners_attempted` is:

| Runner | Result | Why |
|---|---|---|
| `runner:19` (Kilocode GLM 5.1) | rate_limited | Skipped due to cached rate limit until 14:04:40Z |
| `runner:21` (Opencode Kimi K2) | error in 0.2s | "does not match configured runtime model moonshotai/kimi-k2" |
| `claude` | error in 0.1s | "is not compatible with claude (expects anthropic)" |
| `runner:24` (Opencode Deepseek) | error in 0.2s | "does not match configured runtime model deepseek-v4-pro" |
| `runner:27` (Pi MiniMax) | error in 0.1s | "does not match configured runtime model MiniMax-M2.7" |
| `codex` | error in 0.1s | "is not compatible with codex (expects openai)" |

Only `runner:19` could ever have served `glm-5.1`. The other 5 were structurally guaranteed to fail.

## Root cause

`build_runner_order` (`run_agent_activity.rb:719`) returns every container-executable runner. The compatibility-check methods exist but only fire from `selected_runner_runtime()` (`:610-628`) during preflight execution. By then the iterator has already committed to attempting the runner and the failure is recorded in `runners_attempted`.

The provider-key mapping at `app/services/runners/default_tier_model_ids.rb:5-11` (`RUNNER_KEY_TO_MODEL_PROVIDER`) already has all the info needed to filter upfront, plus runners' fixed `direct_outbound_model_id` / configured runtime model.

## Suggested fix

In `build_runner_order`, after the existing `container_executable?` filter, add a compatibility filter against `agent_run.model_selection.llm_model`:

1. If `selected_model.provider` doesn't match `RUNNER_KEY_TO_MODEL_PROVIDER[runner_key]` (when present), drop the runner.
2. If the runner has a fixed runtime model that doesn't equal `selected_model.model_id`, drop the runner.

When the filtered list is empty, classify the failure distinctly (e.g., `no_compatible_runner_available`) so the operator can see "only $RUNNER is compatible with $MODEL and it's currently $REASON" rather than "all runners exhausted." The structurally-incompatible runners should not appear in `runners_attempted` at all.

## Acceptance

- New `Issues::ParseDependencies`-style spec confirming a `glm-5.1` selection yields only `runner:19` (or Kilocode-routing runners) in the attempt order.
- An exhausted run where the only compatible runner is rate-limited emits a distinct, actionable error message rather than the misleading "all runners exhausted" list.
- 24h failure-status counts for `failed` + `started_at IS NULL` measurably drop.

viamin:
## Summary

When the user-selected model has only one compatible runner and that runner is rate-limited or circuit-open, the iterator currently attempts every container-executable runner and surfaces a misleading "All runners exhausted: …" error listing structurally incompatible runners. This change pre-filters the runner order by model compatibility so the operator sees the real cause (e.g. the one compatible runner is rate-limited) instead of noise from runners that could never have served the request.

## Changes

- **`app/temporal/activities/run_agent_activity.rb`**
  - `build_runner_order` gains a model-compatibility pre-filter applied after the existing `container_executable?` check, using the agent run's selected LLM model.
  - New private `runner_compatible_with_model?` enforces two rules: provider-family match via `RUNNER_KEY_TO_MODEL_PROVIDER` (e.g. `claude→anthropic`, `codex→openai`) and, for direct-outbound runners (kilocode/opencode/aider/pi), an exact match against the runner's fixed runtime model.
  - When the only compatible runner is rate-limited or circuit-open, the error message is now `"No compatible runner available: <runner> is the only runner compatible with <model> and it is currently rate limited"` rather than the generic exhausted-list.
- **`spec/temporal/activities/run_agent_activity_spec.rb`**
  - Provider-incompatible runners are excluded (e.g. `codex` dropped for a `zai_coding` model).
  - Fixed-model mismatches are excluded (e.g. a `deepseek-v4-pro` runner dropped when `kimi-k2` is selected).
  - All runners preserved when no model is selected.
  - All runners preserved when the filter would eliminate every candidate (safety-net behavior).

## Design Decisions

- **Safety-net fallback**: if the compatibility filter would empty the candidate list, the unfiltered list is preserved rather than failing the run outright. This keeps the change strictly additive for edge cases where the compatibility metadata is incomplete or stale, at the cost of occasionally still surfacing the old generic error in those cases.
- **Reused existing mapping**: provider matching uses `Runners::DefaultTierModelIds::RUNNER_KEY_TO_MODEL_PROVIDER` rather than introducing a parallel lookup, keeping a single source of truth for runner-key → provider.
- **Pre-filter, don't relocate**: the existing in-preflight checks (`validate_selected_model_matches_runtime!`, `validate_selected_model_runner_compatibility!`) are left intact as a defense-in-depth guard; this change only prevents structurally-incompatible runners from being attempted and recorded in `runners_attempted`.

---

Closes #2121

### DIFF
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -484,7 +484,7 @@ module Activities
           selected_model = agent_run.model_selection&.llm_model
           if selected_model && runners.size == 1
             label = runner_attempt_label(runners.first, agent_run, user_settings.user)
-            reason = all_skipped_rate_limited ? "rate limited" : "unavailable"
+            reason = "rate limited"
             error_message = "No compatible runner available: #{label} is the only runner compatible with #{selected_model.model_id} and it is currently #{reason}"
             error_type = "NoCompatibleRunnerAvailable"
           end
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -453,27 +453,6 @@ module Activities
         agent_run.reload
         return paused_result(agent_run_id) if agent_run.paused?
 
-        # All runners exhausted. Timeout takes precedence over rate_limited
-        # because it indicates an actual execution attempt that should trigger
-        # ProcessRunQueueJob to re-schedule work.
-        if timeout_error.present?
-          timed_out = !agent_run.finished? && agent_run.timeout!(error: timeout_error)
-          Notifications::Rules::ZeroIterationTimeout.call(scope: agent_run) if timed_out
-          # Skip queue processing when cleanup killed the run — the timeout
-          # was not a real runner issue, so there is nothing to re-schedule.
-          # (agent_run was reloaded above, so the model method sees current state)
-          ProcessRunQueueJob.perform_later if timed_out && !agent_run.cancelled_by_cleanup?
-        elsif !agent_run.finished? && (last_error == "rate_limited" || all_skipped_rate_limited)
-          runner_list = runners.any? ? runner_attempt_labels(runners, agent_run, user_settings.user).join(", ") : "none"
-          agent_run.rate_limit!(
-            error: "All runners rate limited: #{runner_list}",
-            reset_at: rate_limit_reset_at
-          )
-        elsif !agent_run.finished?
-          runner_list = runners.any? ? runner_attempt_labels(runners, agent_run, user_settings.user).join(", ") : "none"
-          agent_run.fail!(error: "All runners exhausted: #{runner_list}")
-        end
-
         error_type = "AllRunnersExhausted"
         error_message = "All runners exhausted"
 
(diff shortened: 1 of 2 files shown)


viamin:
## Agent Update

Updated the review-thread bug in [app/temporal/activities/run_agent_activity.rb](/workspace/app/temporal/activities/run_agent_activity.rb:483) so the single-compatible-runner path always reports `rate limited` instead of `unavailable` after an actual rate-limit result. I also added a regression spec in [spec/temporal/activities/run_agent_activity_spec.rb](/workspace/spec/temporal/activities/run_agent_activity_spec.rb:3420) that exercises the filtered single-compatible-runner case and verifies only that runner is attempted.

Validation passed on the final tree: `bundle exec rubocop` and `bundle exec rspec` both completed cleanly, with the existing 7 pending specs still pending and 0 failures. I also retried the commit through the repo hooks after fixing local Bundler path resolution, and the hook’s own lint/test run passed.

`bin/rails db:prepare` did not run successfully in this environment because Rails could not connect to PostgreSQL on the local socket.

Committed as `881d6b66` with message `fix(agent-runs): report single compatible runner as rate limited`.

viamin:
**Escalation Note**

This has been escalated because No meaningful progress for 60 minutes after 3 consecutive provider/infrastructure failures.

**How to resolve:**

viamin:
## Agent Update

Updated the rate-limit exhaustion path so the compatibility-specific message is computed before persistence and reused for `agent_run.rate_limit!`. That means when a single compatible runner is rate-limited, the database record now stores the same `No compatible runner available: ...` message that the Temporal `ApplicationError` raises. I also tightened the existing spec to assert the persisted `error_message`.

Verification passed with `bundle exec rubocop`, `bundle exec rspec`, and the pre-commit hook’s lint/test run. I also ran `bin/rails db:prepare` successfully after exporting `DB_HOST`/`DB_USERNAME`/`DB_PASSWORD` from the provided `DATABASE_URL`, since this repo’s `config/database.yml` only uses TCP settings when `DB_HOST` is present.

Committed on the PR branch as `ea867e8a` with:
`fix(agent-runs): persist compatible-runner rate-limit message`

viamin:

Issue #2083, opened 2026-05-17
Audit post-rename drift after providers->runners migration completes


viamin:
## Summary

Perform a post-migration audit and cleanup sweep for any `provider` naming that should have become `runner` naming as part of the providers-to-runners rename, but may have been introduced or reintroduced by other work that landed during the rename window.

Blocked by #2115

## Why

Phase 1 (#1950) is the application-level rename and is now deployed. Phase 2 (#2115) renames the underlying `providers` / `provider_states` tables to `runners` / `runner_states` (replacing the closed #1952). Both phases intentionally keep the legacy provider-named columns and compatibility wrappers in place so the rename can land incrementally; this issue covers the follow-up audit to remove that scaffolding and fix any drift.

## Scope

Audit code merged after PR #1950 began on 2026-05-12 through the point this issue is worked, and fix stale internal naming that should now use runner terminology. Includes the deferred cleanup explicitly carved out of #2115:

- Drop the legacy provider-named columns from `agent_runs` (`provider_id`, `provider_switches`, `providers_attempted`, `final_provider`), `chat_sessions` (`provider_id`), `user_settings` (`default_agent_provider`, `default_agent_providers_by_goal`, `fallback_providers`, `provider_selection_mode`, `provider_round_robin_state`, `kb_chat_provider`, `kb_chat_fallback_providers`, `kb_embedding_provider`, `kb_embedding_fallback_providers`), and `tenant_settings` (`provider_preferences`, `allowed_provider_keys`).
- Remove the `LegacyAttributeBridge` mappings + `sync_legacy_provider_*` callbacks once the columns are gone.
- Remove the `Provider < Runner` / `ProviderState < RunnerState` compatibility wrappers, the legacy `/providers` route + `app/views/providers/` view bridge, and the `provider_form_controller.js` Stimulus controller (rename to `runner_form_controller` once the legacy view is gone).
- Update migration specs that still reference `:providers` / `:provider_states` literally (e.g. `move_co_author_trailer_from_projects_to_providers_spec.rb`).

## Intentionally excluded naming

Do not rename terms that are intentionally about LLM providers or external provider concepts, including:

- `ProviderApiKey`
- `LlmModel.provider`
- `auth_provider`
- `final_provider` (as an LLM concept; the column on `agent_runs` is being renamed to `final_runner` separately)
- `DIRECT_OUTBOUND_API_PROVIDERS`
- `Automation::Providers::...`

## Acceptance Criteria

[...]

PR #2133, opened 2026-05-18
fix(agent-runs): legacy "provider" JSON key in runners_attempted causes silent nil reads

### LINKED ISSUE #2126: fix(agent-runs): legacy "provider" JSON key in runners_attempted causes silent nil reads
## Summary

The `agent_runs.runners_attempted` jsonb column contains a mix of two entry shapes:

- New rows: `{"runner": "claude", "success": false, …}`
- Legacy rows (and some today's rows that came through pre-rename writers): `{"provider": "claude", "success": false, …}`

Three readers fetch `attempt["runner"]` without falling back to `attempt["provider"]`, so they silently get `nil` on legacy rows. Other readers in the codebase already handle both keys.

## Evidence (live in dev DB today)

Both shapes coexist in 2026-05-18 data:

- AgentRun #18156 (today 13:52) — entries use `"runner"`.
- AgentRun #18079 (today 03:14) — entries use `"provider"`.

This is residue of the rename: migration `20260511171813_rename_providers_to_runners.rb:104-108` copied `providers_attempted` → `runners_attempted` as-is without transforming the inner JSON keys.

## Broken readers (no fallback)

1. `app/views/agent_runs/_detail.html.erb:533, 535, 546, 548` — reads `attempt["runner"]`. On legacy rows the runner display chip / failure-attribution panel renders empty.
2. `app/services/notifications/rules/zero_iteration_timeout.rb:36` — `runners_attempted: Array(...).map { |attempt| attempt["runner"] }.compact`. On legacy rows the metadata field is empty.
3. `app/models/agent_run.rb:1791` — `attempted_runners_by_routing_key` calls `Runner.id_from_routing_key(attempt["runner"])`. Legacy rows contribute no lookups, so the helper returns an incomplete map; the view in (1) feeds off this map.

## Already-safe readers (for reference)

These already do `attempt["runner"] || attempt["provider"]`:

- `app/services/coordination/failure_recovery.rb:180-184`
- `app/services/coordination/failure_recovery_service.rb:121-125`
- `app/controllers/api/secrets_proxy_controller.rb:284-287`
- `app/models/knowledge_run.rb:103-104` (different column, same pattern)

SQL paths route through `AgentRun.normalize_provider_sql` which COALESCEs both keys (`app/models/agent_run.rb:300`).

## Writer

Only one production writer remains, and it writes the new key:

- `app/models/agent_run.rb:1805-1818` — `record_runner_attempt` writes `"runner" => runner`. So if `"provider"` is appearing on today's rows, it's from a code path that bypasses this method (worth grepping for direct `runners_attempted = …` or `<<` assignments).

## Suggested fix

Either:

1. **Backfill the data.** One-shot migration that maps legacy entries to the new key:
   ```ruby
   UPDATE agent_runs SET runners_attempted = (
     SELECT jsonb_agg(
       CASE WHEN entry ? 'provider' AND NOT (entry ? 'runner')
            THEN entry - 'provider' || jsonb_build_object('runner', entry->'provider')
            ELSE entry END
     )
     FROM jsonb_array_elements(runners_attempted) AS entry
   ) WHERE runners_attempted::text LIKE '%"provider"%';
   ```
   Then drop the fallback `|| attempt["provider"]` shims gradually.

2. **Centralize the read shim.** Add `AgentRun#normalized_runners_attempted` returning entries with both keys present (or always-`"runner"`), and have every reader use it. Cheaper but leaves the underlying schema drift in place.

Option 1 is cleaner and fits with the post-#1950 / #2115 cleanup. May fit under the existing audit issue #2083, but it's a specific concrete bug worth its own fix.

## Acceptance

- `psql ... SELECT count(*) FROM agent_runs WHERE runners_attempted::text LIKE '%\"provider\":%';` returns `0`.
- The three broken readers above render correctly on rows older than the rename.

Related: #2083 (broader post-rename drift audit).

viamin:
## Summary

Fixes silent `nil` reads in three call sites that consume `agent_runs.runners_attempted` but only look up the new `"runner"` JSON key, missing legacy entries written under the pre-rename `"provider"` key. The result was empty runner chips, broken failure-attribution panels, and incomplete notification metadata for any agent run whose history predates (or bypassed) the providers→runners rename.

## Changes

- **Database**
  - `db/migrate/20260518210746_backfill_runners_attempted_legacy_provider_key.rb` — one-shot backfill that rewrites legacy `{"provider": …}` entries in `runners_attempted` to `{"runner": …}` via `jsonb_agg` over `jsonb_array_elements`, scoped to rows containing the legacy key.
- **Models**
  - `app/models/agent_run.rb:1794` — `attempted_runners_by_routing_key` now reads `attempt["runner"] || attempt["provider"]` so legacy entries contribute to the routing-key map.
- **Services**
  - `app/services/notifications/rules/zero_iteration_timeout.rb:36` — `runners_attempted` metadata now falls back to `attempt["provider"]`, keeping the notification payload complete for legacy rows.
- **UI**
  - `app/views/agent_runs/_detail.html.erb:533-548` — runner chip and failure-attribution panel both fall back to `attempt["provider"]`.

## Design Decisions

- **Backfill + reader fallbacks (belt and suspenders).** Option 1 from the issue (backfill) is the durable fix, but the reader fallbacks are kept in place to cover any in-flight rows written by a stray code path that bypasses `AgentRun#record_runner_attempt` between deploy and migration. Fallbacks can be removed later once the schema-drift audit (#2083) confirms no remaining writers emit the legacy key.
- **Scoped UPDATE.** The migration filters with `runners_attempted::text LIKE '%"provider"%'` to avoid rewriting rows that are already clean, keeping the migration cheap on large tables.
- **No SQL-path changes needed.** `AgentRun.normalize_provider_sql` already COALESCEs both keys, so query-side consumers were unaffected.

## Operational Notes

- Run `bin/rails db:migrate` on deploy. The migration is a single SQL `UPDATE` and is safe to re-run (idempotent — it only touches rows still containing `"provider"`).
- Post-deploy verification: `SELECT count(*) FROM agent_runs WHERE runners_attempted::text LIKE '%"provider":%';` should return `0`.

## Screenshots

_Reviewer: please attach before/after screenshots of an affected agent run detail page (runner chip + failure-attribution panel) for a row that previously had legacy `"provider"` entries._

Related: #2083 (broader post-rename drift audit).

---

Closes #2126

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -1791,7 +1791,7 @@ class AgentRun < ApplicationRecord
     return {} unless owner
 
     routing_ids = runners_attempted.filter_map do |attempt|
-      Runner.id_from_routing_key(attempt["runner"])
+      Runner.id_from_routing_key(attempt["runner"] || attempt["provider"])
     end
     return {} if routing_ids.empty?
 
--- a/app/services/notifications/rules/zero_iteration_timeout.rb
+++ b/app/services/notifications/rules/zero_iteration_timeout.rb
@@ -33,7 +33,7 @@ module Notifications
             project_id: agent_run.project_id,
             source_pull_request_number: agent_run.source_pull_request_number,
             trigger_type: agent_run.trigger_type,
-            runners_attempted: Array(agent_run.runners_attempted).map { |attempt| attempt["runner"] }.compact
+            runners_attempted: Array(agent_run.runners_attempted).map { |attempt| attempt["runner"] || attempt["provider"] }.compact
           }.compact
         }
       end
--- a/app/views/agent_runs/_detail.html.erb
+++ b/app/views/agent_runs/_detail.html.erb
@@ -530,9 +530,10 @@
               <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M9 5l7 7-7 7" />
             </svg>
           <% end %>
-          <% attempt_runner = attempted_runners_by_routing_key[attempt["runner"]] %>
+          <% attempt_runner_key = attempt["runner"] || attempt["provider"] %>
+          <% attempt_runner = attempted_runners_by_routing_key[attempt_runner_key] %>
           <span class="inline-flex items-center rounded-md px-2 py-1 text-xs font-medium <%= attempt['success'] ? 'bg-green-100 text-green-800' : 'bg-red-100 text-red-800' %>">
-            <%= runner_display_for_identifier(attempt["runner"], runner: attempt_runner) %>
+            <%= runner_display_for_identifier(attempt_runner_key, runner: attempt_runner) %>
             <% if attempt['error_type'].present? %>
               <span class="ml-1 text-xs opacity-75">(<%= attempt['error_type'] %>)</span>
             <% end %>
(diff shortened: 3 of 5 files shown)


viamin:

PR #2154, opened 2026-05-19
perf(webhooks): add composite index on agent_runs(project_id, source_pull_request_number, status) if review-goal feedback volume grows

### LINKED ISSUE #2139: perf(webhooks): add composite index on agent_runs(project_id, source_pull_request_number, status) if review-goal feedback volume grows
Follow-up to #943.

## Summary

`Api::GithubWebhooksController#find_agent_run_by_pr` falls back to looking up review-goal runs by `source_pull_request_number` with `status: "completed"`. The existing partial unique index on `(project_id, source_pull_request_number)` covers active runs only and does not help this completed-status query. At current volume this is fine, but if review-goal feedback volume grows we will need a composite index.

## Site

`app/controllers/api/github_webhooks_controller.rb:232`

## Trigger condition

File a follow-up to add the index when any of the following becomes true:

- The query appears in slow-query logs / DB performance dashboards.
- Review-goal completed-run row counts cross an order of magnitude where unindexed lookups become noticeable.
- p95 webhook latency for review-feedback paths starts trending up.

## Suggested change (when triggered)

Add a non-partial composite index on `agent_runs(project_id, source_pull_request_number, status)`. Consider `INCLUDE (created_at)` if the planner picks Index-Only Scans.

## Acceptance criteria

- Either: (a) index added once the trigger fires, with a migration that backfills cleanly on a hot table, OR (b) issue closed as `wontfix` after volume monitoring confirms the trigger condition will not be met.

viamin:
## Summary

Speeds up the review-goal webhook lookup path by adding a composite index that covers completed-status runs, which the existing partial unique index does not. Addresses the follow-up from #943 to prevent latency growth as review-goal feedback volume scales.

## Changes

- **Database**: New migration `20260519004131_add_composite_index_to_agent_runs_for_review_webhook_lookup.rb` adds `idx_agent_runs_review_feedback_lookup` on `agent_runs(project_id, source_pull_request_number, status)`; reflected in `db/schema.rb:268`.
- **Controllers**: Removed the stale TODO in `app/controllers/api/github_webhooks_controller.rb:223` now that the index is in place to support `find_agent_run_by_pr`'s completed-status fallback lookup.

## Design Decisions

- Chose a non-partial composite index over extending the existing partial unique index because the fallback path queries `status: "completed"`, which the active-runs partial index does not cover.
- Did not add `INCLUDE (created_at)` upfront; this can be revisited if planner data shows Index-Only Scans would benefit, per the trigger guidance in the issue.

## Operational Notes

- `agent_runs` is a hot table — verify the migration uses `algorithm: :concurrently` (or equivalent) and runs outside a transaction to avoid blocking writes during deploy.

---

Closes #2139

### DIFF
--- a/app/controllers/api/github_webhooks_controller.rb
+++ b/app/controllers/api/github_webhooks_controller.rb
@@ -226,11 +226,9 @@ module Api
       # CollectReviewReactionFeedback, which queries source_pull_request_number
       # directly for review-goal runs.
       #
-      # NOTE: the existing partial unique index on (project_id,
-      # source_pull_request_number) covers active runs only and does not help
-      # this completed-status query. At current volume that is fine.
-      # TODO(#943): if review-goal feedback volume grows, consider a composite
-      # index on (project_id, source_pull_request_number, status).
+      # NOTE: completed review-goal lookups use the dedicated
+      # idx_agent_runs_review_feedback_lookup composite index because the
+      # active-run unique index only covers queued/pending/running/paused rows.
       @project.agent_runs
         .where(pull_request_number: pr_number, status: "completed")
         .order(created_at: :desc)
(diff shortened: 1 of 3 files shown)


viamin:

Issue #664, opened 2026-03-31
Support custom pre-commit requirements (configurable at project/user/account level)


viamin:
## Summary

Allow users to configure custom pre-commit checks and requirements that agent runs must satisfy before creating commits or opening PRs. This enables enforcement of team-specific standards like linting rules, type checking, test coverage thresholds, and security scans.

## Motivation

Enterprise codebases typically have mandatory pre-commit checks beyond what the repository's existing git hooks enforce. As an individual developer using paid, being able to configure these requirements ensures that agent-generated code meets the same standards as manually written code, reducing review friction and CI failures.

## Requirements

- Configurable at **project**, **user**, or **account** level
- Define a list of commands/checks that must pass before an agent run commits
- Support common check types:
  - Shell commands (e.g., `npm run lint`, `bundle exec rubocop`, `cargo clippy`)
  - Test suite requirements (e.g., "all tests must pass", "new tests required for new files")
  - Coverage thresholds (e.g., "no decrease in test coverage")
  - Security scanning (e.g., `npm audit`, `brakeman`)
- Configurable failure behavior: block commit, warn and continue, or auto-fix and retry
- Pre-commit check results should be visible in the agent run detail page
- Support ordering/priority of checks

## Implementation Notes

- Add a `PreCommitRequirement` model associated with project/user/account
- Agent harness executes configured checks after code changes, before commit
- Check results are captured in agent run logs with pass/fail status
- Auto-fix mode: if a check fails and provides a fix command (e.g., `rubocop -a`), run it and re-check
- Settings hierarchy: account defaults < user overrides < project overrides (with merge semantics for lists)

## Labels

- enhancement

PR #2155, opened 2026-05-19
fix(pre-commit): bind user-level requirements to the run's initiating user, not project.effective_owner

### LINKED ISSUE #2143: fix(pre-commit): bind user-level requirements to the run's initiating user, not project.effective_owner
Follow-up to #664.

## Summary

`PreCommitRequirements::Evaluate#call` currently resolves requirements with `user: agent_run.project.effective_owner` because `AgentRun` does not track who triggered the run. This means user-level pre-commit requirements always bind to the project owner — wrong for any run triggered by another collaborator.

## Site

`app/services/pre_commit_requirements/evaluate.rb:30`

## Steps

1. Add an `initiating_user_id` (or similar) column to `agent_runs`, populated at run creation time across all entry points (auto-pick, manual trigger, webhook-triggered review, etc.).
2. Update `PreCommitRequirement.resolve` callers (and any other user-scoped settings resolution that piggy-backs on the project owner today) to use the initiating user when present, falling back to `project.effective_owner` only when the run has no associated user (e.g. system-triggered).
3. Remove the TODO.

## Risk / scope

- All paths that create an `AgentRun` must set the new column; missing it on any path silently regresses to the old behavior.
- Migration must be safe to deploy with existing in-flight runs (column nullable; backfill optional or no-op for completed runs).

## Acceptance criteria

- `AgentRun` carries the initiating user.
- User-level pre-commit requirements bind to that user, not the project owner.
- Specs cover the owner-fallback, non-owner-initiator, and system-initiated cases.
- The TODO is removed.

viamin:
## Summary

User-level pre-commit requirements were always binding to the project owner because `AgentRun` had no record of who initiated the run, producing incorrect results whenever a collaborator (not the owner) triggered work. This change tracks the initiating user on each `AgentRun` and routes user-scoped settings resolution through them, falling back to the project owner only for system-triggered runs. Follow-up to #664.

## Changes

**Database**
- New migration adds a nullable, indexed `initiating_user_id` column to `agent_runs` with a foreign key to `users` (`on_delete: :nullify`). Split into three migrations — add column, add unvalidated FK, validate FK — so it is safe to deploy alongside in-flight runs.

**Models**
- `AgentRun`: `belongs_to :initiating_user, optional: true`, account validation, and a `before_validation` callback that auto-populates from `Current.user`. New `settings_user` helper returns `initiating_user` and falls back to `project.effective_owner`.
- `User`: inverse `has_many :initiated_agent_runs` association.

**Services / activities**
- `PreCommitRequirements::Evaluate` now resolves requirements with `agent_run.settings_user`; the `#664` TODO is removed.
- `CreatePullRequestActivity` uses `agent_run.settings_user` for PR template resolution so user-scoped templates follow the same binding rule.

**Run-creation entry points (all updated to set the initiating user)**
- Controllers: manual trigger, retry, and `refresh_auth` paths.
- `Issues::AutoPick`: explicitly passes `nil` to mark system-initiated.
- `CreateFollowup` and `RetryTimedOutIssueGoalJob`: propagate the initiating user from the source run.
- MCP `trigger_agent_run` tool and the Temporal activities/workflows that spawn runs.

**Specs**
- Cover non-owner initiator, owner fallback, system-initiated (`nil`), and propagation through retries and follow-ups.

## Design Decisions

- **Nullable column with explicit `nil` for system runs.** Auto-pick, retries of orphaned runs, and webhook-driven flows are legitimately userless; encoding that as `nil` is clearer than synthesizing a sentinel user and keeps the fallback to `project.effective_owner` honest.
- **`settings_user` as the single resolution point.** Both `PreCommitRequirements::Evaluate` and `CreatePullRequestActivity` go through it, so future user-scoped settings can adopt the same rule without each caller re-implementing the fallback.
- **Three-step FK migration.** Adding the column, adding the FK as `NOT VALID`, and validating it separately avoids long locks on `agent_runs` and is safe to run while runs are being created.
- **`Current.user` auto-population.** Keeps controller and job call sites from having to thread the user manually, while explicit callers (auto-pick, system jobs) still pass `nil` to make system-initiated runs unambiguous.

## Operational Notes

- Deploy the three migrations in order; no backfill is required — existing and in-flight runs remain `nil` and correctly fall back to `project.effective_owner`.
- No configuration or infrastructure changes.

---

Closes #2143


viamin:
## Agent Update

Resolved both review threads. I added the missing blank line between `AgentRun#settings_user` and the following scope in [app/models/agent_run.rb](/workspace/app/models/agent_run.rb:275), and removed the unrelated Bundler checksum hunk from `Gemfile.lock` so the PR diff stays focused.

Verification passed on the final tree: `bundle exec rubocop` succeeded, `bundle exec rspec` finished with `13450 examples, 0 failures, 7 pending` (the same existing pending specs, including the Qdrant and renamed-`providers` migration cases), and the pre-commit hook passed during commit. The new commit is `693eab84` with message `fix(pre-commit): address initiating-user review feedback`.

viamin:

Issue #1121, opened 2026-04-14
Extract auto-continue into strategy modules


viamin:
Depends on #1115
Depends on #1117
Depends on #1118
Depends on #1119

## Summary
Extract auto-continue into a dedicated strategy module that owns PR lifecycle progression decisions.

## Goals
- Move PR-state evaluation out of large workflow/activity branches and into an explicit automation strategy.
- Keep workflows/jobs focused on orchestration and retries rather than detailed policy.
- Preserve current draft/ready/escalated lifecycle behavior while making it easier to plug in other work-item/repository providers later.

## Should Cover
- A decision model for auto-continue actions such as:
  - queue follow-up run
  - request review
  - mark ready
  - escalate to owner
  - dismiss escalation
  - retry review-goal work
- Strategy-level handling of phase progression and gating.
- Clear boundaries between state gathering, policy evaluation, and action execution.

## Deliverable
- Auto-continue logic owned by a modular strategy rather than embedded scanner/workflow logic.

viamin:
Pull request created: https://github.com/viamin/paid/pull/1515

PR #2156, opened 2026-05-19
refactor(pr-scan): stop scan_pr from re-evaluating gates already covered by lifecycle/strategy path

### LINKED ISSUE #2138: refactor(pr-scan): stop scan_pr from re-evaluating gates already covered by lifecycle/strategy path
Follow-up to #1121.

## Summary

`ScanPaidPrsActivity#scan_pr` evaluates several lifecycle gates (`operational_failure_breaker?`, `active_run_exists?`, etc.), and then `build_lifecycle_signals` is invoked separately and re-queries the same gates. With the strategy modules now owning gate evaluation, this re-query is redundant — roughly doubling gate-check DB queries per PR.

## Site

`app/temporal/activities/scan_paid_prs_activity.rb:118`

## Goal

Once strategies fully own gate evaluation, `scan_pr` should stop evaluating gates itself and defer entirely to the signals/strategy path, so each gate is checked at most once per PR per scan.

## Acceptance criteria

- `scan_pr` no longer evaluates lifecycle gates; only the signals path does.
- Per-PR gate-check DB query count is measurably lower (add a spec or benchmark assertion if practical).
- Behavior is unchanged: existing scan specs continue to pass without modification, or modifications reflect only the removed duplication.

viamin:
## Summary

Eliminates redundant lifecycle gate evaluation in `ScanPaidPrsActivity#scan_pr`, which was re-querying gates already checked by `build_lifecycle_signals` on the strategy path — roughly doubling per-PR gate-check DB queries. Follow-up to #1121, completing the migration of gate ownership to strategy modules.

## Changes

- **`app/temporal/activities/scan_paid_prs_activity.rb`** — `scan_pr` no longer evaluates lifecycle gates on the explicit strategy path; defers entirely to `build_lifecycle_signals` and strategy evaluation.
- **Result collection** — adjusted so explicit automation can still emit lifecycle-only decisions without `scan_pr` synthesizing them.
- **Draft / ready / escalated scanners** — threaded the explicit-path behavior through so followup, retry, and dismissal gates are not re-run there either.
- **Specs** — added coverage asserting the active-run lifecycle gate query fires only once per explicit PR scan.

## Design Decisions

- Kept `scan_pr` as the entry point but made it a pure dispatcher on the explicit strategy path — all semantic gate decisions now live in the strategy modules, consistent with the direction set in #1121.
- Preserved the ability for explicit automation to emit lifecycle-only decisions by adjusting result collection, rather than restoring synthesis in `scan_pr` (which would have re-introduced the duplication this PR removes).
- Existing scan specs pass without modification, confirming behavior parity; the new query-count assertion locks in the perf improvement so future regressions surface in tests rather than production.

---

Closes #2138

### DIFF
--- a/app/temporal/activities/scan_paid_prs_activity.rb
+++ b/app/temporal/activities/scan_paid_prs_activity.rb
@@ -125,11 +125,10 @@ module Activities
         issue.update_column(:last_pr_scan_at, Time.current)
         pending_review_states << pending_review_state(issue, result)
         progress_states << serialized_pr_progress_state(project, issue)
-        if result
-          collect_scan_result(issue, result, prs_to_trigger, automation_results,
-            explicit_pr_decisions:,
-            lifecycle: explicit_pr_decisions ? build_lifecycle_signals(project, issue) : nil)
-        end
+        lifecycle = explicit_pr_decisions ? build_lifecycle_signals(project, issue) : nil
+        collect_scan_result(issue, result, prs_to_trigger, automation_results,
+          explicit_pr_decisions:,
+          lifecycle:)
       rescue Temporalio::Error::ApplicationError => e
         raise unless e.type == "RateLimit"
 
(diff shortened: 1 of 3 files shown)


viamin:

PR #2164, opened 2026-05-20
fix(agent-runs): preserve ContainerNotProvisioned error on first attempt

viamin:
## Summary

- Restores the `ContainerNotProvisioned` → `RunnerExecutionError` wrap in `run_agent_with_runner` (dropped during the providers→runners rename in #1950, originally added in #2077).
- Gates the wrap on `agent_run.runners_attempted.present?` so the precise `ContainerNotProvisioned` error still propagates when there's no prior runner attempt to preserve.

## Why

#2077 added the wrap to fix a real bug: when a prior runner attempt failed and cleared `container_id`, the next attempt's `reconnect_container` raised `ContainerNotProvisioned` and overwrote the original failure as the top-level Temporal activity error.

The unconditional wrap, however, regressed the first-attempt case: a fresh run with no container ever provisioned now surfaced `"All runners exhausted"` instead of the more specific `"No container provisioned"`, forcing operators to drill into `runners_attempted` to find the actual cause.

Then during the providers→runners rename refactor (#1950), the rescue block was dropped entirely while restructuring `run_agent_with_provider` into `run_agent_with_runner`, regressing the original #2077 fix on main.

This change restores the wrap and applies the gate in one go.

## Behavior

| Scenario | Before this PR | After this PR |
| --- | --- | --- |
| Fresh run, container_id nil, no prior attempts | `ContainerNotProvisioned` propagates (currently on main) | `ContainerNotProvisioned` propagates (unchanged) |
| Prior runner attempt recorded, container_id cleared | `ContainerNotProvisioned` propagates and masks the real cause | Wrapped as `RunnerExecutionError`; loop records attempt; surfaces `AllRunnersExhausted` with original failure visible in `runners_attempted` |

## Test plan

- [x] `bin/rspec spec/temporal/activities/run_agent_activity_spec.rb` — all 226 specs pass
- [x] Existing `"when no container is provisioned"` spec still asserts raw `ContainerNotProvisioned` propagation
- [x] New regression spec `"wraps ContainerNotProvisioned as RunnerExecutionError when a prior runner attempt is already recorded"` covers the wrap path

### DIFF
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -940,7 +940,23 @@ module Activities
     #
     # @return [Hash] The pre-agent SHA and whether output was present
     def run_agent_with_runner(agent_run, runner_candidate, prompt, user_settings)
-      container_service = reconnect_container(agent_run)
+      container_service = begin
+        reconnect_container(agent_run)
+      rescue Temporalio::Error::ApplicationError => e
+        # When a prior runner attempt has already recorded a failure on this
+        # run, a secondary ContainerNotProvisioned (e.g. because the earlier
+        # failure cleared container_id) would otherwise overwrite the real
+        # root cause at the top level. Wrap it as RunnerExecutionError so the
+        # per-runner rescue records the attempt and the loop surfaces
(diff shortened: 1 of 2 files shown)


viamin:

PR #2179, opened 2026-05-20
fix: Auto-pick allowlist filter is case-sensitive in SQL but case-insensitive in model helper

### LINKED ISSUE #2177: Auto-pick allowlist filter is case-sensitive in SQL but case-insensitive in model helper
## Summary

`Project#allowed_github_usernames` (the auto-pick trusted-creator allowlist) is enforced inconsistently:

- **Model helper** `Project#trusted_github_user?` does a **case-insensitive** comparison.
- **SQL filter** in `Automation::Strategies::AutoPick::DefaultCandidateSource#base_scope` does a **case-sensitive** comparison.

The same allowlist value therefore yields different "trusted?" answers depending on which code path runs, and auto-pick can silently drop every candidate issue when the allowlist entry differs from `github_creator_login` only in case.

## Where

- Case-insensitive helper: `app/models/project.rb:403-407`
  ```ruby
  def trusted_github_user?(login)
    return false if login.blank?

    allowed_github_usernames.any? { |allowed| allowed.downcase == login.downcase }
  end
  ```

- Case-sensitive SQL filter: `app/services/automation/strategies/auto_pick/default_candidate_source.rb:189-190`
  ```ruby
  trusted_usernames = Array(project.allowed_github_usernames).presence
  base = base.where(github_creator_login: trusted_usernames) if trusted_usernames
  ```

`where(github_creator_login: trusted_usernames)` compiles to `github_creator_login IN (...)`, which Postgres evaluates case-sensitively.

## Repro

1. Project with `allowed_github_usernames = ["Viamin"]`.
2. All open issues have `github_creator_login = "viamin"`.
3. `project.trusted_github_user?("viamin")` → `true`.
4. `DefaultCandidateSource.eligible_scope(project).count` → `0`.
5. Auto-pick never creates a run, with no log line explaining why.

## Suggested fix

Pick one source of truth. Either:

- Make the SQL filter case-insensitive — e.g. `where("LOWER(github_creator_login) IN (?)", trusted_usernames.map(&:downcase))` (and add a functional index on `LOWER(github_creator_login)` if hot), or
- Normalize `allowed_github_usernames` to lowercase on write and downcase `github_creator_login` on sync, then keep the exact-match SQL.

Add a regression spec asserting that an allowlist entry of `"Viamin"` matches a `github_creator_login` of `"viamin"` through `DefaultCandidateSource.eligible_scope`.

## How it was found

Diagnosing why a project with `auto_pick_enabled = true` produced zero agent runs. Gate was open, 97 candidate issues passed every other filter, but the allowlist filter dropped them all because the entry was the org name (`HuntHelper`) and creators were `viamin` / `honeybadger[bot]`. The case-sensitivity mismatch was a parallel finding from the same diagnostic — it would have masked legitimate allowlist entries that differed only in case.

viamin:
## Summary

Fixes a case-sensitivity mismatch in the auto-pick trusted-creator allowlist that caused legitimate candidate issues to be silently dropped. `Project#trusted_github_user?` compares case-insensitively, but `DefaultCandidateSource`'s SQL filter compiled to `WHERE github_creator_login IN (...)`, which Postgres evaluates case-sensitively — so an allowlist entry differing only in case (e.g. `"Viamin"` vs `"viamin"`) produced zero candidates with no log line.

## Changes

- **Services**: `app/services/automation/strategies/auto_pick/default_candidate_source.rb` — replaced the case-sensitive `where(github_creator_login: trusted_usernames)` with `where("LOWER(issues.github_creator_login) IN (?)", trusted_usernames.map(&:downcase))` so the SQL filter matches the model helper's behavior.
- **Tests**: `spec/services/automation/strategies/auto_pick/default_candidate_source_spec.rb` — added two regression specs covering the mixed-case allowlist scenario (`"Viamin"` matching `"viamin"`) and confirming non-matching creators are still excluded.

## Design Decisions

- **Picked the SQL-side fix over normalizing on write.** Normalizing `allowed_github_usernames` on write and `github_creator_login` on sync would require a data backfill and changes to the sync path; making the SQL filter case-insensitive is a localized fix that aligns with the existing `Project#trusted_github_user?` semantics already shipping in production.
- **No functional index added.** The allowlist filter runs against an already narrowed candidate set in auto-pick, not a hot OLTP path. If profiling later shows this is a hotspot, a `LOWER(github_creator_login)` functional index can be added without further code changes.

---

Closes #2177

### DIFF
--- a/app/services/automation/strategies/auto_pick/default_candidate_source.rb
+++ b/app/services/automation/strategies/auto_pick/default_candidate_source.rb
@@ -187,7 +187,10 @@ module Automation
               .where.not(id: Issue.open_pull_request_parent_issue_ids(project: project).distinct)
 
             trusted_usernames = Array(project.allowed_github_usernames).presence
-            base = base.where(github_creator_login: trusted_usernames) if trusted_usernames
+            if trusted_usernames
+              downcased = trusted_usernames.map(&:downcase)
+              base = base.where("LOWER(issues.github_creator_login) IN (?)", downcased)
+            end
 
             project.effective_auto_pick_skip_labels.reduce(base) do |scope, label|
               scope.where.not("labels @> ?::jsonb", [ label ].to_json)
--- a/spec/services/automation/strategies/auto_pick/default_candidate_source_spec.rb
+++ b/spec/services/automation/strategies/auto_pick/default_candidate_source_spec.rb
@@ -190,6 +190,24 @@ RSpec.describe Automation::Strategies::AutoPick::DefaultCandidateSource do
 
       expect(described_class.eligible_scope(project).pluck(:labels)).to include([ "planning" ])
     end
+
+    it "matches allowlist entries case-insensitively against github_creator_login" do
+      project.update!(allowed_github_usernames: [ "Viamin" ])
+      issue = create(:issue, project: project, github_creator_login: "viamin")
+
+      scope = described_class.eligible_scope(project)
+
+      expect(scope.pluck(:id)).to contain_exactly(issue.id)
+    end
+
+    it "excludes issues whose creator is not in the allowlist regardless of case" do
+      project.update!(allowed_github_usernames: [ "Viamin" ])
+      create(:issue, project: project, github_creator_login: "otheruser")
+
+      scope = described_class.eligible_scope(project)
+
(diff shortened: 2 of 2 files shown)


viamin:

PR #2194, opened 2026-05-21
fix(agent-runs): resume analyze-issue followups

### LINKED ISSUE #2191: Auto-pick follow-up dispatch should guard on result shape, not coerce nil to enhance_issue
## Summary

The new analyze→follow-up dispatch in `Workflows::AgentExecutionWorkflow` (uncommitted, see local diff in `app/temporal/workflows/agent_execution_workflow.rb:148`) reads `result[:sufficient_context]` from `AnalyzeIssueActivity` and chooses between `create_pr` and `enhance_issue`:

```ruby
followup_goal = result[:sufficient_context] ? "create_pr" : "enhance_issue"
run_activity(Activities::CreateFollowupRunActivity,
  { agent_run_id: agent_run_id, goal: followup_goal },
  timeout: 30, retry_policy: NO_RETRY)
```

If `AnalyzeIssueActivity` returns a malformed hash (e.g. agent crashed mid-parse, JSON missing the key), `result[:sufficient_context]` is `nil`, which is falsy — so the workflow silently enqueues an `enhance_issue` run as if context were insufficient. That's wrong: \"we couldn't tell\" is not the same as \"the agent said context is insufficient,\" and it spends budget on an avoidable enhance round plus an erroneous `enhance_issue_needs_input` label.

## Repro

1. Make `AnalyzeIssueActivity#execute` return `{ agent_run_id: 42 }` (omit `sufficient_context`).
2. Run the workflow.
3. Observe `CreateFollowupRunActivity` is called with `goal: \"enhance_issue\"`.

## Suggested fix

Only enqueue a follow-up when the analysis returned a definitive answer:

```ruby
if result.is_a?(Hash) && result.key?(:sufficient_context)
  followup_goal = result[:sufficient_context] ? \"create_pr\" : \"enhance_issue\"
  run_activity(Activities::CreateFollowupRunActivity, ...)
else
  Rails.logger.warn(message: \"agent_execution.analyze_result_missing_sufficient_context\", agent_run_id: agent_run_id)
end
```

Add a spec case in `spec/temporal/workflows/agent_execution_workflow_spec.rb` covering the missing-key path that asserts `CreateFollowupRunActivity` is NOT invoked.

## How found

Surfaced while tracing why hunthelper/hunthelper had 95 issues end up stuck at `paid_state: \"analyzed\"` and reviewing the uncommitted follow-up dispatch.

### LINKED ISSUE #2192: Backfill issues stuck at paid_state="analyzed" when analyze→follow-up dispatch ships
## Summary

Auto-pick's `DefaultCandidateSource` only considers issues with `paid_state ∈ {new, planning, failed}` (plus the recoverable-completed branch — `app/services/automation/strategies/auto_pick/default_candidate_source.rb:66`). The current production behavior is:

1. Auto-pick picks a `new` issue and fires `analyze_issue`.
2. `MarkAgentRunCompleteActivity` sets the issue to `paid_state: \"analyzed\"` on success (`app/temporal/activities/mark_agent_run_complete_activity.rb:24`).
3. Nothing automatically advances `analyzed` issues — they fall out of the eligible scope and the loop stops.

The uncommitted `CreateFollowupRunActivity` flow (`agent_execution_workflow.rb:148`) fixes the forward path for *future* analyses by enqueueing a `create_pr` or `enhance_issue` follow-up immediately. But it doesn't help any issue that already reached `analyzed` *before* the change ships.

## Evidence

On hunthelper/hunthelper (project id 14) the queue drained yesterday and now sits at:

- 95 open issues at `paid_state: \"analyzed\"`
- 1 at `failed`, 1 at `new`
- `DefaultCandidateSource.eligible_scope.count = 0`

Once the follow-up dispatch lands, only newly analyzed issues will progress; the 95 stuck ones (and every analogous backlog on other projects) need a backfill.

## Suggested fix

Two options, pick one:

1. **One-shot rake task / migration** that scans `issues` with `paid_state = \"analyzed\"`, looks up the most recent completed `analyze_issue` run, and either re-enqueues an `analyze_issue` run (cheap, lets the new dispatch take over) or directly enqueues the right follow-up based on whatever signal the original analysis recorded.
2. **New job** (e.g. `AdvanceAnalyzedIssuesJob`) that runs as part of the auto-pick backfill cadence and routes `analyzed` issues to the appropriate follow-up.

Either should be guarded by `Issues::AutoPickProjectGate` so paused/owner-less projects stay idle.

## How found

Diagnosing why hunthelper/hunthelper showed no new agent runs after the trusted-allowlist fix. The fix worked once (96 analyze runs fired) and then everything stalled at `analyzed`.

viamin:
## Summary
- enqueue follow-up runs after `analyze_issue` instead of stopping at `paid_state = analyzed`
- route follow-ups through the normal queueing path so duplicate protection and `auto_pick` metadata stay consistent
- add a maintenance backfill for legacy analyzed issues that were already stranded before the workflow fix

Closes #2191
Closes #2192

## Testing
- targeted workflow/activity/job specs passed in the original workspace before branch packaging
- pre-commit lint checks passed during commit
- branch-worktree spec rerun hit branch-specific missing test DB setup, not a code failure

### DIFF
--- a/app/jobs/analyze_issue_followup_backfill_job.rb
+++ b/app/jobs/analyze_issue_followup_backfill_job.rb
@@ -1,0 +1,145 @@
+# frozen_string_literal: true
+
+# One-time backfill for issues stranded in paid_state="analyzed" before
+# analyze_issue workflows started enqueueing their own follow-up runs.
+#
+# Without this pass, legacy analyzed issues can remain stuck indefinitely:
+# they are no longer eligible for auto-pick, and unchanged issues are not
+# revisited by incremental sync.
+class AnalyzeIssueFollowupBackfillJob < ApplicationJob
+  include GoodJob::ActiveJobExtensions::Concurrency
+
+  queue_as :maintenance
+
+  good_job_control_concurrency_with(
+    total_limit: 1,
+    enqueue_limit: 1,
+    key: "analyze_issue_followup_backfill"
+  )
+
+  CACHE_NAMESPACE = "analyze_issue_followup_backfill/v1"
+
+  def perform
+    return if completed?
+
+    processed = 0
+
+    eligible_projects.find_each do |project|
+      next if project_backfilled?(project.id)
+      next unless Issues::AutoPickProjectGate.call(project)
+
+      backfill_project(project)
+      mark_project_backfilled(project.id)
+      processed += 1
+    rescue => e
+      Rails.logger.error(
+        message: "analyze_issue_followup_backfill.project_failed",
+        project_id: project.id,
+        error: e.message
+      )
+    end
+
+    remaining = remaining_project_ids
+    mark_completed if remaining.empty?
+
+    Rails.logger.info(
+      message: "analyze_issue_followup_backfill.completed",
+      processed_projects: processed,
+      remaining_projects: remaining.size
+    )
+  end
+
+  private
+
+  def backfill_project(project)
+    analyzed_issues_for(project).find_each do |issue|
+      analysis_run = latest_completed_analysis_run_for(issue)
+      next unless analysis_run
+
+      followup_goal = followup_goal_for(analysis_run)
+      next unless followup_goal
+
+      Activities::CreateFollowupRunActivity.new.execute(
+        agent_run_id: analysis_run.id,
+        goal: followup_goal
+      )
+    end
+  end
+
+  def analyzed_issues_for(project)
+    project.issues
+      .where(github_state: "open", is_pull_request: false, paid_state: "analyzed")
+      .where.not(
+        id: project.agent_runs
+          .where(status: AgentRun::UNFINISHED_STATUSES)
+          .where.not(issue_id: nil)
+          .select(:issue_id)
+      )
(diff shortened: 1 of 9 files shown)


viamin:
## Agent Update

Fixed the failing `test` check by updating the queue-priority guard spec to include `AnalyzeIssueFollowupBackfillJob` in the `maintenance` queue list at [spec/config/application_job_queue_priority_spec.rb](/workspace/spec/config/application_job_queue_priority_spec.rb:24). The branch’s code was fine; CI was failing because the new job class existed under `app/jobs` but wasn’t registered in the spec that enforces queue coverage for every job.

Verified with `bundle exec rubocop`, targeted branch specs, an isolated repro of the transient hook failure, and the full pre-commit hook run. The commit succeeded as `f5a7be13` with message `fix(ci): register analyze issue backfill job queue`. The full `bundle exec rspec` hook run finished with `13631 examples, 0 failures, 7 pending`.

viamin:

PR #2211, opened 2026-05-22
feat(prompts): expose public trusted-comment filter for PR prompts

viamin:
## Summary

- Promote `Prompts::BuildForPr.select_trusted_comments(comments, project:)` to a public class method so REPLs, scripts, and local dogfooding can apply the **same** trusted-user + Paid-generated-comment filter that the live PR prompt builder uses — no more inline reimplementations that risk drifting from the production filter.
- Instance method delegates to the class method (single source of truth). `paid_generated_pr_comment?` stays private — the public surface is just the one filter.
- Update `CLAUDE.md` so the "Read all comments" bullet no longer reads as a blanket policy: human/AI context-gathering can read everything, but any code path that feeds comment bodies into an LLM prompt **must** filter through the allowlist (pointing at `Prompts::BuildForIssue.fetch_trusted_comments` for issues and the new public PR helper). This closes a gap where a developer following the old wording could re-introduce a prompt-injection sink.

## Why now

The CLAUDE.md "Read all comments" directive read like a security policy in tension with the actual production filter (`BuildForIssue#build` raises `UntrustedIssueError`, `fetch_trusted_comments` + `select_trusted_comments` drop non-allowlisted authors). The directive only applies in the devcontainer (CLAUDE.md is not mounted into paid-agent containers — verified), but a dev or AI assistant prototyping new prompt-builder code there could read it and skip the filter. Tightening the wording + giving the PR side a public helper makes the safe path the easy path.

## Test plan

- [x] `bin/rspec spec/services/prompts/build_for_pr_spec.rb` — 74 examples, 0 failures (67 existing + 7 new covering: trusted kept, untrusted dropped, nil-login dropped, nil-body tolerated, agent-update marker dropped, escalation-note marker dropped, input order preserved)
- [x] `bin/rubocop app/services/prompts/build_for_pr.rb spec/services/prompts/build_for_pr_spec.rb` — clean
- [x] Pre-commit hooks pass (RuboCop, markdownlint)
- [ ] CI green on the PR
- [ ] Smoke: open Rails console, call `Prompts::BuildForPr.select_trusted_comments(client.recent_issue_comments("owner/repo", N, pages: 1), project: project)` and confirm allowlisted-only output

## Out of scope (followups)

- `Activities::CompleteExistingPrRunActivity.agent_update_comment?` uses a substring match on the HTML marker (line 18). A trusted user *quoting* a Paid comment in their reply gets silently filtered out. Real UX issue, but pre-existing and changing it shifts production filter semantics — belongs in its own PR with tests for the legacy `## Agent Update` / `Agent pushed updates…` paths.


viamin:
## Agent Update

Restored the dropped `ProjectConventions::InjectIntoPrompt` call in [app/services/prompts/build_for_pr.rb](/workspace/app/services/prompts/build_for_pr.rb:127) so PR prompts again include the repository automation conventions section after style-guide injection. I also added regression coverage in [spec/services/prompts/build_for_pr_spec.rb](/workspace/spec/services/prompts/build_for_pr_spec.rb:153) and updated the no-DB stubbed context there to account for the restored injector.

Validation passed:
- `bin/rails db:prepare`
- `bundle exec rubocop`
- `bundle exec rspec` (`13733 examples, 0 failures, 7 pending`)

Committed on the current branch as `c7175f43` with message `fix(prompts): restore pr project conventions`.

viamin:

PR #2212, opened 2026-05-23
fix: Reject create_pr runs for untrusted issues at queue time (root cause of "No prompt available")

### LINKED ISSUE #2193: Reject create_pr runs for untrusted issues at queue time (root cause of "No prompt available")
## Summary

`Activities::RunAgentActivity` raises `Temporalio::Error::ApplicationError.new(\"No prompt available for agent run\", type: \"MissingPrompt\", non_retryable: true)` (`app/temporal/activities/run_agent_activity.rb:127-130`) whenever `AgentRun#effective_prompt` returns nil. For `create_pr` runs, the most common cause is *not* an actual missing prompt — it's a silently-failed trusted-creator check buried in `prompt_for_issue`:

```ruby
# app/models/agent_run.rb:1695-1700
def prompt_for_issue
  return nil unless issue
  return nil unless issue.trusted?     # silent
  Prompts::BuildForIssue.call(...)
end
```

`issue.trusted?` → `project.trusted_github_user?(github_creator_login)` → checks the project's `allowed_github_usernames` allowlist case-insensitively (model helper at `project.rb:403-407`, separately tracked in #2177 for SQL/Ruby case-sensitivity mismatch).

When the creator is not on the allowlist (e.g. `honeybadger[bot]`, `dependabot[bot]`), the prompt silently becomes nil. The run is created, picked up, scheduled on a runner, and **only** errors out 8+ minutes later inside `RunAgentActivity` — wasting a queue slot, a Temporal activity, and emitting a misleading \"missing prompt\" error message.

Auto-pick's `DefaultCandidateSource#base_scope` already enforces the allowlist at selection time (`default_candidate_source.rb:189-190`), so it never hits this path. But the older automation path in `GitHubPollWorkflow#queue_create_pr_run` (`app/temporal/workflows/git_hub_poll_workflow.rb:321-336`) and any future caller of `QueueAgentRunActivity` has no such guard.

## Evidence

Hunthelper/hunthelper, project id 14:
- AgentRun id 18513, goal=`create_pr`, status=`failed`, `auto_pick: false`, `trigger_type: \"automatic\"`, no `initiating_user`.
- Issue #3622, created by `honeybadger[bot]`. `issue.trusted? == false` against allowlist `[\"HuntHelper\", \"viamin\"]`.
- Run was queued at 17:18:13 UTC, started shortly after, and failed 8m19s later at 17:26:32 UTC with `error_message: \"No prompt available for agent run\"`.

### Primary: reject at the queueing choke point

Add a trust check in `Activities::QueueAgentRunActivity` (which is now also the path taken by the uncommitted `CreateFollowupRunActivity`). For goals that require a trusted issue (`create_pr` at minimum — `analyze_issue` / `enhance_issue` are arguably safer because they only post comments), refuse to enqueue when `issue.untrusted?` and log structured:

```ruby
Rails.logger.info(
  message: \"queue_agent_run.untrusted_issue_skipped\",
  project_id: project.id,
  issue_id: issue.id,
  github_creator_login: issue.github_creator_login,
  goal: goal
)
```

Return a result the caller can act on (e.g. `{ skipped: true, reason: \"untrusted_issue\" }`) instead of creating the AgentRun.

### Defense-in-depth: meaningful error in `prompt_for_issue`

If the trust check still falls through for any future caller, surface why rather than \"No prompt available\":

```ruby
def prompt_for_issue
  return nil unless issue
  unless issue.trusted?
    raise UntrustedIssueError, \"Issue ##{issue.github_number} creator '#{issue.github_creator_login}' is not in the project's allowed_github_usernames\"
  end
  Prompts::BuildForIssue.call(...)
end
```

`RunAgentActivity` can translate `UntrustedIssueError` to a clear `failure_reason`.

## Why P1

This silently burns runner slots and budget on runs that cannot possibly succeed, with an error message that misdirects diagnosis. Anyone whose repo files issues via bots (honeybadger, dependabot, sentry, GitHub Actions bot, etc.) hits this every time the older automation path picks one of those issues.

viamin:
## Summary

Reject `create_pr` agent runs for untrusted issue creators at queue time instead of letting them silently fail 8+ minutes later inside `RunAgentActivity` with a misleading "No prompt available" error. This stops bot-filed issues (honeybadger, dependabot, etc.) from burning runner slots and budget on runs that cannot possibly succeed.

## Changes

- **Queue-time guard (`Activities::QueueAgentRunActivity`)**: Skip untrusted `create_pr` issues before creating an `AgentRun`, log `queue_agent_run.untrusted_issue_skipped` with project/issue/creator/goal context, and return `{ skipped: true, reason: "untrusted_issue" }` so callers can act on the result.
- **Defense-in-depth (`AgentRun#prompt_for_issue`)**: Let `Prompts::BuildForIssue::UntrustedIssueError` propagate instead of silently returning nil.
- **Clearer Temporal failure (`Activities::RunAgentActivity`)**: Translate `UntrustedIssueError` into a non-retryable `UntrustedIssue` application error, distinct from the existing `MissingPrompt` failure.
- **Follow-up runs & orchestration defaults**: Reference the workflow's known failure constants so configuration stays in sync as new failure types are added.

## Design Decisions

- **Guard scoped to `create_pr`**: `analyze_issue` and `enhance_issue` only post comments, so they remain permitted for untrusted issues. Only goals that mutate the repo require trusted creators.
- **Choke point is `QueueAgentRunActivity`, not individual callers**: `DefaultCandidateSource` already filters at auto-pick selection, but the older `GitHubPollWorkflow#queue_create_pr_run` path and the new `CreateFollowupRunActivity` both funnel through `QueueAgentRunActivity` — guarding here covers current and future callers without duplicating checks.
- **Skip vs. raise at queue time**: Returning a `skipped` result (rather than raising) lets the workflow treat untrusted issues as a normal non-event and avoids polluting Temporal failure metrics with expected skips.
- **Case-sensitivity mismatch in the allowlist is out of scope**: Tracked separately in #2177; this PR addresses only the silent-failure path.

## Related

- Root cause of "No prompt available for agent run" errors on bot-filed issues.
- Companion to #2177 (allowlist case-sensitivity).

---

Closes #2193

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -1694,7 +1694,6 @@ class AgentRun < ApplicationRecord
   # @return [String, nil] The built prompt, or nil if no issue is attached
   def prompt_for_issue
     return nil unless issue
-    return nil unless issue.trusted?
 
     Prompts::BuildForIssue.call(issue: issue, project: project, github_client: project.github_token&.client, agent_run: self)
   end
--- a/app/services/orchestration_strategies/defaults.rb
+++ b/app/services/orchestration_strategies/defaults.rb
@@ -196,21 +196,8 @@ module OrchestrationStrategies
         ],
         "planning_outcomes" => planning_mappings[:success] + planning_mappings[:failure].values,
         "parallelization_outcomes" => parallelization_mappings[:success] + parallelization_mappings[:failure].values,
-        "known_failure_types" => %w[
-          AllProvidersExhausted
-          AgentExecutionFailed
-          IssueDraftInvalid
-          McpProvisioningFailed
-          MissingPrompt
-          MissingUser
-          ContainerNotProvisioned
-          ProxyUnavailable
-          RateLimit
-        ],
-        "known_failure_classes" => %w[
-          GithubClient::RateLimitError
-          GithubClient::AuthenticationError
-        ]
+        "known_failure_types" => Workflows::AgentExecutionWorkflow::KNOWN_FAILURE_TYPES,
+        "known_failure_classes" => Workflows::AgentExecutionWorkflow::KNOWN_FAILURE_CLASSES
       }
     end
 
--- a/app/temporal/activities/create_followup_run_activity.rb
+++ b/app/temporal/activities/create_followup_run_activity.rb
@@ -18,7 +18,7 @@ module Activities
         QueueAgentRunActivity.new.execute(queue_input(agent_run, goal))
       end
 
-      {
+      response = {
         agent_run_id: agent_run.id,
         followup_agent_run_id: result[:agent_run_id],
         goal: goal,
@@ -26,6 +26,9 @@ module Activities
         duplicate: result.fetch(:duplicate, false),
         cross_goal: result.fetch(:cross_goal, false)
       }
+      response[:skipped] = true if result[:skipped]
+      response[:reason] = result[:reason] if result[:reason]
+      response
     end
 
     private
--- a/app/temporal/activities/queue_agent_run_activity.rb
+++ b/app/temporal/activities/queue_agent_run_activity.rb
@@ -3,6 +3,7 @@
 module Activities
   class QueueAgentRunActivity < BaseActivity
     activity_name "QueueAgentRun"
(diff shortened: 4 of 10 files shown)


viamin:

Issue #2117, opened 2026-05-18
Allow users to choose which runners are eligible for chat sessions


viamin:
## Problem

Users can currently configure which runners are eligible for agent runs and which runners serve as fallbacks. However, the chat feature does not offer the same level of runner selection — users cannot control which runners are available when using chat.

This creates an inconsistency: users who have configured specific runners for agent work may expect the same control over which runners handle their chat sessions.

## Expected Behavior

Add runner eligibility settings for chat sessions, mirroring the existing agent run and fallback eligibility controls:

- Users should be able to select which runners are eligible for use in chat
- The UI should follow the same pattern as the existing agent run / fallback eligibility selectors
- Chat sessions should respect the configured runner eligibility when selecting a runner

## Current Behavior

Runner eligibility can be configured for:
- Agent runs
- Fallback

But **not** for chat sessions.

## Suggested Implementation

- Add a chat eligibility field to the runner settings (similar to the existing agent run and fallback eligibility fields)
- Update the chat session runner selection logic to filter by chat-eligible runners
- Expose the chat eligibility toggle in the runner configuration UI alongside the existing eligibility controls

## Acceptance Criteria

- Users can configure which runners are eligible for chat from runner settings
- Chat sessions only use runners marked as chat-eligible
- The UI is consistent with the existing agent run and fallback eligibility controls
- Existing chat behavior is preserved as a sensible default (e.g., all currently enabled runners remain chat-eligible after migration)

viamin:
Pull request created: https://github.com/viamin/paid/pull/2258

Issue #2057, opened 2026-05-15
Use agent-harness Z.ai Coding Plan quota classification once upstream support is released


viamin:
Blocked by viamin/agent-harness#215

## Summary

Paid has a tactical local patch for Z.ai Coding Plan quota exhaustion detection so users do not see misleading outcomes when the upstream API returns messages like:

- `Weekly/Monthly Limit Exhausted. Your limit will reset at ...`
- related rate-limit / quota exhaustion wording surfaced through OpenCode, KiloCode, or other wrappers targeting `zai_coding`

Once the upstream `agent-harness` support lands, Paid should switch to using the normalized harness behavior instead of carrying provider-specific wording locally.

## Upstream dependency

Depends on viamin/agent-harness#215

That issue tracks:
- upstream classification of Z.ai Coding Plan exhaustion as quota/rate-limited
- upstream reset-time parsing from `Your limit will reset at ...`
- consistent classification across OpenCode / KiloCode / any `zai_coding` wrapper
- smoke/health-check support for the normalized error category

## Paid follow-up work

After `agent-harness` releases support for this:

1. Upgrade Paid to the released `agent-harness` version.
2. Verify `Providers::TestAgent` receives the normalized `error_category` for Z.ai Coding Plan quota exhaustion.
3. Verify runtime execution paths (`RunAgentActivity`) receive the normalized classification consistently across wrappers.
4. Remove or reduce the tactical provider-specific wording added in Paid for this case, where safe.
5. Keep only generic defense-in-depth logic that belongs in Paid UX/orchestration.
6. Re-run the relevant provider classification specs and update them to assert the upstream-normalized behavior rather than relying on Paid-specific string matching where possible.

## Acceptance criteria

- Paid is upgraded to a version of `agent-harness` that includes viamin/agent-harness#215
- Z.ai Coding Plan exhaustion is classified correctly in test-agent and runtime flows via upstream harness behavior
- Paid no longer needs provider-specific wording here except for any clearly justified fallback safety net
- dependency is removed / issue is unblocked once upstream release is available

viamin:
Pull request created: https://github.com/viamin/paid/pull/2293

Issue #2026, opened 2026-05-15
refactor(pr-review): remove legacy phase-specific breaker/reset plumbing and simplify stuck-state recovery


viamin:
## Summary

Remove the legacy PR lifecycle plumbing that grew around separate draft, review-goal, follow-up, and operational failure counters, and replace it with a simpler recovery model that is easier for both automation and operators to understand.

Depends on #2025.

The current implementation has several stateful reset and dismissal paths spread across scanner logic, issue counters, escalation dismissal, and orchestration policy inputs. Even after behavior is improved, we should follow through and delete the old structure so the code reflects the new simpler model instead of carrying both systems forever.

## Goals

- Reduce scanner and policy complexity by deleting obsolete breaker/counter paths
- Simplify how a PR gets out of a stuck or escalated state
- Prefer progress-triggered recovery over manual counter-reset semantics
- Leave the PR lifecycle code easier to reason about, test, and evolve

## Proposed approach

After #2025 lands:

- remove phase-specific failure accounting code that no longer drives lifecycle decisions
- simplify escalation dismissal/resume behavior so recovery follows from real progress signals rather than bespoke counter resets
- update orchestration-decision and escalation-policy inputs to match the new model
- clean up tests, notifications, and docs that still assume separate draft/ready failure counting regimes

## Acceptance Criteria

- [ ] Remove obsolete or no-longer-authoritative lifecycle counters/breakers and their associated plumbing where possible
- [ ] Simplify issue-level reset methods so they no longer need to juggle multiple PR failure counters for recovery
- [ ] Update escalation dismissal/resume behavior to align with the unified progress/failure model
- [ ] Remove or rewrite tests that encode the legacy draft-vs-ready failure split
- [ ] Update notification/rule logic that still assumes draft-only vs ready-only stuck semantics
- [ ] Update docs/comments to describe the new simpler lifecycle model
- [ ] Confirm the final implementation provides a straightforward answer to: “what makes a PR stuck, and what clears that state?”

## Likely touch points

- `app/temporal/activities/scan_paid_prs_activity.rb`
- `app/models/issue.rb`
- `app/services/automation/strategies/auto_continue.rb`
- `app/services/coordination/escalation_service.rb`
- `app/services/notifications/rules/*`
- specs for PR scanning, escalation, and notifications

## Out of Scope

- Broad UI redesign for PR status presentation
- Unrelated queueing/orchestration cleanup outside the PR lifecycle

## Why P1

[...]

Issue #1897, opened 2026-05-10
fix: upgrade agent-harness once --disallowedTools CLI compat fix lands upstream


viamin:
## Priority

P1 — All non-container LLM calls (`tools: :none`) are broken on Claude CLI v2.1.92+. A local gem patch is applied in the current deployment but will be lost on any `bundle install` or deployment that reinstalls the gem from rubygems.

## Problem

Upstream `agent-harness` v0.18.0 generates `--disallowedTools Agent --disallowedTools Bash ...` which the Claude CLI treats as varargs, consuming the prompt argument. This causes every `AgentHarness.send_message(..., tools: :none, ...)` call to fail with `"No LLM provider produced an issue enhancement"`.

**Affected callers (13 total):**
- `EnhanceIssueActivity`
- `AnalyzeIssueActivity`
- `DecomposeFeatureActivity`
- `AnalyzeKnowledgeGapsActivity`
- `StyleGuides::Compress`
- `StyleGuides::Extract`
- `StrategyEvolution::Mutate`
- `PromptEvolution::Mutate`
- `Models::MetaAgentSelector`
- `Llm::GeneratePrDescription`
- `Llm::GenerateIssueTitle`
- `Knowledge::Decisions::Draft`
- `AgentRuns::DiagnoseError`

## Upstream Issue

Depends on viamin/agent-harness#212

Once that fix is released in a new gem version:

1. Update `Gemfile` to pin `agent-harness` to the patched version (e.g., `~> 0.18.1` or `~> 0.19.0`)
2. Run `bundle update agent-harness`
3. Verify the local gem patch at `/usr/local/rvm/gems/default/gems/agent-harness-0.18.0/lib/agent_harness/providers/anthropic.rb:790` is no longer needed (revert the local edit)
4. Run the full test suite: `bin/rspec spec/temporal/activities/ spec/services/llm/ spec/services/style_guides/ spec/services/prompt_evolution/ spec/services/strategy_evolution/ spec/services/knowledge/decisions/`
5. Run `bin/lint` to confirm no style issues
6. Manually trigger an enhance_issue run against a test issue and verify it completes successfully

## Workaround (Current)

A local patch is applied to the installed gem changing `anthropic.rb:790` from:
```ruby
flags = tool_names.flat_map { |tool| ["--disallowedTools", tool] }
```
to:
```ruby
flags = ["--disallowedTools=#{tool_names.join(",")}"]
```

[...]

PR #2215, opened 2026-05-23
fix(auto-pick): stop parking issues as recommend_close on zero-iteration runs

viamin:
## Summary

Four of six open P1 issues on `viamin/paid` (#1897, #2026, #2057, #2117) were stranded in `paid_state=recommend_close` — invisible to the scheduler — because the no-output classifier in `HandleNoOutputIssueRunActivity` was routing zero-iteration runs to `recommend_close` whenever they incurred even a few cents of prompt-reading cost. A few cents is not evidence of real work.

Three coordinated fixes:

- **Tighten the classifier** (`handle_no_output_issue_run_activity.rb:156`) — `recommend_close` now requires `iterations > 0`. Zero-iteration runs route to `needs_input` (with provider/infrastructure-error overrides applied first).
- **Add a label-removed reset hook** (`upsert_from_github.rb`) — when a GitHub sync observes `paid-recommend-close` was removed from an issue still in `paid_state=recommend_close`, transition `paid_state` back to `new`. The existing `Issue#auto_pick_recheck_needed?` model callback re-enqueues. This turns the label into the user's "I disagree, work on this again" signal, which is how the label was already being used in practice but with no listener.
- **One-shot backfill** (`rake issues:reset_false_positive_recommend_close`) — defaults to `DRY_RUN=true`. Dry-run against the production-equivalent dev DB matches exactly the 4 stranded P1s; no collateral.

## Why now

In production, the four stranded P1s had completed runs with `iterations=0` and `cost_cents` between 3 and 34. For #1897 the agent literally noted "the work is already done." For #2026/#2057 it said "OK." — clearly giving-up signals, not "this should be closed" judgments. The user had manually removed `paid-recommend-close` from each issue 2–6 days ago expecting Paid to retry; nothing in Paid noticed.

## Test plan

- [x] Classifier spec covers the new `iterations=0 + cost>0` → `needs_input` path
- [x] Five new specs cover the label-removed hook (positive, negative, label still present, project override, auto-pick disabled)
- [x] Full upsert + classifier + adjacent specs (569 examples) pass
- [x] Backfill dry-run identifies exactly the 4 stranded P1s

## Followups (out of scope)

- `paid-needs-input` does not have a symmetric label-removed hook. False-positives shifted from `recommend_close` to `needs_input` are still parked. Different semantics (needs_input = "provide info" not "push back") makes the symmetric hook potentially loop-prone — worth separate UX discussion.

### DIFF
--- a/lib/tasks/issues.rake
+++ b/lib/tasks/issues.rake
@@ -32,7 +32,6 @@ namespace :issues do
         next if dry_run
 
         issue.update!(paid_state: "new")
-        Issues::ReenqueueEligibleJob.perform_later(issue.id) if issue.project.auto_pick_enabled?
       end
 
       puts dry_run ? "Dry run only. Re-run with DRY_RUN=false to apply." : "Done."
--- a/lib/tasks/issues.rake
+++ b/lib/tasks/issues.rake
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+namespace :issues do
+  desc "Reset paid_state for issues parked in recommend_close by the false-positive classifier (iterations=0 + cost>0). Use DRY_RUN=false to apply."
+  task reset_false_positive_recommend_close: :environment do
+    dry_run = ENV.fetch("DRY_RUN", "true") != "false"
+    candidates = []
+
+    # Goal allowlist mirrors the workflow path that routes through
+    # HandleNoOutputIssueRunActivity: any issue-based run with no source
+    # PR can reach the classifier. analyze_issue takes a different path
+    # (paid_state: "analyzed") and never sets recommend_close, but listing
+    # both create_pr and enhance_issue here avoids missing future goals
+    # added to the no-output classifier path.
+    classifier_goals = %w[create_pr enhance_issue]
+
+    TenantContext.with_system_access do
+      Issue.where(paid_state: "recommend_close", is_pull_request: false, github_state: "open").find_each do |issue|
+        last = issue.agent_runs
+          .where(goal: classifier_goals, status: "completed")
+          .order(created_at: :desc)
+          .first
+        next unless last
+        next unless last.iterations.to_i.zero?
+
+        candidates << [ issue, last ]
+      end
+
+      puts "Found #{candidates.size} issue(s) with iterations=0 recommend_close runs (DRY_RUN=#{dry_run})"
+      candidates.each do |issue, run|
+        puts "  ##{issue.github_number} project=#{issue.project.full_name} run=#{run.id} iterations=#{run.iterations} cost_cents=#{run.cost_cents}"
+        next if dry_run
+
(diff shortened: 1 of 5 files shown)


lib/tasks/issues.rake
```diff
@@ -0,0 +1,41 @@
...
+        candidates << [ issue, last ]
+      end
+
+      puts "Found #{candidates.size} issue(s) with iterations=0 recommend_close runs (DRY_RUN=#{dry_run})"
+      candidates.each do |issue, run|
+        puts "  ##{issue.github_number} project=#{issue.project.full_name} run=#{run.id} iterations=#{run.iterations} cost_cents=#{run.cost_cents}"
+        next if dry_run
+
+        issue.update!(paid_state: "new")
+        Issues::ReenqueueEligibleJob.perform_later(issue.id) if issue.project.auto_pick_enabled?
```
viamin: This `perform_later` call is redundant. The `issue.update!(paid_state: "new")` on line 34 will trigger the `after_update_commit :enqueue_self_if_became_auto_pick_eligible` callback in the `Issue` model (line 65 of `issue.rb`), which already enqueues `ReenqueueEligibleJob` when `paid_state` transitions to `"new"` and `auto_pick_enabled?` is true.

The result is the job being enqueued twice per issue. Not harmful (presumably idempotent), but unnecessary.

```suggestion
        Issues::ReenqueueEligibleJob.perform_later(issue.id) unless issue.project.auto_pick_enabled?
```

Alternatively, just delete this line entirely and let the model callback handle it in all cases — it already has the same `auto_pick_enabled?` guard.

viamin (commented):
Solid fix for the false-positive `recommend_close` classifier. The three-pronged approach (tighten classifier, add label-removal hook, backfill rake task) is well-structured and the test coverage is thorough.

One minor issue in the backfill rake task — see inline comment.

viamin:
## Agent Update

Removed the redundant `Issues::ReenqueueEligibleJob.perform_later` from [lib/tasks/issues.rake](/workspace/lib/tasks/issues.rake), so the backfill now relies on `Issue#enqueue_self_if_became_auto_pick_eligible` as the single re-enqueue path. That addresses both unresolved review threads directly and avoids double-enqueueing when `paid_state` transitions to `new`.

Validation passed with `DB_HOST=paid-svc-a3-s1-postgres DB_USERNAME=agent DB_PASSWORD=agent`: `bin/rails db:prepare`, `bundle exec rubocop`, `bundle exec rspec`, and the commit hook’s own lint/test run. The full suite finished at `13817 examples, 0 failures, 7 pending`. Commit: `53555c0de7dcf88e306df59da08561f0cd5f172a` (`fix(auto-pick): avoid duplicate reenqueue in reset task`).

viamin:

PR #2216, opened 2026-05-23
fix(auto-pick): unblock cross-project external deps when target is closed

viamin:
## Summary

`Issue.ready_for_work`'s `blocked_by_external` filter excluded every issue with a cross-repo dependency reference (`owner/repo#number`) **unconditionally** — regardless of whether the referenced issue was open, closed, or even resolvable. The parser already promotes cross-repo refs to local `IssueDependency` rows when both repos are synced into the same account at parse time, but deps created before the sibling repo was synced stayed permanently external — and the unconditional exclusion meant a closed external issue could not unblock its dependent.

## Change

Extracted the resolution rule as `IssueDependency.still_blocking_external_for_account`:

- LEFT JOIN against projects in the same account on `(LOWER(owner), LOWER(name))`
- LEFT JOIN the matching `Issue` on `github_number`
- Block the dependent only when the join resolves to nothing (no sibling project synced, or target issue not yet pulled) or when the resolved target is still open
- Mirror `blocked_by_local_open`'s `recommend_close` exception so an external dep parked at `recommend_close` is treated as effectively satisfied

Applied the new rule consistently in all three sites that previously gated on external deps:

1. `Issue.ready_for_work` (scheduler auto-pick filter)
2. `Issue#blocking_external_dependencies` (used by `Issue#ready_to_work?`)
3. `Issue.lifecycle_statuses` (project-page `:blocked`/`:eligible` badge)

Without this consistency, the same issue could appear `:blocked` in the UI while being eligible to the scheduler, or vice versa.

## Index

Adds a functional index on `projects (account_id, LOWER(owner), LOWER(name))`. The existing case-sensitive indexes don't cover the `LOWER()` comparison, so without this the new LEFT JOIN would fall back to a sequence scan of projects per external dep on every auto-pick tick. Created `CONCURRENTLY` since auto-pick queries the table continuously.

## What stays blocked

Conservative default preserved for genuinely unobservable refs — e.g. `rails/rails#12345` from outside Paid's account, where there's no way to observe the target's state. Users who want those to unblock can sync the source repo into Paid.

## Test plan

- [x] 9 new `ready_for_work` specs cover sibling-project resolution (closed/open/recommend_close target, missing issue, case-insensitive owner/repo, different-account)
- [x] New `#ready_to_work?` specs for the same matrix
- [x] New `.lifecycle_statuses` spec confirming UI badge matches scheduler decision
- [x] Full test suite (5814 examples) passes
- [x] Verified against production-equivalent dev DB: only one truly-external dep (`owner/repo#456` placeholder) — correctly stays blocked since no sibling project exists

### DIFF
--- a/app/models/issue.rb
+++ b/app/models/issue.rb
@@ -98,9 +98,14 @@ class Issue < ApplicationRecord
       .where(issues: { project_id: project.id })
       .select(:issue_id)
 
+    # External deps (owner/repo#number) block conservatively when we have
+    # no visibility into the target's state. They unblock once a matching
+    # Issue is observable in any project of the same account — typically
+    # because both repos are synced into Paid. See
+    # IssueDependency.still_blocking_external_for_account for the rule.
     blocked_by_external = IssueDependency
+      .still_blocking_external_for_account(project.account_id)
       .joins(:issue)
-      .where.not(depends_on_owner: nil)
       .where(issues: { project_id: project.id })
       .select(:issue_id)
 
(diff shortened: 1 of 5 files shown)


viamin:

PR #2219, opened 2026-05-23
feat: Add symmetric label-removed hook for paid-needs-input

### LINKED ISSUE #2217: Add symmetric label-removed hook for paid-needs-input
## Context

Follow-up from #2215, which added a label-removed reset hook for `paid-recommend-close`: when a user removes that label from an issue still in `paid_state=recommend_close`, the sync transitions `paid_state` back to `new` and the existing model callback re-enqueues auto-pick. The label is treated as the user's "I disagree, work on this again" signal.

After that fix, zero-iteration false positives that previously routed to `recommend_close` now route to `needs_input`. There is no symmetric label-removed hook for `paid-needs-input`, so any false-positives that land there are still parked indefinitely — the user has no equivalent "I disagree" signal.

## Why P1

This is the same stranding class of bug that #2215 fixed. Without a symmetric hook, the classifier change in #2215 partially shifts the problem rather than eliminating it. Users who manually remove `paid-needs-input` expecting Paid to retry will see nothing happen, mirroring the original incident behavior on `paid-recommend-close`.

## Scope / open questions

The semantics differ from `recommend_close`:

- `recommend_close` = agent thinks the work is done / not needed; removing the label is a clean "push back, try again" signal.
- `needs_input` = agent is blocked waiting for information; removing the label without supplying that information could re-enqueue immediately, hit the same block, and re-park — a tight loop.

Worth a brief UX discussion before implementing. Options to consider:

1. Symmetric hook gated on a body/comment change since the run that parked the issue (i.e., "user provided new info").
2. Symmetric hook with a cooldown / max retry count to bound loop risk.
3. Symmetric hook unconditional — accept that the user removing the label is itself the signal, same as `recommend_close`.

## Acceptance

- `paid_state=needs_input` issues transition back to `new` (or equivalent re-enqueueing state) when `paid-needs-input` is removed, under whichever guard we settle on.
- Specs cover the chosen behavior, including the loop-prevention path if applicable.
- Behavior parity with the `recommend_close` hook where it makes sense.

## References

- PR #2215 — adds the `recommend_close` label-removed hook; followup listed in its description.
- `app/services/issues/upsert_from_github.rb` — where the symmetric hook would live.
- `Issue#auto_pick_recheck_needed?` — existing re-enqueue callback.

viamin:
## Summary

Adds a symmetric label-removed hook for `paid-needs-input` so users can signal "I disagree, retry this" by removing the label, mirroring the existing behavior for `paid-recommend-close` (#2215). Without this, false-positive issues routed to `needs_input` after the #2215 classifier change remain stranded indefinitely with no user-driven recovery path.

## Changes

- **Activities** (`app/temporal/activities/fetch_issues_activity.rb`): Added `detect_needs_input_label_removals`, invoked during GitHub issue sync. When `paid-needs-input` appears in `removed_labels` for an open, non-PR issue in `paid_state=needs_input`, it transitions `paid_state` back to `new`. The existing `Issue#auto_pick_recheck_needed?` callback then re-enqueues auto-pick. The lookup uses `project.label_for_stage("needs_input")` to honor custom label mappings, falling back to the `PAID_NEEDS_INPUT_LABEL` constant.
- **Specs** (`spec/temporal/activities/fetch_issues_activity_spec.rb`): 6 new tests covering the basic transition, auto-pick re-enqueue, and guard clauses for closed issues, PRs, mismatched `paid_state`, and label-still-present cases.

## Design Decisions

- **Unconditional transition (option 3 from the issue).** Treating the user's label removal as the signal — same semantics as `recommend_close` — keeps the two hooks symmetric and the implementation minimal. Options 1 (gate on new body/comment activity) and 2 (cooldown / retry cap) were considered to bound tight-loop risk where `needs_input` was caused by a real information gap; rejected because the immediate re-run will simply re-park the issue with the same blocker, which is recoverable and visible, not destructive. If loop churn shows up in practice, a cooldown can be layered on later without changing the entry point.
- **Guard clauses up front.** Closed/PR/wrong-state issues are filtered before the transition to keep the hook scoped narrowly and avoid touching records outside its concern.

---

Closes #2217

### DIFF
--- a/app/temporal/activities/fetch_issues_activity.rb
+++ b/app/temporal/activities/fetch_issues_activity.rb
@@ -51,6 +51,9 @@ module Activities
         enhance_issue_rechecks = enhance_issue_result[:rechecks]
         sync_changed ||= enhance_issue_result[:changed]
 
+        needs_input_changed = detect_needs_input_label_removals(project, synced_issues)
+        sync_changed ||= needs_input_changed
+
         closed_count = close_stale_issues(project, github_issues, truncated: truncated, incremental: incremental)
         sync_changed ||= closed_count.positive?
 
@@ -394,6 +397,31 @@ module Activities
           paid_state: "needs_input"
         )
       end
+    end
+
+    def detect_needs_input_label_removals(project, synced_issues)
+      needs_input_label = project.label_for_stage("needs_input") ||
+        Activities::HandleNoOutputIssueRunActivity::PAID_NEEDS_INPUT_LABEL
+      changed = false
+
+      synced_issues.each do |issue_data|
+        next unless Array(issue_data[:removed_labels]).include?(needs_input_label)
+
+        issue = project.issues.find(issue_data[:id])
+        next if issue.is_pull_request? || issue.github_state == "closed" || issue.paid_state != "needs_input"
+
+        issue.update!(paid_state: "new")
+        changed = true
+
+        logger.info(
+          message: "github_sync.needs_input_label_removed",
+          project_id: project.id,
+          issue_id: issue.id,
+          issue_number: issue.github_number
+        )
+      end
+
+      changed
     end
 
     # Parses dependency and parent/child relationships from issue comments.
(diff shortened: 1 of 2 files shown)


viamin:

PR #2221, opened 2026-05-23
fix(devcontainer): seed valid JSON in opencode auth.json initializer

viamin:
## Summary

- `.devcontainer/devcontainer.json`'s `initializeCommand` ran `touch ~/.config/opencode/auth.json`, creating an empty file on the host that is then bind-mounted into the container.
- opencode's provider loader does `JSON.parse` on `auth.json`; an empty file throws `SyntaxError: JSON Parse error: Expected ']'`, which surfaces as failures in both `config.providers` and `provider.list`.
- The `touch` exists only to satisfy the moby/26702 workaround (so Docker doesn't create the bind target as a directory). Replaced with `[ -s … ] || echo '{}' > …` so the file is valid JSON when first created, while existing host credentials are left untouched.

## Test plan

- [ ] Rebuild the devcontainer on a host where `~/.config/opencode/auth.json` does not exist and confirm opencode no longer errors on `provider.list` / `config.providers`.
- [ ] Rebuild on a host where `~/.config/opencode/auth.json` already contains real provider auth and confirm the file is not overwritten.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -48,7 +48,7 @@
   },
   // Ensure host paths that are bind-mounted as files (not directories) exist before
   // Docker creates them as directories. See: https://github.com/moby/moby/issues/26702
-  "initializeCommand": "mkdir -p ~/.config/opencode && touch ~/.config/opencode/auth.json",
+  "initializeCommand": "mkdir -p ~/.config/opencode && { [ -s ~/.config/opencode/auth.json ] || echo '{}' > ~/.config/opencode/auth.json; }",
   "mounts": [
     // Git configuration
     "source=${localEnv:HOME}/.gitconfig,target=/home/vscode/.gitconfig,type=bind,consistency=cached",


viamin:
## Agent Update

Let me set up the project and investigate the CI failure.

Let me look at the BranchStorage token configuration and check if there's an issue with how tokens are handled.

Let me look more carefully at the relationship between the workflow files and the recent refactoring. The publish workflow was recently heavily modified. Let me check if there's a missing step that the capture workflow has but the publish workflow doesn't.

Let me try running the test suite to make sure the codebase is healthy, and then focus on making any necessary fixes.

Let me look at one specific potential issue - the `GITHUB_TOKEN` handling between the resolve and publish steps.

Let me run a focused set of tests to verify the codebase health, then commit the change.

All 252 screenshot-related tests pass. Let me now look at the workflow issue from one more angle - the `pr-screenshots.yml` workflow's capture job name in the GitHub API.

The local tests all pass. Let me now verify the devcontainer.json change is correct and commit. The CI failure in the `publish` job appears to be an infrastructure issue that can't be diagnosed without the actual CI logs. The PR's code change (devcontainer.json only) is unrelated to the `pr-screenshots-publish.yml` workflow.

Let me verify the shell command in the `initializeCommand` works correctly:

viamin:

PR #2246, opened 2026-05-23
fix(auto-pick): priority label matching is case-sensitive

### LINKED ISSUE #2233: fix(auto-pick): priority label matching is case-sensitive
## Problem

The priority label matching in `DefaultCandidateSource.priority_label_order_node` uses `jsonb_build_array` with the configured label name, and the `@>` containment operator is case-sensitive. If a GitHub issue is labeled `"p1"` (lowercase) but the project configures priority labels as `"P1"`, the issue won't match and gets the lowest priority rank (4) instead of rank 1.

This doesn't prevent queueing — the issue still gets picked up — but it defeats the purpose of priority labels: a P1 issue competes equally with unlabeled issues and may be starved by older low-priority work.

## Location

`app/services/automation/strategies/auto_pick/default_candidate_source.rb:233-237` — the `@>` operator with `jsonb_build_array`.

The same case sensitivity applies to the queue dispatch ordering in `AgentRun.QUEUE_ORDER` (`app/models/agent_run.rb:961-969`).

## Suggested Fix

Normalize priority label matching to be case-insensitive. Options:
1. Use `LOWER(labels::text)` with a text containment check instead of jsonb `@>`
2. Normalize label casing on sync (upsert all labels to a canonical form)
3. Add a migration to normalize existing labels in the DB

Option 1 is the most localized fix. Option 2 is the most robust long-term.

## Context

viamin:
## Summary

Priority labels configured on a project (e.g. `P1`) failed to match GitHub issue labels with different casing (e.g. `p1`), causing high-priority issues to be ranked as lowest-priority (rank 4) and starved behind older low-priority work. This change makes priority-label matching case-insensitive across both the auto-pick candidate ordering and the queue dispatch ordering, so labels are compared by canonical case at the SQL layer.

## Changes

- **Services**: Updated `Automation::Strategies::AutoPick::DefaultCandidateSource#priority_label_order_node` to compare labels case-insensitively instead of using case-sensitive `jsonb` containment via `@>` with `jsonb_build_array`.
- **Models**: Aligned `AgentRun::QUEUE_ORDER` to apply the same case-insensitive comparison so dispatch ordering matches candidate selection, and updated the Ruby-side `queue_priority_tier` logic so UI/reporting agrees with queue behavior.
- **Tests**: Added regressions in `spec/services/automation/strategies/auto_pick/default_candidate_source_spec.rb` and `spec/models/agent_run_spec.rb` covering label-case mismatches.

## Design Decisions

- Chose the localized SQL-layer normalization (option 1 from the issue) over normalizing labels on sync or backfilling the DB. It fixes the immediate priority-starvation symptom without touching the label storage format or requiring a migration, and keeps the change reviewable. A broader canonicalization on sync remains viable as future work if more label comparisons need the same treatment.
- Kept the fix in both the candidate source and `AgentRun::QUEUE_ORDER` so the two ordering paths stay consistent — divergence between them would reintroduce the same class of bug for in-flight runs.

---

Closes #2233

### DIFF
--- a/app/models/agent_run.rb
+++ b/app/models/agent_run.rb
@@ -808,8 +808,8 @@ class AgentRun < ApplicationRecord
 
   # When a run has both an associated issue AND a source PR, labels from
   # both records are considered (highest configured tier wins). Label
-  # name matching is case-sensitive on both the Ruby and SQL sides; an
-  # issue tagged "p1" will not match a configured tier of "P1".
+  # name matching is case-insensitive on both the Ruby and SQL sides, so
+  # an issue tagged "p1" still matches a configured tier of "P1".
   def compute_label_priority_tier
     return nil unless project
 
@@ -822,7 +822,7 @@ class AgentRun < ApplicationRecord
     Project::PRIORITY_TIERS.find do |tier|
       label_name = project.priority_label_for(tier)
       next false if label_name.blank?
-      label_sources.any? { |labels| labels.include?(label_name) }
+      label_sources.any? { |labels| labels.any? { |label| label.casecmp?(label_name) } }
     end
   end
   private :compute_label_priority_tier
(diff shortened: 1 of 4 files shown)


viamin:

PR #2258, opened 2026-05-23
feat: Allow users to choose which runners are eligible for chat sessions

### LINKED ISSUE #2117: Allow users to choose which runners are eligible for chat sessions

viamin:
## Summary

Adds chat runner eligibility so users can control which runners are available for chat sessions, mirroring the existing agent-run and fallback eligibility controls. This closes the inconsistency where chat sessions ignored user runner preferences.

## Changes

- **Database**: New migration adds an `enabled_for_chat` flag to runners, defaulting to `true` so existing runners stay chat-eligible after deploy.
- **Models**: Runner exposes the new `enabled_for_chat` attribute alongside the existing eligibility flags.
- **Services**: Chat session runner selection now filters to `enabled_for_chat` runners and explicitly rejects a chosen runner that is not chat-eligible.
- **UI**: Runner settings form surfaces the chat eligibility toggle next to the agent-run and fallback toggles, following the same pattern.

## Design Decisions

- **Default to `true` on migration** to preserve current behavior — all enabled runners remain chat-eligible until an operator opts a runner out, satisfying the acceptance criterion that existing chat behavior is unchanged.
- **Explicit rejection** when a caller passes a non-chat-eligible runner, rather than silently falling back, so misconfigurations surface loudly instead of routing chat to an unintended runner.

## Operational Notes

- Requires running the migration on deploy; no backfill is needed because the column defaults to `true`.

## Screenshots

_Please attach before/after screenshots of the runner settings UI showing the new chat eligibility toggle alongside the existing controls._

---

Closes #2117

### DIFF
--- a/app/controllers/runners_controller.rb
+++ b/app/controllers/runners_controller.rb
@@ -168,7 +168,7 @@ class RunnersController < ApplicationController
 
   def runner_params
     raw_params = params[resource_param_key] || params.fetch(:runner)
-    permitted = [ :enabled_for_agent_runs, :enabled_for_fallback, :name, :fallback_role, :agent_co_author_trailer, :weight ]
+    permitted = [ :enabled_for_agent_runs, :enabled_for_chat, :enabled_for_fallback, :name, :fallback_role, :agent_co_author_trailer, :weight ]
     if action_name == "create"
       permitted.push(:runner_key, :provider_key, :auth_type, :provider_api_key_id)
     end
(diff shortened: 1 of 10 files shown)


viamin:

PR #2262, opened 2026-05-24
fix(github-sync): backfilled issues from reconciliation not evaluated for auto-pick

### LINKED ISSUE #2230: fix(github-sync): backfilled issues from reconciliation not evaluated for auto-pick
## Problem

`FetchIssuesActivity#backfill_open_issues` calls `sync_issue` for issues that are open on GitHub but missing from the local DB. However, it does not pass `eager_queue_enabled` or `eligible_issues`, so these newly synced issues are never checked for auto-pick eligibility during that sync cycle.

Since these issues have not been updated on GitHub (they were just discovered), they will not appear in subsequent incremental syncs either. They remain at `paid_state: "new"` but unevaluated until either:
1. Something updates them on GitHub (triggering an incremental sync)
2. The periodic `AutoPickEligibilitySweepJob` picks them up (every 15 min)

## Location

`app/temporal/activities/fetch_issues_activity.rb:806-822` — `backfill_open_issues` calls `sync_issue(project, github_issue)` without passing `eligible_issues`.

## Suggested Fix

Thread the `eligible_issues` array through to `backfill_open_issues` so newly discovered issues are added to the candidate list for the current sync cycle:

```ruby
def backfill_open_issues(project, client, open_issue_numbers, eligible_issues: nil)
  # ...
  missing_numbers.each do |number|
    github_issue = client.issue(project.full_name, number)
    sync_issue(project, github_issue, eligible_issues: eligible_issues)
  end
end
```

## Context

viamin:
## Summary

Backfilled issues discovered during reconciliation are now evaluated for auto-pick in the same incremental sync cycle, instead of waiting up to 15 minutes for the periodic eligibility sweep. This closes a gap where P1 issues that existed on GitHub but were missing locally would remain unevaluated at `paid_state: "new"` after being synced.

## Changes

- **Activities**: Threaded `eager_queue_enabled` and `eligible_issues` through `reconcile_open_issues` into `backfill_open_issues` in `app/temporal/activities/fetch_issues_activity.rb`, so `sync_issue` receives the same eligibility-tracking context used by the main incremental sync path.
- **Tests**: Added a regression spec in `spec/temporal/activities/fetch_issues_activity_spec.rb` covering the incremental auto-pick case for a reconciliation-backfilled issue.

## Design Decisions

- Kept the fix surgical — passed the existing parameters through the call chain rather than restructuring reconciliation or introducing a separate post-sync eligibility pass. The candidate list already exists for this purpose; backfill just wasn't contributing to it.
- Relies on the existing `AutoPickEligibilitySweepJob` as a backstop, so issues discovered outside an incremental sync window remain covered.

---

Closes #2230

### DIFF
--- a/app/temporal/activities/fetch_issues_activity.rb
+++ b/app/temporal/activities/fetch_issues_activity.rb
@@ -62,7 +62,12 @@ module Activities
           stale_pr_count = stale_pr_result[:closed_count]
           sync_changed ||= stale_pr_result[:changed]
 
-          stale_issue_result = reconcile_open_issues(project, client)
+          stale_issue_result = reconcile_open_issues(
+            project,
+            client,
+            eager_queue_enabled: eager_queue_enabled,
+            eligible_issues: eligible_issues
+          )
           stale_issue_count = stale_issue_result[:closed_count]
           sync_changed ||= stale_issue_result[:changed]
         end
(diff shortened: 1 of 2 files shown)


viamin:

PR #2292, opened 2026-05-25
fix: Chat interface UI is cluttered, repetitive, and breaks on mobile

### LINKED ISSUE #2283: fix: Chat interface UI is cluttered, repetitive, and breaks on mobile
## Summary

The chat interface has several usability and layout issues that make it feel busy and harder to use than it should be.

### 1. Duplicate runner and model selectors
Runners and models are selectable in multiple places in the UI. This adds visual clutter and creates confusion about which selector is authoritative.

### 2. System prompt takes up too much screen space
The system prompt occupies a large portion of the visible area, pushing the actual conversation content down and reducing the usable chat space.

### 3. Chat interface overflows on mobile
The chat interface is wider than the screen on mobile devices, which makes the project selector inaccessible. Users cannot select a project on smaller screens.

### 4. Empty chats are saved to archive
Chats appear to be saved in the archive even when no messages have been sent. Only chats with at least one sent message should be persisted.

## Expected behavior

- Each control (runner, model) should appear in exactly one place
- The system prompt should be collapsible or otherwise take up less default space
- The chat layout should be responsive and fully usable on mobile screens
- Chats should only be archived after at least one message is sent

viamin:
## Summary

This change streamlines the chat experience so the interface is easier to use, works cleanly on mobile, and only preserves conversations that contain actual user activity. It removes redundant controls, reduces default visual noise from system prompts, fixes responsive layout issues that blocked key actions on small screens, and prevents empty chats from being archived.

## Changes

- **UI**
  - Removed the duplicate runner/model configuration bar from the chat page so each control appears in a single authoritative location.
  - Made system messages collapsible by default, reducing the amount of vertical space consumed before a conversation starts.
  - Extracted shared message bubble markup into a dedicated partial to keep the message rendering path simpler after introducing the system-message toggle.

- **Responsive layout**
  - Updated the chat layout containers to prevent horizontal overflow on mobile.
  - Ensured the header/content column can shrink correctly on smaller screens so controls like the project selector remain accessible.

- **Chat session lifecycle**
  - Changed chat session closing behavior so sessions with no user messages are destroyed instead of archived.
  - Preserved normal close/archive behavior for sessions that contain actual user conversation.

- **Tests and maintenance**
  - Updated close-service specs and related chat tests to reflect the new empty-session lifecycle behavior.
  - Adjusted view structure to satisfy template linting constraints introduced by the collapsible system prompt UI.

## Design Decisions

- The duplicate runner/model selectors were removed rather than synchronized across multiple surfaces. Keeping one authoritative control reduces ambiguity and avoids state-consistency problems between parallel selectors.
- System prompts use native `<details>`/`<summary>` behavior, which keeps the implementation lightweight and accessible while defaulting to a less cluttered layout.
- Empty sessions are deleted at close time instead of being filtered later in archive views. This keeps persistence rules aligned with product expectations and avoids accumulating meaningless records.
- The message bubble markup was extracted into a partial to accommodate the new conditional rendering without fighting ERB/HTML linting rules.

## Operational Concerns

- No database migration or infrastructure change is required.
- Deployment risk is low and concentrated in chat UI rendering and chat session close behavior.

## Screenshots

Attach before/after screenshots of:
- Chat page with duplicate selectors removed
- Collapsed system prompt treatment
- Mobile layout showing the project selector remains accessible

---

Closes #2283

### DIFF
--- a/app/services/chat_sessions/close.rb
+++ b/app/services/chat_sessions/close.rb
@@ -20,16 +20,12 @@ module ChatSessions
     def call
       validate!
 
-      if chat_session.messages.where(role: "user").none?
-        cleanup_workspace if chat_session.mode == "workspace"
-        chat_session.destroy!
-        return chat_session
-      end
+      return destroy_empty_session unless user_messages?
 
       ActiveRecord::Base.transaction do
         compute_totals
-        transition_to_closed
-        cleanup_workspace if chat_session.mode == "workspace"
+        cleanup_workspace_resources if workspace_session?
+        transition_to_closed(workspace_cleanup_attributes)
       end
 
       chat_session
@@ -41,6 +37,20 @@ module ChatSessions
       return if %w[active idle].include?(chat_session.status)
 
       raise ArgumentError, "chat session must be active or idle to close (current: #{chat_session.status})"
+    end
+
+    def destroy_empty_session
+      cleanup_workspace_resources if workspace_session?
+      chat_session.destroy!
+      chat_session
+    end
+
+    def user_messages?
+      chat_session.messages.where(role: "user").exists?
+    end
+
+    def workspace_session?
+      chat_session.mode == "workspace"
     end
 
     def compute_totals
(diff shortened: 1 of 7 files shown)


viamin:
## Agent Update

Addressed both unresolved review threads and committed the fixes as `9965c429` (`fix(chat): address review feedback on session close`).

In [app/views/chat_messages/_message.html.erb](/workspace/app/views/chat_messages/_message.html.erb:2), the system-message `<details>` now has `class="group"` so the `group-open:rotate-180` chevron state works. In [app/services/chat_sessions/close.rb](/workspace/app/services/chat_sessions/close.rb:20), empty sessions now go through a dedicated destroy path that still performs resource cleanup logging but avoids the redundant `container_id` / `workspace_volume` updates before `destroy!`; for non-empty workspace sessions, cleanup logging remains and the row is cleared in the close update.

Verification passed with `bundle exec rubocop`, targeted chat specs, and a full `bundle exec rspec` run (`13622 examples, 0 failures, 7 pending`). `bin/rails db:prepare` could not be used directly with the provided default DB env because the supplied `DATABASE_URL` points at the guarded superuser role and the secondary DB config falls back to a local socket, so I ran verification against the existing `paid_test` database with explicit env overrides for the hook and manual test runs.

viamin:

PR #2320, opened 2026-05-27
fix(runners): improve recovery for circuit-breaker, credit, and large-diff failures

viamin:
## Summary

Four targeted fixes surfaced by analyzing the past ~30 agent-run failures, all clustering on real user-impacting recovery gaps:

- **Bug 1 — All-circuit-open requeues instead of failing terminally.** When every runner attempt is skipped because circuits are open (zero real attempts), the run now transitions to `rate_limited` with an earliest-recovery `reset_at` (parallels the existing all-rate-limited path) so the failure-recovery decision job can retry instead of leaving the agent run terminally `failed`.
- **Bug 2 — Credit/quota errors mark the runner unavailable + raise the existing user notification.** Insufficient-balance / out-of-credits responses now call `state.mark_rate_limited!(reset_at: 1.hour.from_now)` so subsequent runs skip the empty-balance runner instantly (instead of burning ~3–4s/run of preflight). The existing `Notifications::Rules::RunnerQuotaExhausted` rule fires automatically after its 15-minute warning threshold — no new notification code.
- **Bug 5 — Auto-commit file cap relaxed for `focus=merge_conflict` runs (100 → 500).** Real merge-conflict resolutions on this codebase legitimately touch 200+ files; the artifact-pattern check (forbidden directories + binary extensions) remains the real safety net. Two successful agent runs were being terminally killed in the data with 225- and 232-file legitimate diffs.
- **Codex bare-key subscription guard.** `selected_runner_runtime` already skips `--model` for Codex subscription auth via `runner_entry_for(...)`, but `runner_entry_for` short-circuits on non-routing-key candidates. Fallback chains pass the bare key `"codex"` rather than `"runner:NN"`, so the guard was missing — leaking stale `tier_model_ids` (e.g. `gpt-4o`, which the `/v1/responses` subscription endpoint rejects) into `--model`. Now a dedicated bare-key lookup (memoized per user) catches that case.

## Code review fixes applied

Self-review surfaced these and they're included:
- `circuit_recovery_at` defaults `opened_at` to `Time.current` when state is `circuit_open?` but `circuit_opened_at` is nil, so reset never returns nil.
- Final fallback for `rate_limit_reset_at` when every runner_state lookup misses, so the requeue path always has a recovery point.
- "No compatible runner available: X is the only runner compatible with Y" message now also fires for circuit-open (was rate-limit-only) so single-pinned-runner users get the actionable message in both cases.
- Renamed `raise_preflight_credit_exhausted!` → `raise_credit_exhausted!` (called from preflight *and* main-execution paths); inlined neutral `agent_execution.credit_exhausted` log tag.

## Test plan

- [x] `bin/rspec spec/temporal/activities/run_agent_activity_spec.rb spec/services/containers/git_operations_spec.rb` — 371 examples, 0 failures, including:
  - new regression: agent_run goes to `rate_limited` (not `failed`) when every runner is circuit-open
  - new regression: preflight credit error marks runner state rate-limited with `INSUFFICIENT_CREDITS_BACKOFF` reset
  - new regression: `selected_runner_runtime` returns nil for Codex subscription auth referenced by bare key (`"codex"`)
  - new regression: bare-key Codex subscription lookup is memoized per user
  - new regression: merge_conflict focus allows 250-file diffs; still rejects 501-file diffs
  - updated: existing exit-0 quota-error test now asserts rate-limited handling instead of terminal fail
- [x] `bin/rubocop` clean on all four touched files

## Follow-ups (out of scope)

The structural blind spot that let Paid's pattern detector miss "Codex `gpt-4o` not supported" entirely is being tracked as a separate stack of three P2 issues (linked below once filed). Filing those separately because the design is general — making the diagnostic engine read `runners_attempted`, replacing the regex classifier with a ZFC-style LLM diagnosis, and adding a bounded auto-apply remediation framework — and shouldn't be coupled to this hotfix.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -98,16 +98,15 @@
     // to scripts/extract-provider-install-contract.rb. This means versions and
     // install commands stay in sync with the agent image without separate pins.
     // The script handles npm, uv_tool, and artifact-based install sources
-    // transparently. All npm-based installs get --ignore-scripts enforced.
+    // transparently and now evaluates the contract's npm install command
+    // directly so provider-specific postinstall handling can flow through.
     "codex": "bash scripts/install-from-contract.sh codex",
     "gemini-cli": "bash scripts/install-from-contract.sh gemini",
     "github-copilot-cli": "bash scripts/install-from-contract.sh copilot",
     "kilocode": "bash scripts/install-from-contract.sh kilocode",
     "cursor-agent": "bash scripts/install-from-contract.sh cursor",
     "aider": "bash scripts/install-from-contract.sh aider",
-    // opencode is intentionally NOT converted to contract-driven install yet.
-    // Tracked in #2297 — keeping the hardcoded npm entry until that lands.
-    "opencode": "npm i -g --ignore-scripts opencode-ai@latest",
+    "opencode": "bash scripts/install-from-contract.sh opencode",
     // Dev tools
     "markdownlint": "npm install -g --ignore-scripts markdownlint-cli",
     "git-curate": "env PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:$PATH gem install git_curate",
--- a/app/controllers/accounts/compliance_dashboards_controller.rb
+++ b/app/controllers/accounts/compliance_dashboards_controller.rb
@@ -1,0 +1,96 @@
+# frozen_string_literal: true
+
+module Accounts
+  class ComplianceDashboardsController < ApplicationController
+    before_action :set_tenant_setting
+    before_action :load_compliance_dashboard, only: :show
+
+    def show
+      authorize current_account, :show?
+    end
+
+    def update
+      authorize current_account, :update?
+
+      ActiveRecord::Base.transaction do
+        @tenant_setting.deployment_assurance_configuration = deployment_assurance_params
+        @tenant_setting.save!
+
+        if @tenant_setting.saved_changes.except("updated_at").any?
+          Accounts::RecordActivity.call(
+            account: current_account,
+            actor: current_user,
+            action: "compliance.assurance_updated",
+            subject: @tenant_setting,
+            metadata: { changed_fields: [ "deployment_assurance" ] }
+          )
+        end
+      end
+
+      redirect_to account_compliance_dashboard_path, notice: "Compliance and deployment assurance settings saved."
+    rescue ActiveRecord::RecordInvalid
+      load_compliance_dashboard
+      flash.now[:alert] = "Unable to save compliance settings."
+      render :show, status: :unprocessable_content
+    end
+
+    def export
+      authorize current_account, :update?
+
+      pack = Accounts::Compliance::EvidencePack.call(
+        account: current_account,
+        tenant_setting: @tenant_setting,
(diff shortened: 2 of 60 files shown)


viamin:
## Agent Update

OK

OK

All conflicts are add/add. Since the PR branch contains the bug fixes we need to preserve, I'll resolve all conflicts by taking the PR branch's version.

All conflicts resolved. Now let me set up dependencies and run tests.

viamin:
## Agent Update

OK

3 conflicts to resolve. Let me read each conflicted file.

The conflicts are all straightforward: the PR branch (`HEAD`) has no additions in these areas, while `origin/main` added ROI/compliance dashboard entries. I need to accept the `origin/main` additions.

Now resolve the accounts show view conflict. Let me see the full conflict:

Now resolve the routes conflict:

Now verify no conflict markers remain:

All conflicts resolved. Now let me install dependencies and set up the database.

Disk space issue. Let me free up some space.

viamin:
## Agent Update

Updated [spec/system/runners/tier_model_assignment_spec.rb](/workspace/spec/system/runners/tier_model_assignment_spec.rb) so the system regression exercises the current runner form and selection path: it now uses a `cursor` runner, drives `runner_tier_model_ids_mid`, and asserts against `tier_model_ids` plus `Models::Select`.

Verification:
- `bundle exec bin/rspec spec/system` passed: 65 examples, 0 failures.
- `bundle exec rspec` still fails on this branch with 22 unrelated failures in existing `accounts/compliance` and install-contract spec clusters.
- `git commit` was attempted twice, but the pre-commit hook runs full RSpec and rejected the commit for those unrelated failures.

Current branch state:
- Staged change only: `spec/system/runners/tier_model_assignment_spec.rb`
- No commit was created because the hook blocked it.

viamin:
## Agent Update

I fixed the CI `test` failure by restoring branch regressions that were breaking the account/compliance request path, the contract-install script behavior, and test cleanup for the new `tenant_settings` foreign key. The key code fixes were in [app/controllers/concerns/account_administration_page.rb](/workspace/app/controllers/concerns/account_administration_page.rb), [app/models/tenant_setting.rb](/workspace/app/models/tenant_setting.rb), [app/models/account_activity_event.rb](/workspace/app/models/account_activity_event.rb), [scripts/install-from-contract.sh](/workspace/scripts/install-from-contract.sh), and the related extractor specs/helpers.

Verification passed. `bundle exec rubocop` was clean, and the pre-commit hook completed the full `bundle exec rspec` run successfully: `13886 examples, 0 failures, 7 pending` in about 10 minutes. The changes are committed as `30a28811` with message `fix(ci): restore test and contract regressions`.

viamin:
## Agent Update

Merged `origin/main` into the PR branch, resolved the conflict in [spec/requests/accounts_spec.rb](/workspace/spec/requests/accounts_spec.rb:17), and committed the result as `8806e8e3` with `fix(runners): merge main into failure recovery branch`.

Verification passed:
- `bundle exec rubocop`
- `bundle exec rspec` with repo-local Bundler path and DB host vars derived from `DATABASE_URL`
- Pre-commit hook also passed on commit, including its lint/test gate

The worktree is clean on `fix/agent-run-failure-recovery`.

viamin:

PR #2325, opened 2026-05-27
feat(self-heal): widen failure-pattern detection to read per-attempt errors

### LINKED ISSUE #2321: feat(self-heal): widen failure-pattern detection to read per-attempt errors
## Problem

`AgentRunPatterns::Detect` currently reads only `agent_run.error_message` — the aggregated outer wrapper (e.g. `"All runners exhausted: Codex, Claude, ..."`). The provider-level error messages that actually identify the root cause live in `runners_attempted` JSONB and never reach the classifier.

Concrete evidence: a real run-data sweep against the current detector produces clusters like `"All runners exhausted: Codex, Kilocode GLM <N>, Opencode Kimi K2, ..."` → `Diagnose` classifies as **Unknown** with remediation `"Manual investigation required"`. Per-attempt errors (`"Model metadata for X not found"`, `"Insufficient Balance"`, `"Auth token expired: refresh_token_reused"`) exist in `runners_attempted` and would have matched existing `ROOT_CAUSE_CATEGORIES` patterns if they had reached the classifier.

A related miscategorization in the live data: a cluster matching `"No compatible runner available: runner:NN is the only runner compatible with glm-X.Y and it is currently rate limited"` gets bucketed as **GitHub API Error** because the regex `/rate.?limit/i` is shared between `github_api` and `llm_provider` categories. Per-attempt context would disambiguate.

## Scope

Refactor `AgentRunPatterns::Detect` to produce an **`EvidenceBundle`** (new value object) per failure cluster instead of a flat string, drawing from multiple sources and sanitized through existing secret-redaction patterns:

- `agent_run.error_message` (outer wrapper, today's input)
- Per-attempt errors from `runners_attempted` JSONB (`runner`, `error_type`, `error_message`, `diagnostics`)
- Last N stdout/stderr lines from `agent_run_logs` (already capped)
- Snapshot of relevant runner config (`tier_model_ids`, `auth_type`, presence of `provider_api_key_id` — never the key itself)
- Aggregate stats (n runs, time window, distinct runners failing the same way)

`Diagnose` is updated to consume the bundle (not just the message string) so existing regex patterns operate over the richer surface. **No new classifier patterns required for this issue** — widening the input alone meaningfully improves accuracy because the existing patterns are stronger matches than the aggregated wrapper.

## Out of scope

- Replacing the regex classifier with an LLM (tracked separately).
- Any auto-apply / remediation action (tracked separately).
- Adding new pattern categories beyond what already exists.

## Acceptance

- `Detect` returns `Pattern` records whose `details` include an `evidence_bundle` field with the multi-source content.
- `Diagnose` consumes the bundle; existing categories now match against per-attempt error text in addition to the outer wrapper.
- Existing rubocop clean; existing `agent_run_pattern_detector` cron continues to operate.
- A regression spec that loads a fixture run with `runners_attempted` containing `"Model metadata for `gpt-4o` not found"` is classified as `llm_provider` (today: Unknown).
- Sensitive-field allowlist for the config snapshot is explicit; redaction patterns are unit-tested.

## Privacy & safety

- The evidence bundle never includes provider API keys or GitHub tokens. The runner-config snapshot is a fixed allowlist of fields known to be safe.
- Existing `RUNNER_ATTEMPT_SECRET_PATTERNS` redaction is reused on any free-text strings included in the bundle.

viamin:
## Summary

Implemented the self-heal refactor for #2321 and committed it as `7b9bcf5` with message `feat(self-heal): widen failure pattern evidence collection`.

`AgentRunPatterns::Detect` now attaches a sanitized `details[:evidence_bundle]` built from wrapper errors, `runners_attempted`, stdout/stderr tails, an explicit safe runner-config snapshot, and aggregate cluster stats. `AgentRunPatterns::Diagnose` now classifies against that richer bundle through the new [`app/services/agent_run_patterns/evidence_bundle.rb`](/workspace/app/services/agent_run_patterns/evidence_bundle.rb), which lets existing regex categories match per-attempt provider failures instead of only the outer wrapper.

I added regression coverage for the `Model metadata for \`gpt-4o\` not found` case and redaction coverage using a fixture at [`spec/fixtures/files/agent_run_patterns/model_metadata_not_found_runners_attempted.json`](/workspace/spec/fixtures/files/agent_run_patterns/model_metadata_not_found_runners_attempted.json). Verification completed with `bundle exec rubocop` clean and `bundle exec rspec` passing: `13897 examples, 0 failures, 7 pending`.

---

Closes #2321

### DIFF
--- a/app/services/agent_run_patterns/detect.rb
+++ b/app/services/agent_run_patterns/detect.rb
@@ -8,6 +8,11 @@ module AgentRunPatterns
     ERROR_CLUSTER_MIN_SIZE = 3
     DEFAULT_WINDOW_HOURS = 6
     BASELINE_WINDOW_DAYS = 30
+    EVIDENCE_LOG_TAIL_LINE_LIMIT = 20
+    EVIDENCE_MESSAGE_LIMIT = 5
+    EVIDENCE_RUNNER_ATTEMPT_LIMIT = 10
+    EVIDENCE_RUNNER_CONFIG_LIMIT = 5
+    EVIDENCE_SAFE_RUNNER_CONFIG_FIELDS = %w[runner_key auth_type tier_model_ids].freeze
     DETECTION_FINISHED_STATUSES = (AgentRun::FINISHED_STATUSES - [ "retried" ]).freeze
 
     Pattern = Data.define(:type, :goal, :severity, :details)
@@ -42,7 +47,8 @@ module AgentRunPatterns
     attr_reader :account, :window_hours
 
     def fetch_recent_finished_runs
-      AgentRun.joins(:project)
+      AgentRun.includes(:runner, :agent_run_logs)
+        .joins(:project)
         .where(projects: { account_id: account.id })
         .where(status: DETECTION_FINISHED_STATUSES)
         .where(completed_at: window_hours.hours.ago..)
@@ -64,6 +70,7 @@ module AgentRunPatterns
           total_runs: runs.size,
           failure_rate: streak.size.to_f / runs.size,
           error_messages: streak.filter_map(&:error_message).first(5),
+          evidence_bundle: build_evidence_bundle(streak),
           run_ids: streak.map(&:id),
           started_at: streak.last.completed_at,
           ended_at: streak.first.completed_at
@@ -92,6 +99,7 @@ module AgentRunPatterns
           baseline_rate: baseline&.round(4),
           error_messages: runs.select { |r| AgentRun::FAILURE_STATUSES.include?(r.status) }
                               .filter_map(&:error_message).first(5),
+          evidence_bundle: build_evidence_bundle(runs.select { |r| AgentRun::FAILURE_STATUSES.include?(r.status) }),
           run_ids: runs.select { |r| AgentRun::FAILURE_STATUSES.include?(r.status) }.map(&:id)
         }
       ) ]
@@ -118,6 +126,7 @@ module AgentRunPatterns
             occurrence_count: error_runs.size,
             total_failures: failed_runs.size,
             sample_messages: error_runs.map(&:error_message).compact.uniq.first(3),
+            evidence_bundle: build_evidence_bundle(error_runs),
             run_ids: error_runs.map(&:id),
             statuses: error_runs.group_by(&:status).transform_values(&:count)
           }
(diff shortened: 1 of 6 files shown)


viamin:
## Agent Update

OK.

Now I understand both issues. Let me fix them:

**Thread 1**: Remove `.includes(:agent_run_logs)` and load logs on-demand in `extract_log_tail` only for the few runs that need them.

**Thread 2**: Extract `failed_runs` in `detect_high_failure_rate` to avoid triple iteration.

Now fix `extract_log_tail` to load logs on-demand instead of relying on eager-loaded association:

Now update the rest of `extract_log_tail` to work with the already-filtered/limited records:

Now fix **Thread 2** — extract `failed_runs` in `detect_high_failure_rate`:

Let me review the full file for similar issues:

The proactive scan shows `detect_error_clusters` already extracts `failed_runs` properly. The `detect_failure_streak` uses `streak` which is computed once via `take_while`. No other similar issues found.

Now let me check the tests to see if they need updating due to the removal of the eager load:

The tests create logs in the database, so the new query-based `extract_log_tail` will find them. Let me now install dependencies and run lint + tests.

viamin:
## Agent Update

Addressed both unresolved review threads and committed the fix as `16f2392b` with message `fix(self-heal): resolve review feedback on evidence detection`.

`AgentRunPatterns::Detect` now batch-loads sampled `stdout`/`stderr` tails in a single `AgentRunLog` query and partitions them in Ruby, preserving the existing evidence payload while removing the per-run/per-stream query pattern. `AgentRunPatterns::Diagnose` now restores binary per-document scoring, so confidence remains “fraction of evidence documents that matched” instead of being inflated by multiple regex hits in one document.

I added regression coverage for both changes: one spec locks the log-tail path to a single query, and another ensures a document matching multiple patterns in the same category still contributes only `1` toward confidence.

Verification:
- `bin/rails db:prepare`
- `bundle exec rspec spec/services/agent_run_patterns/detect_spec.rb spec/services/agent_run_patterns/diagnose_spec.rb`
- `bundle exec rubocop app/services/agent_run_patterns/detect.rb app/services/agent_run_patterns/diagnose.rb spec/services/agent_run_patterns/detect_spec.rb spec/services/agent_run_patterns/diagnose_spec.rb`
- `bundle exec rspec`
- pre-commit hook passed on `git commit`

Notes:
- Bundler had to be installed into a temporary path because the system gem cache was not writable.
- `yarn install` hit an environment extraction/disk-space issue, but it was not needed for this Ruby-only change and did not affect lint/test/commit verification.

viamin:
## Agent Update

Addressed the unresolved review thread in [`app/services/agent_run_patterns/detect.rb`](/workspace/app/services/agent_run_patterns/detect.rb) by changing log-tail collection to fetch only the last `EVIDENCE_LOG_TAIL_LINE_LIMIT` rows per `agent_run_id`/`log_type` in SQL via `ROW_NUMBER()`, and by memoizing fetched tails per sampled run set so repeated evidence-bundle builds do not re-read the same logs. I also added regression coverage in [`spec/services/agent_run_patterns/detect_spec.rb`](/workspace/spec/services/agent_run_patterns/detect_spec.rb) for cache reuse and for keeping only the most recent tail lines.

Verification passed with `bundle exec rubocop` and `bundle exec rspec` (`13900 examples, 0 failures, 7 pending`). The fix is committed as `55546d5d` with message `fix(self-heal): cache bounded evidence log tails`.

viamin:
## Agent Update

Addressed the unresolved review thread by excluding `runner_configs` from `AgentRunPatterns::EvidenceBundle#documents` while keeping them in `to_payload` for contextual inspection. That removes non-matchable metadata from `Diagnose`’s confidence denominator. I also tightened the existing regression in `spec/services/agent_run_patterns/diagnose_spec.rb` to assert the corrected `0.5` confidence.

Verification passed with `bundle exec rubocop` and `bundle exec rspec` (`13924 examples, 0 failures, 7 pending`). `bin/rails db:prepare` also succeeded once the provided Postgres service was mapped through `DB_HOST`/`DB_USERNAME`/`DB_PASSWORD` for this app’s multi-DB config. `yarn install` hit `ENOSPC` in this container, but it was not needed for this Ruby-only change or the passing hook run.

Committed as `7ce00c4b85fa6db4fee25372a5e2c11ffb2b6514` with message `fix(self-heal): exclude runner configs from diagnosis evidence`.

viamin:

PR #2334, opened 2026-05-28
feat(self-heal): bounded auto-apply for reversible remediation actions + revert UI

### LINKED ISSUE #2323: feat(self-heal): bounded auto-apply for reversible remediation actions + revert UI
## Dependencies

Depends on #2322
Depends on #2321

## Problem

After #2322, Paid has a structured proposed action for every diagnosed failure cluster, persisted in `remediation_decisions`, but every action still requires a human to act on the notification. The system observes and reasons but doesn't close the loop. Configuration-level failures (runners with empty balances, stale tier_model_ids on subscription-auth runners, persistently broken runners that nobody disables) continue to consume capacity and trigger notifications indefinitely.

### 3a. Action handlers

For each enum value from #2322, implement a corresponding handler that performs the mutation atomically and writes `revert_data` to the existing `remediation_decisions` row before completing:

- `mark_runner_unavailable` → `state.mark_rate_limited!(reset_at: …)` (already reversible via auto-expiry)
- `clear_runner_field` → null out a named field on a Runner record (only allowlisted fields: `tier_model_ids`; new fields require code review + a migration to the allowlist)
- `disable_runner_fallback` → flip `enabled_for_fallback` to false
- `file_issue` → existing `ExceptionHandler::IssueFiler` path
- `notify` → existing `Notifications::Publish` (already in #2322)

Field allowlist is a constant in code, not a database setting — the LLM cannot widen it.

### 3b. Policy & user control

Per-account settings (existing `Account` columns or new `Account#remediation_policy` jsonb):

- For each action enum value: `notify_only` (default) | `auto_apply` | `disabled`.
- Per-action minimum confidence threshold for auto-apply (default 0.8).
- Per-action filing threshold (how many distinct occurrences across cycles before escalating).

Hard rules in code, not configurable:

- No action ever mutates project settings, API keys, or anything outside the runner/runner_state scope.
- Same fingerprint cannot trigger the same auto-applied action more than once per 24h.
- An auto-applied action whose `post_remediation_failure_count` ≥ `pre_remediation_failure_count` after one detection cycle is marked `outcome: unchanged`; same-fingerprint runs after that downgrade the action to notify-only.

### 3c. Revert UI

On each in-app notification with an applied remediation:

- "Revert" button → enqueues a job that reads `revert_data` and re-applies the prior state, then sets `status: reverted` on the row.
- Decision history view per runner: timeline of proposed/applied/reverted actions with `Why` evidence pointers.
- Required because we're mutating user data: every change must be one click to undo.

## Out of scope

- New action enum values. Each addition is its own PR with security review and an entry in the allowlist constant.
- Cross-account or admin-level remediations.
- Anything that costs the user money (e.g. provisioning replacement runners).

## Acceptance

- All four mutating actions (`mark_runner_unavailable`, `clear_runner_field`, `disable_runner_fallback`, `file_issue`) implemented as handlers with `revert_data`.
- Per-action policy persisted on `Account`; default policy for every action is `notify_only` (no behavior change unless the user opts in).
- Revert button on the notification re-applies prior state and marks the decision `reverted`.
- "Decision history" view on the runner detail page shows the timeline.
- Outcome tracking: a cron updates `outcome` for each applied decision based on next-cycle failure counts.
- Specs: end-to-end test of "fingerprint → proposed → auto-applied → outcome marked unchanged → downgraded to notify-only" for `mark_runner_unavailable`; revert spec for `clear_runner_field`.
- Audit log entry written for every auto-apply, visible in the existing account audit log.

## Privacy & safety

- Every mutating action writes to a `RemediationDecision` row with `revert_data` BEFORE the mutation commits (transactional).
- Auto-apply opt-in is per-action so a user can opt in to e.g. `mark_runner_unavailable` (lowest blast radius) without opting in to `clear_runner_field`.
- The action enum is the closed-set safety boundary; even if a future Diagnose regression produces a wrong target, the action can only do one of the enumerated safe mutations and is one-click reversible.

viamin:
## Summary

Implemented bounded self-heal auto-apply with reversible handlers, per-account remediation policy, revert jobs/UI, runner decision history, outcome tracking, and audit-log entries. The flow now supports auto-applying safe runner-scoped actions, downgrading repeat-unchanged fingerprints back to notify-only, and one-click reverts from notifications and decision detail pages.

Verification passed with `bundle exec rubocop` and `bundle exec rspec` under `RAILS_ENV=test` using the provided DB host/user env overrides in this workspace. Commit created: `8976440` with message `feat(self-heal): auto-apply reversible remediations`.

---

Closes #2323

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -34,6 +34,10 @@ class AccountsController < ApplicationController
   private
 
   def account_params
-    params.require(:account).permit(:name, :default_max_tokens_per_run)
+    params.require(:account).permit(
+      :name,
+      :default_max_tokens_per_run,
+      remediation_policy: RemediationDecision::PROPOSED_ACTIONS.index_with { %i[mode minimum_confidence filing_threshold] }
+    )
   end
 end
--- a/app/controllers/remediation_decisions_controller.rb
+++ b/app/controllers/remediation_decisions_controller.rb
@@ -5,4 +5,17 @@ class RemediationDecisionsController < ApplicationController
     @remediation_decision = policy_scope(RemediationDecision).find(params[:id])
     authorize @remediation_decision
   end
+
+  def revert
+    @remediation_decision = policy_scope(RemediationDecision).find(params[:id])
+    authorize @remediation_decision, :update?
+
+    unless @remediation_decision.revertable?
+      redirect_to remediation_decision_path(@remediation_decision), alert: "This remediation cannot be reverted."
+      return
+    end
+
+    RevertRemediationDecisionJob.perform_later(@remediation_decision.id, actor_id: current_user.id)
+    redirect_to remediation_decision_path(@remediation_decision), notice: "Revert queued."
+  end
 end
--- a/app/controllers/runners_controller.rb
+++ b/app/controllers/runners_controller.rb
@@ -59,10 +59,12 @@ class RunnersController < ApplicationController
 
   def edit
     authorize @runner
+    load_remediation_history
   end
 
   def update
     authorize @runner
+    load_remediation_history
     @runner.assign_attributes(runner_params)
     validate_container_executable!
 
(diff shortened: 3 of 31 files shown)


viamin:

PR #2310, opened 2026-05-27
feat(quality): RDR-036 phase 3b — nightly mutation sweep job + dashboard widget

viamin:
## Summary

Closes the loop on RDR-036 Phase 3b by adding a nightly full-suite mutation sweep job that catches indirect-change blind spots missed by incremental mode, and surfacing kill-rate trends on the quality dashboard so teams can monitor mutation testing health over time.

### Database
- Add `source` column to `quality_metrics` table to distinguish scheduled sweep results (`scheduled_mutation_sweep`) from per-agent-run metrics

### Services
- New `MutationSweeps::Run` service that enumerates projects with an enabled `mutation_test` `PreCommitRequirement` for Ruby, spins up an isolated container against `main`, runs `bundle exec mutant run` (full sweep, no `--since`), and records the kill rate as a `QualityMetric`

### Jobs
- New `ScheduledMutationSweepJob` (GoodJob recurring) that iterates eligible projects nightly, throttled to one project per worker to avoid resource spikes
- Triggers `QualityAlerts::CheckGate` evaluation when the score breaches a configured threshold

### Dashboard
- Add "Mutation kill-rate" panel to `Projects::QualityDashboardsController` view showing:
  - Current sweep score from the most recent scheduled sweep metric
  - 30-day trend sparkline of sweep scores
  - Per-agent-run kill-rate distribution histogram (last 100 runs)
- Panel is conditionally rendered — hidden when the project has no enabled `mutation_test` requirement

### Specs
- Job correctly enumerates eligible projects and skips ineligible ones
- Job records `QualityMetric` with the correct `source` tag
- Dashboard view renders only for projects with mutation testing enabled
- Trend sparkline data shape matches front-end chart component expectations

## Design Decisions

- **Throttle to one project per worker per night**: avoids resource spikes from concurrent full-suite mutation runs, staying within the < 90 min per-project budget from RDR-036's validation section
- **Separate `source` tag rather than separate table**: reuses the existing `QualityMetric` pipeline and gate evaluation path while keeping scheduled sweeps distinguishable from incremental per-agent-run metrics
- **Conditional panel rendering**: projects that haven't opted into mutation testing see no empty widget, keeping the dashboard clean

## Dependencies

- Depends on #2281 (Phase 3a) which introduced `QualityMetric#mutation_kill_rate`

---

Closes #2282

PR #2335, opened 2026-05-29
fix(migrations): make AddSourceToQualityMetrics safe and prevent recurrence

viamin:
## Summary

- `AddSourceToQualityMetrics` was rejected by `strong_migrations` on the first developer to run `bin/setup` (`add_index` ran non-concurrently). Fixed with `disable_ddl_transaction!`, `algorithm: :concurrently`, and idempotency guards.
- Add a **Database Schema Workflow** section to the agent prompt for Ruby projects with a Postgres service container, instructing the agent to use `bin/rails generate migration` + `db:migrate` and never hand-edit `db/schema.rb`. Prevents the same class of bug at its source.
- Hoist `language` resolution in `Prompts::ServiceContainerSections.service_environment_section_for` so the new check is safe when `include_setup_instruction: false` is passed (latent `NameError`). Regression spec added.

## Why this slipped through

The agent that produced the original migration (PR #2310) hand-edited `db/schema.rb` to match the new column/index instead of running `db:migrate`. The agent's verify loop runs `db:prepare` once at setup time — before the migration file exists — and tests use `db:schema:load`, never executing the new migration. CI does the same (`db:create db:schema:load`). `strong_migrations` is a runtime check inside `db:migrate`, so it never fired in any environment until a real developer hit it locally.

The prompt update closes the agent-side gap. (The CI-side gap — running `db:migrate` against a real Postgres on PRs that touch `db/migrate/**` — is worth doing separately as a defense-in-depth layer.)

## Test plan

- [x] `bin/setup --skip-server` runs clean against a fresh DB
- [x] `bin/rspec spec/services/prompts/build_for_issue_spec.rb spec/services/prompts/build_for_pr_spec.rb` — 125 examples, 0 failures
- [x] `bin/rubocop` clean on modified files
- [ ] CI green


viamin:

PR #2374, opened 2026-05-29
feat(chat): enforce "chat ≤ user" authorization invariant in tool framework

### LINKED ISSUE #2349: feat(chat): enforce "chat ≤ user" authorization invariant in tool framework
## Context

Today every MCP tool in `app/mcp/tools/` independently calls `Pundit.policy(user, record)` to authorize ([base_tool.rb](app/mcp/tools/base_tool.rb), [registry.rb:41-49](app/mcp/tools/registry.rb#L41-L49), [trigger_agent_run.rb:25-47](app/mcp/tools/trigger_agent_run.rb#L25-L47)). This works, but it is per-tool discipline — one forgotten `authorize!` or one accidental `TenantContext.with_system_access` inside a tool would silently grant chat callers privileges their UI session does not have.

This becomes load-bearing as we expand the chat tool surface (admin actions, operator Avo actions, source-code reads, multi-repo containerized chat). We need the invariant **enforced by the framework**, not by per-tool diligence.

## Goal

Establish a single chokepoint such that every chat-driven tool call:

1. Resolves a `current_user` (no anonymous tool calls)
2. Executes the entire tool body inside `TenantContext.with(user.account)` — `TenantContext.with_system_access` is forbidden inside tool bodies (only allowed in the controller/job bootstrap that resolves the session, which already exists in [chat_messages_controller.rb:120-125](app/controllers/chat_messages_controller.rb#L120-L125) and [process_message_job.rb:86](app/jobs/chat_sessions/process_message_job.rb#L86))
3. Has performed at least one Pundit authorization check before performing any DB write or external call — declared via a class-level `authorize :action_on, ->(args) { ... }` DSL or equivalent, enforced by `BaseTool#call` wrapping

## Proposed shape

- `Tools::BaseTool` gains a `dispatch` wrapper that:
  - asserts `user.present?`
  - opens `TenantContext.with(user.account)` for the whole call
  - tracks whether `authorize!` was called and raises `Tools::UnauthorizedError` if no authorization happened before mutation
- A RuboCop cop or simple grep-based spec rejects `TenantContext.with_system_access` inside `app/mcp/tools/**`
- `Tools::Registry.dispatch` becomes the only entry path; `PaidMcpServer#call_tool` ([paid_mcp_server.rb:52-60](app/mcp/paid_mcp_server.rb#L52-L60)) routes through it

## Tests

The critical test is the *parity* spec — for every registered tool:

- Construct a user without the needed permission for a sample target record
- Call the tool through the MCP path; assert it raises/returns an authorization error
- Call the equivalent Servo/controller path directly with the same user; assert it also raises/returns an authorization error
- Both paths must agree

A failure here means chat can do something the UI cannot.

## Out of scope

- Adding new tools (covered by sibling issues for admin/operator/source-code surfaces)
- Operator-role tool gating (covered by the OperatorTools issue)

## Acceptance

- [ ] `BaseTool#dispatch` enforces user, tenant context, and authorization-was-called
- [ ] Spec asserting `TenantContext.with_system_access` does not appear under `app/mcp/tools/**`
- [ ] Parity spec walks every registered tool and asserts UI-denial ⇒ chat-denial
- [ ] Existing tools refactored to use the dispatch wrapper without behavior change

## Related

- Discussion in conversation that produced sibling issues (admin tools, operator tools, source-code tools, containerized chat RDR, dangerous_mode audit)

viamin:
## Summary

Enforce a "chat ≤ user" authorization invariant at the framework level so that every MCP tool call is guaranteed to resolve a user, run inside the correct tenant context, and perform at least one Pundit authorization check — eliminating reliance on per-tool discipline that could silently grant chat callers privileges their UI session doesn't have.

### MCP Tool Framework (`app/mcp/tools/`)

- **`BaseTool#dispatch`** — new chokepoint wrapper that asserts `user.present?`, opens `TenantContext.with(user.account)` for the entire tool execution, runs class-level declared authorizations before the tool body, and raises `Tools::UnauthorizedError` if no authorization occurred before mutation
- **Class-level `authorize` DSL** — tools now declare their authorization contract declaratively rather than scattering inline `Pundit.policy` calls
- **`#perform` convention** — tool bodies moved from `#call` to `#perform`, with `#dispatch` owning the lifecycle

### MCP Server & Registry

- **`Tools::Registry.dispatch`** — becomes the single execution entrypoint for all tool calls
- **`PaidMcpServer#call_tool`** — routes through `Registry.dispatch` instead of invoking tools directly

### Existing Tool Refactors

- All current tools refactored to declare their authorization contract via the new DSL and implement `#perform` — no behavior change, only structural alignment with the new framework

### Tests

- **Base wrapper spec** — validates user requirement, tenant scoping, and authorization-was-called enforcement
- **System access prohibition spec** — asserts `TenantContext.with_system_access` does not appear anywhere under `app/mcp/tools/**`
- **Parity spec** — walks every registered tool, constructs an unauthorized user, and verifies that chat-path denial matches direct Servo/controller-path denial
- **Server spec** — updated to assert registry dispatch routing

## Design Decisions

- **Framework-enforced, not convention-enforced**: The prior model relied on each tool author remembering to call `authorize!`. A single forgotten check would silently escalate privileges. The `dispatch` wrapper makes this structurally impossible — tools that skip authorization raise at runtime.
- **`with_system_access` banned in tool bodies**: System-access bypass is only legitimate in the controller/job bootstrap that resolves the session (which already exists). A grep-based spec enforces this boundary, preventing future tools from accidentally escaping tenant isolation.
- **Parity spec as the critical safety net**: Rather than testing tools in isolation, the parity spec ensures chat authorization and UI authorization always agree. A failure means chat can do something the UI cannot — the exact class of bug this change prevents.

---

Closes #2349

### DIFF
--- a/app/mcp/paid_mcp_server.rb
+++ b/app/mcp/paid_mcp_server.rb
@@ -51,12 +51,7 @@ class PaidMcpServer
   end
 
   def call_tool(name:, arguments:)
-    tool_class = Tools::Registry.find(name)
-    raise ArgumentError, "Unknown tool: #{name}" unless tool_class
-    raise ArgumentError, "Tool arguments must be a JSON object" unless arguments.is_a?(Hash)
-
-    tool = tool_class.new(user:, session:)
-    tool.call(**arguments.symbolize_keys)
+    Tools::Registry.dispatch(name:, arguments:, user:, session:)
   end
 
   class RateLimitExceeded < StandardError; end
--- a/app/mcp/tools/base_tool.rb
+++ b/app/mcp/tools/base_tool.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 module Tools
+  class UnauthorizedError < Pundit::NotAuthorizedError; end
+
   class BaseTool
     include Pundit::Authorization
 
@@ -11,8 +13,25 @@ module Tools
       @session = session
     end
 
-    def call(**_args)
-      raise NotImplementedError, "#{self.class}#call must be implemented"
+    def call(**args)
+      dispatch(**args)
+    end
+
+    def dispatch(**args)
+      raise UnauthorizedError, "Tool calls require an authenticated user" if user.blank?
+
+      TenantContext.with(account) do
+        reset_authorization_tracking!
+        run_declared_authorizations!(args)
+
+        perform(**args).tap do
+          raise UnauthorizedError, "#{self.class.tool_name} must authorize before execution" unless authorized?
+        end
+      end
+    end
+
+    def perform(**_args)
+      raise NotImplementedError, "#{self.class}#perform must be implemented"
     end
 
     def self.tool_name
@@ -39,7 +58,29 @@ module Tools
       false
     end
 
+    def self.authorize(query, resolver = nil, policy_class: nil, &block)
+      resolver ||= block
+      raise ArgumentError, "#{name}.authorize requires a resolver" unless resolver
+
+      own_authorization_rules << {
+        query: query,
+        resolver: resolver,
+        policy_class: policy_class
+      }
+    end
+
+    def self.authorization_rules
+      inherited_rules = superclass.respond_to?(:authorization_rules) ? superclass.authorization_rules : []
+      inherited_rules + own_authorization_rules
+    end
+
     private
+
+    def authorize(record, query = nil, policy_class: nil)
+      @authorization_performed = true
+      super
+    end
+    alias_method :authorize!, :authorize
 
     # Pundit expects current_user
     def current_user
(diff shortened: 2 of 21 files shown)


viamin:
## Agent Update

Addressed the unresolved review threads and committed the fixes in `482a7ca` (`fix(chat): require preflight tool authorization`).

`BaseTool#dispatch` now requires a declared authorization to complete before `perform` begins, so a post-hoc `authorize` inside the tool body can no longer satisfy the guard after side effects. I also memoized resolver-backed `project_for`/`run_for` lookups by ID across the affected MCP tools so the preflight authorization lookup is reused during `perform` instead of issuing duplicate scoped queries.

I added spec coverage for both behaviors: `perform` is blocked when authorization is missing or only attempted post hoc, and the affected project/run tools verify that the scoped finder is hit only once per request.

Checks run:
- `bundle install`
- `yarn install`
- `bin/rails db:prepare`
- `bundle exec rubocop`
- `bundle exec rspec`
- pre-commit hook on `git commit` also passed (`14131 examples, 0 failures, 7 pending`)

viamin:

PR #2425, opened 2026-05-31
feat: Chat continuation incorrectly requires configured API-key runner for existing previous chat

### LINKED ISSUE #2422: Chat continuation incorrectly requires configured API-key runner for existing previous chat
## Summary
Continuing a chat from the previous chats list can show an incorrect configuration error in the chat UI:

> Chat requires a configured API-key runner. Add a chat-enabled runner with an API key and select it for this session.

Based on the observed behavior, this appears to be wrong for at least some existing chats and blocks or misdirects the user when trying to resume a prior conversation.

## Steps to Reproduce
1. Open the app.
2. Go to the previous chats list.
3. Select an existing chat to continue it.
4. Observe the message shown in the chat UI.

## Actual Behavior
The UI shows:

> Chat requires a configured API-key runner. Add a chat-enabled runner with an API key and select it for this session.

## Expected Behavior
Opening or continuing an existing chat should not show this message unless a runner with an API key is actually required for that specific chat/session state. If the chat can be resumed with the existing configuration, the UI should allow continuation normally.

## Notes
- The reported message appears to be incorrect rather than a valid configuration requirement.
- This may indicate that the runner requirement check is being applied too broadly when loading existing chats from the previous chats list.
- The issue should be investigated in the chat session restore/resume path and the conditions that decide when to surface runner configuration errors.

viamin:
## Summary

Adjusted chat client resolution so resumed legacy chats no longer fail just because the stored session runner is missing or subscription-backed. In [app/services/chat_sessions/build_llm_client.rb](/workspace/app/services/chat_sessions/build_llm_client.rb), the resolver now falls back to the chat creator’s current chat-enabled API-key runner when the session has no runner or only a subscription runner, while still raising the explicit misconfiguration error if a selected API-key runner itself has no secret. I updated [spec/services/chat_sessions/build_llm_client_spec.rb](/workspace/spec/services/chat_sessions/build_llm_client_spec.rb) to cover both fallback cases and the retained hard failure for a broken API-key runner.

Validation passed with `bin/rails db:prepare`, `bundle exec rubocop`, and `bundle exec rspec` using the provided Postgres service via explicit `DB_HOST`/`DB_USERNAME`/`DB_PASSWORD` env vars. The suite finished with `14242 examples, 0 failures, 7 pending` from existing repo pending specs. Commit created: `3da51ad` (`fix(chat): resume legacy chats with api runner fallback`).

---

Closes #2422

### DIFF
--- a/app/services/chat_sessions/build_llm_client.rb
+++ b/app/services/chat_sessions/build_llm_client.rb
@@ -13,7 +13,7 @@ module ChatSessions
     end
 
     def call
-      provider = chat_session.runner
+      provider = resolved_runner
       raise LlmClientConfigurationError, missing_runner_message unless provider
 
       api_key = provider.effective_api_secret
@@ -32,6 +32,18 @@ module ChatSessions
     private
 
     attr_reader :chat_session
+
+    def resolved_runner
+      provider = chat_session.runner
+      return provider if provider&.api_key?
+      return fallback_runner if provider.nil? || provider.subscription?
+
+      provider
+    end
+
+    def fallback_runner
+      Runner.kept_only.api_key.for_chat.where(user: chat_session.created_by).ordered.first
+    end
 
     def anthropic_client(api_key)
       transport = AgentHarness::TextTransport.new(api_key: api_key)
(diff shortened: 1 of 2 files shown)


viamin:

Issue #2435, opened 2026-06-01
chore(agent-harness): remove --mcp-config variadic flag-form workaround once upstream ships =value form


viamin:
## Summary

Paid carries a temporary patch in `config/initializers/agent_harness.rb` that forces the Claude `--mcp-config` flag into the `--flag=value` form, because the Claude CLI treats `--mcp-config <configs...>` as **variadic** and otherwise swallows the positional prompt as a second config path.

This was the root cause of widespread Claude fallback failures:

```
Error: Invalid MCP configuration:
MCP config file not found: /workspace/Reply with exactly OK.
```

(The `Reply with exactly OK.` text is the preflight prompt being consumed as a config path.)

## Why this is its own issue (not folded into #2364 / #2365)

- #2365 (suppress `.mcp.json` autodiscovery) shipped upstream in agent-harness **0.19.0**.
- But upstream implemented suppression with the **same buggy space-form** placement, and it is still present in **v0.20.0** (latest). See upstream **viamin/agent-harness#229**.
- Our patch is version-gated `< 0.19.0`. The variadic-safe flag form was split into its own **ungated, self-deactivating** module (`PaidAgentHarnessAnthropicMcpConfigFlagFormPatch`) precisely so that adopting 0.19.0+ does NOT silently reintroduce the prompt-swallow.

## Cleanup criteria

Remove `PaidAgentHarnessAnthropicMcpConfigFlagFormPatch` once agent-harness ships the `--mcp-config=VALUE` form (upstream **viamin/agent-harness#229**) AND Paid's pinned version is at/after that release.

The module's guard already makes it a no-op once the gem emits a single equals-form token, so removal is safe-by-default — this issue just tracks deleting the dead code.

## Related

- Blocks-aware-of: viamin/agent-harness#229 (upstream flag-form fix)
- See also #2364 (adopt native suppression) — that adoption is **blocked** on #229 and on container materialization (upstream `build_execution_preparation` returns `nil`).

PR #2419, opened 2026-05-31
fix: Claude CLI auto-discovers .mcp.json from working directory, contaminating stdout

viamin:
## Summary

Prevent the Claude CLI from auto-discovering `.mcp.json` files in the working directory, which contaminates the `--output-format=json` stdout envelope and causes agent run failures. This adds a Paid-side patch to force an explicit empty MCP config when no MCP servers are configured, suppressing the CLI's fallback auto-discovery behavior until agent-harness ships native suppression (viamin/agent-harness#225).

## Changes

**Initializer patch** (`config/initializers/agent_harness.rb`)
- Prepend a module to the Anthropic provider that injects `mcp_servers: []` and passes `--mcp-config` with an empty `{"mcpServers": {}}` tempfile when no MCP servers are explicitly configured
- Ensures the empty config is materialized through `ExecutionPreparation` for both host and container execution paths
- Follows the existing pattern used for Pi API-key and embedding patches in the same initializer

**Tests**
- Added regression coverage in `spec/services/runners/harness_execution_plan_spec.rb` for the zero-server MCP config case
- Updated container execution expectations in `spec/temporal/activities/run_agent_activity_spec.rb`

## Design Decisions

- **Paid-side patch over waiting for upstream**: The CLI auto-discovery caused ~15+ consecutive agent run failures on host execution and presents a supply-chain risk for container execution (malicious `.mcp.json` in target repos). A Paid-side patch provides immediate protection while viamin/agent-harness#225 tracks the permanent fix.
- **Empty tempfile approach**: Passing `--mcp-config` with an empty `{"mcpServers": {}}` file is the simplest way to suppress auto-discovery without modifying CLI behavior — the CLI sees an explicit config and skips its directory walk.
- **Remove when upstream ships**: This patch should be removed when agent-harness >= the version shipping native suppression is adopted. Tracked by #2364.

---

Closes #2365

Issue #2364, opened 2026-05-29
feat: adopt agent-harness MCP auto-discovery suppression


viamin:
Depends on viamin/agent-harness#225

## Summary

When agent-harness v0.18.2+ adds support for suppressing the Claude CLI's `.mcp.json` auto-discovery (via an empty MCP config tempfile or `--no-mcp` flag), Paid should adopt it to prevent streaming session events from contaminating CLI stdout.

## Current Workaround

Paid PR #TBD adds defensive parsing in `AnalyzeIssueActivity`:
- `cli_streaming_only?` detects streaming-only output in `call_llm`
- `extract_result_from_jsonl` handles mixed JSONL (streaming events + result envelope) in `parse_response!`

This workaround handles the symptom. The upstream fix addresses the root cause.

## Adoption Steps

1. Bump agent-harness dependency to the version that includes the fix
2. Remove the `CLI_STREAMING_EVENT_PREFIX` / `cli_streaming_only?` detection in `call_llm` if the upstream handles it in `parse_response`
3. Consider removing `extract_result_from_jsonl` if the upstream strips streaming events before returning `Response.output`
4. Keep the JSONL fallback as defense-in-depth if the upstream only fixes the auto-discovery case (not all possible stdout contamination scenarios)
5. Update tests accordingly

## Context

- Upstream issue: viamin/agent-harness#225
- ~15+ consecutive agent runs failed because the Claude CLI auto-discovered the workspace's `.mcp.json` (Playwright MCP server) and emitted `{"type":"session.mcp_servers_loa...` to stdout instead of the expected JSON envelope

viamin:
## Clarifying questions
1. Has viamin/agent-harness#225 been merged and released? Which version includes the fix (the issue mentions v0.18.2+ speculatively)?
2. What is the final upstream API — does it use an empty MCP config tempfile, a `--no-mcp` CLI flag, or something else? The adoption steps hinge on knowing which mechanism was implemented.
3. Steps 3–4 are conditional: "Consider removing `extract_result_from_jsonl` **if** the upstream strips streaming events before returning `Response.output`" / "Keep the JSONL fallback **if** the upstream only fixes the auto-discovery case." Which of these conditions is true for the shipped upstream fix?
4. Which PR introduced the current workaround (`cli_streaming_only?` / `extract_result_from_jsonl`)? The issue references "Paid PR #TBD" — is it merged to `main` yet?

## Current context
[...]

PR #2436, opened 2026-06-01
fix(agent-runs): stop Claude --mcp-config from swallowing the prompt

viamin:
## Problem

Claude fallback runs were failing fleet-wide with:

```
Error: Invalid MCP configuration:
MCP config file not found: /workspace/Reply with exactly OK.
```

`Reply with exactly OK.` is the preflight **prompt** — the CLI was treating it as an MCP config path.

## Root cause

The Claude CLI declares `--mcp-config <configs...>` as a **variadic** flag (confirmed in `claude --help`, v2.1.92). agent-harness emits the space-form `["--mcp-config", path]` immediately before the positional prompt, so the CLI greedily consumes the prompt as a *second* config path, resolves it against the container CWD (`/workspace`), and dies before the run starts.

PR #2419 started passing `--mcp-config` on **every** no-server Claude run (to suppress `.mcp.json` autodiscovery), which turned this latent bug into a regression affecting essentially every Claude run.

Verified against the real CLI:

```sh
printf '{"mcpServers":{}}' > /tmp/empty.json
claude --print --output-format=json --mcp-config /tmp/empty.json "Reply with exactly OK."
# => Error: MCP config file not found: /tmp/Reply with exactly OK.   (BROKEN)
claude --print --output-format=json --mcp-config=/tmp/empty.json "Reply with exactly OK."
# => {"result":"OK", ...}                                            (WORKS)
```

## Fix

Emit the single-token `--mcp-config=<path>` form, which captures exactly one path and keeps the prompt positional.

- **Empty-config suppression path** (`PaidAgentHarnessAnthropicMcpSuppressionPatch#build_command`): builds the `=value` flag by hand, because agent-harness 0.18.2 emits no flag at all when the server list is empty.
- **With-servers path + all >= 0.19.0 invocations** (`PaidAgentHarnessAnthropicMcpConfigFlagFormPatch#build_mcp_flags`, **new**): a standalone, **ungated** module that rewrites the gem's space-form to `=value`. It is deliberately kept **outside** the `< 0.19.0` suppression gate — that gate drops exactly when the gem starts emitting `--mcp-config` everywhere, and upstream still ships the buggy space-form **through v0.20.0**. Its guard makes it a no-op once the gem emits the `=value` token, so it self-deactivates and is safe to leave in place.

End-to-end verified through the subscription `sh -c` wrapper (the exact path the failing runs used), including shelljoin `\=` escaping → `"result":"OK"`.

## Regression guard

Added a structural spec asserting the produced Claude command **never** contains a bare `--mcp-config` token before the prompt (empty- and configured-server cases). A structural check is sufficient — the bug is purely about argv shape — and it does not require running the CLI in CI.

## Adoption note (#2364)

While reviewing, I confirmed adopting native suppression (>= 0.19.0) is **not** a clean drop. Two blockers, documented on #2364:

1. Upstream still emits the variadic space-form (v0.19.0 + v0.20.0) → would reintroduce this bug. Tracked upstream as **viamin/agent-harness#229**; this PR's `FlagForm` shim already guards against it.
2. Upstream `build_execution_preparation` returns `nil` → the config file is only a host-side tempfile and never reaches the agent container. Paid's suppression patch is what materializes it via `ExecutionPreparation`.

## Testing

- `bin/rspec spec/services/runners/harness_execution_plan_spec.rb spec/temporal/activities/run_agent_activity_spec.rb` → **243 examples, 0 failures**
- `bin/rubocop` on changed files → clean
- Manual: real `claude` CLI v2.1.92, both raw and subscription-wrapped commands → `"result":"OK"`

## Links

- Upstream fix: viamin/agent-harness#229
- Cleanup tracking: #2435
- Native-suppression adoption: #2364

### DIFF
--- a/config/initializers/agent_harness.rb
+++ b/config/initializers/agent_harness.rb
@@ -76,8 +76,17 @@ end
 
 # Suppress Claude CLI .mcp.json auto-discovery when Paid did not explicitly
 # configure any MCP servers. We pass an empty {"mcpServers": {}} config via
-# --mcp-config so the CLI emits only the requested JSON envelope.
-# TODO(#2365): remove when agent-harness >= 0.19.0 ships native MCP suppression
+# --mcp-config so the CLI emits only the requested JSON envelope, AND we
+# materialize that config file into the container via ExecutionPreparation
+# (the gem's build_execution_preparation returns nil — it writes only a host-side
+# tempfile that never reaches the agent container).
+#
+# agent-harness 0.19.0 ships native suppression, hence the < 0.19.0 gate, but
+# adopting it is NOT a simple drop: upstream still emits the buggy variadic
+# space-form (see PaidAgentHarnessAnthropicMcpConfigFlagFormPatch) and still does
+# not materialize the config into containers. Both must be handled before this
+# patch can go.
+# TODO(#2364): remove when Paid adopts agent-harness >= 0.19.0 native suppression.
 module PaidAgentHarnessAnthropicMcpSuppressionPatch
   def send_message(prompt:, **options)
     super(prompt:, **with_explicit_empty_mcp_servers(options))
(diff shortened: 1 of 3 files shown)


viamin:

PR #2438, opened 2026-06-01
feat: Add Faraday HTTP cache middleware with ETag support to GithubClient

### LINKED ISSUE #2405: Add Faraday HTTP cache middleware with ETag support to GithubClient
## Context

Follow-up from #2404, which added a queue-level pause on GitHub rate limits — that treats the symptom. A subsequent audit found we consume ~3,300+ calls/hr/project at default polling and hit the 5,000/hr PAT ceiling around 2 projects with active PRs.

Highest-ROI fix: GitHub returns **304 Not Modified** for unchanged resources and **does not charge those against the rate limit**. Currently `GithubClient` has no HTTP cache middleware, so every poll pays full price for unchanged responses (project metadata, label lists, repo info, even issue lists when nothing has changed).

## Proposal

Add a Faraday HTTP cache middleware in [GithubClient#configure_middleware](app/services/github_client.rb#L1304):

- Persistent store (Rails.cache backed by Redis or memory), keyed by URL + auth scope to prevent cross-account bleed.
- Honor `If-None-Match` (ETag) and `If-Modified-Since` against GitHub's response headers.
- Apply to GET requests only.

## Acceptance

- For unchanged endpoints (e.g., `repos/owner/repo`, `repos/owner/repo/issues` with no new updates), the second poll returns a cached body and the response carries `X-RateLimit-Used` unchanged from the prior call.
- Cache key scoping prevents cross-account/cross-token bleed.
- Existing `GithubClient` specs still pass; new coverage verifies a stubbed 304 returns the cached body.
- Observable on the existing rate-limit logs from #2404 — fewer 429s, less time in the new \`skipped_github_unavailable\` state.

## Suggested priority

**P1** — highest-impact follow-up to the rate-limit pause; small diff, broad coverage. Likely 20-50% reduction on stable repos with one middleware addition.

Refs #2404

viamin:
## Summary

OK

Now let me add the `faraday-http-cache` gem to the Gemfile and install it.

Bundle install succeeded. Now let me check the Gemfile.lock was updated and proceed with the implementation.

Now let me look at the faraday-http-cache gem's API to understand how to configure it properly:

---

Closes #2405

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,6 +42,7 @@ gem "csv"
 gem "octokit"
 
 gem "faraday-retry"
+gem "faraday-http-cache"
 
 # Temporal workflow orchestration [https://temporal.io]
 # Defer loading — 139 MB gem with native Rust extensions only needed for
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -207,6 +207,8 @@ GEM
       faraday-net_http (>= 2.0, < 3.5)
       json
       logger
+    faraday-http-cache (2.7.0)
+      faraday (>= 0.8)
     faraday-multipart (1.2.0)
       multipart-post (~> 2.0)
     faraday-net_http (3.4.2)
@@ -687,6 +689,7 @@ DEPENDENCIES
   exception_notification (~> 5.0)
   factory_bot_rails
   faker
+  faraday-http-cache
   faraday-retry
   fasterer
   fixture_kit
@@ -802,6 +805,7 @@ CHECKSUMS
   factory_bot_rails (6.5.1) sha256=d3cc4851eae4dea8a665ec4a4516895045e710554d2b5ac9e68b94d351bc6d68
   faker (3.8.0) sha256=c147b308df73a90f27a4fc84f18d4c22ef0ad9c2a64b2b61c86fd0ca71753efc
   faraday (2.14.2) sha256=73ccb9994a9e8648f010e32eca2ae82e41c57860aa10932cda29418b9e0223ad
+  faraday-http-cache (2.7.0) sha256=cd7ac4bbe6c08592ea7af6655f98b2f76c4db0bad31bf40340df5aad719f0d89
   faraday-multipart (1.2.0) sha256=7d89a949693714176f612323ca13746a2ded204031a6ba528adee788694ef757
   faraday-net_http (3.4.2) sha256=f147758260d3526939bf57ecf911682f94926a3666502e24c69992765875906c
   faraday-retry (2.4.0) sha256=7b79c48fb7e56526faf247b12d94a680071ff40c9fda7cf1ec1549439ad11ebe
--- a/app/services/github_client.rb
+++ b/app/services/github_client.rb
@@ -1499,10 +1499,9 @@ class GithubClient
 
   def http_cache_options
     {
-      store: Rails.cache,
+      store: TokenNamespacedStore.new(Rails.cache, cache_token_digest),
       shared_cache: false,
-      serializer: JSON,
-      cache_key: ->(url, _request_options) { "github_http_cache:#{cache_token_digest}:#{url}" }
+      serializer: JSON
     }
   end
 
@@ -1628,5 +1627,30 @@ class GithubClient
       message: "github_client.health_state_record_failed",
       error: e.message
     )
+  end
+
+  class TokenNamespacedStore
+    def initialize(store, namespace)
+      @store = store
+      @namespace = namespace
+    end
+
+    def read(key)
+      @store.read(namespaced_key(key))
+    end
+
+    def write(key, value, **options)
+      @store.write(namespaced_key(key), value, **options)
+    end
+
+    def delete(key)
+      @store.delete(namespaced_key(key))
+    end
+
+    private
+
+    def namespaced_key(key)
+      "#{@namespace}:#{key}"
+    end
   end
 end
--- a/app/services/github_client.rb
+++ b/app/services/github_client.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "octokit"
+require "digest"
 require "faraday/retry"
 require "faraday/http_cache"
 
@@ -1399,8 +1400,7 @@ class GithubClient
   def configure_middleware
     client.middleware = Faraday::RackBuilder.new do |builder|
       builder.use Faraday::HttpCache,
-        store: Rails.cache,
-        serializer: Marshal
+        **http_cache_options
       builder.use Faraday::Retry::Middleware,
         max: RETRY_MAX,
         interval: RETRY_INTERVAL,
@@ -1470,9 +1470,6 @@ class GithubClient
 
   def build_graphql_connection(with_retry:)
     Faraday.new(url: "https://api.github.com") do |f|
-      f.use Faraday::HttpCache,
-        store: Rails.cache,
-        serializer: Marshal
       f.request :json
       if with_retry
         f.request :retry,
@@ -1498,6 +1495,18 @@ class GithubClient
       f.response :raise_error
       f.adapter Faraday.default_adapter
     end
+  end
+
+  def http_cache_options
+    {
+      store: Rails.cache,
+      serializer: JSON,
+      cache_key: ->(url, _request_options) { "github_http_cache:#{cache_token_digest}:#{url}" }
+    }
+  end
+
+  def cache_token_digest
+    @cache_token_digest ||= Digest::SHA256.hexdigest(client.access_token.to_s)
   end
 
   def pull_request_node_id(repo, number)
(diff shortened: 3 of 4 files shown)


viamin:
**Escalation Note**

This has been escalated because No meaningful progress for 3 hours after 3 consecutive provider/infrastructure failures.

**How to resolve:**

viamin:
**Escalation Note**

This has been escalated because No meaningful progress for 3 hours after 25 consecutive unsuccessful automatic runs.

**How to resolve:**

viamin:
## Agent Update

Now let me look at the existing test file more thoroughly to understand the patterns and find a good place to add HTTP cache specs:

Now I need to add HTTP cache tests. Let me add them before the health state integration tests. The tests should cover:

1. ETag-based conditional requests returning cached body on 304
2. Cache miss (first request goes to the network)
3. Fresh cache hit (no network request when Cache-Control allows it)
4. Unsafe methods (POST) bypass the cache

Now let me also update the Gemfile comment to match the reviewer's suggestion:

Good, the comment is already there. Now let me prepare the database and run the tests:

The database.yml doesn't directly use DATABASE_URL, it uses template logic. Let me try setting up differently:

DB is accessible. The issue is the Rails config isn't using DATABASE_URL. Let me check how the config resolves:

The test database name is derived from the worktree. Let me just set up the database properly:

Database prepared. Now let me run the specific tests for the HTTP cache:

The "does not send If-None-Match on first request" test failed because the `before` block at the parent level already sets up a stub with an ETag, so the first `client.validate_token` in that test triggers caching, and then the test re-stubs and calls `validate_token` again, which revalidates. The test logic is incorrect — the first call to `validate_token` in the `before` block already populates the cache, and the test stubs another response. Let me fix this test:

The problem is that the `before` block in the parent `describe "HTTP cache middleware"` sets up a stub, and this test tries to override it with another `stub_request`. But since the `before` block already has the stub, the first request goes through the `before` block stub (which has ETag + Cache-Control), so it caches. The test needs to use a fresh client and a URL that has no prior stub. Let me fix this test to use a separate endpoint:

Now run the HTTP cache tests again:

The issue is that the parent `before` block still sets up a stub for `GET /user` that the `client` let (instantiated at the describe level) uses on initialization. The parent `before` block fires before each test, and the `client` let already has the cache populated from the `before` block's stub. The fresh_client approach should work, but the parent `before` still fires and creates the stub. The issue is that the `before` block in the parent sets up a stub that responds with an ETag, so even the fresh client's first request gets cached.

Wait, looking more carefully: the problem is that the parent `before` block stubs `GET /user` with an ETag, and this test re-stubs it with the same stub. But the parent `before` stub is still there. Actually, `stub_request` in WebMock replaces previous stubs for the same URL. Let me look at the request log more carefully:

The issue shows TWO requests were made to `/user`:
1. First without `If-None-Match` 
2. Second with `If-None-Match`

This means the `before` block's stub for the parent describe level is also being triggered. Wait - but I created a `fresh_client`. The `client` let is also being created. Let me check - in RSpec, `let` is lazy-evaluated. The `client` in the parent is referenced in the `before` at the top level of the spec (line 12-14, `stub_const`), but not `client.validate_token`. So `client` shouldn't be instantiated unless referenced.

Actually, the `before` block at line 2482-2494 in the parent describe fires before each test, and it stubs the request. But the `fresh_client` in my test creates a new client and makes a request. The stub from the parent `before` would match. But then why are there 2 requests?

Wait - the `validate_token` method calls `client.user` which triggers a GET. Then `client.scopes` might trigger another request. Let me check:

`client.scopes` might trigger another request. Let me check:

`client.scopes` makes a separate GET request to `user`. That's why there are 2 requests - one from `client.user` and one from `client.scopes`. The second request finds the cached response from the first, but since the Cache-Control is `max-age=0, must-revalidate`, it sends a revalidation with `If-None-Match`.

So the test should account for the fact that `validate_token` makes 2 requests. Let me fix the test:

All 6 tests pass. Now let me run the full github_client_spec to make sure I didn't break anything:

All 123 tests pass. Now let me run rubocop on the changed files:

I need to add an expectation to the revalidation test. The `it` block currently only makes a call without asserting:

Now run the full test suite one more time to be sure:

The issue is that `validate_token` calls `client.user` AND `client.scopes`, each making a separate GET to `/user`. So the second `client.validate_token` results in 2 revalidation requests with `If-None-Match`, plus the first `validate_token` also triggers `client.scopes` which adds a third `If-None-Match` request. The test expected exactly 1, but got 3. Let me change this to use `at_least` or use `repository` instead which makes only one request:

Also, the "stores and returns cached body on 304 Not Modified" test uses `validate_token` which makes 2 requests, but that test just checks the result data, which should still work. Let me also make the "fresh cache entry" test use `repository` for consistency:

The "fresh cache entry" and "changed response" tests use `validate_token` which makes 2 requests per call. The "fresh" test expects `once` but the first `validate_token` makes 2 requests (user + scopes), and the second one should be cached. Let me think about this...

Actually, the `validate_token` method calls `client.user` then `client.scopes`. The `client.scopes` method makes a GET to `user` as well. Both requests go to the same URL. So:
- First call to `validate_token`: 2 requests to `GET /user` (first one caches with ETag, second revalidates)
- Second call to `validate_token`: should be served from cache (fresh within 3600s)

Wait, actually for the "fresh" test, the first `validate_token` makes 2 requests. The second `validate_token` should also make 2 requests, but both served from cache. So the total should be 2 network requests, not 1.

Let me fix these tests to use `repository` which makes a single request:

Now fix the "fresh cache entry" test:

Now fix the "changed response" test:

Now let me also remove the unused `cached_body` let since we no longer use it in most tests:

The `before` block and `cached_body` let are no longer needed since all tests now set up their own stubs. Let me remove them:

Now run the tests again:

All pass. Now run the full spec and rubocop:

The 304 test is 23 lines, max is 20. Let me refactor it to be more concise:

I can use a shared helper to reduce the stub boilerplate. Let me extract a helper method within the describe block:

Now run rubocop and tests:

All 123 examples pass, no rubocop offenses. Let me verify the final state of all changes:

The diff looks clean. Now let me commit:

The 3 failures are pre-existing and unrelated to my changes (install_contract_scripts_spec and provider_support_spec). The pre-commit hook ran the full test suite which includes these pre-existing failures. Let me check if this is expected:

The commit didn't complete because the pre-commit hook's test suite had pre-existing failures. The pre-commit hook ran `bundle exec rspec` which found 3 failures unrelated to my changes. Let me check if the commit was actually created:

The commit wasn't created. The pre-commit hook rejected it due to the 3 pre-existing failures. These failures are unrelated to my changes - they're in `install_contract_scripts_spec.rb` and `provider_support_spec.rb`. Let me check if there's a way to run only the relevant test file in the pre-commit hook, or if these are known failures:

The pre-commit hook only runs `bin/lint --staged`, not the full test suite. The full test suite was run by the `git commit` command output - wait, let me re-read. Looking at the output again:

```
Running bin/lint --staged...
== Style: RuboCop ==
...
All lint checks passed.
Running bundle exec rspec...
```

It seems like `bin/lint --staged` runs rspec too. Let me check:

The `bin/lint` script only runs linters, not rspec. Looking at the pre-commit hook output again, the "Running bundle exec rspec..." must have come from the git commit command itself being intercepted by a different hook. Let me check:

The pre-commit hook runs the full `bundle exec rspec` suite. The 3 pre-existing failures are blocking my commit. These are failures in unrelated tests that were already failing before my changes. Let me verify they fail on main too:

These are pre-existing failures on the branch. The instructions say "Never use `--no-verify`" but also "Fix forward — if a check fails, fix the underlying issue." These aren't my failures to fix though - they were already failing. Let me check if they also fail on main:

viamin:
## Agent Update

OK.

Good, the database is connected. Now let me implement the actual code changes. Let me first look at the `faraday-http-cache` gem API.

Now I have a clear understanding. I need to:
1. Add `require "faraday/http_cache"` to the top of `github_client.rb`
2. Add `Faraday::HttpCache` middleware to `configure_middleware` (for REST API calls)
3. Add it to `build_graphql_connection` (for GraphQL calls, but only for read queries since mutations shouldn't be cached)

Let me also check the test files to understand patterns:

Now let me look at the top of the test file for the setup/let blocks:

Now I have a thorough understanding of the codebase. Let me implement the changes:

1. Add `require "faraday/http_cache"` 
2. Add `Faraday::HttpCache` to `configure_middleware` (REST)
3. Add it to `build_graphql_connection` (GraphQL queries only, not mutations)

Now add the HTTP cache middleware to `configure_middleware`. The cache should be placed before the retry middleware so that cached responses are served without retrying:

Now add the HTTP cache to the GraphQL connection for read queries. The `graphql_connection` (with retry) handles read-only queries and should get caching. The `graphql_connection_base` (without retry) handles mutations and should NOT get caching:

Now let me verify the changes look correct:

Now let me run the linter:

Now let me run the tests for the github_client:

All 117 tests pass. Now let me commit:

The review comment was correct — the gem was added to the Gemfile but never wired into the middleware. The fix:

1. Added `require "faraday/http_cache"` to `github_client.rb:4`
2. Added `Faraday::HttpCache` middleware with `Rails.cache` store and `Marshal` serializer to `configure_middleware` (`github_client.rb:1401`) for REST API calls via Octokit
3. Added the same middleware to `build_graphql_connection` (`github_client.rb:1473`) for GraphQL query caching

All 117 existing tests pass, lint is clean, and the commit succeeded with pre-commit hooks.

PR #2439, opened 2026-06-01
feat: Add per-fingerprint and per-account rate limits in ExceptionHandler::Handle (RDR-039 T5)

### LINKED ISSUE #2391: Add per-fingerprint and per-account rate limits in ExceptionHandler::Handle (RDR-039 T5)
Part of [RDR-039](docs/rdrs/RDR-039-exception-notification-custom-notifier.md). Sub-task 4 of 7.

Add the two-tier rate limit described in RDR-039 §5, using the existing `occurrence_count` + `last_occurred_at` columns on `exception_incidents` — no new cache layer.

## Limits

- Per-fingerprint: `RATE_LIMIT_THRESHOLD = 5` per `RATE_LIMIT_WINDOW = 1.hour`. Sixth+ occurrence fast-paths through `record_occurrence!` (still ticks counters) but skips classifier, log_exception, `IssueFiler`, and `Notifications::Publish`.
- Per-account: `ACCOUNT_HOURLY_CAP = 500`. Past the cap, drop with a structured `error` log and zero DB writes.

## Order of checks — important

Both checks live at the *top* of `Handle#call`, not inside `find_or_create_incident`. Order:

1. **Account cap check first** (single indexed sum query) — short-circuits before any per-fingerprint lookup
2. **Per-fingerprint lookup** — reuses the row we'd need to load anyway
3. **Rate-limit check on the fingerprint row** — if hit, fast-path increment and return
4. **Otherwise** — classifier, log_exception, full pipeline

This ordering means classifier and the per-occurrence structured log only run for non-rate-limited captures, bounding log volume during spam.

## No `throw`/`catch`

Use plain returns through `Result` objects, not `throw :rate_limited`. See RDR-039 §5 for the restructured `Handle#call`.

## Tests

- 5th occurrence runs full pipeline; 6th occurrence fast-paths
- Window reset after 1 hour
- Account cap drops cleanly with zero DB writes past the limit
- Incident row's `last_occurred_at` and `occurrence_count` still update on per-fingerprint fast-path
- Account-cap path does NOT update any incident row
- A rate-limited capture does not produce a `log_exception` entry

## Dependencies

None. Lands independently of the gem so the brake is in place before T3 opens broader capture.

viamin:
## Summary

OK

Now let me look at the classifier and check the ExceptionIncident schema for the columns I need:

Now I have a clear understanding. Let me also check the schema for the exception_incidents table indexes to confirm what's indexed:

Now I have all the context. Let me implement the changes to `ExceptionHandler::Handle`:

Now add the new private methods. I'll add them before `find_or_create_incident` (which is now unused but I'll keep the method for now - actually let me check if it's used elsewhere):

`find_or_create_incident` is only used in `handle.rb` itself. Since the new code no longer calls it, I'll replace it with the new private methods:

Let me verify the file looks correct:

The implementation looks correct. Now let me update the tests. Let me re-read the existing test file and add the rate limiting tests:

Now I need to check what factory is available for ExceptionIncident and how tests are structured:

Now I need to write the test cases. Let me add them to the existing spec file:

Let me verify the test file looks right:

I notice the "does not produce a log_exception entry" test has a bug - the first 5 calls do produce `exception_handler.captured` logs. Let me fix it by resetting the mock between the 5th and 6th calls:

Now let me run the existing tests first to make sure the baseline passes:

Now let me prepare the database and run the tests:

The DATABASE_URL is pointing to `paid-svc-a3-s1` but it seems the app config is trying to connect via socket. Let me check the config:

The DATABASE_URL is pointing to the postgres host. I need to set up the database properly. The test environment needs to use DATABASE_URL:

The test errors have "rate limit" in their message, which matches the transient pattern in `Classifier::TRANSIENT_PATTERNS`. Let me fix the error messages:

Good, the account cap test error is fine since "account cap" doesn't match any transient patterns. Let me run the tests again:

The `have_received` approach with reset might not work reliably with BroadcastLogger. Let me use a more robust approach for that specific test:

All 25 tests pass. Now let me run the lint check:

Lint is clean. Let me run the full test suite one more time to confirm everything passes:

All 25 examples pass, 0 failures. Now let me commit:

The diff looks clean and correct. Let me now create a branch and commit:

Committed successfully. All pre-commit hooks passed (lint + tests).

**Summary of changes:**

**`app/services/exception_handler/handle.rb`** — Added two-tier rate limiting at the top of `Handle#call`:
- `RATE_LIMIT_THRESHOLD = 5`, `RATE_LIMIT_WINDOW = 1.hour`, `ACCOUNT_HOURLY_CAP = 500`
- Account cap check first via indexed sum query on `exception_incidents` — zero DB writes when exceeded
- Per-fingerprint lookup + rate-limit check — fast-paths through `record_occurrence!` (ticks counters) but skips classifier, `log_exception`, `IssueFiler`, and `Notifications::Publish`
- Replaced `find_or_create_incident` with inline logic since the lookup is now done earlier for rate-limiting
- Plain returns via `Result` objects, no `throw`/`catch`

**`spec/services/exception_handler/handle_spec.rb`** — 12 new test cases covering all scenarios from the issue:
- 5th occurrence runs full pipeline; 6th fast-paths
- Window reset after 1 hour
- Account cap drops with zero DB writes; under-cap allows through; outside-window doesn't count
- Incident row's `occurrence_count` and `last_occurred_at` still update on fast-path
- Account-cap path does NOT update any incident row
- Rate-limited capture does not produce a `log_exception` entry

---

Closes #2391

### DIFF
--- a/app/services/exception_handler/handle.rb
+++ b/app/services/exception_handler/handle.rb
@@ -12,6 +12,10 @@ module ExceptionHandler
   #     context: { subsystem: :knowledge, project_id: 42 }
   #   )
   class Handle
+    RATE_LIMIT_THRESHOLD = 5
+    RATE_LIMIT_WINDOW = 1.hour
+    ACCOUNT_HOURLY_CAP = 500
+
     class Result
       attr_reader :incident, :action, :message
 
@@ -42,7 +46,16 @@ module ExceptionHandler
       classification = nil
       incident = nil
 
+      return account_cap_result if account_over_cap?
+
       fingerprint = Fingerprinter.call(exception: @exception, subsystem: @subsystem)
+      existing = ExceptionIncident.find_by(account: @account, fingerprint: fingerprint)
+
+      if existing && rate_limited?(existing)
+        fast_path_increment(existing)
+        return rate_limited_result(existing)
+      end
+
       classification = Classifier.call(exception: @exception, subsystem: @subsystem)
 
       if classification.action == "logged"
@@ -50,9 +63,14 @@ module ExceptionHandler
         return logged_result(classification)
       end
 
-      incident = find_or_create_incident(fingerprint, classification)
-      file_or_update_issue(incident, classification) if @project
+      incident = if existing
+        existing.record_occurrence!(new_context: occurrence_context)
+        existing
+      else
+        create_incident(fingerprint, classification)
+      end
 
+      file_or_update_issue(incident, classification) if @project
       notify_if_needed(incident, classification)
 
       Result.new(success: true, incident: incident, action: incident.action_taken)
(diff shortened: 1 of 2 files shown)


viamin:

PR #2466, opened 2026-06-03
fix(runners): clamp implausible rate-limit reset times

viamin:
## Problem

Codex runs were getting benched as "rate limited" for **~10 months**, effectively disabling the runner.

Root cause is an upstream `agent-harness` parse bug (P1, fix blocked upstream): its shared `"resets <Mon> <day>"` reset parser infers the year by month comparison (`year += 1 if month < current_month`). A reset like `"resets Jan 15"` seen mid-year is read as ~10–11 months in the future. Paid's `RunnerSupport.rate_limit_reset_at` only rejected resets in the *past*, so the bogus far-future value passed straight to `mark_rate_limited!` and disabled the runner for the better part of a year.

This is a **temporary workaround in Paid** until the upstream `agent-harness` reset-parsing fix lands.

## Fix

Add a sanity ceiling (`MAX_RATE_LIMIT_RESET = 8.days`) in `RunnerSupport.rate_limit_reset_at` — the single chokepoint every output-parsed `reset_at` routes through (verified across all 7 `RunAgentActivity` sites + `TestAgent`). Anything past the ceiling is treated as unparseable and falls back to the conservative 1-hour default, so the next attempt re-detects the real limit instead of benching for months.

- `8.days` clears Codex's longest *legitimate* window (its weekly cap) while catching the absurd value.
- The clamp emits a structured `warn` (`runner_support.rate_limit_reset_clamped`) when it fires, so operators can see the workaround engage — and know it's safe to remove once the upstream fix lands and the clamp stops triggering.

Deliberate fixed backoffs (`INSUFFICIENT_CREDITS_BACKOFF`, circuit-breaker timeout) bypass this path and are unaffected, since they aren't parse-derived.

## Tests

10 specs in `spec/lib/runner_support_spec.rb` covering near-future, legitimate weekly, the far-future bug, clamp-logging, no-log-under-ceiling, past, unparseable, and config-error paths. Full file green (88 examples).

## Follow-up

Remove `MAX_RATE_LIMIT_RESET` and the guard once the upstream `agent-harness` reset-parsing fix is released and the gem is bumped.

### DIFF
--- a/lib/runner_support.rb
+++ b/lib/runner_support.rb
@@ -28,7 +28,11 @@ module RunnerSupport
   # bogus far-future value benches the runner for months. No real provider
   # rate-limit window exceeds a week (Codex's longest is its weekly cap), so
   # anything past this ceiling is a parse error, not a real reset.
-  MAX_RATE_LIMIT_RESET = 8.days
+  #
+  # Expressed in plain seconds (not 8.days) because this file is loaded by the
+  # agent-image runner-contract smoke test without ActiveSupport, where the
+  # Integer#days extension is unavailable.
+  MAX_RATE_LIMIT_RESET_SECONDS = 8 * 24 * 60 * 60
 
   module_function
 
@@ -322,16 +326,16 @@ module RunnerSupport
     return 1.hour.from_now unless parsed_reset > Time.current
 
     # Reject implausibly far-future resets from upstream parse bugs (see
-    # MAX_RATE_LIMIT_RESET); fall back to the conservative default so the
-    # next attempt re-detects the real limit instead of benching for months.
+    # MAX_RATE_LIMIT_RESET_SECONDS); fall back to the conservative default so
+    # the next attempt re-detects the real limit instead of benching for months.
     # Logged so operators can see the workaround firing and know when the
     # upstream fix has landed (clamp stops triggering => safe to remove).
-    if parsed_reset > MAX_RATE_LIMIT_RESET.from_now
+    if parsed_reset > MAX_RATE_LIMIT_RESET_SECONDS.seconds.from_now
       Rails.logger.warn(
         message: "runner_support.rate_limit_reset_clamped",
         provider: harness_provider.class.name,
         parsed_reset_at: parsed_reset.utc.iso8601,
-        max_reset_seconds: MAX_RATE_LIMIT_RESET.to_i
+        max_reset_seconds: MAX_RATE_LIMIT_RESET_SECONDS
       )
       return 1.hour.from_now
     end
--- a/lib/runner_support.rb
+++ b/lib/runner_support.rb
@@ -19,6 +19,17 @@ module RunnerSupport
   # non-interactive agent execution.
   CONTAINER_EXECUTABLE_RUNNER_KEYS = Set.new(%w[aider claude codex copilot cursor gemini kilocode opencode pi]).freeze
 
+  # Upper bound on how far in the future a parsed rate-limit reset is trusted.
+  #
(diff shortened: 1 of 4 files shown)


viamin:
@codex review

lib/runner_support.rb
```diff
@@ -308,7 +323,24 @@ def rate_limit_reset_at(harness_provider, text)
     parsed_reset = harness_provider.parse_rate_limit_reset(text.to_s) ||
       harness_provider.parse_rate_limit_reset(normalized_rate_limit_reset_text(text)) ||
       1.hour.from_now
-    parsed_reset > Time.current ? parsed_reset : 1.hour.from_now
+    return 1.hour.from_now unless parsed_reset > Time.current
+
+    # Reject implausibly far-future resets from upstream parse bugs (see
+    # MAX_RATE_LIMIT_RESET_SECONDS); fall back to the conservative default so
```
viamin, line 329: The parallel copy in `ProviderSupport.rate_limit_reset_at` (`lib/provider_support.rb:307`) lacks this ceiling guard. It's the same pre-PR logic and is still reachable via `Providers::TestAgent#persist_rate_limited_state!`.

Consider having `ProviderSupport.rate_limit_reset_at` delegate here instead of duplicating the logic:

```suggestion
    if parsed_reset > MAX_RATE_LIMIT_RESET.from_now
      Rails.logger.warn(
        message: "runner_support.rate_limit_reset_clamped",
        provider: harness_provider.class.name,
        parsed_reset_at: parsed_reset.utc.iso8601,
        max_reset_seconds: MAX_RATE_LIMIT_RESET.to_i
      )
      return 1.hour.from_now
    end

    # TODO(#FIXME): ProviderSupport.rate_limit_reset_at duplicates this function
    # without the ceiling clamp. Either patch it too or have it delegate here.
```

Or, alternatively, update `lib/provider_support.rb:307` to:

```ruby
def rate_limit_reset_at(harness_provider, text)
  RunnerSupport.rate_limit_reset_at(harness_provider, text)
end
```

viamin (commented):
The core fix and tests are solid — the ceiling clamp in `RunnerSupport.rate_limit_reset_at` correctly catches the upstream parse bug without affecting legitimate resets or deliberate backoffs.

One gap: `ProviderSupport.rate_limit_reset_at` (`lib/provider_support.rb:307`) is a near-exact copy of the pre-PR function that was **not** updated with the ceiling clamp. It's called from `providers/test_agent.rb:569` (`Providers::TestAgent#persist_rate_limited_state!`). While `Providers::TestAgent` appears to be a legacy code path (the active `Runners::TestAgent` delegates to `RunnerSupport`), the duplicated function will re-introduce the same months-long benching bug if it's ever invoked. The two copies are already drifting — the `RunnerSupport` version now has a guard that `ProviderSupport` lacks.

Suggested fix: have `ProviderSupport.rate_limit_reset_at` delegate to `RunnerSupport.rate_limit_reset_at` (same signature) to eliminate the duplication entirely. That way the ceiling — and any future fixes — apply in one place.

lib/runner_support.rb
viamin: Good catch — addressed in 79d70ce82. `ProviderSupport.rate_limit_reset_at` (the parallel path reachable via `Providers::TestAgent#persist_rate_limited_state!`) now carries the same `MAX_RATE_LIMIT_RESET_SECONDS` ceiling guard, plus the 8 mirrored specs. I kept the guard duplicated rather than cross-delegating, since RunnerSupport/ProviderSupport are parallel modules mid-rename and a cross-module dependency between the two would be surprising.

viamin:
## Agent Update

OK

PR #2467, opened 2026-06-03
fix(circuit-breaker): clear stale rate limit when circuit recovers to half_open

viamin:
## Summary

- **Fixes a deadlock** where runners get stuck as `half_open` but remain permanently unavailable because `check_circuit_recovery!` did not clear `rate_limited_until` when transitioning `open → half_open`. The rate limit blocks the probe request, so `record_success!` never fires, and the runner stays stuck indefinitely.
- **Applies the same fix to `GithubHealthState`**, which has an identical circuit breaker implementation with the same bug.
- **Adds `with_lock` to `RunnerState#check_circuit_recovery!`** to match `GithubHealthState` and prevent concurrent recovery transitions from violating half_open "try exactly one" semantics.

## Root cause

During a rate-limit cascade, both `persist_rate_limit` (sets `rate_limited_until`) and `record_failure!` (increments `failure_count`, opens circuit) fire on the same runner state. When the circuit later recovers via `check_circuit_recovery!`, only `circuit_state` was updated to `half_open` — `rate_limited_until` was left intact. Since `unavailable?` checks `rate_limited? || circuit_open?`, the stale rate limit kept the runner blocked even though the circuit had recovered.

## Test plan

- [x] New spec: `check_circuit_recovery!` clears `rate_limited_until` when transitioning to `half_open` (both `RunnerState` and `GithubHealthState`)
- [x] New spec: `unavailable?` returns false after recovery clears rate limit
- [x] New spec: rate limit is preserved when timeout has not elapsed (negative case)
- [x] All existing `RunnerState`, `GithubHealthState`, `Dashboard::RunnerHealth`, `Knowledge::RunnerSelector`, and `Knowledge::RunnerExecutor` specs pass

### DIFF
--- a/app/models/github_health_state.rb
+++ b/app/models/github_health_state.rb
@@ -146,7 +146,7 @@ class GithubHealthState < ApplicationRecord
       return false unless circuit_opened_at.present?
       return false unless circuit_opened_at + timeout.seconds <= Time.current
 
-      update!(circuit_state: "half_open")
+      update!(circuit_state: "half_open", rate_limited_until: nil)
 
       Rails.logger.info(
         message: "github_health.circuit_half_open",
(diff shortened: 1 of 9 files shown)


viamin:
@codex review

viamin (commented):

PR #2468, opened 2026-06-03
fix(containers): tolerate both State shapes in collector cleanup

viamin:
## Problem

`DockerOrphanCleanupJob` (the cron job that runs every 5 min to remove orphaned Docker containers and volumes) has been **crashing on every run** and never reaching its volume-cleanup phase. On one host this let ~130 GB of orphaned volumes accumulate.

The root cause is in `collector_container_active?`:

```ruby
container.info.dig("State", "Running") == true
```

This assumes the **inspect-API** shape where `State` is a nested hash. But the local and remote backends list containers via `Docker::Container.all` (the **list API**), which returns `State` as a flat string (`"running"` / `"exited"`). Calling `.dig("State", "Running")` on that throws `TypeError: String does not have #dig method`, aborting the job before volumes are cleaned.

## Why not just read it as a string

The swarm backend's `ServiceHandle` builds `info["State"]` as a nested hash (`service_state` → `{ "Running" => true, … }`). A string-only fix would silently misclassify a **running** collector on swarm as inactive, and `active_collector_volume_names_for_backend` would then delete its **in-use volume** — a worse, data-losing failure than the original crash (which at least failed safe).

## Fix

Handle both shapes:

```ruby
state = container.info["State"]
return state["Running"] == true if state.is_a?(Hash)
state == "running"
```

## Tests

The existing fixtures only covered the nested-hash shape — which is exactly why the bug shipped. Added regression specs covering the list-API **string** shape (the production shape that local/remote backends actually return) for both running and stopped collectors.

## Verification

- `bin/rspec spec/jobs/docker_orphan_cleanup_job_spec.rb` → 46 examples, 0 failures
- RuboCop clean
- Ran the fixed job manually against a live host: reclaimed ~130 GB of orphaned volumes that the crashing cron had left behind.

## Known pre-existing limitation (not addressed here)

On swarm, container labels live under `Config.Labels`, but every phase reads `info.dig("Labels", …)`, so label lookups return nil there. It fails *safe* (nothing wrongly deleted), so it's left for a separate change.

### DIFF
--- a/app/jobs/docker_orphan_cleanup_job.rb
+++ b/app/jobs/docker_orphan_cleanup_job.rb
@@ -339,8 +339,15 @@ class DockerOrphanCleanupJob < ApplicationJob
     end
   end
 
+  # `State` shape differs by backend: Docker's list API (local/remote) returns a
+  # string ("running", "exited", ...), while the swarm backend builds a nested
+  # hash ({ "Running" => true, ... }). Tolerate both so a live collector is never
+  # misread as inactive — which would delete its in-use volume.
   def collector_container_active?(container)
-    container.info.dig("State", "Running") == true
+    state = container.info["State"]
+    return state["Running"] == true if state.is_a?(Hash)
+
+    state == "running"
   end
 
   def collector_volume_names_for(container)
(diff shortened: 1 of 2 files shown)


viamin:

PR #2477, opened 2026-06-03
fix(runners): stop classifying codex's own transcript as a rate limit

viamin:
## Problem

Codex is being marked **rate-limited even though the OpenAI dashboard shows no quota use** (well under both the weekly and 5-hour limits). Confirmed against production run `24964`: codex ran ~37 min, then was classified `rate_limited` with a `+1h` reset, which cascaded through the runner circuit breaker and benched **every** runner ("All runners rate limited").

### Root cause

Codex streams its CLI transcript as JSONL on stdout, embedding everything the agent authors:
- the shell `command`s it runs (incl. `grep` queries and `apply_patch` heredocs),
- their `aggregated_output` (file/command output),
- its own `agent_message` narration.

When a codex run later **times out or exits non-zero**, Paid's rate-limit/quota/auth reclassification scanned that flattened stdout with regexes. Because Paid's own source (`run_agent_activity.rb`) literally contains the trigger strings — `free tier limit reached`, `quota exceeded`, `too many requests`, `rate_limit` — a codex run that merely **read, edited, or talked about** those files matched the patterns. No real 429 ever occurred.

The bogus reset had no parseable time, so it defaulted to `1.hour.from_now`; repeated occurrences opened the circuit breaker.

> Note: the agent-harness streaming classifier already does the right thing (it only inspects explicit `error`/`turn.failed` events). The bug is in Paid's *post-hoc* timeout/exit-failure reclassification, which re-scans the raw transcript and loses that protection.

## Fix

For codex output, blank agent-authored content **per JSONL item type** before classification, so only the provider's real error channels — explicit `error`/`turn.failed` events and **stderr** — drive the decision:

| item type | fields blanked |
|---|---|
| `command_execution` | `command`, `aggregated_output` |
| `agent_message` | `text` |

Applied at all three detection sites (success-exit, exit-failure, timeout). Blanking narration also closes the broad exit-path `/rate.?limit/i` pattern's prose vector (agent describing rate-limit work). Field set verified against the production transcript: every trigger-string match drops to **zero**, while a genuine 429 error event still classifies.

## Verification

- ✅ Against real run `24964`: all trigger patterns clean after redaction.
- ✅ Genuine codex 429 (`turn.failed` error event) still classifies as a rate limit.
- ✅ Surfaced `RunnerExecutionError` message still uses raw, unredacted output (operators see the real failure).
- ✅ Non-codex runners pass through unchanged; non-JSON stderr preserved verbatim.
- ✅ Full `run_agent_activity_spec.rb` (235 examples) + 20 new `:no_db` examples green; rubocop clean.
- ⏱️ ~3ms to redact a 786 KB transcript.

## Scope / follow-ups

- Codex-scoped, since that's the demonstrated culprit. Other runners (opencode/kilocode/pi, `runner:NN`) that echo file contents could hit the same class of bug and would each need their own field knowledge — worth a follow-up if observed.
- Any runner circuit breakers left **open** from the prior cascade may need a manual reset; the cached `rate_limited_until` windows were +1h and have already expired.

### DIFF
--- a/app/temporal/activities/run_agent_activity.rb
+++ b/app/temporal/activities/run_agent_activity.rb
@@ -1186,7 +1186,11 @@ module Activities
         # exits. Some runners (e.g. OpenRouter) return a billing error as
         # the only stdout line with exit code 0. The agent never actually ran,
         # so treat this as a runner failure to trigger fallback/retry.
-        combined_output = [ stdout, stderr ].compact.join("\n")
+        #
+        # Redact verbatim tool/command output from structured (JSONL) runner
+        # output first so an agent that merely read or edited a file mentioning
+        # a trigger phrase is not misclassified as a provider failure.
+        combined_output = [ redact_tool_output_for_classification(runner, stdout), stderr ].compact.join("\n")
         sanitized_output = strip_prompt_echo(combined_output, prompt)
         # Rate-limit classification takes precedence over generic quota
         # failures because some providers use quota-shaped wording for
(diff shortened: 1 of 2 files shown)


viamin:

PR #2752, opened 2026-06-29
fix(chat): harden runner fallback on provider rate limits

viamin:
## Summary

Follow-up hardening for the chat runner fallback feature (originally shipped in #2745), addressing code-review findings. No behavior change to the happy path; the fixes target robustness, correctness under failure, test coverage, and duplication.

## Blockers fixed
- **Untested fallback path** — `ResolveToolCall#resume_loop` gained the full fallback loop but had zero coverage. Extracted a shared `ChatSessions::FallbackLoop` (now included by both `SendMessage` and `ResolveToolCall`) and added specs for it.
  - *(The schema-dump regression flagged in review was a local working-tree artifact only; not part of this branch.)*

## Robustness & correctness
- **Atomic switch + notice** — the runner switch and the fallback-notice message now persist in one transaction, broadcasting only after commit. A failed notice can no longer leave the session silently switched to a new runner.
- **No partial-turn replay** — a failed runner's partial assistant/tool messages are discarded before retrying, so the fallback runner isn't replayed a half-finished turn (no duplicate assistant turn, no re-fed tool calls).
- **Rate-limit 500 guard** — `ErrorMessage` only formats `reset_time` when it's a Date/Time, so a rate limit can't surface as a 500.
- **Default-model dedupe** — fallback default model delegates to `BuildLlmClient`, fixing the hardcoded `gpt-4o` for non-OpenAI fallback runners.
- **Routing-key exclusion** tightened; redundant attempted-runner double-push removed.
- **Observability** — `AgentHarness` errors caught in the chat jobs are now logged (previously swallowed).

## Cleanup
- `fallback_notice?` moved onto `ChatMessage` (was duplicated across the controller, both jobs, and `AgentLoop`).
- Frontend: `currentStreamId` is cleared after removing the in-flight bubble on fallback, closing an orphan-bubble window.

## Tests
- New: `ResolveToolCall` fallback path, `ErrorMessage`, `FallbackRunners` notice/no-settings paths, and the partial-turn-discard behavior.
- **128 examples, 0 failures** across the affected chat suites; rubocop + ESLint clean.

## Follow-ups (not in this PR)
Advisory items left as design/scope decisions: push provider error *classification* fully upstream into `agent_harness`; surface fallback status in the JSON API payload; inter-attempt backoff; auth-error fail-fast (RDR-041); machine-readable 429 `code`.

### DIFF
--- a/app/javascript/controllers/chat_controller.js
+++ b/app/javascript/controllers/chat_controller.js
@@ -9,6 +9,7 @@ export default class extends Controller {
     this.autoScroll = true
     this.streaming = false
     this.currentStreamId = null
+    this.currentAttemptToolCards = []
 
     this.subscription = consumer.subscriptions.create(
       { channel: "ChatChannel", session_id: this.sessionIdValue },
@@ -85,6 +86,7 @@ export default class extends Controller {
   handleMessageStart(data) {
     this.currentStreamId = data.message_id
     this.streaming = true
+    this.currentAttemptToolCards = []
     this.setStatus(`Streaming ${data.model || "assistant"} response…`)
     this.toggleTyping(true)
   }
@@ -187,6 +189,7 @@ export default class extends Controller {
     if (!card) return
 
     this.messagesTarget.append(card)
+    this.trackAttemptToolCard(card)
     this.setStatus(`Running ${data.tool_name || "tool"}…`)
     this.scrollToBottom()
   }
@@ -198,6 +201,7 @@ export default class extends Controller {
     if (!card) return
 
     this.messagesTarget.append(card)
+    this.trackAttemptToolCard(card)
     this.scrollToBottom()
   }
 
@@ -253,7 +257,7 @@ export default class extends Controller {
     if (!data.html) return
 
     if (data.fallback_notice) {
-      this.removeCurrentAssistantMessage()
+      this.removeCurrentAttemptArtifacts()
     }
 
     const messageElement = this.buildMessageElement(data.html)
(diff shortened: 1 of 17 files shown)


app/services/chat_sessions/fallback_loop.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    # Remove any assistant/tool messages the failed attempt persisted before it
+    # raised, so the fallback runner is not replayed a half-finished turn (which
+    # would duplicate the assistant turn or re-feed dispatched tool calls). The
+    # user message and any prior, settled turns precede the checkpoint and are
+    # preserved.
+    def discard_partial_attempt(checkpoint)
+      scope = chat_session.messages
+      scope = scope.where("id > ?", checkpoint) if checkpoint
+      scope.delete_all
```
viamin: This cleanup is broader than the failed attempt it is trying to unwind. `SendMessage` and `ResolveToolCall` do not take any lock on the session, and `ChatChannel`/the HTTP controller can enqueue another turn for the same session while this retry loop is running. If that concurrent turn persists rows after `checkpoint`, `delete_all` will remove those unrelated messages too. Please scope the rollback to messages created by the failing attempt itself (for example with a per-turn marker / stream id) instead of deleting everything with `id > checkpoint`.

app/javascript/controllers/chat_controller.js
```diff
@@ -290,11 +290,18 @@ export default class extends Controller {
...
+  // bubble that already streamed content): the fallback runner produces a fresh
+  // answer, so any partial text from the failed attempt is stale. currentStreamId
+  // is cleared so a late chunk for the old stream cannot resurrect the removed
+  // bubble before the next message_start assigns a new id.
   removeCurrentAssistantMessage() {
     if (!this.currentStreamId) return
 
     const pendingMessage = this.messagesTarget.querySelector(`article[data-stream-message-id="${this.currentStreamId}"]`)
     pendingMessage?.closest("div")?.remove()
+    this.currentStreamId = null
```
viamin, line 335: The fallback notice only removes the streaming assistant bubble. That does not clean up any `message_tool_call` / `message_tool_result` cards that were already appended by the failed attempt, so the UI can still show stale tool activity even though `FallbackLoop#discard_partial_attempt` deletes those rows before retrying. Please clear or reconcile those tool cards when `fallback_notice` arrives, and add a regression test that covers a tool round followed by fallback.

viamin (commented):
I found two correctness regressions in the new fallback path.

The retry cleanup now deletes every message inserted after the checkpoint, but chat turns are not serialized anywhere in this stack. A concurrent send/resolve request on the same session can append its own rows during the retry window, and this cleanup will remove those unrelated messages. The rollback needs to be scoped to rows created by the failing attempt only.

On the client side, the fallback notice only removes the in-flight assistant bubble. If the failed attempt already emitted tool-call or tool-result cards, those remain rendered even though the server deletes the corresponding rows before retrying, so the browser can show stale tool activity until a full reload.

PR #2765, opened 2026-06-30
[codex] Fix worktree post-checkout database setup

viamin:
## Summary

Fixes the local post-checkout migration guard for linked git worktrees that derive disposable database names.

The hook now ensures worktree-scoped development, cable, and test databases exist before running migration rollback or migrate commands. It only performs that ensure step when a worktree database suffix is present, so primary checkouts do not need CREATEDB privileges just to run ordinary branch-switch migrations.

## Root Cause

`config/database.yml` derives unique database names for linked worktrees, but `.githooks/post-checkout` could be the first Rails command to touch the derived database. When a branch switch introduced migrations, the hook ran `bin/rails db:migrate` before the disposable database existed.

## Validation

- `bin/rspec spec/scripts/post_checkout_hook_spec.rb spec/scripts/ensure_worktree_databases_script_spec.rb`
- `bin/rubocop spec/scripts/post_checkout_hook_spec.rb`
- `shellcheck -x .githooks/post-checkout`
- `bin/lint --changed`
- `git diff --check -- .githooks/post-checkout spec/scripts/post_checkout_hook_spec.rb`

### DIFF
--- a/.githooks/post-checkout
+++ b/.githooks/post-checkout
@@ -36,6 +36,27 @@ if [ -z "$removed" ] && [ -z "$added" ]; then
   exit 0
 fi
 
+echo ""
+echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+echo "  Migration changes detected on branch switch"
+echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
+
+worktree_database_suffix() {
+  ruby <<'RUBY'
+require File.expand_path("config/worktree_database_names", Dir.pwd)
+
+puts Paid::WorktreeDatabaseNames.suffix.to_s
+RUBY
+}
+
+if command -v bin/ensure-worktree-databases &>/dev/null && [ -n "$(worktree_database_suffix)" ]; then
+  echo "  Ensuring worktree databases..."
+  if ! bin/ensure-worktree-databases 2>&1; then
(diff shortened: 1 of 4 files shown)


.githooks/post-checkout
```diff
@@ -36,6 +36,27 @@ if [ -z "$removed" ] && [ -z "$added" ]; then
...
+  ruby <<'RUBY'
+require File.expand_path("config/worktree_database_names", Dir.pwd)
+
+puts Paid::WorktreeDatabaseNames.suffix.to_s
+RUBY
+}
+
+if command -v bin/ensure-worktree-databases &>/dev/null && [ -n "$(worktree_database_suffix)" ]; then
+  echo "  Ensuring worktree databases..."
+  if ! bin/ensure-worktree-databases 2>&1; then
```
viamin, line 54: This now turns every `bin/ensure-worktree-databases` failure into `exit 0`, which regresses linked worktrees that already have their derived databases but do not have CREATEDB/admin credentials. The helper in `bin/ensure-worktree-databases` chooses creator credentials before it checks `pg_database`, so it exits 1 even when nothing needs to be created. In that case the old hook would still run `db:migrate`, but this patch skips rollback/migrate entirely. Please either make the ensure script succeed when the databases already exist without creator privileges, or only skip when you can prove the databases are actually missing.

viamin (commented):
The new worktree database guard introduces one regression: in linked worktrees where the databases already exist but the application role lacks CREATEDB (and no admin credentials are configured), the hook now skips the entire migration guard. `bin/ensure-worktree-databases` exits before it ever checks whether the derived databases already exist, so a previously working worktree can stop running checkout migrations after this patch.

PR #3043, opened 2026-07-28
feat(db): auto-back-up populated DBs before destructive ops

viamin:
## What

Adds `Database::SafetyBackup`, a dev/test guardrail that snapshots any populated database **before** destructive Rails db operations, so a mistyped `bin/rails db:reset` / `db:drop` / `db:schema:load` / `db:purge` — or a stray `DatabaseTasks.load_schema` / `purge` / `truncate_tables` call from a console/runner — can never silently destroy data.

Motivated by a real incident: a diagnostic `DatabaseTasks.load_schema` call wiped a populated dev database, and **no backup existed** to recover from.

## How it works

Two complementary entry points (wired in `config/initializers/database_safety_backup.rb`):

- **Rake-level** (`after_initialize`): backs up before the DROP-path tasks `db:drop` / `db:reset` / `db:migrate:reset` and their `:all` + per-DB variants (whose data is gone before any schema-loading hook can fire).
- **Method-level** (prepended on `ActiveRecord::Tasks::DatabaseTasks#load_schema`, `#purge`, `#truncate_tables`): also covers `db:schema:load`, `db:purge`, `db:truncate_all`, `db:seed:replant`, the `db:test:*` rebuild path, `db:prepare`'s internal reload, and direct console/runner misuse.

Safety properties:
- **Fail-closed** — only `NoDatabaseError` is treated as "safely empty"; any transient connection error aborts the destructive task. `populated?` runs `ANALYZE` first so planner estimates reflect current data (avoids the stale-`n_live_tup` false-negative trap).
- **Backup validated** — each snapshot is checked for non-empty size **and** `pg_restore --list` shows restorable data items before the task proceeds.
- `SKIP_DB_BACKUP=1` bypass prints a **loud warning** so a leaked env var can't silently disable the guard. Skipped automatically in `production` and `CI=true`.
- `bin/db-dump`: dumps locked to `0600`, `backups/` to `0700` (dumps contain every tenant's data with RLS disabled).

## Review

Reviewed by an 8-persona pass (correctness, adversarial, reliability, data-integrity, security, performance, testing, maintainability). Key defects found and fixed:
- `populated?` used planner estimates as ground truth → false negatives (verified: `pg_stat_reset()` → 0) → no backup → data loss. Now `ANALYZE`-first.
- The `schema_migrations` precondition was *exactly* the wipe trigger — removed.
- `db:purge` / `db:truncate_all` / `db:test:*` / per-DB variants / direct `purge`/`truncate_tables` calls all bypassed the guard — now covered via prepend + prefix matching.
- Connection errors failed **open** — now fail closed.
- Dumps were world-readable (0644) — now 0600.
- No backup validation; silent `SKIP_DB_BACKUP`; `bin/db-dump` re-booting Rails; no dump timeout — all fixed.

Out of scope (documented): raw SQL `DROP TABLE`/`TRUNCATE` from migrations or console can't be intercepted.

## Verification
- 21 new specs green (incl. wiring regression tests that fail if the prepend detaches); RuboCop clean; pre-commit hooks pass.
- `db:prepare` (the originally broken flow) unchanged; a live `snapshot_if_populated!` produced a mode-0600, 266-item restorable archive.

## Docs
`CLAUDE.md` gains a "Destructive Database Operations" section documenting the policy and the guard's actual coverage and limits.


viamin:

PR #3059, opened 2026-07-29
fix(agent-runs): recover parked rate_limited runs, stop duplicate churn

viamin:
## Summary
Parked `rate_limited` agent runs froze the entire account queue **and** accumulated hundreds of duplicate runs. Two symptoms, one root cause: `rate_limited` was invisible to both duplicate-dedup and runner-availability recovery.

## Problem 1 — duplicate-run churn
`QueueAgentRunActivity#find_existing_run` deduped only against `UNFINISHED_STATUSES` (`queued/running/paused`), so a parked `rate_limited` run didn't count as in-flight. Re-triggering pumps (e.g. the PR CI-fix scanner) minted a fresh run every cycle while the prior one sat parked — one issue accumulated 78 duplicate runs.

**Fix:** `DEDUP_BLOCKING_STATUSES = UNFINISHED_STATUSES + [rate_limited]`, used in `find_existing_run`. A parked run now blocks duplicates until it re-queues or terminally fails.

## Problem 2 — queue frozen until a rate limit cleared
Parked runs recovered only on a *time* trigger (`StaleRunDetectorJob`, once `rate_limited_until` elapsed). When new runner capacity appeared (e.g. a fallback runner added while the only compatible runner was rate-limited), nothing woke them — the whole account's queue stayed parked for the original runner's full rate-limit window even though a fallback could serve the work immediately.

**Fix:** `Runners::RecoverParkedRunsJob` plus a `Runner` `after_commit` (`became_available_for_agent_runs?`) that fires on create / re-enable / undiscard to make the owner's parked runs due now, so `StaleRunDetectorJob` re-queues them. If the runner is still genuinely unavailable it just re-parks; the `MAX_RATE_LIMITED_REQUEUES` budget is preserved so a truly stuck run still terminally fails rather than looping forever.

## Notes
- **Root-cause follow-up (not in this PR):** `idx_agent_runs_unique_active_issue` excludes `rate_limited`, so `AgentRuns::CreateFollowup` (one-shot, post-analysis) can still mint a single *bounded* duplicate. Adding `rate_limited` to that index is the structural fix for every creation path and deserves its own migration with a dedup backfill.
- **Out-of-band data remediation already applied:** 437 orphaned `rate_limited` runs sitting on already-completed issues were cancelled (they would have re-queued on the rate-limit reset date and dispatched agents for finished work), and the account's parked runs were re-queued onto the newly-added fallback runners.

## Test plan
- `Runners::RecoverParkedRunsJob` (9 new) including the `Runner` callback firing on create / re-enable / undiscard and *not* on disable or unrelated updates.
- `QueueAgentRunActivity` rate_limited dedup for both issue and PR (2 new); full file green.
- `AgentRun`, `Runner`, `StaleRunDetectorJob`, `ScanPaidPrsActivity`, `CreateFollowup` suites green; `bin/lint --changed` clean.

### DIFF
--- a/app/jobs/runners/recover_parked_runs_job.rb
+++ b/app/jobs/runners/recover_parked_runs_job.rb
@@ -26,16 +26,21 @@ module Runners
       key: ->(user_id) { "runners_recover_parked_runs/#{user_id}" }
     )
 
+    def self.parked_runs_for(account)
+      account_project_ids = Project.where(account_id: account.id).select(:id)
+
+      AgentRun.rate_limited
+        .where("agent_runs.issue_id IS NOT NULL OR agent_runs.source_pull_request_number IS NOT NULL")
+        .where("agent_runs.rate_limited_until > ?", Time.current)
+        .where(project_id: account_project_ids)
+    end
+
     def perform(user_id)
       TenantContext.with_system_access do
         account = User.find_by(id: user_id)&.account
         return unless account
 
-        account_project_ids = Project.where(account_id: account.id).select(:id)
-        parked = AgentRun.rate_limited
-          .where.not(issue_id: nil)
-          .where("agent_runs.rate_limited_until > ?", Time.current)
-          .where(project_id: account_project_ids)
+        parked = self.class.parked_runs_for(account)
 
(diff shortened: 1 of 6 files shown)


viamin:

